PR #36, opened 2019-10-29
Fix missed material filter parser defintion

### LINKED ISSUE #33: Material filter needed
The filter for materials (`<material/>`) is needed to be added back in. Afaik the filter was supposed to have been changed from `<block/>` in proto 1.4.0, but as of right now, `<material>` is an unknown filter. Not having this filter breaks a lot of maps.

Much appreciated.

AustinLMayes:
Fixes #33


Electroid:
Thanks for catching this, there was a bad refactor a while back that got `world` and `material` mixed up.

PR #37, opened 2019-10-29
Add custom MOTD support

### LINKED ISSUE #28: Customizable MoTD
MoTD should be customizable so you can not only choose the colors you prefer, but also the format, and for example be able to add a newline.

https://github.com/Electroid/PGM/blob/master/src/main/java/tc/oc/pgm/listeners/PGMListener.java#L44

AustinLMayes:
Resolves #28

### DIFF
--- a/src/main/java/tc/oc/pgm/listeners/MotdListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/MotdListener.java
@@ -16,7 +16,7 @@ public class MotdListener implements Listener {
 
   private static final String MAP_NAME_KEY = "map.name";
   private static final String MAP_VERSION_KEY = "map.version";
-  private static final String STAtE_COLOR_KEY = "state.color";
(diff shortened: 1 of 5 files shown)


src/main/java/tc/oc/pgm/listeners/MotdListener.java
```diff
@@ -0,0 +1,76 @@
...
+import org.bukkit.event.server.ServerListPingEvent;
+import tc.oc.pgm.Config;
+import tc.oc.pgm.events.MatchLoadEvent;
+import tc.oc.pgm.events.MatchStateChangeEvent;
+
+public class MotdListener implements Listener {
+
+  private static final String MAP_NAME_KEY = "map.name";
+  private static final String MAP_VERSION_KEY = "map.version";
+  private static final String STAtE_COLOR_KEY = "state.color";
```
Pugzy: `STAtE_COLOR_KEY `

PR #42, opened 2019-10-31
Fix MOTD coloring

AustinLMayes:
Turns out `toString` contains the section symbol.


AustinLMayes:
@Brottweiler while this is open, and since you made the original suggestion issue, are there any other placeholders you would like to be able to use?

Brottweiler:
@AustinLMayes Not that I can think if. I was mainly interested in just being able to use the current map in the motd, so all the functions you added were more than enough! :)

PR #45, opened 2019-11-01
Pull out Match into Match and MatchImpl

Electroid:


deploy/server/server.properties
```diff
@@ -15,7 +15,7 @@ online-mode=true
 gamemode=1
 force-gamemode=true
 
-# Nether and end are not supported
+# Nether and finish are not supported
```
cswhite2000: This should stay as end

src/main/java/tc/oc/pgm/blitz/BlitzMatchModule.java
```diff
@@ -146,10 +150,10 @@ private void handleElimination(final MatchPlayer player, Competitor competitor)
               public void run() {
                 match.callEvent(eliminatedEvent);
                 if (player.getParty() instanceof Competitor) {
-                  match.setPlayerParty(player, match.getDefaultParty());
+                  match.setParty(player, match.getDefaultParty());
                 }
-                match.invalidateCompetitorRanking();
-                match.checkEnd();
+                match.calculateVictory();
```
cswhite2000: Is there a reason this is called twice?

src/main/java/tc/oc/pgm/commands/AdminCommands.java
```diff
@@ -101,7 +99,7 @@ public void cancelRestart(CommandSender sender) {
   }
 
   @Command(
-      aliases = {"end", "finish"},
+      aliases = {"finish", "finish"},
```
cswhite2000: "finish", "finish"

src/main/java/tc/oc/pgm/crafting/CraftingMatchModule.java
```diff
@@ -34,16 +35,16 @@ public void enable() {
...
     for (Recipe recipe : customRecipes) {
-      getMatch().getServer().addRecipe(recipe);
+      PGM.get().getServer().addRecipe(recipe);
     }
   }
 
   @Override
   public void disable() {
-    // Recipe changes affect all worlds on the server, so we make changes at match start/end
+    // Recipe changes affect all worlds on the server, so we make changes at match start/finish
```
cswhite2000: SportPaper has per world recipe changes, the server recipe methods loop through all of the worlds and set the recipes for each world.

deploy/server/server.properties
Electroid: Yea, thanks for catching, this was because I changed #end() to #finish() so it matches MatchPhase#FINISHED

src/main/java/tc/oc/item/Items.java
```diff
@@ -2,12 +2,7 @@
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
Pablete1234: Should probably not be changing the whole project over to wildcard imports. Wildcard imports are to be avoided, especially when they're gonna create such a big diff that it's gonna be a nightmare to deal with

src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
```diff
@@ -46,7 +46,7 @@
   //
   // This WILL leak a few entities now and then, because there are ways they can
   // die that do not fire an event e.g. the tick age limit, but this should be
-  // rare and they will only leak until the end of the match.
+  // rare and they will only leak until the finish of the match.
```
Pablete1234: This should stay as "end" instead of "finish"

src/main/java/tc/oc/pgm/cycle/CycleConfig.java
```diff
@@ -31,7 +31,7 @@ public Auto matchEmpty() {
   }
 
   public Auto matchEnd() {
-    return new Auto("match-end");
+    return new Auto("match-finish");
```
Pablete1234: Not sure if the config change is intended, probably the user-facing side should keep this as match-end, "finish" being more of an internal name

src/main/java/tc/oc/item/Items.java
Electroid: Hmm, there are wild cards in many places in the code.. I think I'll fix that in a separate commit.

Electroid:
Likely there might be some new bugs, but going to merge this before it gets too gigantic.

PR #50, opened 2019-11-04
Add channels package

PotatoStealer:
With love from lim_bo56 for resolving my issue.
(Note: This PR is unsigned because fuck windows)


Electroid:
Thanks for working on this! My concern is that this is adding a lot of boiler plate for the same functionality of `ChatCommands`. Why do we need to keep track of channel objects?

PotatoStealer:
I'm thinking objects are easier to work with than the things in the ChatCommands, but if there isn't anything else to these channels other than to chat, then I think this PR doesn't stand.

Electroid:
Yea, I just think as it stands right now, we don't need it.

PR #65, opened 2019-11-09
Fixed Timed Restarts

### LINKED ISSUE #9: Fix inaccurate /restart countdowns
`RestartManager` schedules a repeatable task to check if the server was queued to restart, then restarts. Instead of a repeatable task, it should queue a delayed task to run at the time it was scheduled restart. If the restart is "deferred" (see code) or cancelled, the task gets canceled.

ThatOneTqnk:
Uses a delayed task and only does its job if there are no deferrals. Task is cancelled if `RestartManager#cancelRestart` is called. Resolves #9 .

Renames `/restart` to `/timedrestart` (Otherwise, it uses the default `/restart` from Spigot)

Signed-off-by: ThatOneTqnk <nodeadbatteries@gmail.com>

### DIFF
--- a/src/main/java/tc/oc/pgm/restart/RestartManager.java
+++ b/src/main/java/tc/oc/pgm/restart/RestartManager.java
@@ -98,6 +98,10 @@ public class RestartManager implements Runnable, Listener {
   }
 
   public void requestTimedRestart(Duration duration) {
+    if (queuedRestartTask > 0) {
+      this.plugin.getServer().getScheduler().cancelTask(queuedRestartTask);
+      queuedRestartTask = 0;
+    }
     Instant queueStamp = Instant.now();
     long ticks = duration.getStandardSeconds() * 20;
     queuedRestartTask =
--- a/src/main/java/tc/oc/pgm/restart/RestartManager.java
+++ b/src/main/java/tc/oc/pgm/restart/RestartManager.java
@@ -35,6 +35,8 @@ public class RestartManager implements Runnable, Listener {
   private Instant queuedAt;
   private String reason;
 
+  private int timedQueuedRestartTaskID;
+
   public RestartManager(Plugin plugin) {
     checkState(instance == null);
     instance = this;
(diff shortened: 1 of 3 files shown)


src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -35,6 +35,8 @@ public static RestartManager get() {
   private Instant queuedAt;
   private String reason;
 
+  private int timedQueuedRestartTaskID;
```
Pablete1234: This needs a better name, probably `queuedRestartTask`?

src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -94,6 +96,28 @@ public boolean isRestartDeferredBy(RequestRestartEvent.Deferral deferral) {
     return this.deferrals.contains(deferral);
   }
 
+  public void requestTimedRestart(Duration duration) {
+    Instant queueStamp = Instant.now();
+    long ticks = duration.getStandardSeconds() * 20;
+    timedQueuedRestartTaskID =
+        this.plugin
+            .getServer()
+            .getScheduler()
```
Pablete1234: `Bukkit.getScheduler()` would make this more concise

src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -94,6 +96,28 @@ public boolean isRestartDeferredBy(RequestRestartEvent.Deferral deferral) {
...
+            .runTaskLater(
+                this.plugin,
+                () -> {
+                  if (!this.deferrals.isEmpty()) {
+                    timedQueuedRestartTaskID = 0;
+                    return;
+                  }
+                  timedQueuedRestartTaskID = 0;
```
Pablete1234: If you're going to reset the taskId anyways, just place it once before the conditional

src/main/java/tc/oc/pgm/restart/RestartManager.java
ThatOneTqnk: only did that to stay consistent with https://github.com/Electroid/PGM/blob/2c5a496911b9d3ca8e77ddbb62415c208a70e3b4/src/main/java/tc/oc/pgm/restart/RestartManager.java#L126-L128

src/main/java/tc/oc/pgm/restart/RestartManager.java
ThatOneTqnk: renamed

src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -94,6 +97,23 @@ public boolean isRestartDeferredBy(RequestRestartEvent.Deferral deferral) {
     return this.deferrals.contains(deferral);
   }
 
+  public void requestTimedRestart(Duration duration) {
```
Electroid: What happens if this method was already called? will the previous task get cancelled? Doesn't seem to do that right now.

ThatOneTqnk: Ah true, fixed.

Issue #56, opened 2019-11-06
Incorrect page display


ChemistryX:
![스크린샷 2019-11-06 오후 11 57 01](https://user-images.githubusercontent.com/10433434/68309667-1a352e80-00f2-11ea-8ceb-0b79dd17d81d.png)
The title says it has 7 pages of maps, but when I type `/maps 7` it says `Unknown page selected! 6 total pages`.
My server has 49 maps in total.

Electroid:
Fixed https://github.com/Electroid/PGM/commit/0f27708b02034030cfa87a1cb76bb5faf7c2bf5f

PR #67, opened 2019-11-09
Fix page display

xChevy:
#56

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/MapCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/MapCommands.java
@@ -73,8 +73,11 @@ public class MapCommands {
       throws CommandException {
     final Set<PGMMap> maps = ImmutableSortedSet.copyOf(library.getMaps());
 
-    int resultsPerPage = 8;
-    int pages = library.getMaps().size() / resultsPerPage + 1;
+    double resultsPerPage = 8;
+    int pages =
+        library.getMaps().size() / resultsPerPage <= 0
+            ? 1
+            : (int) Math.ceil(library.getMaps().size() / resultsPerPage);
 
     String listHeader =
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -73,8 +73,11 @@ public static void maplist(
       throws CommandException {
     final Set<PGMMap> maps = ImmutableSortedSet.copyOf(library.getMaps());
 
-    int resultsPerPage = 8;
-    int pages = library.getMaps().size() / resultsPerPage + 1;
+    double resultsPerPage = 8;
```
PotatoStealer: Keep this an int; why should this be a double?

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -73,8 +73,11 @@ public static void maplist(
...
+    int pages =
+        library.getMaps().size() / resultsPerPage <= 0
+            ? 1
+            : (int) Math.ceil(library.getMaps().size() / resultsPerPage);
```
PotatoStealer: This looks unnecessary, none of the variables you're dealing with needs to be a double. 
Note that division between two integers in Java returns the resultant value with the fractional part truncated away.

xChevy: I also though it was unnecessary but it is needed so the ceiling returns the upper number else it will just return the integer value. Example: 9 / 8 as integer returns 1 but using one value as double returns 2

Pablete1234: All being ints: `(size + resultsPerPage - 1) / resultsPerPage`

PR #70, opened 2019-11-10
DMs now send to both sender and receiver

### LINKED ISSUE #69: Cannot see your messages in /msg
When you /msg someone you can see their responses but not your own messages.

ThatOneTqnk:
Resolves #69 

Signed-off-by: ThatOneTqnk <nodeadbatteries@gmail.com>

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/ChatCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/ChatCommands.java
@@ -83,6 +83,12 @@ public class ChatCommands {
         message,
         "[" + ChatColor.GOLD + "DM" + ChatColor.WHITE + "] {0}: {1}",
         viewer -> viewer.getBukkit().equals(receiver));
+
+    send(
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/commands/ChatCommands.java
```diff
@@ -83,6 +83,12 @@ public void direct(MatchPlayer sender, Player receiver, @Text String message) {
         message,
         "[" + ChatColor.GOLD + "DM" + ChatColor.WHITE + "] {0}: {1}",
         viewer -> viewer.getBukkit().equals(receiver));
```
Pablete1234: Could probably just change this predicate to match both sender & receiver

ThatOneTqnk: That is actually what I did at first, although the message doesn't make too much sense then in the perspective of the sender

Me: does /msg yikes woah
[DM] ThatOneTqnk: woah

PR #72, opened 2019-11-10
Fix MatchManagerImpl.matches duplicated values

TheMolkaPL:
For some reason, matches were indexed by their ID and world name in the
same map. MatchManager.getMatches() returned values twice. This has been
seen in "Currently playing" announcer which broadcasted current map
twice.

### DIFF
--- a/src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
+++ b/src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
@@ -30,7 +30,8 @@ public class MatchManagerImpl implements MatchManager {
   private final MapLibrary library;
   private final MapLoader loader;
 
-  private final Map<String, Match> matches;
+  private final Map<String, Match> matchById;
+  private final Map<String, String> matchIdByWorldName;
   private final AtomicInteger count;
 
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -30,7 +30,8 @@
   private final MapLibrary library;
   private final MapLoader loader;
 
-  private final Map<String, Match> matches;
+  private final Map<String, Match> matchesById;
+  private final Map<String, Match> matchesByWorldName;
```
Pablete1234: Having 2 different maps that keep references to something as relevant as a Match, sounds like asking for trouble & memory leaks down the line. I think there has to be a better way to handle this than using 2 maps.

TheMolkaPL: I'm open to any suggestions.

Electroid: @TheMolkaPL How about `Map<String, String> matchIdByWorldName` ?

Issue #59, opened 2019-11-07
TNT doesn't hurt players, or destroys dropped items


Brottweiler:
Players cannot get hurt by TNT, and igniting TNT next to dropped items doesn't destroy them.

Electroid:
Fixed in #73

PR #73, opened 2019-11-10
Fix TNT not dealing damage

TheMolkaPL:
5a5ad2e4423ee588683b213dd37ed59c655be7d5 has changed how interactions are checked. Before, all non-players entities could interact to the match. This commit adds more checks to verify that all entities are players before cancelling interaction events.

Signed-off-by: TheMolkaPL <themolkapl@gmail.com>


TheMolkaPL:
Fixes Electroid/PGM#59

Issue #58, opened 2019-11-07
Game is broken


ChemistryX:
```
[17:04:31] [Server thread/ERROR]: Could not pass event PlayerJoinEvent to PGM v1.8-SNAPSHOT
org.bukkit.event.EventException
	at org.bukkit.plugin.java.JavaPluginLoader$1.execute(JavaPluginLoader.java:302) ~[sportpaper.jar:git-SportPaper-"374af6d"]
	at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:74) ~[sportpaper.jar:git-SportPaper-"374af6d"]
	at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:65) ~[sportpaper.jar:git-SportPaper-"374af6d"]
	at org.bukkit.plugin.SimplePluginManager.callEventHandler(SimplePluginManager.java:569) [sportpaper.jar:git-SportPaper-"374af6d"]
	at org.bukkit.plugin.SimplePluginManager.dispatchEvent(SimplePluginManager.java:534) [sportpaper.jar:git-SportPaper-"374af6d"]
	at org.bukkit.plugin.SimplePluginManager.callEventSynchronously(SimplePluginManager.java:527) [sportpaper.jar:git-SportPaper-"374af6d"]
	at org.bukkit.plugin.SimplePluginManager.callEvent(SimplePluginManager.java:521) [sportpaper.jar:git-SportPaper-"374af6d"]
	at org.bukkit.plugin.SimplePluginManager.callEvent(SimplePluginManager.java:496) [sportpaper.jar:git-SportPaper-"374af6d"]
	at net.minecraft.server.v1_8_R3.PlayerList.onPlayerJoin(PlayerList.java:319) [sportpaper.jar:git-SportPaper-"374af6d"]
	at net.minecraft.server.v1_8_R3.PlayerList.a(PlayerList.java:177) [sportpaper.jar:git-SportPaper-"374af6d"]
	at net.minecraft.server.v1_8_R3.LoginListener.b(LoginListener.java:161) [sportpaper.jar:git-SportPaper-"374af6d"]
	at net.minecraft.server.v1_8_R3.LoginListener.c(LoginListener.java:62) [sportpaper.jar:git-SportPaper-"374af6d"]
	at net.minecraft.server.v1_8_R3.NetworkManager.a(NetworkManager.java:232) [sportpaper.jar:git-SportPaper-"374af6d"]
[...]

PR #81, opened 2019-11-12
Fix CME in MatchManagerImpl.getMatches()

TheMolkaPL:
Resolves https://github.com/Electroid/PGM/issues/58#issuecomment-551157118


Electroid:
I would rather find where the CMEs are coming from rather than creating lots of `Immutable`s, but for now this is fine.

Issue #79, opened 2019-11-12
Avoid cycling to the same map


ChemistryX:
When match ends, it sometimes cycles to the exact same map

Electroid:
It's that random? Either way, we should *try* to avoid playing the same map.

PR #83, opened 2019-11-12
Avoid cycling to the same map

TheMolkaPL:
Resolves Electroid/PGM#79

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/MapCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/MapCommands.java
@@ -8,7 +8,9 @@ import java.lang.ref.WeakReference;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -46,10 +49,19 @@ public static PGMMap peekNextMap() {
...
+
     List<PGMMap> maps = new ArrayList<>(PGM.get().getMapLibrary().getMaps());
-    Collections.shuffle(maps);
-    setNextMap(maps.get(0));
-    return nextMap.get();
+    PGMMap next;
+    do {
+      Collections.shuffle(maps);
+      next = maps.get(0);
+    } while (maps.size() > 1 && Objects.equals(current, next));
```
Electroid: We typically haven't used the `do..while` convention in PGM, can we just use `while`?

TheMolkaPL: do/while and while are two different types of loops. do..while is the simplest way to loop at least once. A while condition will be complex here.

Electroid: Ok, that's fine

PR #87, opened 2019-11-13
Remove edit wand

TheMolkaPL:
It seems that edit wand isn't used anywhere.


Electroid:
It's used for WorldEdit for moderators.

TheMolkaPL:
Should be somehow configurable then.

Brottweiler:
@TheMolkaPL it is, in the worldedit config.

TheMolkaPL:
I don't even have WorldEdit. Teleportation tool and edit wand (or any other item you want) should be imo configurable in PGM.

Brottweiler:
There's no reason not to have WorldEdit though; the compass is useful for any player.

Electroid:
@TheMolkaPL To remove the wand, you can add a negative permission for the wand.

PR #106, opened 2019-12-08
Fix NPE when using an EnderPearl with ModifyBowProjectile module

### LINKED ISSUE #57: Enderpearl bow projectile error
Read: https://github.com/Electroid/PGM/issues/57#issuecomment-551205212

~~Add custom projectiles so maps like Ozone works properly, and so we can have cool snowball bows on Glacial Impact;~~

```xml
<modifybowprojectile>
    <projectile>EnderPearl</projectile>
    <velocityMod>2.5</velocityMod>
</modifybowprojectile>
```
```xml
<modifybowprojectile>
    <projectile>Snowball</projectile>
    <velocityMod>2.0</velocityMod>
</modifybowprojectile>
```

botinator:
Resolves #57

### DIFF
--- a/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
+++ b/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
@@ -4,6 +4,7 @@ import java.util.Random;
 import java.util.Set;
 import org.bukkit.World;
 import org.bukkit.entity.Arrow;
+import org.bukkit.entity.EnderPearl;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Projectile;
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
```diff
@@ -40,44 +41,52 @@ public ModifyBowProjectileMatchModule(
...
-    Plugin plugin = PGM.get();
-    Entity newProjectile;
-
-    if (this.cls == Arrow.class && event.getProjectile() instanceof Arrow) {
-      // Don't change the projectile if it's an Arrow and the custom entity type is also Arrow
-      newProjectile = event.getProjectile();
-    } else {
-      // Replace the projectile
-      World world = event.getEntity().getWorld();
+    if (event.getProjectile() instanceof Arrow) {
```
Pablete1234: You could avoid indenting the whole method by negating this if:
`if (!(event.getProjectile() instanceof Arrow)) return;`

src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
```diff
@@ -40,44 +41,52 @@ public ModifyBowProjectileMatchModule(
...
-      newProjectile.setFireTicks(oldEntity.getFireTicks());
-
-      if (newProjectile instanceof Projectile) {
-        ((Projectile) newProjectile).setShooter(oldEntity.getShooter());
-        ((Projectile) newProjectile).setBounce(oldEntity.doesBounce());
+      // EnderPearl isn't an entity type in world#spawn so this is a workaround
+      if (EnderPearl.class.isAssignableFrom(this.cls)) {
+        event.getEntity().launchProjectile(EnderPearl.class).setVelocity(oldEntity.getVelocity());
+        event.setCancelled(true);
+        return;
```
Pablete1234: This workarround should still follow the logic of the function, not return straight away without setting shoother etc.
Just replace `world.spawn(..)` with the `entity.launchProjectile(...)` method, and let other things be handled the same

src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
botinator: I don't really like this code block in general, because the projectile in theory should always be an instanceof an arrow because the event is originating from `EntityShootBowEvent` right?

src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
botinator: good catch

PR #107, opened 2019-12-09
Fix NPE on projectile source within ModifyBowProjectile module, Resolves #57

### LINKED ISSUE #57: Enderpearl bow projectile error

botinator:

This is related to PR #106, this is the iteration based on review feedback. Used the web editor for that version so decided to just split it into another PR

### DIFF
--- a/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
+++ b/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
@@ -2,11 +2,7 @@ package tc.oc.pgm.modules;
 
 import java.util.Random;
 import java.util.Set;
-import org.bukkit.World;
-import org.bukkit.entity.Arrow;
-import org.bukkit.entity.Entity;
-import org.bukkit.entity.LivingEntity;
-import org.bukkit.entity.Projectile;
+import org.bukkit.entity.*;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.entity.EntityDamageByEntityEvent;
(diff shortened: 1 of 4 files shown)


botinator:
This also fixes an unreported bug where custom projectiles weren't reporting the appropriate ownership on death. See Discord for more information.

src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
```diff
@@ -25,13 +21,13 @@
 
 @ListenerScope(MatchScope.RUNNING)
 public class ModifyBowProjectileMatchModule extends MatchModule implements Listener {
-  protected final Class<? extends Entity> cls;
+  protected final Class<? extends Projectile> cls;
```
Pablete1234: PGM allows for entities to be shot, not necessarily just projectiles, I don't think that functionality should be removed.

botinator: If this is something we're interested in keeping I can address it in the next couple of days, my thought process here was to simplify the solution using a native method that we know works. A lot of these custom entities have problems, for example: deaths caused by explosions, splash potion events, etc. I left the projectile definitions as is (entities) so this limit is just for this specific module.

Electroid: Although entities were technically supported, it was usually used to do weird stuff like creeper bows -- which caused other sorts of bugs.

To fix the ender pearl issue, I'm okay merging this fix.

PR #109, opened 2019-12-11
Fix NMS issue that prevented villager trades, Resolves #101

### LINKED ISSUE #101: Villagers don't work
The gui won't open, u can't trade with them.
(Clicking a villager will cancel your spriting.)

botinator:


Electroid:
Nice catch, thanks!

PR #111, opened 2019-12-12
Fetch and cache UUIDs from Contributor, Resolves Electroid#18

Meeples10:

I apologize for the quality of my code and the number of reviews it needed.


src/main/java/tc/oc/pgm/modules/InfoModule.java
```diff
@@ -117,6 +135,48 @@ public static InfoModule parse(MapModuleContext context, Logger logger, Document
...
+                          response.append(line.trim());
+                        }
+                        br.close();
+                        String n =
+                            new Gson()
+                                .fromJson(response.toString(), JsonObject.class)
+                                .get("username")
+                                .getAsString();
+                        Contributor.cacheName(uuid, n);
+                        System.out.println("CACHED: " + n + " (" + uuid.toString() + ")");
```
Electroid: No print statements please

src/main/java/tc/oc/pgm/modules/InfoModule.java
```diff
@@ -117,6 +135,48 @@ public static InfoModule parse(MapModuleContext context, Logger logger, Document
...
+                      } catch (Exception e) {
+                        // TODO proper exception handling
```
Electroid: I would just silence on error for now.

src/main/java/tc/oc/pgm/modules/InfoModule.java
```diff
@@ -117,6 +135,48 @@ public static InfoModule parse(MapModuleContext context, Logger logger, Document
...
+                        e.printStackTrace();
+                      }
+                    });
+          }
+        }
+        // end low quality code
```
Electroid: don't need to comment about that

src/main/java/tc/oc/pgm/modules/InfoModule.java
```diff
@@ -117,6 +135,48 @@ public static InfoModule parse(MapModuleContext context, Logger logger, Document
...
+                        String line = null;
+                        while ((line = br.readLine()) != null) {
+                          response.append(line.trim());
+                        }
+                        br.close();
+                        String n =
+                            new Gson()
+                                .fromJson(response.toString(), JsonObject.class)
+                                .get("username")
+                                .getAsString();
```
Electroid: Can this method just be in the `Contributor` class? `public boolean resolve()`

src/main/java/tc/oc/pgm/modules/InfoModule.java
```diff
@@ -117,6 +135,48 @@ public static InfoModule parse(MapModuleContext context, Logger logger, Document
...
+                .runTaskAsynchronously(
+                    PGM.get(),
+                    () -> {
+                      try {
+                        HttpURLConnection con =
+                            (HttpURLConnection)
+                                new URL("https://api.ashcon.app/mojang/v2/user/" + uuid.toString())
+                                    .openConnection();
+                        con.setRequestMethod("GET");
+                        con.setRequestProperty("Accept", "application/json");
```
Electroid: Need to add a timeout? Also would add a user-agent.

Meeples10: What would you prefer that the user-agent be?

src/main/java/tc/oc/pgm/modules/InfoModule.java
```diff
@@ -117,6 +129,22 @@ public static InfoModule parse(MapModuleContext context, Logger logger, Document
           throw new InvalidXMLException("Contributor must have either a name or UUID", child);
         }
 
+        if (uuid != null) {
+          if (Contributor.isUUIDCached(uuid)) {
+            name = Contributor.getCachedName(uuid);
+          } else {
+            Bukkit.getScheduler()
+                .runTaskAsynchronously(
```
Pablete1234: This is creating a new bukkit scheduled task for every contributor UUID, and if there are any repeated ones you may be querying for them multiple times.

Consider doing this once (instead of per-map & per-author) with all the unique different uuids loaded by all maps

src/main/java/tc/oc/pgm/map/Contributor.java
```diff
@@ -103,4 +113,37 @@ public boolean hasContribution() {
...
+
+  public static String getCachedName(UUID uuid) {
+    return cachedNames.get(uuid);
+  }
+
+  public static void cacheName(UUID uuid, String name) {
+    cachedNames.put(uuid, name);
+  }
+
+  public static void resolve(UUID uuid) throws IOException {
```
Pablete1234: The Contributor class shouldn't be handling this logic, move it to a name resolver or name cache util

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public static void writeCacheToDisk() throws IOException {
+    try (FileWriter out = new FileWriter(new File(PGM.get().getDataFolder(), "uuidcache"))) {
+      for (UUID uuid : cachedNames.keySet()) {
+        out.write(uuid.toString() + ":" + cachedNames.get(uuid) + "\n");
+      }
+    } catch (IOException e) {
+      throw new IOException(e);
+    }
+  }
```
Pablete1234: This is quite a dodgy way to persist the data, ideally we'd want bukkit to store it for us, as OfflinePlayers in it's user cache, not sure that there's an API for it.
If there's no easy way for bukkit to do it for us, just store & retreive it as a JSON file using GSON. 

Also i'd say, consider storeing more than just uuid-name, use a Map<UUID, UserCache> with a custom object that contains playername & last time it was updated, that way you can refresh usernames that are the oldest before you reach rate limit, if any.

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,102 @@
...
+    return new Gson().fromJson(response.toString(), JsonObject.class).get("username").getAsString();
+  }
+
+  public static void resolveAll() {
+    Bukkit.getScheduler()
+        .runTaskAsynchronously(
+            PGM.get(),
+            () -> {
+              for (UUID uuid : cachedNames.keySet()) {
+                if (cachedNames.get(uuid) == null) {
```
Pablete1234: Don't iterate the keyset and run a get for every key.
Either iterate the entryset, or better, make a set of pending names, and iterate (and remove) from it when resolveAll is called. That way you can keep track of resolved and not resolved names separately.

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,102 @@
...
+    while ((line = in.readLine()) != null) {
+      data.add(line);
+    }
+    in.close();
+
+    for (String s : data) {
+      String[] split = s.split(":", 2);
+      cachedNames.put(UUID.fromString(split[0]), split[1]);
+    }
+  }
```
Pablete1234: Same as above, this is quite a dodgy way to read the data

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,102 @@
...
+    con.setReadTimeout(5000);
+    BufferedReader br =
+        new BufferedReader(new InputStreamReader(con.getInputStream(), StandardCharsets.UTF_8));
+    StringBuilder response = new StringBuilder();
+    String line;
+    while ((line = br.readLine()) != null) {
+      response.append(line.trim());
+    }
+    br.close();
+    return new Gson().fromJson(response.toString(), JsonObject.class).get("username").getAsString();
```
Pablete1234: I'm pretty sure you can and should reuse the Gson object, declare it as a final static constant and reuse it elsewhere

src/main/java/tc/oc/identity/IdentityBase.java
```diff
@@ -64,7 +64,8 @@ public boolean isSelf(Player player) {
 
   @Override
   public boolean isSelf(CommandSender sender) {
-    return Bukkit.getPlayer(getPlayerId()).equals(sender);
+    Player p = Bukkit.getPlayer(getPlayerId());
```
Electroid: `p` -> `player`

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,122 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map.Entry;
+import java.util.UUID;
+import org.bukkit.Bukkit;
+import tc.oc.pgm.api.PGM;
+
+public class NameCacheUtil {
```
Electroid: Let's move this to `tc.oc.util`.

src/main/java/tc/oc/pgm/map/CachedPlayer.java
```diff
@@ -0,0 +1,36 @@
+package tc.oc.pgm.map;
+
+import java.util.UUID;
+
+public class CachedPlayer {
```
Electroid: I would move this to inside `NameCacheUtil` as an inner static class. eg.
```java
public class NameCacheUtil {
  static class NameCacheEntry {
    public final UUID uuid;
    public final long timestamp;
    public String name;
  }
  // No need for get/set methods, just do NameCacheEntry.name = "...";
}
```

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,122 @@
...
+
+  private static final Gson GSON = new Gson();
+
+  private static List<UUID> unresolved = new ArrayList<UUID>();
```
Pablete1234:
```suggestion
  private static Set<UUID> unresolved = new HashSet<UUID>();
```
Make this a set, otherwise there's no purpose to it since several maps with same author would be added multiple times here.

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,122 @@
...
+  private static HashMap<UUID, CachedPlayer> cache = new HashMap<UUID, CachedPlayer>();
```
Pablete1234:
```suggestion
  private static Map<UUID, CachedPlayer> cache = new HashMap<UUID, CachedPlayer>();
```

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,122 @@
...
+    br.close();
+    return GSON.fromJson(response.toString(), JsonObject.class).get("username").getAsString();
+  }
+
+  public static void resolveAll() {
+    Bukkit.getScheduler()
+        .runTaskAsynchronously(
+            PGM.get(),
+            () -> {
+              for (int i = 0; i < unresolved.size(); i++) {
```
Pablete1234:
```suggestion
              for (UUID uuid : unresolved) {
```

There's no reason to iterate with a for-i loop, and you won't be able to do that after  you use a set

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,122 @@
...
+    writer.beginArray();
+    for (Entry<UUID, CachedPlayer> entry : cache.entrySet()) {
+      writer.beginObject();
+      writer.name("uuid").value(entry.getKey().toString());
+      writer.name("name").value(entry.getValue().getName());
+      writer.name("timestamp").value(entry.getValue().getTimestamp());
+      writer.endObject();
+    }
+    writer.endArray();
+    writer.close();
```
Pablete1234: When using GSON, you don't need to handle this yourself, you can just let gson do it for you. Also if you use try-with-resources syntax you don't need to close it yourself.

```java
        try (FileWriter writer = new FileWriter(new File(...))) {
            GSON.toJson(cache, writer);
        }
```

src/main/java/tc/oc/pgm/map/NameCacheUtil.java
```diff
@@ -0,0 +1,122 @@
...
+              UUID.fromString(obj.get("uuid").getAsString()),
+              obj.get("name").getAsString(),
+              obj.get("timestamp").getAsLong());
+      // If the player's name has been fetched within the last seven days
+      if (System.currentTimeMillis() - player.getTimestamp() < 604800000L) {
+        cache.put(player.getUUID(), player);
+      } else {
+        unresolved.add(player.getUUID());
+      }
+    }
```
Pablete1234: Same applies here, you can directly load the cache from GSON.
```java
    try (FileReader reader = new FileReader(new File(...))) {
        cache = GSON.fromJson(reader, new TypeToken<HashMap<UUID, NameCacheEntry>>(){}.getType());
    }
```

Also, you shouldn't avoid old names getting in the cache (or remove them from the cache), you can just both keep them, and add to unresolved to have them updated. And try to give `604800000L` a better name over a comment
```java
    long ONE_WEEK_AGO = System.currentTimeMillis() - (7 * 24 * 60 * 60 * 1000);
    cache.values().stream().filter(ce -> ce.timestamp < ONE_WEEK_AGO).forEach(ce -> unresolved.add(ce.uuid));
```

src/main/java/tc/oc/pgm/util/NameCacheUtil.java
```diff
@@ -13,26 +12,27 @@
...
+import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 import org.bukkit.Bukkit;
 import tc.oc.pgm.api.PGM;
 
 public class NameCacheUtil {
 
   private static final Gson GSON = new Gson();
+  private static final long ONE_WEEK_AGO = 7 * 24 * 60 * 60 * 1000;
```
Pablete1234: This can't be a constant as time moves on, it needs to be inside `readCacheFromDisk`, maybe keep this as `ONE_WEEK` and have the method use `long ONE_WEEK_AGO = System.currentTimeMillis() - ONE_WEEK`;

src/main/java/tc/oc/pgm/util/NameCacheUtil.java
```diff
@@ -85,38 +84,32 @@ public static void resolveAll() {
...
+    try (FileReader reader =
+        new FileReader(new File(PGM.get().getDataFolder(), "uuidcache.json"))) {
+      cache = GSON.fromJson(reader, new TypeToken<HashMap<UUID, NameCacheEntry>>() {}.getType());
+    }
+    cache.values().stream()
+        .filter(entry -> entry.timestamp < ONE_WEEK_AGO)
+        .forEach(ce -> unresolved.add(ce.uuid));
+  }
+
+  public static class NameCacheEntry {
```
Pablete1234: Make this private, it's internal data that i don't think needs to be exposed

src/main/java/tc/oc/pgm/util/NameCacheUtil.java
```diff
@@ -13,26 +12,27 @@
...
+  private static Set<UUID> unresolved = new HashSet<UUID>();
+  private static Map<UUID, NameCacheEntry> cache = new HashMap<UUID, NameCacheEntry>();
 
   public static boolean isUUIDCached(UUID uuid) {
     return cache.containsKey(uuid);
   }
 
-  public static CachedPlayer getCachedPlayer(UUID uuid) {
+  public static NameCacheEntry getCachedPlayer(UUID uuid) {
     return cache.get(uuid);
```
Pablete1234: Just return a String as cached name, the NameCacheEntry is an internal thing:
```
NameCacheEntry entry = cache.get(uuid);
return entry == null ? null : entry.name;
```

Pablete1234:
Just some final small changes, and i think this is good to go

PR #114, opened 2019-12-13
Add Rotations 

BGMP:
Did my best to implement them correctly.


src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,127 @@
...
+  public boolean isEnabled() {
+    return enabled;
+  }
+
+  public void setEnabled(boolean enabled) {
+    this.enabled = enabled;
+  }
+
+  public List<PGMMap> getMaps() {
+    return maps;
```
TheMolkaPL: Could we return something immutable here?

src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,127 @@
...
+    return position;
+  }
+
+  public void load() {
+    configurationSection
+        .getStringList("maps")
+        .forEach(
+            map -> {
+              if (PGM.GLOBAL.get().getMapLibrary().getMapByNameOrId(map).isPresent()) {
+                maps.add(PGM.GLOBAL.get().getMapLibrary().getMapByNameOrId(map).get());
```
TheMolkaPL: Just save the map in a variable.

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,218 @@
...
+  public RotationManager(MatchManager matchManager, File rotationsFile) {
+    this.matchManager = matchManager;
+    this.rotationsFile = rotationsFile;
+
+    if (!rotationsFile.exists()) {
+      try {
+        FileUtils.copyInputStreamToFile(
+            PGM.GLOBAL.get().getResource("rotations.yml"), rotationsFile);
+      } catch (IOException e) {
+        e.printStackTrace();
```
TheMolkaPL: Should use a logger.

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+
+    loadRotations();
+    setActiveRotation(getRotationByName("default"));
```
TheMolkaPL: I wouldn't do these things in a construtor. Can execute these from outside instead? Also `setActiveRotation` is already called in `loadRotation()`. https://github.com/Electroid/PGM/pull/114/files#diff-485c495f5a184b3547785959ac5276d4R64

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,218 @@
...
+  private void saveCurrentPosition() {
+    FileConfiguration rotationFileConfiguration =
+        YamlConfiguration.loadConfiguration(rotationsFile);
+    rotationFileConfiguration.set(
+        activeRotation.getName() + ".position", activeRotation.getPosition());
+
+    try {
+      rotationFileConfiguration.save(rotationsFile);
+    } catch (IOException e) {
+      e.printStackTrace();
```
TheMolkaPL: Should use a logger.

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,218 @@
...
+                    + ChatColor.GOLD
+                    + "Rotations"
+                    + ChatColor.WHITE
+                    + "] "
+                    + ChatColor.GREEN
+                    + "Rotation has been set to "
+                    + ChatColor.AQUA
+                    + rotation.getName()
+                    + ChatColor.GREEN
+                    + " in order to better adjust to the current player count.");
```
TheMolkaPL: This should be translatable.

src/main/resources/rotations.yml
```diff
@@ -0,0 +1,18 @@
...
+default:
+  enabled: true
+  players: 1
+  position: 0                 # This helps the server resume rotations in the position they were before a server shutdown/restart
+  maps:
+    - Airship Battle
+    - Harb
+    - Race for Victory
+    - The Fenland
+    - Warlock
```
TheMolkaPL: No line ending.

src/main/java/tc/oc/pgm/commands/RotationCommands.java
```diff
@@ -0,0 +1,247 @@
...
+            + " ("
+            + ChatColor.AQUA
+            + rotation.getName()
+            + ChatColor.DARK_AQUA
+            + ")"
+            + " ("
+            + ChatColor.AQUA
+            + page
+            + ChatColor.DARK_AQUA
+            + " of "
```
TheMolkaPL: Should be translatable.

BGMP: Should I just plain create a translatable string for "of"? It wasn't translated on the MapCommands either, so I figured I'd just follow it's guide. https://github.com/Electroid/PGM/blob/ae0ad4750e80ce6266da1d914fdc89e69ad61283/src/main/java/tc/oc/pgm/commands/MapCommands.java#L94

src/main/java/tc/oc/pgm/commands/RotationCommands.java
```diff
@@ -0,0 +1,247 @@
...
+                ? ChatColor.DARK_AQUA.toString() + index
+                : String.valueOf(index);
+        return (indexString) + ". " + ChatColor.RESET + map.getInfo().getShortDescription(sender);
+      }
+    }.display(audience, maps, page);
+  }
+
+  @Command(
+      aliases = {
+        "rotationcheckout",
```
Electroid: Avoid too many command aliases. Only need 1 with a short cut.

src/main/java/tc/oc/pgm/commands/RotationCommands.java
```diff
@@ -0,0 +1,247 @@
...
+            + ChatColor.AQUA
+            + pages
+            + ChatColor.DARK_AQUA
+            + ") "
+            /*
+             * This below ensures that the string remains always colored. For some reason the logic behind this de-colored
+             * the text if it were to get put into a second line of chat when the header exceeded the maximum length
+             * of the first one
+             */
+            + ChatColor.translateAlternateColorCodes('&', "&9&m-----------");
```
Electroid: I think there is some Pagination or HeaderComponent code that you could use.

src/main/java/tc/oc/pgm/commands/RotationCommands.java
```diff
@@ -0,0 +1,247 @@
...
+            rotation.getPosition() + 1 == index
+                ? ChatColor.DARK_AQUA.toString() + index
+                : String.valueOf(index);
+        return (indexString) + ". " + ChatColor.RESET + map.getInfo().getShortDescription(sender);
+      }
+    }.display(audience, maps, page);
+  }
+
+  @Command(
+      aliases = {"rotations", "rots", "rts"},
```
Electroid: I'd say support `rotation` and `rot` command aliases, not `rt`. (this is for all of the commands)

src/main/java/tc/oc/pgm/commands/RotationCommands.java
```diff
@@ -0,0 +1,247 @@
...
+            + "["
+            + ChatColor.AQUA
+            + matchManager.getRotationManager().getActiveRotation().getName()
+            + ChatColor.WHITE
+            + "] "
+            + "Skipped a total of "
+            + ChatColor.AQUA
+            + positions
+            + ChatColor.WHITE
+            + " positions.");
```
Electroid: Translations please!

src/main/java/tc/oc/pgm/rotation/PGMMapOrderProvider.java
```diff
@@ -0,0 +1,13 @@
+package tc.oc.pgm.rotation;
+
+import tc.oc.pgm.map.PGMMap;
+
+public interface PGMMapOrderProvider {
```
Electroid: Please provide at least some basic documentation.

src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,130 @@
...
+    return position;
+  }
+
+  public void load() {
+    configurationSection
+        .getStringList("maps")
+        .forEach(
+            map -> {
+              final Optional<PGMMap> mapByNameOrId =
+                  PGM.GLOBAL.get().getMapLibrary().getMapByNameOrId(map);
```
Electroid: Do not use `PGM.Global` instead do `PGM.get()`

src/main/java/tc/oc/pgm/rotation/PGMMapOrderProvider.java
Electroid: Also the name of this interface is a little ambiguous.

src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,130 @@
...
+
+import com.google.common.base.Optional;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.bukkit.configuration.ConfigurationSection;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.map.PGMMap;
+
+public class Rotation implements PGMMapOrderProvider {
```
Electroid: How about this:
`PGMMapOrderProvider` -> `Rotation`
`Rotation` -> `SingleRotation implements Rotation`
`RotationManger` -> `RotationManager implements Rotation`

src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,130 @@
...
+    this.position = configurationSection.getInt("position");
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
```
Electroid: Are the `setName` `setMaps` and `setPlayers` methods necessary? When would this ever be used?

src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private void rotate() {
+    if (position + 1 == maps.size() || position + 1 > maps.size()) {
+      position = 0;
+    } else {
+      movePosition(1);
+    }
+  }
+
+  public void movePosition(int positions) {
```
Electroid: Should probably be `private`

src/main/java/tc/oc/pgm/rotation/Rotation.java
```diff
@@ -0,0 +1,130 @@
...
+    return nextMap;
+  }
+
+  @Override
+  public PGMMap getNextMap() {
+    return maps.get(position);
+  }
+
+  @Override
+  public PGMMap popFallbackMap() {
```
Electroid: Not sure what this method is suppose to do.

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,220 @@
...
+import tc.oc.pgm.map.PGMMap;
+
+public class RotationManager implements PGMMapOrderProvider {
+
+  private MatchManager matchManager;
+
+  private File rotationsFile;
+  private List<Rotation> rotations = new ArrayList<>();
+  private Rotation activeRotation;
+  private boolean isEvaluatingPlayerCount = true;
```
Electroid: What does this mean? Name is not clear, is this dynamic rotations?

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,220 @@
...
+    for (int i = 1; i < playerCounts.length; i++) {
+      int idistance = Math.abs(playerCounts[i] - activePlayers);
+      if (idistance < distance) {
+        accuratePlayerCountIndex = i;
+        distance = idistance;
+      }
+    }
+
+    updateRotation(playerCounts, accuratePlayerCountIndex);
+  }
```
Electroid: This code is very cryptic, and hard to understand exactly what's going on. What does this do exactly?

src/main/java/tc/oc/pgm/rotation/RotationManager.java
```diff
@@ -0,0 +1,220 @@
...
+  public PGMMap getNextMap() {
+    if (activeRotation != null) {
+      if (overriderMap == null) {
+        return activeRotation.getNextMap();
+      } else return overriderMap;
+    } else return popFallbackMap();
+  }
+
+  @Override
+  public PGMMap popFallbackMap() {
```
Electroid: Isn't this just: getRandomMap?

src/main/resources/rotations.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Each new rotation should follow the same configuration structure, within this file
+#
+# NOTE: Map names are case-sensitive
+#
+
+default:
+  enabled: true
+  players: 1
+  position: 0                 # This helps the server resume rotations in the position they were before a server shutdown/restart
```
Electroid: How about instead of storing an index, we store the map name of the first map. So:
`next_map: Harb`

That makes it easier for users to configure this.

src/main/java/tc/oc/pgm/rotation/RotationManager.java
BGMP: I was encouraged not to use getRandomMap anymore... I also throught it would be. @Pablete1234 ):

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: I mainly avoided `PGM.get()` because it figures as `@Depcrecated`. I guess I'll use it here anyways, but for future reference,  when should I use which?

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: I'll follow this scheme and apply it for the next update I push, and I'll also document it hopefully, so the comment above should be resolved by then.

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: Well the `setName` & `setMaps` I will remove, makes no sense to keep them there. But I was planning on creating a command for changing rotation's player count trigger from in-game

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: The skip command uses it [here](https://github.com/BGMP/PGM/blob/c7dea5d5ace460df0e7362570ae493469ee28f13/src/main/java/tc/oc/pgm/commands/RotationCommands.java#L228). Should I make it so it does so somehow else?

src/main/java/tc/oc/pgm/rotation/Rotation.java
Electroid: `PGM.get()`

src/main/java/tc/oc/pgm/rotation/RotationManager.java
Pablete1234: This method should probably not exist at all.
If the rotation failed to load, pgm should just load a RandomMapOrder (pick random map in the map pool) or a DefaultMapOrder (single map) for pgm to work

Pablete1234:
I had already told him this, but basically PGMMapOrderProvider is the only interface that the general code should be using. methods like "getRotationManager" assume there's a rotation, when there may not be one (eg: random map ordering, or voting based next map).

The rotation commands should be testing if matchManager.getMapOrder() is instance of a RotationManager, and if it isn't, just inform the user a rotation isn't in use.

Then commands like /setnext or /next can use directly the PGMMapOrderProvider interface as that's all they need to interact with

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: huh, I know I should use `PGM.get()` in this case,  but did you read my question? x.x

TheMolkaPL: You shouldn't use `GLOBAL` at all. This is probably just because it is an interface so the field is public. https://github.com/Electroid/PGM/blob/ae0ad4750e80ce6266da1d914fdc89e69ad61283/src/main/java/tc/oc/pgm/api/PGM.java#L61

src/main/java/tc/oc/pgm/rotation/PGMMapOrderProvider.java
Pablete1234: The point of the name is it should be open to basically whatever kind of ordering that can tell you what map should play next. getNextMap can be null, and pollNextMap must return a map.

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: Okay, I will avoid using `GLOBAL` at all. Thanks for the clarification.

src/main/java/tc/oc/pgm/commands/RotationCommands.java
Electroid: Sure, that's fine for now.

src/main/java/tc/oc/pgm/rotation/PGMMapOrderProvider.java
Electroid: Makes sense, ok with `PGMMapOrder`

src/main/java/tc/oc/pgm/rotation/Rotation.java
Electroid: Ok with `PGMMapOrder`, `FixedPGMMapOrder`, and `PGMMapOrderManager` @BGMP

src/main/java/tc/oc/pgm/rotation/Rotation.java
Electroid: Oh, then never mind.

BGMP: I guess I'm marking this as resolved then qwq.

src/main/resources/rotations.yml
BGMP: Resolved in https://github.com/Electroid/PGM/pull/114/commits/8047f78e56826a1ca8430811204fda8e6f95447c

src/main/java/tc/oc/pgm/rotation/RotationManager.java
BGMP: Made it less cryptic I guess, https://github.com/Electroid/PGM/pull/114/commits/223f8e1eaea620bb129a78e3b6c0eb629a5bc193

src/main/java/tc/oc/pgm/commands/RotationCommands.java
BGMP: Translated. https://github.com/Electroid/PGM/pull/114/commits/5f38fbe052b876db7f73496c14bdfdee0ed892a0

src/main/java/tc/oc/pgm/commands/RotationCommands.java
BGMP: These ones should be fine. https://github.com/Electroid/PGM/pull/114/commits/c569a6d24c1e086bf9568de2e589ac3f3930161d

src/main/java/tc/oc/pgm/commands/RotationCommands.java
BGMP: https://github.com/Electroid/PGM/pull/114/commits/c569a6d24c1e086bf9568de2e589ac3f3930161d

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: Solved here: https://github.com/Electroid/PGM/pull/114/commits/223f8e1eaea620bb129a78e3b6c0eb629a5bc193

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: Removed. https://github.com/Electroid/PGM/pull/114/commits/99cce8ad7f45acaedc80accac8a3329cf30dca1d

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: Does not exist anymore. https://github.com/Electroid/PGM/pull/114/commits/fe83e3a18ec9419e8184475a0ccd8c4fb5a5e93a

src/main/java/tc/oc/pgm/rotation/RotationManager.java
BGMP: Should be clearer now, with the classes being refactored. https://github.com/Electroid/PGM/pull/114/commits/fe83e3a18ec9419e8184475a0ccd8c4fb5a5e93a

src/main/java/tc/oc/pgm/rotation/RotationManager.java
BGMP: I have created a type of `PGMMapOrder` called `RandomPGMMapOrder`, hopefully that will do for this. https://github.com/Electroid/PGM/pull/114/commits/fe83e3a18ec9419e8184475a0ccd8c4fb5a5e93a

src/main/java/tc/oc/pgm/rotation/PGMMapOrderProvider.java
BGMP: I have added some basic documentation to all of the now refactored classes. Tell me if more is required. https://github.com/Electroid/PGM/pull/114/commits/fe83e3a18ec9419e8184475a0ccd8c4fb5a5e93a

BGMP: I have added `PGMMapOrder`

src/main/java/tc/oc/pgm/rotation/Rotation.java
BGMP: Done. https://github.com/Electroid/PGM/pull/114/commits/fe83e3a18ec9419e8184475a0ccd8c4fb5a5e93a

BGMP:
> 
> 
> I had already told him this, but basically PGMMapOrderProvider is the only interface that the general code should be using. methods like "getRotationManager" assume there's a rotation, when there may not be one (eg: random map ordering, or voting based next map).
> 
> The rotation commands should be testing if matchManager.getMapOrder() is instance of a RotationManager, and if it isn't, just inform the user a rotation isn't in use.
> 
> Then commands like /setnext or /next can use directly the PGMMapOrderProvider interface as that's all they need to interact with

Oke, I have implemented them the way you have requested me to: https://github.com/Electroid/PGM/pull/114/commits/fe83e3a18ec9419e8184475a0ccd8c4fb5a5e93a

src/main/java/tc/oc/pgm/commands/RotationCommands.java
```diff
@@ -0,0 +1,269 @@
...
+              + " ("
+              + ChatColor.AQUA
+              + page
+              + ChatColor.DARK_AQUA
+              + " of "
+              + ChatColor.AQUA
+              + pages
+              + ChatColor.DARK_AQUA
+              + ") "
+              + ChatColor.translateAlternateColorCodes('&', "&9&m-----------");
```
Electroid: TBH, there is probably a simpler way to do this, but I'm okay with this for now.

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
```diff
@@ -0,0 +1,161 @@
...
+                        "[Rotation] "
+                            + "["
+                            + name
+                            + "] "
+                            + map
+                            + " not found in map repo. Ignoring...");
+              }
+            });
+
+    Optional<PGMMap> next_map =
```
Electroid: In the future, Java variables should not have `_` in them. So this would be `nextMap`. Don't need to change this now.

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private int getMapPosition(PGMMap map) {
+    int count = 0;
+
+    for (PGMMap pgmMap : maps) {
+      if (pgmMap.getName().equals(map.getName())) break;
+      count++;
+    }
+
+    System.out.println(count);
```
Electroid: Oops, can't merge with this. Please remove!

BGMP: oops, forgot to remove it, it was a debug thingie

src/main/java/tc/oc/pgm/commands/RotationCommands.java
BGMP: ;_;

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
BGMP: Fixed https://github.com/Electroid/PGM/pull/114/commits/c6f69e3ed9c438b0e9c81ae6eaeb17762ba9463b

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -158,7 +164,7 @@ public void unloadMatch(@Nullable String id) {
   @Override
   public Optional<Match> cycleMatch(@Nullable Match oldMatch, PGMMap nextMap, boolean retry) {
     // Pop map out
-    MapCommands.popNextMap();
+    pgmMapOrder.popNextMap();
```
Pablete1234: this is popping the map off twice, as the nextMap parameter passed here, should be the popped map.

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
```diff
@@ -0,0 +1,160 @@
...
+      movePosition(1);
+    }
+  }
+
+  public void movePosition(int positions) {
+    if (((position + positions) + 1) > maps.size()) {
+      position = ((position + positions) + 1) % maps.size();
+    } else {
+      position = position + positions;
+    }
```
Pablete1234: Why isn't this just
`position = (position + positions) % maps.size();` ?
If you add an extra +1, you'll end at position 1 after finishing the rotation instead of position 0

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }
+
+  @Override
+  public PGMMap getNextMap() {
+    if (activeRotation != null) {
+      if (overriderMap == null) {
+        return activeRotation.getNextMap();
+      } else return overriderMap;
+    } else return null;
```
Pablete1234:
```
if (overriderMap != null) return overriderMap;
if (activeRotation != null) return activeRotation.getNextMap();
return null;
```

src/main/java/tc/oc/pgm/commands/provider/PGMMapProvider.java
```diff
@@ -45,10 +44,10 @@ public PGMMap get(CommandSender sender, CommandArgs args, List<? extends Annotat
         map = mapLibrary.getMapByNameOrId(fuzzyName).orNull();
       }
     } else if (isGoToNext(annotations)) {
-      map = MapCommands.peekNextMap();
+      map = matchManager.getMapOrder().getNextMap();
     }
 
-    if (map == null) {
+     if (map == null && !isGoToNext(annotations)) {
```
Pablete1234:
```suggestion
    if (map == null && !isGoToNext(annotations)) {
```

Electroid:
@BGMP Go ahead and look at @Pablete1234's suggestions first

BGMP:
@Electroid I am. Almost done

BGMP:
Blame DCO for the mess

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
```diff
@@ -0,0 +1,225 @@
...
+    int count = 0;
+    while (count < playerCounts.size()) {
+      newRotation = getRotationByPlayerCount(playerCounts.get(count));
+
+      if (playerCounts.get(count) >= activePlayers) {
+        break;
+      }
+
+      count++;
+    }
```
Pablete1234: Seems like a weird mix of list, foreach, sort, and while, that could be simplified to a single stream
```java
FixedPGMMapOrder newRotation = rotations.stream()
   .filter(FixedPGMMapOrder::isEnabled)
    .map(FixedPGMMapOrder::getPlayers)
    .sorted()
    .filter(playerCount -> playerCount >= activePlayers)
    .map(this::getRotationByPlayerCount)
    .findFirst().orElse(null);
```

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
```diff
@@ -0,0 +1,225 @@
...
+    AtomicReference<FixedPGMMapOrder> matchByPlayerCount = new AtomicReference<>();
+
+    rotations.forEach(
+        rotation -> {
+          if (rotation.getPlayers() == playerCount) {
+            matchByPlayerCount.set(rotation);
+          }
+        });
+
+    return matchByPlayerCount.get();
```
Pablete1234: Using an atomic reference to edit a value from inside a stream, is a bad idea. 
If you want to do the edit-variable loop, do a normal `for (FixedPgmMapOrder rot : rotations) {`
If you wanna do lambda-style, do make a proper stream:
```java
return rotations.stream()
    .filter(rotation -> rotation.getPlayers() == playerCount)
    .findFirst().orElse(null);
```

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
```diff
@@ -0,0 +1,225 @@
...
+
+    getRotations()
+        .forEach(
+            rotation -> {
+              if (rotation.getName().equals(name)) {
+                matchByName.set(rotation);
+              }
+            });
+
+    return matchByName.get();
```
Pablete1234: Same as before:
```java
return rotations.stream()
    .filter(rot -> rot.getName().equals(name))
    .findFirst().orElse(null);
```

src/main/java/tc/oc/pgm/cycle/CycleCountdown.java
```diff
@@ -40,6 +40,9 @@ protected Component formatText() {
   @Override
   public void onEnd(Duration total) {
     super.onEnd(total);
-    this.mm.cycleMatch(this.getMatch(), nextMap, false);
+    // For some reason, popNextMap() isn't being executed from within this function below
+    this.mm.cycleMatch(this.getMatch(), this.mm.getMapOrder().popNextMap(), false);
+    // This remains here for the above stated:
+    this.mm.getMapOrder().popNextMap();
```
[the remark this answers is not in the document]
BGMP: For some reason it doesn't. I was told by Electroid that he would let it pass.

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
```diff
@@ -105,25 +109,15 @@ private int getMapPosition(PGMMap map) {
...
-      position = position + positions;
-    }
+    position = (position + positions) % maps.size();
   }
 
   PGMMap getMapInPosition(int position) {
-    if (position == maps.size()) {
-      return getMapInPosition(0);
-    } else if (position > maps.size()) {
+    if (position > maps.size()) {
```
Pablete1234:
```suggestion
    if (position >= maps.size()) {
```
To properly restrict this, it should be invalid if == to size too.

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
```diff
@@ -134,7 +128,9 @@ PGMMap getMapInPosition(int position) {
                   + maps.size()
                   + " has been issued. Returning map in position 0 instead.");
       return getMapInPosition(0);
-    } else return maps.get(position);
+    }
+
+    return maps.get(position % maps.size());
```
Pablete1234:
```suggestion
    return maps.get(position);
```
Since the input is restricted, the % for normalization isn't needed

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
BGMP: Fixed! https://github.com/Electroid/PGM/pull/120/commits/ee1be6fa4580db48098150a6d586a056038b5afd

src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrder.java
BGMP: Fixed! https://github.com/Electroid/PGM/pull/120/commits/ee1be6fa4580db48098150a6d586a056038b5afd

Issue #103, opened 2019-12-05
Only the first defined spawn position is used


RuedigerLP:
i noticed it on "Toss Up" it also happened on other ffa maps, i dont know if it only happens on ffa maps.

Xerocoles:
Yeah I mentioned this in a previous issue but it was closed because the first part of what I was mentioning then just turned out to just be a mistake on my part, but this part was still very true.  FFAs, and especially Rage FFAs, are practically unplayable until this is fixed.

Pugzy:
Just investigated this with the "Toss Up" map and spawns seem to be working as intended.

In FFA all spawn points are checked with a `SpreadPointProvider`.
https://github.com/Electroid/PGM/blob/de9bf2b10e1d6e8ff9758166ad1d707f68974dff/src/main/java/tc/oc/pgm/points/SpreadPointProvider.java#L20 

This finds the best spawn point based on the distance of enemies away from them, the one with the highest distance is used. When testing this with no other players the same point `x=-22.5,y=11.0,z=-86.5,pitch=0.0,yaw=-85.0` is always used as it is the first tested and remains greater than/equal to subsequent spawn distance checks.

When testing this with 2+ players points are selected based on the distance which changes as players move around the map so different spawn points are used.

Do you have more info about your testing environment?
Have you tested this with multiple players?

Xerocoles:
Seems to work now, not sure what happened cause I think I had been using the updated code, but it seems to work for me now so that's good (or maybe it doesn't work when it's just one person, which wouldn't matter anyways)

RuedigerLP:
I tested alone.

Electroid:
Is this an issue or not?

Brottweiler:
I tested this now again on Iteritus. Spawns work as intended, but the OP refers to FFA maps and Pugzy tested the affected map mentioned in the OP and it seems to work there as well, and Xerocoles confirms it. Does not seem like an issue.

Xerocoles:
Yea now i can definitely clarify - it only happens when you're in the FFA match alone, which is how I originally saw it from xml stuff ofc

Brottweiler:
Sounds like a bug even if it's very low priority if it only affects FFA maps when there's 1 player.

Pablete1234:
How is it a bug? I feel like it's intended behaviour. Like @Pugzy mentioned, pgm finds the "safest" point (the one furthest away from other players), if you're alone that point doesn't change, and this feels like expected behaviour to me.
[...]

PR #115, opened 2019-12-15
Randomize FFA spawns of equal distance

Pugzy:
When multiple spawn points return the same distance from an enemy randomly select one.

This only really applies when there is a single player as described in #103.

### DIFF
--- a/src/main/java/tc/oc/pgm/points/SpreadPointProvider.java
+++ b/src/main/java/tc/oc/pgm/points/SpreadPointProvider.java
@@ -1,11 +1,14 @@
 package tc.oc.pgm.points;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/points/SpreadPointProvider.java
```diff
@@ -37,14 +40,14 @@ public Location getPoint(Match match, @Nullable Entity entity) {
           }
         }
 
-        if (bestDistance < nearest) {
+        if (bestDistance <= nearest) {
           bestDistance = nearest;
-          bestPoint = pos;
+          bestPoints.add(pos);
         }
```
Pablete1234: `if (bestDistance < nearest)`, you need to clear the list of bestPoints, otherwise you will be adding to the list all points wich become better, but without removing previous worse points.

PR #121, opened 2019-12-19
Fix KotF issues (sidebar, respawning, etc.), resolves Electroid#102

Meeples10:


Meeples10:
> 
> 
> You seem to be replacing every in-line if statement arround the code, i don't think you should be replacing them, it doesn't help anything, just makes commits bigger with more changes to review, and hides original source from git history in the future under a commit name that doesn't have any relation to it.

The formatter seems to be doing that. I haven't touched the if statements.

PR #126, opened 2019-12-21
Fix NPE on cycle

Meeples10:


src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -157,7 +157,7 @@ public void unloadMatch(@Nullable String id) {
       match.unload();
     }
 
-    matchIdByWorldName.remove(match.getWorld().getName());
+    matchIdByWorldName.remove("match-" + id);
```
TheMolkaPL: There should be a separate method to format world names, used by this and the `WorldCreator`.

PR #127, opened 2019-12-21
Add a method to get the world name associated with a match

Meeples10:
As suggested by @TheMolkaPL

### DIFF
--- a/src/main/java/tc/oc/pgm/match/MatchImpl.java
+++ b/src/main/java/tc/oc/pgm/match/MatchImpl.java
@@ -841,9 +841,4 @@ public class MatchImpl implements Match, Comparable<Match> {
         .append("state", getPhase())
         .build();
   }
-
-  @Override
-  public String getWorldName() {
-    return MatchManagerImpl.getWorldName(getId());
-  }
 }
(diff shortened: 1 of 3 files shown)


Electroid:
Maybe this should be `Match#getWorldName`?

src/main/java/tc/oc/pgm/match/MatchImpl.java
```diff
@@ -841,4 +841,9 @@ public String toString() {
         .append("state", getPhase())
         .build();
   }
+
+  @Override
+  public String getWorldName() {
+    return MatchManagerImpl.getWorldName(getId());
```
Electroid: Yea.. that actually won't work. We don't want `Match` to call `MatchManager`

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -329,7 +329,7 @@ public MatchPlayer getPlayer(@Nullable Player player) {
     return getMatches();
   }
 
-  private static String getWorldName(String matchID) {
-    return "match-" + matchID;
+  public static String getWorldName(String matchId) {
```
Electroid: Just keep this method, and keep it private, that's fine.

PR #132, opened 2019-12-22
Add a Datastore API for persistent data

Electroid:
* Stores data in a SQLite3 database
* Replaces NameCacheUtil with Username
* Allows for future expansions like settings, stats, and discord


src/main/java/tc/oc/pgm/util/UsernameResolver.java
```diff
@@ -61,15 +63,10 @@ public static void resolve(UUID id, @Nullable Consumer<String> callback) {
...
-
-    // If the queue is reaching capacity, try to queue an asynchronous resolve
-    if (QUEUE.size() >= 10) {
-      resolveAll();
-    }
   }
 
   private static void resolveAllSync() {
-    if (!LOCK.tryLock()) return;
+    LOCK.tryAcquire();
```
Pablete1234: why not just using java's synchronized keyword?

Electroid: Lock acquisition here is async. We actually try to acquire the both before the async task is scheduled and before the actual task starts. This is to prevent a dead lock state if someone calls the sync resolve without the async thread.

PR #134, opened 2019-12-23
Fix players not being frozen, and riding bossbar on death

### LINKED ISSUE #89: Dead players can freely move around
When a player dies, you are not bound to a minecart (or similar) and you can freely move around in creative mode. 

This also means that if you die in the void, you keep falling and take damage even after you died.

Edit: and on TDMs, you are sitting down but you are flying around at very high velocity the direction you are looking.

Pablete1234:
Fixes #89


Electroid:
Yea will need to move the code around, but let’s fix this

PR #136, opened 2019-12-24
Fix upper bounds issue with dynamic rotation scaling

botinator:

There was an issue in production on the `oc.tc` server where when the player count exceeded the last rotation option in the dynamic scaling rotation list it would set `updateActiveRotation` to null and freeze the server.

### DIFF
--- a/src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
+++ b/src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
@@ -120,10 +120,9 @@ public class FixedPGMMapOrderManager implements PGMMapOrder {
     rotations.stream()
         .filter(FixedPGMMapOrder::isEnabled)
         .map(FixedPGMMapOrder::getPlayers)
-        .sorted()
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/rotation/FixedPGMMapOrderManager.java
```diff
@@ -117,17 +117,14 @@ public void saveCurrentPosition() {
...
-            .map(this::getRotationByPlayerCount)
-            .findFirst()
-            .orElse(null);
-
-    updateActiveRotation(newRotation);
+    rotations.stream()
+        .filter(FixedPGMMapOrder::isEnabled)
+        .map(FixedPGMMapOrder::getPlayers)
+        .sorted()
+        .filter(playerCount -> playerCount >= activePlayers)
```
Pablete1234: Filter should be applied before sorting, as sorting is a stateful operation that will have to do less work if some rots are filtered out beforehand. Also, should probably not be sorting and just use a `min` reductor

botinator: Fixed and tested

PR #137, opened 2019-12-24
Add basic flair functionality, resolves Electroid#53

Meeples10:


src/main/java/tc/oc/pgm/Config.java
```diff
@@ -274,4 +280,65 @@ public static ChatColor finished() {
...
     }
   }
+
+  public static class Flairs implements Listener {
+    private boolean configMode;
+    private Map<String, Flair> flairs = new HashMap<String, Flair>();
+
+    @EventHandler
+    public void onConfigLoad(ConfigLoadEvent event) throws InvalidConfigurationException {
+      this.configMode = event.getConfig().getBoolean("flair.config-mode", true);
```
Electroid: `config-mode` -> `enabled`

Edit: See my idea for `groups`, this would no longer be needed. If no `flair` is defined, flairs aren't enabled.

src/main/java/tc/oc/pgm/Config.java
```diff
@@ -274,4 +280,65 @@ public static ChatColor finished() {
...
+      for (String key : section.getKeys(false)) {
+        flairs.add(
+            new Flair(
+                key,
+                section.getInt(key + ".priority"),
+                section.getString(key + ".symbol"),
+                ChatColor.valueOf(
+                    section.getString(key + ".color").trim().toUpperCase().replace(' ', '_'))));
+      }
+      flairs.sort((a, b) -> b.compareTo(a));
```
Electroid: Why sort if they get put in an un-sorted map?

src/main/java/tc/oc/pgm/Config.java
```diff
@@ -274,4 +280,65 @@ public static ChatColor finished() {
...
+      public ChatColor color;
+
+      public Flair(String name, int priority, String symbol, ChatColor color) {
+        this.name = name;
+        this.priority = priority;
+        this.symbol = symbol;
+        this.color = color;
+      }
+
+      public int compareTo(Flair other) {
```
Electroid: `Flair` should implement `Comparable<Flair>`

src/main/java/tc/oc/pgm/PGMImpl.java
```diff
@@ -249,6 +250,8 @@ public void onEnable() {
       return;
     }
 
+    FlairUtils.setMatchManager(matchManager);
```
Electroid: This isn't going to work, also for name renderer.

Instead, there should be a `PGM.getFlairRegistry()`

src/main/java/tc/oc/pgm/util/FlairUtils.java
```diff
@@ -0,0 +1,92 @@
...
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.tablist.PlayerTabEntry;
+import tc.oc.tablist.TabManager;
+
+public class FlairUtils {
+
+  private static Map<UUID, String> flairCache = new HashMap<UUID, String>();
+  private static MatchManager matchManagerInstance;
+  private static NameRenderer nameRendererInstance;
+  private static TabManager tabManagerInstance;
```
Electroid: This isn't going to work. Instead of a `FlairUtils` class, make a `FlairRegistry` api interface and `FlairRegistryImpl` class. `PGM` can have `getFlairRegistry`.

src/main/java/tc/oc/pgm/util/FlairUtils.java
```diff
@@ -0,0 +1,92 @@
...
+import tc.oc.pgm.Config.Flairs.Flair;
+import tc.oc.pgm.api.match.MatchManager;
+import tc.oc.pgm.api.party.Party;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.tablist.PlayerTabEntry;
+import tc.oc.tablist.TabManager;
+
+public class FlairUtils {
+
+  private static Map<UUID, String> flairCache = new HashMap<UUID, String>();
```
Electroid: Don't use `static` here, also this should probably be a `Cache<UUID, String> cache = CacheBuilder.newBuilder() ...`

src/main/java/tc/oc/pgm/util/FlairUtils.java
```diff
@@ -0,0 +1,92 @@
...
+    matchManagerInstance = instance;
+  }
+
+  public static void setNameRenderer(NameRenderer instance) {
+    nameRendererInstance = instance;
+  }
+
+  public static void setTabManager(TabManager instance) {
+    tabManagerInstance = instance;
+  }
```
Electroid: For similar reasons stated before, no need for this.

src/main/resources/config.yml
```diff
@@ -90,4 +90,17 @@ motd:
...
+                                                         # otherwise use flairs sent from an API
+  definitions:
+    admin:
+      priority: 1
+      symbol: ❖
+      color: gold
+    mod:
+      priority: 2
+      symbol: ❖
+      color: red
```
Electroid: So this may not make full sense now, but with some upcoming changes it would be better to do this:

```yml
groups:
   admin:
      priority: 1
      flair:
         symbol: ❖
         color: gold
     # Future additions (not to be implemented)
     #
     # permissions:
     #   - "my.other.plugin" 
     # discord:
     #   group: ...
```

src/main/java/tc/oc/named/NicknameRenderer.java
```diff
@@ -90,8 +94,11 @@ public Component getComponentName(Identity identity, NameType type) {
           new PersonalizedTranslatable("tip.teleportTo", dupe).render());
     }
 
-    if (type.style.showFlair && type.online && type.reveal) {
-      // rendered = <get flair> + rendered
+    if (type.style.showFlair && type.online && type.reveal && identity.isOnline(null)) {
```
Electroid: Why the `identity.isOnline(null)` ?

src/main/java/tc/oc/component/render/MatchNameRenderer.java
```diff
@@ -42,7 +43,7 @@ public void onPartyChange(PlayerPartyChangeEvent event) {
     Names.invalidate(player);
     final Party party = event.getNewParty();
     if (party != null) {
-      player.setDisplayName(/* flair + */ party.getColor() + player.getName() + ChatColor.WHITE);
+      player.setDisplayName(FlairUtils.getFlairedName(player, party));
```
Electroid: Maybe we need to implement a `MatchPlayer#getFlair() -> String` method?

src/main/java/tc/oc/pgm/Config.java
Meeples10: Could you expand on this?

Electroid: `groups` is just the syntax for `Config`, keep everything else flair related for your code. No mention of groups.

Pablete1234:
Should't flairs just be known as prefixes indtead? i think it's a more inclusive name for both flairs and tags

src/main/java/tc/oc/pgm/Config.java
Meeples10: So if there are no groups defined in the config, flairs are completely disabled? Or flairs are only used from the API?

src/main/java/tc/oc/pgm/util/FlairUtils.java
Meeples10: How would a `CacheBuilder` work when this exists just to store the value given by the API?

src/main/java/tc/oc/pgm/Config.java
Electroid: Yep. No groups, no flairs. No flair for a group, that group has no flairs.

Meeples10:
How does it look now?

src/main/java/tc/oc/pgm/flair/FlairRegistryImpl.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public void removeFlair(UUID uuid) {
+    flairCache.remove(uuid);
+  }
+
+  private String getFlairs(Player player) {
+    return Config.Flairs.enabled()
+        ? getConfigFlairs(player)
+        : (getAPIFlairs(player) != null ? getAPIFlairs(player) : "");
+  }
```
Pablete1234: I don't think this kind of behaviour should exist.
On the PGM side, there should just be some FlairProvider, wich is config-based. If a 3rd party API wants to override such provider, they should be able to set it. But making PGM's logic ignore all mentions of such api.

src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -125,6 +132,7 @@ public void removePlayerOnDisconnect(PlayerQuitEvent event) {
     }
 
     match.removePlayer(event.getPlayer());
+    PGM.get().getFlairRegistry().removeFlair(event.getPlayer().getUniqueId());
```
Pablete1234: `removeFlair` does not seem like a good name for this. FlairRegistry#removePlayer would make alot more sense

Brottweiler:
> Should't flairs just be known as prefixes indtead? i think it's a more inclusive name for both flairs and tags

Even though I prefer the name flairs, the name prefixes are used literally *everywhere* in the server community.

Meeples10:
Until the API is updated to set a `PrefixProvider`, PGM will fall back to the prefixes defined in the config file, or else no prefixes if none are defined.

src/main/java/tc/oc/pgm/prefix/ConfigPrefixProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public void removePlayer(UUID uuid) {
+    prefixMap.remove(uuid);
+  }
+
+  private String getPrefixFromConfig(UUID uuid) {
+    if (!Bukkit.getPlayer(uuid).isOnline()) {
+      return "";
+    }
```
Pablete1234: Just because a player isn't online doesn't mean they can't have flairs, right? Does bukkit not allow getting the permissions of offline players?

src/main/java/tc/oc/pgm/prefix/ConfigPrefixProvider.java
```diff
@@ -0,0 +1,46 @@
...
+    StringBuilder prefix = new StringBuilder();
+    for (Entry<String, Prefix> entry : Prefixes.getPrefixes().entrySet()) {
+      if (Bukkit.getPlayer(uuid).hasPermission("pgm.flair." + entry.getKey())) {
```
Pablete1234:
```suggestion
      if (Bukkit.getPlayer(uuid).hasPermission("pgm.group." + entry.getKey())) {
```
Will probably make more sense to just call this permission pgm.group.*

src/main/java/tc/oc/pgm/api/PGM.java
```diff
@@ -52,6 +57,12 @@
    */
   MapErrorTracker getMapErrorTracker();
 
+  PrefixRegistry getPrefixRegistry();
+
+  void setDefaultPrefixProvider(PrefixProvider prefixProvider);
+
+  PrefixProvider getDefaultPrefixProvider();
```
Pablete1234: Why `get/setDefaultPrefixProvider` and not just `get/setPrefixProvider` ? Also, i think they should go inside PrefixRegistry : `PGM.get().getPrefixRegistry().setPrefixProvider(provider)`

src/main/java/tc/oc/pgm/PGMImpl.java
```diff
@@ -147,6 +150,9 @@
   private IdentityProvider identityProvider;
   private NameRenderer nameRenderer;
 
+  private PrefixProvider defaultPrefixProvider;
```
Pablete1234: Why not keep the prefixProvider inside the prefix registry?

src/main/resources/config.yml
```diff
@@ -90,4 +90,16 @@ motd:
...
+  admin:
+    priority: 1
+    prefix:
+      symbol: ❖
+      color: gold
+  mod:
+    priority: 2
+    prefix:
+      symbol: ❖
+      color: red
```
Pablete1234: Maybe this should go in a different `groups.yml` config file, but for now it's probably fine if it stays here

Electroid: Ok for now.

src/main/java/tc/oc/pgm/prefix/ConfigPrefixProvider.java
iangbb: As far as I can see, it doesn't. Generally you can only obtain a Player object using the Bukkit API for a player who is online; I think Bukkit#getPlayer(UUID) usually returns null for offline players (and so this should probably be changed to reflect that).

For offline players, you need to get an OfflinePlayer, which doesn't extend Permissible and so doesn't have permissions. Furthermore, permissions aren't attached to players until they have joined, and attachments should be removed when they disconnect.

src/main/java/tc/oc/pgm/prefix/ConfigPrefixProvider.java
Meeples10: Why? This determines which prefixes a player should have by testing permissions, e.g. `pgm.group.admin` or `pgm.group.mod`.

src/main/java/tc/oc/pgm/api/PGM.java
Meeples10: The plugin loads its config before PGMImpl creates the PrefixRegistry, so this is the only way I could think of to set the PrefixProvider to a ConfigPrefixProvider.

PR #138, opened 2019-12-24
Optimize cycle

Pablete1234:


src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -159,6 +159,11 @@ public void unloadMatch(@Nullable String id) {
 
     matchIdByWorldName.remove(getWorldName(id));
     matchById.remove(id);
+    // Destroy the match 5 seconds after unload, to let the server catch-up
```
Electroid: Can we delay destroy until the server TPS recovers to a certain threshold? 5 seconds is arbitrary especially depending on machine type. Maybe it can wait a max of 1 minute?

src/main/java/tc/oc/pgm/match/MatchImpl.java
```diff
@@ -809,6 +797,23 @@ public void unload() {
     world.enqueue();
 
     loaded.compareAndSet(true, false);
+  }
+
+  @Override
+  public void destroy() {
+    if (isLoaded())
+      throw new IllegalStateException("Can't destroy a match that has not been unloaded");
```
Electroid: How about it unloads if it isn't already. I don't think destroy should ever fail, similarly with unload.

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
Pablete1234: This is all just still testing, we really don't really know the impact of this. I want this to be tested/profiled before actually deciding what exactly we do with this

src/main/java/tc/oc/pgm/match/MatchImpl.java
Pablete1234: That seems fine, but since destroy is called asyncronously, it may cause other issues. I'll log it as severe and call unload

Pablete1234:
This is a WIP thing, but i think it works well enough that it could  be reviewed and merged. Has settings to disable most if not all of the new behaviour and has been tested. Further refining could be done on a latter commit, as some things are starting to depend on this.

src/main/i18n/templates/pgm/PGMUI.properties
```diff
@@ -268,7 +268,9 @@ match.scoreboard.gs.title = Ghost Squadron
 tablist.authors.tooMany = type /map for info
 
 countdown.cycle.message = Cycling to {0} in {1}
+countdown.cycle.message.no_map=Cycling in {0}
```
Electroid: Nit: `=` needs proper spacing

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -31,15 +37,32 @@
...
-  private final Map<String, String> matchIdByWorldName;
+  private final Map<String, Match> matchById = new HashMap<>();
+  private final Map<String, String> matchIdByWorldName = new HashMap<>();
+  private final LoadingCache<PGMMap, String> preMatch =
+      CacheBuilder.newBuilder()
+          .expireAfterWrite(10, TimeUnit.SECONDS)
+          // If after 10 secs of writing the match it hasn't been loaded, unload & destroy it
+          .removalListener(
+              (RemovalNotification<PGMMap, String> r) -> {
+                if (!matchById.get(r.getValue()).isLoaded()) unloadMatch(r.getValue());
```
Electroid: Don't you need to check `r.isEvicked()` ?

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -50,16 +73,41 @@ public MatchManagerImpl(Server server, MapLibrary library, MapLoader loader)
...
     this.count = new AtomicInteger(0);
 
     loadNewMaps();
   }
 
-  @Override
-  public Match createMatch(@Nullable String id, PGMMap map) throws Throwable {
-    if (id == null || id.trim().isEmpty()) id = Integer.toString(count.get());
+  public void createPreMatchAsync(final PGMMap map) {
+    PGM.get().getLogger().fine("Creating pre-match for " + map.getName() + " async");
```
Electroid: This class has its own `logger`

Edit: Not just this line, but several in this class need this change.

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -139,6 +196,44 @@ private World createMatchWorld(String worldName, MapInfo info, TerrainModule ter
...
+    if (server.isPrimaryThread()) return server.createWorld(creator);
+
+    // FIXME: This is pretty dirty, there must be a better way to run & wait on main
+    final Object LOCK = new Object();
+    final AtomicReference<World> world = new AtomicReference<>();
+    final AtomicBoolean isDone = new AtomicBoolean();
+    PGM.get()
+        .getServer()
+        .getScheduler()
+        .runTask(
```
Electroid: This makes me think we need a `Scheduler` for `PGM` (outside `Match` scope) and also have it support async tasks. Because we've been using `getServer#getScheduler` a lot. This is a side note.

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
Pablete1234: I tried to find such a method, but all i had found is `r.getCause()` wich has a non-public `wasEvicted`. Turns out RemovalNotification exposes this via `r.wasEvicted()`.
I think however, checking if the map was loaded at all, is safer in regards to not leaving non-used matches floating arround and never unable to fully get removed.

PR #140, opened 2019-12-25
Remove WeakReference wrap on Location in MatchPlayerStateImpl

botinator:

This was causing the reference to be lost when called inside the `MatchPlayerDeathEvent` and subsequently causing NPE errors pertaining to the Proximity module. See #contributing channel for the initial report and explanation


Electroid:
We were seeing elevated problems with this, so I went ahead and committed a fix.

PR #141, opened 2019-12-25
Cleanup rotations

Pablete1234:
General cleanup of the rotations, trying to simplify some things and fix some bugs:
 - Renamed `FixedPGMMapOrderManager` to `RotationManager`
 - Renamed `FixedPGMMapOrder` to `Rotation`
 - Made `Rotation` immutable with the exception of `position` (current map)
 - Removed `/rotcheck`, `/rotationcheckout` command in favor of `/rot -r name`
 - Replaced check for rotation manager, using a CommandException simplifies it
 - Refactored MapLibrary's guava Optional to java's Optional
 - Added a more generic `matchEnded` optional method to PGMMapOrder, instead of the specific check for dynamic rotations
 - Added more documentation to`PGMMapOrder`
 - RandomPGMMapOrder will onlu shuffle the maps once
 - Removed all mentions of `isEvaluatingPlayerCount`, wich made no sense and was always true
 - Each `Rotation` is in charge of updating it's own map, instead of `RotationManager` doing it


src/main/java/tc/oc/pgm/rotation/PGMMapOrder.java
```diff
@@ -1,15 +1,40 @@
...
+  /**
+   * Returns the next map that will be played. Returning a null value is allowed and specifies the
+   * map order hasn't picked the next map yet.
+   *
+   * @return The next map to play, if present.
+   */
   PGMMap getNextMap();
 
+  /**
+   * Forces a specific map to be played next. The underlying PGMMapOrder may ignore this, but it is
```
Electroid: Use `{@link PGMMapOrder}` helps with renaming in the future.

Pablete1234: Fixed both there and on another comment

Issue #17, opened 2019-10-26
Add map voting


Electroid:
Similar to the `PickerMatchModule` there should be a module that displays 3-4 maps for players to vote on to play for the next match. Given the size of this feature, it would be best to discuss some options and approach.

Brottweiler:
I think voting is better than rotations (even if you can have both) since the players online can choose between a few maps and decide themselves what they want to play and not be forced by a rotation where they have to end the map quickly "because it sucks". 

Also makes for a more varied experience when you don't know exactly what map will be played next.

>When and how should the map voting be displayed?

At the game end. Show a GUI and you press wool blocks which correspond to maps, similar how the "rating" feature worked. Or just show it in chat. Also have a countdown so if no one votes it chooses a random map. Or if it's a tie, choose one of them maps at random.

Pablete1234:
Now that rotations have been added, i feel it's a good time to discuss this. I've been thinking about ways to implement this, and the best one that comes to mind, is:

Abstract "Rotations" to "MapPools", as in, config files, wich define a list of maps, and how many players should be on the match  to use such map pool.
Make it so that "DynamicRotations" become "DynamicMapPools", depending on how many players are online, it will use one map pool or another one. Allows for stuff like defined rotation for up to 20 players, and voting-based pool after that.

Adding a type of MapPool, that is a VotedMapPool, wich displays 5 maps to play after match end (during cycle time) to players. 

How this maps should be picked is an open topic, but i think a system based on previous votes makes sense, something along the lines of:
Show 5 random maps, using a weighted random based on the score of the map, the score is determined by players's votes.
The winner (1st map) is the one played, and it's votes/score is reset.
The losers get however many votes they got added to their score, so that next round they'll be more likely to be in the vote.

Such scoring system, allows for liked maps to be played even if they didn't make the cut first time (and since it's still random, any map could be up for voting and get a chance)

Electroid:
It would be nice if voting was not based on pure majority, but by increasing chances. So if 50% vote for A, 40% vote for B, and 10% vote for C -- those percentages are actually random weights.

Xerocoles:
[...]

PR #143, opened 2019-12-26
Add map pools & map voting

Pablete1234:
Includes #17 

Still WIP, only started laying out the classes for it


src/main/java/tc/oc/pgm/rotation/DisabledMapPool.java
```diff
@@ -0,0 +1,25 @@
+package tc.oc.pgm.rotation;
+
+import org.bukkit.configuration.ConfigurationSection;
+import tc.oc.pgm.map.PGMMap;
+
+public class DisabledMapPool extends MapPool {
```
Electroid: Isn't this basically a `RandomMapPool` would that be more descriptive of what this class does?

src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.*;
+import org.bukkit.configuration.ConfigurationSection;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.map.PGMMap;
+
+@SuppressWarnings("unchecked")
+public class VotingPool extends MapPool {
+
+  // Number of maps in the vote, unless not enough maps in pool
+  private static final int MAX_VOTE_OPTIONS = 5;
```
Electroid: We have to be careful about paradox of choice. What about 3 maps with an additional random map option?

src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private final double DEFAULT_WEIGHT;
```
Electroid: Would be nice if we gave `pgm.premium` double weight, possibly configurable.

Pablete1234: This weight is a map weight, not a vote weight.

If there are 10 maps, the default weight is 0.1 (10% chance)

src/main/java/tc/oc/pgm/rotation/VotingPool.java
Pablete1234: The idea is, show 5 maps, all of them at random from the pool, but weighted with their "how liked were they before"

src/main/java/tc/oc/pgm/rotation/DisabledMapPool.java
Pablete1234: This is not a random map pool, this is an always disabled map pool, it's used when you specify an invalid type of map pool, so it's not a rotation nor a voted pool, so it's an "always disabled map pool fallback", the random is only an ok-ish-enough default if for any reason someone calls pollNextMap, wich should never happen.

src/main/java/tc/oc/pgm/rotation/VotingPool.java
Electroid: I'm not sure I understand

Pablete1234: Imagine we had just 3 vote options, and a map pool with 4 maps:
 - Map 1: 20%
 - Map 2: 20%
 - Map 3: 20%
 - Map 4: 20%
 - Map 5: 20%

Randomly but based on weight, 3 options are picked (imagine map 1,2,3 are picked).
10 players vote, and give 8, 7, 3 votes to the maps (multiple choice is allowed)
The maps are updated to be:
 - Map 1: ~~90%~~ 0% (since it won the vote, it's picked to play, resets to 0%)
 - Map 2: 70%
 - Map 3: 30%
 - Map 4: 20%
 - Map 5: 20%

Randomly based on weight,3 options are picked (imagine map 3, 4 and 5 are picked)
10 players vote, and give 4 5 1 votes to the maps
 - Map 1: 5% (0 * 0.75 + 20 * 0.25)
 - Map 2: 57.5% (70 * 0.75 + 20 * 0.25)
 - Map 3: 40%
 - Map 4: ~~50%~~ 0% (won the vote, even if map 2 has higher score)
 - Map 5: 10%

With this type of system, there's no non-random picks, all are always polled at random, but the ones who were previously voted are more likely to get picked for voting, but not necessarily what must happen. Also if a map is not voted at all, it will land even lower than the avg, and is less likely to be picked than any other random map that hasn't yet been voted.

src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -0,0 +1,158 @@
...
+
+  @Override
+  public void matchEnded(Match match) {
+    if (manager.getOverriderMap() != null) return;
+    currentPoll = new Poll();
+    match.getPlayers().forEach(currentPoll::sendMessage);
+  }
+
+  /** Represents a polling process, with a set of options. */
+  public class Poll {
```
Electroid: If we're exposing this, how about we make it a separate class?

Pablete1234: Can't really make it a separate class, it's not a static inner class, it takes info from the VotingPool directly. Also it's not a class that i think will be reused, so it's fine if it stays there.

src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -0,0 +1,207 @@
...
+     * double votes.
+     *
+     * @param uuids The players who voted
+     * @return The number of votes counted
+     */
+    private int countVotes(Set<UUID> uuids) {
+      return uuids.stream()
+          .map(Bukkit::getPlayer)
+          // Count disconnected players as 1, can't test for their perms
+          .mapToInt(p -> p == null || !p.hasPermission("pgm.premium") ? 1 : 2)
```
Electroid: Use `Permissions` class

src/main/java/tc/oc/pgm/rotation/MapPoolManager.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private void updateActiveMapPool(MapPool mapPool) {
+    if (mapPool == activeMapPool) return;
+
+    setActiveMapPool(mapPool);
+
+    mapPoolFileConfig.set("last_active", activeMapPool.getName());
+    saveMapPools();
+
+    Bukkit.broadcastMessage(
```
Electroid: How about `MatchManager#getMatches()#sendMessage` ?

Pablete1234: I inherited this from older code, i'll actually pass the proper match to here and broadcast to the players on it

PR #145, opened 2019-12-26
Fix incorrect condition and use of deprecated methods in PrefixRegistryImpl

Meeples10:


src/main/java/tc/oc/pgm/prefix/PrefixRegistryImpl.java
```diff
@@ -28,19 +28,18 @@ public void onPrefixChange(PrefixChangeEvent event) {
...
       return;
     }
-    final MatchPlayer matchPlayer = PGM.get().getMatchManager().getPlayer(event.getUUID());
-    if (matchPlayer == null
-        || matchPlayer.getBukkit() == null
-        || matchPlayer.getBukkit().isOnline()) {
+    final Player player = Bukkit.getPlayer(event.getUUID());
+    if (player == null || !player.isOnline()) {
       return;
     }
```
Electroid: Why is the `Bukkit#getPlayer` check nessecary, can't you just use the `MatchManager#getPlayer` check?

src/main/java/tc/oc/pgm/prefix/PrefixRegistryImpl.java
```diff
@@ -28,19 +28,15 @@ public void onPrefixChange(PrefixChangeEvent event) {
     if (event.getUUID() == null) {
       return;
     }
-    final MatchPlayer matchPlayer = PGM.get().getMatchManager().getPlayer(event.getUUID());
-    if (matchPlayer == null
-        || matchPlayer.getBukkit() == null
-        || matchPlayer.getBukkit().isOnline()) {
+    final Player player = Bukkit.getPlayer(event.getUUID());
+    final MatchPlayer matchPlayer = PGM.get().getMatchManager().getPlayer(player);
```
Electroid: `final MatchPlayer player = PGM.get().getMatchManager().getPlayer(event.getUUID());` ?

Meeples10: Does it matter if I use `Player#getBukkit`? I wrote it like this to avoid using that.

Electroid: What I'm saying is use what I put :)

PR #149, opened 2019-12-27
[WIP] Refactor restarts

botinator:
Branch is dirty, this still needs a fair amount of polishing but just publishing the branch to get some initial feedback and see if this is a direction we want to head in before finishing it off.


botinator:
I'm probably going to end up renaming RestartTask back to RestartManager. I thought task was a better name for it, but as I started working on commands I realized they're rather intertwined unfortunately

src/main/java/tc/oc/pgm/Config.java
```diff
@@ -44,22 +44,21 @@ public static String getPath() {
...
-    public Duration time() {
-      return Duration.standardSeconds(this.config.getInt("restart.time", 30)); // seconds
+    public static int matchLimit() {
+      return getConfiguration().getInt("restart.match-limit", 30);
     }
 
-    public int matchLimit() {
-      return this.config.getInt("restart.match-limit", 30);
+    public static Duration taskInterval() {
+      return Duration.standardSeconds(getConfiguration().getInt("restart.interval", 60)); // seconds
```
Electroid: TBH, not sure why PGM ever let this be configured, there's no reason to change it. This is okay to be hard coded.

src/main/java/tc/oc/pgm/match/MatchPlayerStateImpl.java
```diff
@@ -27,15 +26,15 @@
   private final Match match;
   private final Identity identity;
   private final Party party;
-  private final WeakReference<Location> location;
```
Electroid: Rebase to master, because this is now outdated.

PR #150, opened 2019-12-27
Add a death message setting, defaults to own

Electroid:
Signed-off-by: Ashcon Partovi <ashcon@partovi.net>


botinator:
Can we make it default to `ALL`? Seeing death messages is actually really useful for knowing when someone has gotten behind you and killed defenders, etc - and most won't know the command to toggle this without assistance.

Electroid:
Chat gets *really* spammy with death messages, probably not the best first experience.

I agree the command is basically unknown, we'll need to implement a GUI someday down the road.

Pablete1234:
The vanilla behaviour of always showing death messages may be what players expect by default, but i agree, it gets spammy. 

One thing that i'd like to see tho, is getting a message if someone your participated on the kill dies. If i hit someone who a bit later gets shot by someone else, i'd like to see the player was killed. I have not looked into if the current tracker could allow for such behaviour, or if it's only ever saving latest hit for each player.

Xerocoles:
I'd agree that _by default_ it should be at 'all', I think it can be very valuable information even if you aren't defending, and it's good to get players using it, and then just, if they're actually annoyed, they can stop that.  I think they'd be a bit confused if they're only seeing their own, as well.

Electroid:
After discussion a good compromise is making team chat, global chat after the match ends. I'll add that in.

PR #155, opened 2019-12-29
Refactor restarts to be more concise, fix bugs, add countdown

botinator:


src/main/java/tc/oc/pgm/PGMImpl.java
```diff
@@ -339,7 +340,10 @@ public Identity getIdentity(UUID playerId, String username, @Nullable String nic
     prefixRegistry = new PrefixRegistryImpl();
     registerEvents(prefixRegistry);
 
-    new RestartManager(this);
+    if (Config.AutoRestart.enabled()) {
+      new RestartManager();
```
TheMolkaPL: `RestartManager.setInstance(new RestartManager());`?

Pugzy: The restart manager constructor does this anyway.

TheMolkaPL: Yes, but setting static variables from objects (non-static context) isn't right. If you don't want to call `setInstance` you can always use static initializer.

```java
private static final Kebabtallrik MEAL;

static {
    MEAL = new Kebabtallrik();
}
```

src/main/java/tc/oc/pgm/restart/RestartListener.java
```diff
@@ -0,0 +1,108 @@
...
+public class RestartListener implements Listener {
+
+  private final Plugin plugin;
+  private final Logger logger;
+
+  private int matchCount;
+  private @Nullable Integer matchLimit;
+  private @Nullable RequestRestartEvent.Deferral deferral;
+
+  public RestartListener(Plugin plugin) {
```
Electroid: Pass `MatchManager` here and don't use the `PGM.get()` elsewhere in this class.

src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -1,171 +1,92 @@
...
+import static com.google.common.base.Preconditions.checkState;
+
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.Nullable;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+import tc.oc.pgm.api.event.RequestRestartEvent;
+
+public class RestartManager {
```
Electroid: How about instead of having a static `RestartManager`, change all of its methods and fields to be static.

src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -1,171 +1,92 @@
...
+
+  private static RestartManager instance;
+  private final Set<RequestRestartEvent.Deferral> deferrals = new HashSet<>();
```
Electroid: Use `ConcurrentHashSet` so this is thread safe

src/main/java/tc/oc/pgm/restart/RestartManager.java
```diff
@@ -1,171 +1,79 @@
...
+import java.util.Set;
+import javax.annotation.Nullable;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+
+public class RestartManager {
+
+  private static final Set<RequestRestartEvent.Deferral> _deferrals = new HashSet<>();
+
+  private static Instant _queuedAt;
```
Electroid: No `_` variable please.

src/main/java/tc/oc/pgm/restart/RestartListener.java
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+  }
+
+  @EventHandler
+  public void onRequestRestart(RequestRestartEvent event) {
+    if (this.plugin.getServer().getOnlinePlayers().isEmpty()) {
+      Bukkit.getServer().shutdown();
+    } else {
+      Match match = this.matchManager.getMatches().iterator().next();
```
Electroid: Need to check if the match list is not empty, this will throw an exception if no matches are there.

src/main/java/tc/oc/pgm/restart/RestartManager.java
botinator: What should it be changed to? These are all internal variables

hugmanrique: deferrals and queuedAt without the underline (camelCase basically)

PR #167, opened 2019-12-31
Append extra plugin data in server list ping

TheMolkaPL:
PGM can now append extra information about itself in [server list ping](https://wiki.vg/Server_List_Ping). ~~This feature is toggleable in the configuration - switch `server-ping-data.enabled` to `false`.~~

SportPaper implementation: https://github.com/Electroid/SportPaper/pull/15
Related: https://github.com/ProjectEnyo/projectenyo.github.io/pull/4#issuecomment-569471250

Complete server ping response from my local server:

```json
{
    "description": "§7Â» §bAirship Battle §7Â«",
    "players": {
        "max": 20,
        "online": 0
    },
    "version": {
        "name": "SportPaper 1.8.8",
        "protocol": 47
    },
    "bukkit_extra": {
        "pgm": {
            "0": {
                "state": "idle",
                "duration": 0,
                "observers": 0,
                "participants": 0,
                "max_players": 100,
                "map": {
                    "slug": "airship_battle",
                    "name": "Airship Battle",
                    "version": "1.0.4",
                    "objective": "Leak lava from the enemy\u0027s obsidian core into the void.",
                    "authors": [
                        {
                            "uuid": "30e27366-0b14-4076-8f55-0819ece49ce3",
                            "name": "dewtroid"
                        }
                    ]
                },
                "next_map": {
                    "slug": "harb",
                    "name": "Harb",
                    "version": "1.3.5",
                    "objective": "Be the team with the least amount of deaths after 10 minutes.",
                    "authors": [
                        {
                            "uuid": "1379cb6e-f291-4498-9807-e636f9674ac0",
                            "name": "SH4D0W_HAWK"
                        },
                        {
                            "uuid": "ef4ea031-998f-4ec9-b7b6-1bdd428bcef8",
                            "name": "Plastix"
                        }
                    ]
                }
            }
        }
    }
}

```

### DIFF
--- a/src/main/java/tc/oc/pgm/api/match/MatchPhase.java
+++ b/src/main/java/tc/oc/pgm/api/match/MatchPhase.java
@@ -1,5 +1,6 @@
 package tc.oc.pgm.api.match;
 
+import java.util.Locale;
 import tc.oc.pgm.api.party.Competitor;
 import tc.oc.pgm.api.player.MatchPlayer;
 
@@ -49,5 +50,10 @@ public enum MatchPhase {
       default:
         throw new IllegalStateException("Unknown transition state for " + next);
     }
+  }
+
+  @Override
+  public String toString() {
+    return this.name().toLowerCase(Locale.ROOT);
   }
 }
--- a/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
@@ -5,10 +5,11 @@ import static com.google.common.base.Preconditions.*;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.server.ServerListPingEvent;
-import tc.oc.pgm.Config;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchManager;
@@ -21,20 +22,21 @@ import tc.oc.pgm.rotation.PGMMapOrder;
 public class ServerPingDataListener implements Listener {
 
   private final MatchManager matchManager;
-  private boolean ready;
+  private final AtomicBoolean ready;
 
   public ServerPingDataListener(MatchManager matchManager) {
     this.matchManager = checkNotNull(matchManager);
+    this.ready = new AtomicBoolean();
   }
 
   @EventHandler
   public void onMatchLoad(MatchLoadEvent event) {
-    ready = true;
+    ready.compareAndSet(false, true);
   }
 
   @EventHandler
   public void onServerListPing(ServerListPingEvent event) {
-    if (!Config.ServerPingData.enabled() || !ready) return;
+    if (!ready.get()) return;
 
     JsonObject root = event.getOrCreateExtra(PGM.get());
     for (Match match : this.matchManager.getMatches()) {
--- a/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
@@ -67,7 +67,7 @@ public class ServerPingDataListener implements Listener {
       try {
         root.add(matchId, this.matchCache.get(match));
       } catch (ExecutionException e) {
-        this.logger.log(Level.SEVERE, "Could not cache match " + matchId, e);
+        this.logger.log(Level.SEVERE, "Could not load server ping data for match: " + matchId, e);
       }
     }
   }
--- a/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
@@ -35,7 +35,7 @@ public class ServerPingDataListener implements Listener {
 
   public ServerPingDataListener(MatchManager matchManager, Logger parentLogger) {
     this.matchManager = checkNotNull(matchManager);
-    this.logger = ClassLogger.get(parentLogger, ServerPingDataListener.class);
+    this.logger = ClassLogger.get(checkNotNull(parentLogger), ServerPingDataListener.class);
     this.ready = new AtomicBoolean();
     this.matchCache =
         CacheBuilder.newBuilder()
--- a/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
@@ -34,7 +34,7 @@ public class ServerPingDataListener implements Listener {
     this.ready = new AtomicBoolean();
     this.matchCache =
         CacheBuilder.newBuilder()
-            .expireAfterWrite(1L, TimeUnit.SECONDS)
+            .expireAfterWrite(5L, TimeUnit.SECONDS)
             .build(
                 new CacheLoader<Match, JsonObject>() {
                   @Override
--- a/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
@@ -13,7 +13,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.server.ServerListPingEvent;
@@ -40,6 +39,7 @@ public class ServerPingDataListener implements Listener {
     this.ready = new AtomicBoolean();
     this.matchCache =
         CacheBuilder.newBuilder()
+            .weakKeys()
             .expireAfterWrite(5L, TimeUnit.SECONDS)
             .build(
                 new CacheLoader<Match, JsonObject>() {
--- a/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
@@ -1,0 +1,126 @@
+package tc.oc.pgm.listeners;
+
+import static com.google.common.base.Preconditions.*;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import java.util.UUID;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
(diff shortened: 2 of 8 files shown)


TheMolkaPL:
It failed to build because it was depending on SportPaper prior to https://github.com/Electroid/SportPaper/commit/d85121f17590656a46750a405e824a2824b3e122

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/runner/work/PGM/PGM/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java:[39,28] cannot find symbol
  symbol:   method getOrCreateExtra(tc.oc.pgm.api.PGM)
  location: variable event of type org.bukkit.event.server.ServerListPingEvent
```

src/main/resources/config.yml
```diff
@@ -92,6 +92,9 @@ motd:
     running: green                                       # Color designating a match which is currently running
     finished: red                                        # Color designating a match which has ended
 
+server-ping-data:
+  enabled: true
+
```
Electroid: TBH, we don't need a toggle for this. There's really no reason to disable it. Let's just always do it.

src/main/java/tc/oc/pgm/api/match/MatchPhase.java
```diff
@@ -50,4 +51,9 @@ public boolean canTransitionTo(MatchPhase next) {
         throw new IllegalStateException("Unknown transition state for " + next);
     }
   }
+
+  @Override
+  public String toString() {
+    return this.name().toLowerCase(Locale.ROOT);
```
Electroid: This seems like a breaking change? Why the change?

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,126 @@
...
+  private final MatchManager matchManager;
+  private boolean ready;
+
+  public ServerPingDataListener(MatchManager matchManager) {
+    this.matchManager = checkNotNull(matchManager);
+  }
+
+  @EventHandler
+  public void onMatchLoad(MatchLoadEvent event) {
+    ready = true;
```
Electroid: Use an `AtomicBoolean`

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,126 @@
...
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchManager;
+import tc.oc.pgm.api.match.event.MatchLoadEvent;
+import tc.oc.pgm.map.Contributor;
+import tc.oc.pgm.map.MapInfo;
+import tc.oc.pgm.map.PGMMap;
+import tc.oc.pgm.rotation.PGMMapOrder;
+
+public class ServerPingDataListener implements Listener {
```
Electroid: Isn't there already some `PingListener`? Can we have the Motd and this data all in one listener?

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,126 @@
...
+  @EventHandler
+  public void onServerListPing(ServerListPingEvent event) {
+    if (!Config.ServerPingData.enabled() || !ready) return;
+
+    JsonObject root = event.getOrCreateExtra(PGM.get());
+    for (Match match : this.matchManager.getMatches()) {
+      JsonObject matchObject = new JsonObject();
+      this.serializeMatch(match, matchObject);
+      root.add(match.getId(), matchObject);
+    }
```
Electroid: I think we should be caching this. We want server pings to be super fast.

Use a `Cache`

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -45,7 +46,7 @@
   private final MapLibrary library;
   private final MapLoader loader;
 
-  private final Map<String, Match> matchById = new HashMap<>();
+  private final Map<String, Match> matchById = new ConcurrentHashMap<>();
```
Electroid: I presume this is because the `PingListener` is getting this on an async thread?

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
TheMolkaPL: There is `MotdListener`. https://github.com/TheMolkaPL/PGM-1/blob/de337077ef99a96c0fc4051bc6359061ad049d73/src/main/java/tc/oc/pgm/listeners/MotdListener.java

I felt that it is cleaner to keep them separated. https://github.com/TheMolkaPL/PGM-1/blob/de337077ef99a96c0fc4051bc6359061ad049d73/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
TheMolkaPL: exactly

src/main/java/tc/oc/pgm/api/match/MatchPhase.java
TheMolkaPL: I checked all usages of `MatchPhase` and this is not breaking. I use this [here](https://github.com/TheMolkaPL/PGM-1/blob/de337077ef99a96c0fc4051bc6359061ad049d73/src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java#L53) to serialize `MatchPhase` (`starting` instead of `STARTING`, etc). I could do this in `ServerPingDataListener`, but I felt that doing this here is more universal.

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
TheMolkaPL: done

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,147 @@
...
+  @EventHandler
+  public void onServerListPing(ServerListPingEvent event) {
+    if (!ready.get()) return;
+
+    JsonObject root = event.getOrCreateExtra(PGM.get());
+    for (Match match : this.matchManager.getMatches()) {
+      try {
+        root.add(match.getId(), this.matchCache.get(match));
+      } catch (ExecutionException e) {
+        throw new RuntimeException(e);
```
Electroid: Log this and don't throw the error.

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,147 @@
...
+  private final MatchManager matchManager;
+  private final AtomicBoolean ready;
+  private final LoadingCache<Match, JsonObject> matchCache;
+
+  public ServerPingDataListener(MatchManager matchManager) {
+    this.matchManager = checkNotNull(matchManager);
+    this.ready = new AtomicBoolean();
+    this.matchCache =
+        CacheBuilder.newBuilder()
+            .expireAfterWrite(1L, TimeUnit.SECONDS)
```
Electroid: Too aggressive, let's cache for 5 seconds.

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,147 @@
...
+import tc.oc.pgm.map.Contributor;
+import tc.oc.pgm.map.MapInfo;
+import tc.oc.pgm.map.PGMMap;
+import tc.oc.pgm.rotation.PGMMapOrder;
+
+public class ServerPingDataListener implements Listener {
+
+  private final MatchManager matchManager;
+  private final AtomicBoolean ready;
+  private final LoadingCache<Match, JsonObject> matchCache;
```
Electroid: I'd rather not store a strong reference to the match, how about a string, which is the match id.

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,154 @@
...
+public class ServerPingDataListener implements Listener {
+
+  private final MatchManager matchManager;
+  private final Logger logger;
+  private final AtomicBoolean ready;
+  private final LoadingCache<Match, JsonObject> matchCache;
+
+  public ServerPingDataListener(MatchManager matchManager, Logger parentLogger) {
+    this.matchManager = checkNotNull(matchManager);
+    this.logger = ClassLogger.get(parentLogger, ServerPingDataListener.class);
```
Electroid: checkNotNull(parentLogger)

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,154 @@
...
+  public void onServerListPing(ServerListPingEvent event) {
+    if (!ready.get()) return;
+
+    JsonObject root = event.getOrCreateExtra(PGM.get());
+    for (Match match : this.matchManager.getMatches()) {
+      String matchId = match.getId();
+      try {
+        root.add(matchId, this.matchCache.get(match));
+      } catch (ExecutionException e) {
+        this.logger.log(Level.SEVERE, "Could not cache match " + matchId, e);
```
Electroid: `Could not load server ping data for match: `

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
TheMolkaPL: It's nullable in `ClassLogger`.

https://github.com/TheMolkaPL/PGM-1/blob/b3b5a18b7be80789e3a6b8187606a9de9dc91eff/src/main/java/tc/oc/util/logging/ClassLogger.java#L62-L70

Electroid:
Don't want to merge this yet, since if people don't update SportPaper things will break. Thinking about what's best to do here.

TheMolkaPL:
Merged changes from master 👌

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -0,0 +1,154 @@
...
+  @EventHandler
+  public void onMatchLoad(MatchLoadEvent event) {
+    ready.compareAndSet(false, true);
+  }
+
+  @EventHandler
+  public void onServerListPing(ServerListPingEvent event) {
+    if (!ready.get()) return;
+
+    JsonObject root = event.getOrCreateExtra(PGM.get());
```
Pablete1234: Can't you just try/catch this for the method not found exception/throwable, and return if it throws? AFAIK it's the only thing that uses a new sportbukkit method

PR #170, opened 2019-12-31
Hapy New Year 2020

TheMolkaPL:
Happy New Year guys!

I wish you more kebabs from me haha

### DIFF
--- a/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -16,7 +16,7 @@ import java.util.List;
 public enum SettingKey {
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
   DEATH(
-      Arrays.asList("death", "dms", "dm"),
+      Arrays.asList("death", "dms"),
       DEATH_ALL,
       DEATH_OWN); // Changes which death messages are seen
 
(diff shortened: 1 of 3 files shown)


src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -1,21 +1,35 @@
...
 /**
  * A toggleable setting with various possible {@link SettingValue}s.
  *
  * @see SettingValue
  */
 public enum SettingKey {
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
-  DEATH("death", DEATH_OWN, DEATH_ALL); // Changes which death messages are seen
+  DEATH(
+      Arrays.asList("death", "dms", "dm"),
```
Brottweiler: `/toggle dm` sounds like you are toggling private (direct) messages because when you get a private message it says `[DM]`.

TheMolkaPL: I have used old aliases from the 1.11 PGM.

https://github.com/OvercastNetwork/ProjectAres/blob/079d13ce62aee9a85f6871384a53356bd25735a8/PGM/src/main/java/tc/oc/pgm/death/DeathMessageSetting.java#L11

Electroid: Let's only support `death` and `dms`

Electroid:
@TheMolkaPL With the new settings, this needs git conflict resolution.

Issue #154, opened 2019-12-29
/map information


Xerocoles:
I'd like to request several things relating to /map

- On stratus we had two useful flags for _/maps,_ one of which was -a where it filtered by map author, and another I forget the letter, but it filtered by gamemode.

- Let's say we've got a 26v26 map. Instead of showing the max players number like it does now, 52, it makes sense to just say "26v26."  Because that's how map limits are thought of, people don't think "oh this is a 52 player map," it's the two team numbers.  This might be extra prevalent on Stratus because of rotation sizes and stuff, but I think it just makes sense everywhere.  And if you're a mod setnexting maps, you want to check the map sizes -- but when you check how many people are online, it's easier to just see "oh we're at a 27v29, probably shouldn't setnext this 26v26" rather than doing a bit of math to see.  I sort of overexplained that, but it just makes sense to me.

- Remove the genre thing, idek what it is or how it works but nobody uses it

RuedigerLP:
it would be cool if u could just do /maps -tags (playername,gamemode,future tags that could be added) future tags for tnt or 4-team or what ever could be useful to find a specific map

Xerocoles:
> it would be cool if u could just do /maps -tags (playername,gamemode,future tags that could be added) future tags for tnt or 4-team or what ever could be useful to find a specific map

That would be cool, but I don't think many people would be aware of it / use it practically ever, and it would require going through every single map's xml, which would be kinda painful

RuedigerLP:
we dont need to add any extra tags now, but it would allow for more tags in the future and also it would be easier to search if u can search for any tags and u dont need to remember the specific flag cus there is only one

TheMolkaPL:
How about something like #hashtags? Most of them would be automatic, but map authors could define their own. Let's say RFV2 would have `#wool` (automatic based wool module) and `#rfw`. You can do `/maps #rfw` to list all race for wool maps.

Other hashtags would be `#tnt` if TNT module is defined, #dte, #dtm, #4team, #ffa, etc.

RuedigerLP:
it should be #ctw not #wool, also would this also inplement searching for map author names?

also id say there should be categories of hashtags that cant be combined like:
- teams: ffa, 1team, 2team, etc.

other hashtags
- gamemodes
- supply-stuff: tnt, diamond, iron, gold etc.

Xerocoles:
I'm pretty sure that hashtags like that would go practically un-used by people in-game, and it's not worth our time.

[...]

PR #172, opened 2020-01-01
Map tags

TheMolkaPL:
Related: #154


src/main/java/tc/oc/pgm/maptag/MapTag.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Locale;
+import java.util.Objects;
+
+public class MapTag implements Comparable<MapTag> {
+  public static final char SYMBOL = '#';
+
+  private final String name;
+
+  protected MapTag(String name) {
+    this.name = checkNotNull(name).toLowerCase(Locale.US);
```
Pablete1234:
```suggestion
    this.name = checkNotNull(name).toLowerCase(Locale.ROOT);
```

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
```diff
@@ -0,0 +1,105 @@
...
+  StandardMapTag TEAMS = create("teams", TeamModule.class);
+  StandardMapTag TDM =
+      create(
+          "tdm",
+          ScoreModule.class,
+          score -> score.getConfig().deathScore != 0 || score.getConfig().killScore != 0);
+  StandardMapTag TIMELIMIT =
+      create("timelimit", TimeLimitModule.class, timeLimit -> timeLimit.getTimeLimit().isPresent());
+  StandardMapTag VANILLAWORLD =
+      create("vanillaworld", TerrainModule.class, terrain -> terrain.getOptions().vanilla);
```
Pablete1234: Maybe call this `#worldgen` ?

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
```diff
@@ -0,0 +1,105 @@
...
+
+public interface StandardMapTags {
+  StandardMapTag _4TEAMS = create("4teams", TeamModule.class, team -> team.getTeams().size() == 4);
+  StandardMapTag BLITZ = create("blitz", BlitzModule.class, blitz -> !blitz.isDisabled(null));
+  StandardMapTag CLASSES = create("classes", ClassModule.class);
+  StandardMapTag CORE = create("core", CoreModule.class);
+  StandardMapTag CP = create("cp", ControlPointModule.class);
+  StandardMapTag CRAFTING =
+      create("crafting", CraftingModule.class, crafting -> !crafting.getCustomRecipes().isEmpty());
+  StandardMapTag DESTROYABLE = create("destroyable", DestroyableModule.class);
```
Pablete1234: no one calls them destroyables, `#monument` ?

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
```diff
@@ -0,0 +1,105 @@
...
+  StandardMapTag EVENTEAMS =
+      create("eventeams", TeamModule.class, team -> team.shouldRequireEven().orElse(false));
+  StandardMapTag FFA = create("ffa", FreeForAllModule.class);
+  StandardMapTag FLAG = create("flag", FlagModule.class);
+  StandardMapTag FRIENDLYFIRE =
+      create("friendlyfire", InfoModule.class, info -> info.getMapInfo().friendlyFire);
+  StandardMapTag INTERNAL = create("internal", InternalModule.class);
+  StandardMapTag INSTANTTNT =
+      create("instanttnt", TNTModule.class, tnt -> tnt.getProperties().instantIgnite);
```
Pablete1234: Sounds misleading, maybe `#autotnt` or `#autoignote` ?

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
```diff
@@ -0,0 +1,105 @@
...
+import tc.oc.pgm.tnt.TNTModule;
+import tc.oc.pgm.wool.WoolModule;
+import tc.oc.pgm.worldborder.WorldBorderModule;
+
+public interface StandardMapTags {
+  StandardMapTag _4TEAMS = create("4teams", TeamModule.class, team -> team.getTeams().size() == 4);
+  StandardMapTag BLITZ = create("blitz", BlitzModule.class, blitz -> !blitz.isDisabled(null));
+  StandardMapTag CLASSES = create("classes", ClassModule.class);
+  StandardMapTag CORE = create("core", CoreModule.class);
+  StandardMapTag CP = create("cp", ControlPointModule.class);
```
Pablete1234: Even tho there's an xml difference between control points and hills, i think `#hill` may  be more user-friendly?

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
TheMolkaPL: Isn't `vanillaworld` more user friendly?

image

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
TheMolkaPL: DTE (Destroy The Enemy) uses destroyables to define islands. If we aren't calling them `#destroyable` I would go for `#dtm` instead.

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
TheMolkaPL: How about `#ctp` instead?

Pablete1234: Yeah i think `#ctp` would be nice

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
Pablete1234: You have `#core` and `#wool`, basically missing `#monument`, otherwise, rename them all to `#dtc`, `#ctw`, `#dtm`, but those names may be a more cryptic for users.

Destroy the enemy is just a type of map that uses a big monument, but it's still a monument. Code-wise they're called destroyables, but they're only called that to the user when it's destroy the enemy, and never on any other monument map (eg: warlock, snowy wars).

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
Pablete1234: i'm not sure, if i didn't know all the tags and i had to guess, i don't think i would've ever guessed `vanillaworld` to find a map that had vanilla world generation

TheMolkaPL: Actually none of maps in [maps.oc.tc](https://github.com/OvercastNetwork/maps.oc.tc) repo uses this feature. I'll make it longer and call it `#vanillaworldgen`.

src/main/java/tc/oc/pgm/PGMImpl.java
```diff
@@ -265,6 +268,8 @@ public void onEnable() {
     MapLoader mapLoader = new MapLoader(this, logger, registry);
     mapLibrary = new MapLibrary(logger);
 
+    StandardMapTags.registerDefaults(logger);
```
TheMolkaPL: I'm not sure if this should be called directly from `onEnable()`. Static initializers in interfaces aren't possible.
Related: https://github.com/TheMolkaPL/PGM-1/blob/a2cecd355b3894b8d4e22f712716f503d197b428/src/main/java/tc/oc/pgm/maptag/StandardMapTags.java#L41

src/main/java/tc/oc/pgm/commands/provider/MapTagsConditionProvider.java
```diff
@@ -0,0 +1,87 @@
...
+        mapTags.put(MapTag.forName(query), allow);
+      }
+    }
+
+    return new MapTagsCondition(mapTags);
+  }
+
+  @Override
+  public List<String> getSuggestions(
+      String prefix, Namespace namespace, List<? extends Annotation> modifiers) {
```
TheMolkaPL: This method isn't called after I put a space in the parameters. The expected behavior is that this should suggest maptags in each parameter. `/maps #tdm #<-- it's not suggesting here`

@Electroid Any ideas how I can fix this?

TheMolkaPL:
Everything else seems to be working fine. I have tested this on my local server.

src/main/java/tc/oc/pgm/commands/provider/MapTagsConditionProvider.java
TheMolkaPL: This issue seems to be deeper. I'll leave it as it is for now.

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
```diff
@@ -0,0 +1,136 @@
...
+import tc.oc.pgm.score.ScoreModule;
+import tc.oc.pgm.teams.TeamModule;
+import tc.oc.pgm.terrain.TerrainModule;
+import tc.oc.pgm.timelimit.TimeLimitModule;
+import tc.oc.pgm.tnt.TNTModule;
+import tc.oc.pgm.wool.WoolModule;
+import tc.oc.pgm.worldborder.WorldBorderModule;
+import tc.oc.util.logging.ClassLogger;
+
+public final class StandardMapTags {
```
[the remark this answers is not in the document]
TheMolkaPL: It allows to make private static methods and load the registry in static initalizer block.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -81,12 +72,26 @@ public void map(Audience audience, CommandSender sender, @Text PGMMap map) {
...
+            .createComponent(
+                (mapTag, component) -> {
+                  component.clickEvent(
+                      ClickEvent.Action.RUN_COMMAND, "/maplist " + mapTag.toString());
+                  component.hoverEvent(
+                      HoverEvent.Action.SHOW_TEXT,
+                      new PersonalizedTranslatable(
+                              "command.map.mapInfo.mapTag.hover", mapTag.toString())
+                          .render());
+                })
```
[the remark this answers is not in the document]
TheMolkaPL: How should I call the translation key then?

src/main/java/tc/oc/pgm/maptag/StandardMapTags.java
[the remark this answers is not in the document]
Pablete1234: Couldn't you just add a `boolean LOADED = someLoadingLogicMethod();`, wich will basically be your static "initializer block" ? (the method always returns true)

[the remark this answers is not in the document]
TheMolkaPL: Yes, but it exposes those static methods which should be kept private.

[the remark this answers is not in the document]
Pablete1234: You can prevent the method from running by checking for LOADED's value on the initializer method, but yeah, it starts to be a bit of a hack

PR #176, opened 2020-01-02
Implement custom sidebar messages, resolves Electroid#157

Meeples10:


src/main/resources/config.yml
```diff
@@ -108,4 +108,11 @@ experiments:
...
-  destroy-match-delay: 100
\ No newline at end of file
+  destroy-match-delay: 100
+
+sidebar-message:
+  enabled: false
+  # Text may be formatted with '&'
+  text: ""
+  # If the message should be positioned at the top or bottom of the sidebar
+  top: false
```
TheMolkaPL: Missing new line

src/main/java/tc/oc/pgm/Config.java
```diff
@@ -384,4 +384,19 @@ public int getDestroyMatchDelay() {
...
+    }
+
+    public static String format() {
+      return ChatColor.translateAlternateColorCodes(
+          '&', getConfiguration().getString("sidebar-message.text"));
+    }
+
+    public static boolean atTop() {
+      return getConfiguration().getBoolean("sidebar-message.top", false);
+    }
```
Pablete1234: Would maybe be easier to have message-top and message-bottom, with the ability to leave either of them blank?
That way you convert 3 settings into 2 (remove enabled & atTop, keep message-top and message-bottom), it gives more flexibility for people trying to configure it, and it's (maybe?) easier to code the condition.

src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -463,6 +463,17 @@ public int compare(Goal a, Goal b) {
...
+        if (Config.SidebarMessage.atTop()) {
+          rows.add(0, Config.SidebarMessage.format());
+        } else {
+          if (rows.size() >= MAX_ROWS) {
+            rows.set(MAX_ROWS - 1, Config.SidebarMessage.format());
+          } else {
+            rows.add(Config.SidebarMessage.format());
+          }
+        }
+      }
```
Pablete1234: Can we have a player setting to disable scoreboard messages?

TheMolkaPL: I think there should a setting `overwrite-existing` default to `false`. If they scoreboard if full already, we can lose important information about the objectives/match.

Meeples10: This wouldn't be possible with the way sidebars are currently rendered. The sidebar doesn't know which player is viewing it, only the player's team.

TheMolkaPL: Don't add the setting then lol ;p
https://github.com/Electroid/PGM/pull/176/commits/98246c32928f7490406181d41dd0c4a0d7004323#diff-f08f8cdb20626aaae1393d6a6ea2a79fR13-R16

TheMolkaPL:
Probably because it would require refactor of the whole sidebar rendering logic.

src/main/resources/config.yml
```diff
@@ -108,4 +108,12 @@ experiments:
   # Defer tab rendering to group more updates at once, in ticks
   tab-render-delay: 5
   # Ticks to wait to destroy older match
-  destroy-match-delay: 100
\ No newline at end of file
+  destroy-match-delay: 100
+
+sidebar-message:
```
Electroid: `sidebar`

src/main/resources/config.yml
```diff
@@ -108,4 +108,12 @@ experiments:
...
+  # Text may be formatted with '&'
+  # Leave blank to disable
+  message-top: ""
```
Electroid: `top`

src/main/resources/config.yml
```diff
@@ -108,4 +108,12 @@ experiments:
...
+  message-bottom: ""
```
Electroid: `bottom`

src/main/resources/config.yml
```diff
@@ -108,4 +108,12 @@ experiments:
...
+  # If the sidebar message should overwrite lines in the sidebar when it is full
+  overwrite-existing: false
```
Electroid: `overwrite`

PR #178, opened 2020-01-02
Cleanup sync createWorld mess

Pablete1234:
As suggested by a commit message, the CountDownLatch is a cleaner way to do this type of syncronization

### DIFF
--- a/src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
+++ b/src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
@@ -12,8 +12,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -197,39 +197,30 @@ private World createMatchWorld(String worldName, MapInfo info, TerrainModule ter
...
-      synchronized (LOCK) {
-        try {
-          LOCK.wait();
-        } catch (InterruptedException ignore) {
-        }
-      }
+    try {
+      latch.await();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
Electroid: Logger?

Pablete1234: This is the main thread, if it's getting interrupted it's likely shutting down the server anyways, but will add the logger

Electroid: `logger.log(Level.SEVERE, "...", e)`

We want to always use the logger for when I add Sentry/other logging support.

Pablete1234: Had already changed it, was failing to push stuff properly

PR #180, opened 2020-01-02
Add basic report command with alert and event

Pugzy:
Adds a basic report command which takes a player and a reason. This triggers a `PlayerReportEvent` and sends a formatted message to all players with the permission `Permissions.ADMINCHAT`.

e.g. image below (both player names are clickable, ignore my messed up flairs)

image

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
@@ -1,0 +1,70 @@
+package tc.oc.pgm.commands;
+
+import app.ashcon.intake.Command;
+import app.ashcon.intake.parametric.annotation.Text;
+import net.md_5.bungee.api.ChatColor;
+import org.bukkit.Bukkit;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+import tc.oc.component.Component;
+import tc.oc.component.types.PersonalizedText;
+import tc.oc.component.types.PersonalizedTranslatable;
+import tc.oc.named.NameStyle;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.chat.Audience;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.PlayerReportEvent;
+
+public class ModerationCommands {
+
+  @Command(
+      aliases = {"report"},
+      usage = "<player> <reason>",
+      desc = "Report a player who is breaking the rules")
+  public static void report(
+      CommandSender commandSender,
+      MatchPlayer matchPlayer,
+      Match match,
+      Player player,
+      @Text String reason) {
+    MatchPlayer accused = match.getPlayer(player);
+    PlayerReportEvent event = new PlayerReportEvent(commandSender, accused, reason);
+    match.callEvent(event);
+
+    if (event.isCancelled()) {
+      if (event.getCancelMessage() != null) {
+        commandSender.sendMessage(event.getCancelMessage());
+      }
+      return;
+    }
+
+    commandSender.sendMessage(
+        new PersonalizedText(
+            new PersonalizedText(new PersonalizedTranslatable("misc.thankYou"), ChatColor.GREEN),
+            new PersonalizedText(" "),
+            new PersonalizedText(
+                new PersonalizedTranslatable("command.report.acknowledge"), ChatColor.GOLD)));
+
+    final Component component =
+        new PersonalizedTranslatable(
+            "command.report.notify",
+            matchPlayer == null
+                ? new PersonalizedText("Console", ChatColor.AQUA, ChatColor.ITALIC)
+                : matchPlayer.getStyledName(NameStyle.FANCY),
+            accused.getStyledName(NameStyle.FANCY),
+            new PersonalizedText(reason.trim(), ChatColor.WHITE));
+
+    final Component prefixedComponent =
+        new PersonalizedText(
+            new PersonalizedText("["),
+            new PersonalizedText("A", ChatColor.GOLD),
+            new PersonalizedText("] "),
+            new PersonalizedText(component, ChatColor.YELLOW));
+
+    match.getPlayers().stream()
+        .filter(viewer -> viewer.getBukkit().hasPermission(Permissions.ADMINCHAT))
+        .forEach(viewer -> viewer.sendMessage(prefixedComponent));
+    Audience.get(Bukkit.getConsoleSender()).sendMessage(component);
+  }
+}
--- a/src/main/java/tc/oc/pgm/events/PlayerReportEvent.java
+++ b/src/main/java/tc/oc/pgm/events/PlayerReportEvent.java
@@ -1,0 +1,45 @@
+package tc.oc.pgm.events;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import org.bukkit.command.CommandSender;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.event.ExtendedCancellable;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+/** Called immediately AFTER a player runs the report command. */
+public class PlayerReportEvent extends ExtendedCancellable {
+
+  private final CommandSender sender;
+  private final MatchPlayer player;
+  private final String reason;
+
+  public PlayerReportEvent(CommandSender sender, MatchPlayer player, String reason) {
+    this.sender = checkNotNull(sender);
+    this.player = checkNotNull(player);
+    this.reason = checkNotNull(reason);
+  }
+
+  public CommandSender getSender() {
+    return sender;
+  }
+
+  public MatchPlayer getPlayer() {
+    return player;
+  }
+
+  public String getReason() {
+    return reason;
+  }
+
+  private static final HandlerList handlers = new HandlerList();
+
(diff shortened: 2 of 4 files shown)


src/main/java/tc/oc/pgm/commands/ReportCommands.java
```diff
@@ -0,0 +1,51 @@
...
+                    + "] "
+                    + ChatColor.DARK_AQUA
+                    + "{0}"
+                    + ChatColor.YELLOW
+                    + " \u2794 "
+                    + ChatColor.DARK_AQUA
+                    + "{1}"
+                    + ChatColor.YELLOW
+                    + ": "
+                    + "{2}",
```
TheMolkaPL: I think this should be defined in a constant?

src/main/java/tc/oc/pgm/events/PlayerReportEvent.java
```diff
@@ -0,0 +1,37 @@
+package tc.oc.pgm.events;
+
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.player.event.MatchPlayerEvent;
+
+/** Called immediately AFTER a player runs the report command. */
+public class PlayerReportEvent extends MatchPlayerEvent {
```
TheMolkaPL: Should be cancelable.

src/main/java/tc/oc/pgm/events/PlayerReportEvent.java
```diff
@@ -0,0 +1,37 @@
...
+
+  private final MatchPlayer sender;
+  private final String reason;
+
+  public PlayerReportEvent(MatchPlayer sender, MatchPlayer player, String reason) {
+    super(player);
+    this.sender = sender;
+    this.reason = reason;
```
TheMolkaPL:
```java
this.sender = checkNotNull(sender);
this.reason = checkNotNull(reason);
```

src/main/java/tc/oc/pgm/commands/ReportCommands.java
```diff
@@ -0,0 +1,51 @@
...
+                sender.getStyledName(NameStyle.FANCY),
+                accused.getStyledName(NameStyle.FANCY),
+                new PersonalizedText(reason.trim())));
+
+    match.getPlayers().stream()
```
TheMolkaPL: Before sending to players please log this in a logger.

TheMolkaPL:
Imo there should also be `/reports` which prints last reports. In case all admins were offline and you want to see who got reported during that time.

Brottweiler:
> Imo there should also be `/reports` which prints last reports. In case all admins were offline and you want to see who got reported during that time.

I do not think we need to store reports, it isn't *that* useful for us. Simply display them to all online staff.

Pablete1234:
The event is launched precisely to allow a 3rd party plugin persisting the data if they want to. I don't think PGM needs to store reports for them to be effective, rather treat the base /report as a way to message online staff.

Also, i'm not convinced that it should be prefixed with `[R]` instead of `[A] PlayerA reported PlayerB: reason`, for admin chat

src/main/java/tc/oc/pgm/commands/ReportCommands.java
```diff
@@ -0,0 +1,51 @@
...
+import tc.oc.component.Component;
+import tc.oc.component.types.PersonalizedText;
+import tc.oc.named.NameStyle;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.PlayerReportEvent;
+import tc.oc.util.components.Components;
+
+public class ReportCommands {
```
Pablete1234: I think this would fit better as just ModerationCommands, for when other commands are added

src/main/java/tc/oc/pgm/commands/ReportCommands.java
Pablete1234: I'm pretty sure this would be shorter if you used the text component stuff instead of string concat:
```java
new PersonalizedComponent(
  new PersonalizedComponent("["),
  new PersonalizedComponent("A", ChatColor.GOLD),
  new PersonalizedComponent("] "),
  sender.getStyledName(NameStyle.FANCY),
  new PersonalizedComponent(" \u2794 ", ChatColor.YELLOW),
...
```

src/main/java/tc/oc/pgm/commands/ReportCommands.java
Electroid: Sending to console is good for this

src/main/java/tc/oc/pgm/commands/ReportCommands.java
Pugzy: Changed to components.

src/main/java/tc/oc/pgm/commands/ReportCommands.java
Pugzy: Renamed

src/main/java/tc/oc/pgm/commands/ReportCommands.java
Pugzy: Moved to components rather than a string.

Pugzy:
New formatted screenshot: 
image

How can I prevent console from running the command?
Should the sender receive a message if the event is cancelled?

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -0,0 +1,55 @@
...
+  @Command(
+      aliases = {"report"},
+      usage = "<player> <reason>",
+      desc = "Report a player who is breaking the rules")
+  public static void report(MatchPlayer sender, Match match, Player player, @Text String reason) {
+    MatchPlayer accused = match.getPlayer(player);
+    PlayerReportEvent event = new PlayerReportEvent(sender, accused, reason);
+    match.callEvent(event);
+
+    if (!event.isCancelled()) {
```
TheMolkaPL: It would be cleaner with

```java
if (event.isCancelled()) return;
```

TheMolkaPL:
> How can I prevent console from running the command?

Why would you prevent?

> Should the sender receive a message if the event is cancelled?

Hmm, how about using `ExtendedCancellable` instead of the standard `Cancellable`?

src/main/java/tc/oc/pgm/commands/ReportCommands.java
Pugzy: Now sending to console with `[A]` prefix removed.

Pugzy:
Replaced sender in the event to be a `CommandSender` to allow for Console reports.

Updated event to extend `ExtendedCancellable` and relayed message to the sender.

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -0,0 +1,70 @@
...
+      aliases = {"report"},
+      usage = "<player> <reason>",
+      desc = "Report a player who is breaking the rules")
+  public static void report(
+      CommandSender commandSender,
+      MatchPlayer matchPlayer,
+      Match match,
+      Player player,
+      @Text String reason) {
+    MatchPlayer accused = match.getPlayer(player);
```
Pablete1234: What will happen if the player is in another match? i don't think the Match should be a parameter for this command, you should make use of the MatchManager to get the MatchPlayer

Electroid: I think the report should be scoped to the Match. The issue here is you need to check that `accused != null`

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -0,0 +1,70 @@
...
+    PlayerReportEvent event = new PlayerReportEvent(commandSender, accused, reason);
+    match.callEvent(event);
+
+    if (event.isCancelled()) {
+      if (event.getCancelMessage() != null) {
+        commandSender.sendMessage(event.getCancelMessage());
```
Electroid: `matchPlayer.sendMessage`

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -0,0 +1,70 @@
...
+      }
+      return;
+    }
+
+    commandSender.sendMessage(
```
Electroid: `matchPlayer.sendMessage`

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
Pugzy: > What will happen if the player is in another match?

We don't have the ability to run multiple matches at the moment and I don't think every new change needs to unnecessary pander to this idea. 

Regardless you shouldn't be reporting players from alternative matches because as you said yourself they should act independently. The audiences that it broadcast the notification message to could be server-wide as moderators may wish to switch match to investigate.

> you need to check that `accused != null`

When would the accused not be a player? The command framework seems pretty hell-bent on making sure the correct arguments are passed to the function.

image

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
Pugzy: The use of `CommandSender` allows for console reports. I can remove this feature if someone lets me know how to prevent Console from running the command as requested previously in a discussion with Molka.

PR #181, opened 2020-01-02
Add a setting to toggle displaying the picker

applenick:
**Added a setting to toggle the picker visibility.** 

Personally I believe players should have the choice to toggle the picker GUI off. With a large majority of players not having the ability to choose their team, I think many will find this useful. However I did make it so the “displaying” option is on by default.

~~On a side note, I wanted to make the setting values easy to understand (“hide” and “display”). Though if there are any better recommendations I am open to suggestions.~~

Edit:
We’ve updated “display” -> “show”, as it’s more user friendly. 
(I’m not going to update the screenshot, sorry 😆)

Edit 2 (1/3/2020): Please note screenshot is now outdated. See below comments to view updated setting values.

Example of setting values:
Screen Shot 2020-01-02 at 3 13 45 PM

Example of when picker is hidden:
Screen Recording 2020-01-02 at 12 48 PM

### DIFF
--- a/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
@@ -19,8 +19,9 @@ public enum SettingValue {
   CHAT_ADMIN("chat", "admin"), // Send to all server operators
   DEATH_OWN("death", "own"), // Only send death messages involving self
   DEATH_ALL("death", "all"), // Send all death messages, highlight your own
-  PICKER_DISPLAY("picker", "show"), // Display the picker GUI when clicked
-  PICKER_HIDE("picker", "hide"); // Hide the picker GUI, will auto join a team when clicked
+  PICKER_AUTO("picker", "auto"), // Display after cycle, or with permissions.
+  PICKER_ON("picker", "on"), // Display the picker GUI always
+  PICKER_OFF("picker", "off"); // Never display the picker GUI
 
   private final String key;
   private final String name;
--- a/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
@@ -18,7 +18,9 @@ public enum SettingValue {
   CHAT_GLOBAL("chat", "global"), // Send to all players in the same match
   CHAT_ADMIN("chat", "admin"), // Send to all server operators
   DEATH_OWN("death", "own"), // Only send death messages involving self
-  DEATH_ALL("death", "all"); // Send all death messages, highlight your own
+  DEATH_ALL("death", "all"), // Send all death messages, highlight your own
+  PICKER_DISPLAY("picker", "display"), // Display the picker GUI when clicked
+  PICKER_HIDE("picker", "hide"); // Hide the picker GUI, will auto join a team when clicked
 
   private final String key;
   private final String name;
--- a/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
@@ -19,7 +19,7 @@ public enum SettingValue {
   CHAT_ADMIN("chat", "admin"), // Send to all server operators
   DEATH_OWN("death", "own"), // Only send death messages involving self
   DEATH_ALL("death", "all"), // Send all death messages, highlight your own
-  PICKER_DISPLAY("picker", "display"), // Display the picker GUI when clicked
+  PICKER_DISPLAY("picker", "show"), // Display the picker GUI when clicked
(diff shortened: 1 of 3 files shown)


src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -18,7 +18,9 @@
   CHAT_GLOBAL("chat", "global"), // Send to all players in the same match
   CHAT_ADMIN("chat", "admin"), // Send to all server operators
   DEATH_OWN("death", "own"), // Only send death messages involving self
-  DEATH_ALL("death", "all"); // Send all death messages, highlight your own
+  DEATH_ALL("death", "all"), // Send all death messages, highlight your own
+  PICKER_DISPLAY("picker", "display"), // Display the picker GUI when clicked
```
TheMolkaPL: `show` maybe?

applenick: Thanks for the suggestion. That’s more user friendly 😄

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -18,7 +18,9 @@
   CHAT_GLOBAL("chat", "global"), // Send to all players in the same match
   CHAT_ADMIN("chat", "admin"), // Send to all server operators
   DEATH_OWN("death", "own"), // Only send death messages involving self
-  DEATH_ALL("death", "all"); // Send all death messages, highlight your own
+  DEATH_ALL("death", "all"), // Send all death messages, highlight your own
+  PICKER_DISPLAY("picker", "show"), // Display the picker GUI when clicked
+  PICKER_HIDE("picker", "hide"); // Hide the picker GUI, will auto join a team when clicked
```
Electroid: Why not `on` and `off`?

Brottweiler:
What if a user with `pgm.premium` don't want the picker to show up on every cycle, but still want to use the leather cap to choose team rather than having to do `/join [team]` since the leather cap autojoins you?

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
applenick: To be honest, `on` and `off` did not immediately come to mind. 🤦‍♂ 

I’ll make the adjustment, thanks 👍

applenick:
@Brottweiler 
This setting makes it so the Picker is on by default, the player will have to toggle the setting in order to hide the picker. Perhaps it won’t be confusing for premium players, as they would have to know how to use the setting in order to achieve that scenario. However I understand the problem you stated.

A possible solution that I could implement if you think it would help:
If a player has `pgm.premium` and picker is `off`, then clicking on the leather cap will send a chat message saying `Would you like to join [Red] or [Blue]?`
The message of course would list as many teams as the map is capable of holding with their respective color/formatting. Then clicking the team name would basically be the same as running /join [team name]

Let me know if that solution is ideal for what you had in mind, and if there are any suggestions for the message formatting.

Pablete1234:
What happens after match cycle? Is the GUI automatically opened for you to pick? I feel like it's 2 different settings if it should show the gui after cycle or if it should hide it completely. 

Maybe it makes more sense as: off, auto, on. Auto being the default.
Off : never show the gui, click joins you without picking, use /join [team] to select
Auto: Show after cycle, if clicking show only if you have the pgm.premium perm
On: Always show both when cycling and when clicking the hat yourself

src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
```diff
@@ -90,8 +90,16 @@ public PickerMatchModule(Match match) {
     this.isBlitz = match.hasMatchModule(BlitzMatchModule.class);
   }
 
-  protected boolean settingEnabled(MatchPlayer player) {
-    return player.getSettings().getValue(SettingKey.PICKER) == SettingValue.PICKER_DISPLAY;
+  protected boolean settingEnabled(MatchPlayer player, boolean isJoinButton) {
```
TheMolkaPL: maybe `playerTriggered`?

applenick:
So here’s a basic overview of the last commit:
- Updated `hide` and `display` to `on` and `off`
- Added a `auto` option, which is now the default value

Now the values for the Picker are `auto`, `on` and `off`

I believe the `auto` option is a new great default value. As @Pablete1234 stated, when the match cycles and `auto` is selected it will display the picker. From my testing everything works as intended. However instead of just checking for the `pgm.premium` node, I made it `pgm.join.choose`.

src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
applenick: Great suggestion thanks! 👍

PR #182, opened 2020-01-03
Fix incomplete symbol used on simple goals

Pugzy:
Reverts a change made by Pablete in 6aee67bffac1b54a30f14f6a795b6901d19143f2 where the symbol was changed to one that is not supported in-game.


Pablete1234:
I want to point out, it is supported in-game and works just fine in the 1.15.1 vanilla client. Some resource packs or older versions may not support it though.

Issue #184, opened 2020-01-03
Rage maps show as Blitz in the scoreboard


Xerocoles:
Here's an example on BlockRAGE -- it's assuming the rage maps are blitz I guess, but most aren't.
<img width="165" alt="Screen Shot 2020-01-03 at 2 58 20 PM" src="https://user-images.githubusercontent.com/54342797/71745889-83f35f80-2e39-11ea-971e-58519a9e0c35.png">

TheMolkaPL:
Fixed fe7008c0df8645f4ac2c4a1b414a658eb39fcf98 👌

PR #185, opened 2020-01-03
Fix rage game component

TheMolkaPL:
That's because `BlitzModule` is always parsed and therefore always exists.

https://github.com/TheMolkaPL/PGM-1/blob/0989c7431b5159fedd2b68ba3327ec3b5a71ea99/src/main/java/tc/oc/pgm/blitz/BlitzModule.java#L55-L67

Related: https://github.com/Electroid/PGM/issues/184

### DIFF
--- a/src/main/java/tc/oc/pgm/rage/RageModule.java
+++ b/src/main/java/tc/oc/pgm/rage/RageModule.java
@@ -41,7 +41,7 @@ public class RageModule extends MapModule {
   public static RageModule parse(MapModuleContext context, Logger logger, Document doc) {
 
     if (doc.getRootElement().getChild("rage") != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/tc/oc/pgm/rage/RageModule.java
```diff
@@ -41,7 +41,8 @@ public MatchModule createMatchModule(Match match) {
   public static RageModule parse(MapModuleContext context, Logger logger, Document doc) {
 
     if (doc.getRootElement().getChild("rage") != null) {
-      return new RageModule(context.hasModule(BlitzModule.class));
+      BlitzModule blitzModule = context.getModule(BlitzModule.class);
```
Electroid: Is there a `needModule`? If so, use that.

PR #186, opened 2020-01-04
Cancel damage for dead players, resolves Electroid#179

Meeples10:

### DIFF
--- a/src/main/java/tc/oc/pgm/damage/DamageMatchModule.java
+++ b/src/main/java/tc/oc/pgm/damage/DamageMatchModule.java
@@ -159,6 +159,11 @@ public class DamageMatchModule extends MatchModule implements Listener {
     MatchPlayer victim = getVictim(event.getEntity());
     if (victim == null) return;
 
+    if (victim.isDead()) {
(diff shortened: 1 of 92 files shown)


Meeples10:
Is there a better place to cancel the event?

TheMolkaPL:
This completly wrong place to do this. Trackers are meant to keep track of death reasons. Before you do the coding please try reproduce this bug before refacators. I belive one of them completly broke respawn system together with this and flying aroung bug.

TheMolkaPL:
@Meeples10 can you check your Discord, please?

src/main/java/tc/oc/pgm/damage/DamageMatchModule.java
```diff
@@ -159,6 +159,11 @@ public void onDamage(EntityDamageEvent event) {
     MatchPlayer victim = getVictim(event.getEntity());
     if (victim == null) return;
 
+    if (victim.isDead()) {
+      event.setCancelled(true);
+      return;
+    }
+
```
Pablete1234: I think this doesn't belong here, right?

PR #187, opened 2020-01-04
Fix dead players' names not turning gray

Meeples10:


src/main/java/tc/oc/world/DeathOverride.java
```diff
@@ -35,7 +36,12 @@ static boolean isDead(Entity player) {
     if (value != null) {
       return value.asBoolean();
     } else {
-      return player.isDead();
+      MatchPlayer matchPlayer = PGM.get().getMatchManager().getPlayer(player);
+      if (matchPlayer != null) {
+        return matchPlayer.isDead();
+      } else {
+        return player.isDead();
+      }
```
Pablete1234: The correct way of doing this, is setting the metadata key by calling setDead from MatchPlayer when setDead is called there, instead of here, checking for the match player. Notice the setDead event on this class is unused

PR #188, opened 2020-01-04
Better join messages

TheMolkaPL:
They are now toggleable as requested by @Brottweiler.
I've added an additional `auto` option which disables join messages when player count in a match exceeds 30.

### DIFF
--- a/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -19,10 +19,7 @@ public enum SettingKey {
       Arrays.asList("death", "dms"), DEATH_ALL, DEATH_OWN), // Changes which death messages are seen
   PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF), // Changes when the picker is displayed
   JOIN_MESSAGES(
-      Arrays.asList("join-messages", "jms"),
-      JOIN_MESSAGES_AUTO,
-      JOIN_MESSAGES_ALL,
-      JOIN_MESSAGES_NONE), // Changes which join messages are seen
+      Arrays.asList("join", "jms"), JOIN_ON, JOIN_OFF), // Changes which join messages are seen
   ;
 
   private final List<String> aliases;
--- a/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -17,7 +17,13 @@ public enum SettingKey {
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
   DEATH(
       Arrays.asList("death", "dms"), DEATH_ALL, DEATH_OWN), // Changes which death messages are seen
-  PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF); // Changes when the picker is displayed
+  PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF), // Changes when the picker is displayed
+  JOIN_MESSAGES(
+      Arrays.asList("join-messages", "jms"),
+      JOIN_MESSAGES_AUTO,
+      JOIN_MESSAGES_ALL,
+      JOIN_MESSAGES_NONE), // Changes which join messages are seen
+  ;
 
   private final List<String> aliases;
   private final SettingValue[] values;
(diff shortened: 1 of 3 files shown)


Pugzy:
What is the purpose of the auto option? It doesn't seem very intuitive.

TheMolkaPL:
Auto doesn't show join messages when they are shown too often.

Pablete1234:
Doesn't seem to be if it's "too often", but rather, if there are over 30 players on the match. This doesn't seem like a good thing.
What would be best to fix join-spam, is to have a guava expiring cache (maybe 5-10 seconds) with joins, and another one with leaves. If the UUID is already in the cache ignore the message, otherwise show it & add it to the cache. This will make it so at most you'll see one join and one leave message per player. 

The issue that can occur with that, is that if someone joins, leaves, and joins, you will only see in chat that they joined and left, but they will actually be online wich can be confusing.

Pugzy:
> Auto doesn't show join messages when they are shown too often.

I don't think this is necessary and may cause confusion. I'd either want to see all the join messages or none, if the same setting 'feature' was applied on death messages it would would be confusing as to why nobody was getting kills any more.

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -17,7 +17,13 @@
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
   DEATH(
       Arrays.asList("death", "dms"), DEATH_ALL, DEATH_OWN), // Changes which death messages are seen
-  PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF); // Changes when the picker is displayed
+  PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF), // Changes when the picker is displayed
+  JOIN_MESSAGES(
+      Arrays.asList("join-messages", "jms"),
```
Electroid: Similar with `death`, how about `join` and `jms`

TheMolkaPL: I took `jms` from `OCN/ProjectAres` repo, in case people were used to it.

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -17,7 +17,9 @@
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
   DEATH(
       Arrays.asList("death", "dms"), DEATH_ALL, DEATH_OWN), // Changes which death messages are seen
-  PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF); // Changes when the picker is displayed
+  PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF), // Changes when the picker is displayed
+  JOIN(Arrays.asList("join", "jms"), JOIN_ON, JOIN_OFF), // Changes which join messages are seen
```
Pablete1234:
```suggestion
  JOIN(Arrays.asList("join", "jms"), JOIN_ON, JOIN_OFF), // Changes if join messages are seen
```

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -17,11 +17,17 @@
...
   DEATH_OWN("death", "own"), // Only send death messages involving self
   DEATH_ALL("death", "all"), // Send all death messages, highlight your own
+
   PICKER_AUTO("picker", "auto"), // Display after cycle, or with permissions.
   PICKER_ON("picker", "on"), // Display the picker GUI always
-  PICKER_OFF("picker", "off"); // Never display the picker GUI
+  PICKER_OFF("picker", "off"), // Never display the picker GUI
+
+  JOIN_ON("join", "all"), // Send all join messages
+  JOIN_OFF("join", "none"), // Never send join messages
```
Pablete1234:
```suggestion
  JOIN_OFF("join", "none") // Never send join messages
```

src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -33,13 +38,19 @@
...
 import tc.oc.pgm.events.PlayerPartyChangeEvent;
 import tc.oc.pgm.gamerules.GameRule;
 import tc.oc.pgm.gamerules.GameRulesModule;
 import tc.oc.pgm.modules.TimeLockModule;
 
 public class PGMListener implements Listener {
+
+  /** If player count exceeds this number, viewers with "AUTO" setting won't receive any message. */
+  private static final int JOIN_MESSAGE_ONLINE_THRESHOLD = 30;
+
```
Pablete1234: Unused

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
TheMolkaPL: Why this change?

Pablete1234: Doesn't make sense to leave a trailing comma at the end

TheMolkaPL: If somebody adds more settings later, less lines are going to be affected. Makes resolving conflicts easier.

PR #191, opened 2020-01-04
Allow to toggle observers

TheMolkaPL:


src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -21,7 +21,9 @@
   DEATH_ALL("death", "all"), // Send all death messages, highlight your own
   PICKER_AUTO("picker", "auto"), // Display after cycle, or with permissions.
   PICKER_ON("picker", "on"), // Display the picker GUI always
-  PICKER_OFF("picker", "off"); // Never display the picker GUI
+  PICKER_OFF("picker", "off"), // Never display the picker GUI
+  OBSERVERS_SHOWN("observers", "shown"), // Show observers
+  OBSERVERS_HIDDEN("observers", "hidden"); // Hide observers
```
Pablete1234: Why not just on/off? I think it's better to be somewhat consistent with other settings if players want to be able to find/explore them easily.

src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -172,7 +174,13 @@ public boolean canInteract() {
...
   @Override
   public boolean canSee(MatchPlayer other) {
-    return (isObserving() || other.isParticipating()) && other.isVisible();
+    if (!other.isVisible()) return false;
+    if (other.isParticipating()) return true;
+    if (isObserving()) {
+      SettingValue observers = getSettings().getValue(SettingKey.OBSERVERS);
+      return observers.equals(SettingValue.OBSERVERS_SHOWN);
+    }
+    return false;
```
Pablete1234: This 4th branch should never occur, or am i missing something? If that's the case, maybe leave a comment saying that shouldn't happen.

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
TheMolkaPL: It's currently more understandable what it means.

src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
TheMolkaPL: Maybe throw `IllegalStateException`?

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
Pablete1234: Are you sure? I think obs on or obs off is just as easy to understeand as obs shown and obs hidden. It's just a matter of consistency with other settings tho, i am not too concerned about that as long as it's easy to tab-complete tho

Electroid: Let's stick to on/off. Eventually we'll have a Gui with an explanation of what the setting does

src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
Pablete1234: Maybe add a logger, but don't throw an exception, it's better not to crash PGM because of something that can easily be recoverable by just returning an arbitrary false

Electroid: Change to:
```java
return getSettings().getValue(SettingKey.OBSERVERS) == SettingValue.OBSERVERS_SHOWN;
```

Electroid: `canX` methods should never return an illegal state, see my comment below.

Pablete1234: Actually, if you make it that, players inside the match will be able to see obs.

src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -172,7 +174,9 @@ public boolean canInteract() {
 
   @Override
   public boolean canSee(MatchPlayer other) {
-    return (isObserving() || other.isParticipating()) && other.isVisible();
+    if (!other.isVisible()) return false;
+    if (other.isParticipating()) return true;
+    return getSettings().getValue(SettingKey.OBSERVERS) == SettingValue.OBSERVERS_ON;
```
Pablete1234:
```suggestion
    return isObserving() && getSettings().getValue(SettingKey.OBSERVERS) == SettingValue.OBSERVERS_ON;
```

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -21,7 +21,9 @@
   DEATH_ALL("death", "all"), // Send all death messages, highlight your own
   PICKER_AUTO("picker", "auto"), // Display after cycle, or with permissions.
   PICKER_ON("picker", "on"), // Display the picker GUI always
-  PICKER_OFF("picker", "off"); // Never display the picker GUI
+  PICKER_OFF("picker", "off"), // Never display the picker GUI
+  OBSERVERS_ON("observers", "shown"), // Show observers
+  OBSERVERS_OFF("observers", "hidden"); // Hide observers
```
Pablete1234: make the string constants on/off?

TheMolkaPL:
Player visibility is now updated after the setting has changed.

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -18,10 +19,12 @@
...
-  OBSERVERS(
-      Arrays.asList("observers", "obs"),
-      OBSERVERS_ON,
-      OBSERVERS_OFF); // Changes if observers are visible
+  OBSERVERS(Arrays.asList("observers", "obs"), OBSERVERS_ON, OBSERVERS_OFF) {
+    @Override
+    public void update(MatchPlayer player) {
+      player.resetVisibility();
+    }
+  }; // Changes if observers are visible
```
Pablete1234: A way to heavily simplify this, is to have a constructor overload:
`SettingKey(List<String> aliases, Consumer<MatchPlayer> update, SettingValue... values)`

Then this just becomes:
`OBSERVERS(Arrays.asList("observers", "obs"), MatchPlayer::resetVisibility, OBSERVERS_ON, OBSERVERS_OFF)`

As obvious, you'll need to keep the consumer on a variable, and on the update method call it if it's not null

TheMolkaPL: The current version is imo simpler, and I don't think there are going to be many settings with custom update logic.

Pablete1234: For now, either way works

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -78,4 +85,11 @@ public SettingValue getDefaultValue() {
   public String toString() {
     return getName();
   }
+
+  /**
+   * Called whether setting has changed and is ready to be updated internally.
+   *
+   * @param player owner of the setting
+   */
+  public void update(MatchPlayer player) {}
```
Electroid: For future reference, I think this will be a stop gap solution. In making `Datastore` async, I think I'll add support for a `Listener` to be registered.

TheMolkaPL:
Fixed conflicts 👌

PR #192, opened 2020-01-04
Don't suggest setting key aliases

TheMolkaPL:
Currently, after typing `/toggle ` it suggests all available aliases, which is confusing, as it looks like there are more settings.

### DIFF
--- a/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -42,9 +42,9 @@ public enum SettingKey {
   }
 
   /**
-   * Get all aliases of this {@link SettingKey}.
+   * Get all aliases of this {@link SettingKey}. First index is always equal to {@code #getName}.
    *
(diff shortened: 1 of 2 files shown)


Pablete1234:
Instead of suggesting only the main name, consider suggesting the first available alias for each setting

TheMolkaPL:
First alias is always the main name.

https://github.com/Electroid/PGM/blob/1a604f129d510489aa9ec0a398a31b3c787c139d/src/main/java/tc/oc/pgm/api/setting/SettingKey.java#L35-L42

Pablete1234:
@TheMolkaPL what i mean is the first alias that matches your search.

Example: "death" or "dms":
`/toggle d` it should autocomplete "death", and no more aliases of that setting
`/toggle dm` it should autocomplete "dms", and no more aliases of that setting

Electroid:
@TheMolkaPL Is something wrong with this PR? You only seem to have changed the java doc?

Pablete1234:
@Electroid the java-doc and the added `break;`, wich is all needed for this fix

Issue #169, opened 2019-12-31
Map voting enhancements


Pablete1234:
- [x] Map voting should make use of a **Book GUI**, because people use chat and your voting messages move arround, can get annoying. 
- [x] **Bossbar** & cycle messages give away what the **highest** voted **map** is. May want to make it so it no longer does, so people don't vote based on that.
- [x] Try to make it more **obvious** that you can **vote for multiple maps**, several users assumed it was single vote even if the messages are layed out as a list with a corss or check each.
 - [x] Add a **flag** to **/pool** that shows **map scores**, to see how they evolve
 - [x] Consider tweaking/**adjusting map scoring** if it's not being too representative of the playerbase
 - [x] Consider more **mechanisms to pick the voting maps**, like gamemode rotation based approach. Decreasing map chance based on other selected maps of the same type in the same vote adds variety.
 - [x] Playercount-based picking. One large pool is created and maps selected for voting depend on current playercount.

Xerocoles:
At OCC it might not matter as much for gamemode variation, although they can speak for themselves, but at Stratus it's super important that we get different pools of maps for each gamemode, and so we can choose the gamemode flow ordering.  Otherwise it's gonna be really unbalanced and conquest will hardly _ever_ play, just as an example.  So yeah, although we aren't sure if we want votations yet, gamemode pools is a must for us, and I'm guessing it would still benefit OCC.

rafibaum:
@Pablete1234 I think this should be closed unless can flesh out an actual concept for an extension to map voting.

Pablete1234:
@rafibaum There is one, just kind of abandoned.

The idea behind it is that the weight of maps is able to be influenced by current playerbase and by other selected maps in the same vote. That means the more tags of a certain type you share with the already picked maps, the lower weight the map will have. It will make it so less maps of the same type will be on the vote.

rafibaum:
You can either update the issue title/description to reflect that concept then or just close this and make a new issue.

Pablete1234:
Finished in #1010

PR #193, opened 2020-01-05
Add several enhancments to map voting, including book gui

Pablete1234:
From the things in #169 :
 - Added book GUI (opens 5s after match end)
 - The next map is no longer shown before the actual cycle
 - The book gui informs you about multi-map voting, it also keeps re-opening the book as you vote until you close.
 - You can see the map scores with -s (score) on /pool
 - Map scoring is now linear instead of exponential towards default score. This means map scores will be generally slower to get more/less significant over time, and way less likely to re-appear on a vote after they've played
 - Book GUI shows map tags on hover, should be helpful if you don't know the map.

### DIFF
--- a/src/main/i18n/templates/strings.properties
+++ b/src/main/i18n/templates/strings.properties
@@ -548,7 +548,6 @@ command.pools.skip.noNegative = You may not skip negative positions!
 
 command.pool.vote.book.title = Map Vote
 command.pool.vote.book.header = Click to select all maps you'd like to play:
-command.pool.vote.hover = Click to toggle vote
 command.pool.vote.noVote = There is no poll to vote in currently.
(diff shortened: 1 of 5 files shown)


src/main/i18n/templates/strings.properties
```diff
@@ -546,7 +546,8 @@ command.pools.mapPoolList.title = Loaded Map Pools
 command.pools.skip.message = Skipped a total of {0} positions.
 command.pools.skip.noNegative = You may not skip negative positions!
 
-command.pool.vote.hover = Click to toggle vote
+command.pool.vote.book.title = Map Vote
+command.pool.vote.book.header = Click to select all maps you'd like to play:
```
Electroid: For translations avoid contractions: `you would like`

PR #197, opened 2020-01-05
Modernize max players in MapPersistentContext

TheMolkaPL:
image

Finally resolves Electroid/PGM#154

### DIFF
--- a/src/main/i18n/templates/strings.properties
+++ b/src/main/i18n/templates/strings.properties
@@ -529,6 +529,7 @@ command.map.mapInfo.authorPlural = Authors
 command.map.mapInfo.contributors = Contributors
 command.map.mapInfo.rules = Rules
 command.map.mapInfo.playerLimit = Max players
+command.map.mapInfo.playerLimit.vs = vs
 command.map.mapInfo.xml = XML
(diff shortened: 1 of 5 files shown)


Pablete1234:
Well this will already have a 20 vs 20 vs 20 as the format for 3+team maps. Using a 3-team map was just an example. The "normal" behaviour it will see is usually going to be `50 (25 vs 25)`, but i think it makes sense to keep both the sum and the split amounts.

Electroid:
On second thought.. what if we did something like this?

```
// When all team sizes are the same
2 teams of 50 players
3 teams of 20 players
// When there are different size teams
1 team of 10 players, 2 teams of 20 players
```

Pablete1234:
I think such format just gets needlessly verbose, probably best to keep it somewhat short

Electroid:
When the team sizes are even, it's not verbose?

Pablete1234:
`2 teams of 20 players` is still longer than `40 (20 vs 20)`, without including the total

Only for maps where there's many teams (probably 4+) it may take up less space

TheMolkaPL:
image

PR #203, opened 2020-01-05
Add option for gaps between sidebar messages and objectives

Meeples10:

As requested by Stu.

### DIFF
--- a/src/main/java/tc/oc/pgm/Config.java
+++ b/src/main/java/tc/oc/pgm/Config.java
@@ -404,6 +404,14 @@ public class Config {
           '&', getConfiguration().getString("sidebar.top"));
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/tc/oc/pgm/Config.java
```diff
@@ -404,6 +404,14 @@ public static String formatTop() {
           '&', getConfiguration().getString("sidebar.top"));
     }
 
+    public static boolean topGap() {
+      return getConfiguration().getBoolean("sidebar.gap-top");
+    }
+
+    public static boolean bottomGap() {
+      return getConfiguration().getBoolean("sidebar.gap-bottom");
+    }
```
TheMolkaPL: Can you pass default `false` values for consistency?

PR #209, opened 2020-01-08
Add an auto-respawn setting, resolves Electroid#208

Meeples10:

### DIFF
--- a/src/main/java/tc/oc/pgm/spawns/states/Dead.java
+++ b/src/main/java/tc/oc/pgm/spawns/states/Dead.java
@@ -11,6 +11,8 @@ import tc.oc.component.types.PersonalizedText;
 import tc.oc.component.types.PersonalizedTranslatable;
 import tc.oc.pgm.api.party.Competitor;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.setting.SettingKey;
+import tc.oc.pgm.api.setting.SettingValue;
 import tc.oc.pgm.events.PlayerJoinPartyEvent;
 import tc.oc.pgm.spawns.Spawn;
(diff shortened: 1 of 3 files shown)


src/main/java/tc/oc/pgm/spawns/states/Dead.java
```diff
@@ -79,6 +81,11 @@ public void tick() {
       player.resetGamemode();
     }
 
+    if (player.getSettings().getValue(SettingKey.RESPAWN) == SettingValue.RESPAWN_AUTO
+        && ticksUntilRespawn() <= 0) {
+      super.requestSpawn();
+    }
+
```
Pablete1234: I don't think there's much of a benefit to calculating this each tick. Just set `spawnRequested` to true if this setting matches on the constructor of `Dead`

src/main/java/tc/oc/pgm/spawns/states/Dead.java
```diff
@@ -122,7 +129,9 @@ protected Component getSubtitle() {
     long ticks = ticksUntilRespawn();
     if (ticks > 0) {
       return new PersonalizedTranslatable(
-          spawnRequested ? "death.respawn.confirmed.time" : "death.respawn.unconfirmed.time",
+          player.getSettings().getValue(SettingKey.RESPAWN) == SettingValue.RESPAWN_AUTO
+              ? "death.respawn.confirmed.time"
+              : spawnRequested ? "death.respawn.confirmed.time" : "death.respawn.unconfirmed.time",
```
Pablete1234: This would be unnecessary too

Electroid:
Going to pass on this for now, although it seems nice it may introduce some other issues (like kill farming people with auto respawn, or bypassing map intentions).

PR #216, opened 2020-01-10
Add the ability to forfeit a match

applenick:
# Forfeit Command

## Background
The other day I witnessed a scenario in which having a /forfeit command could have been useful. The scenario in question involved players unable to finish a match due to an element being broken. Due to this unintentional error, players were confused and had to wait for a staff member to come online and end the match. Had a /forfeit command been available, they would have easily been able to solve this problem without intervention. 

## Overview
Forfeit as implemented  will allow for users to vote to end a match without completing the main objective. Also included are multiple config options to customize the functionality of a forfeit, such as whether the votes are broadcasted. Also restrictions are in place to prevent the command from being used in situations where it wouldn’t make sense, such as a match with no teams or a time limit. I did perform extensive testing and found no issues regarding the command. 

### Command
The command syntax is simply `/forfeit` 

Staff (players with `pgm.staff`), can use the `/forfeit -t` to toggle whether forfeit is enabled/disabled for the current match.

In order for the forfeit vote to go through, all activate participants must use the `/forfeit` command. This decision was made to make sure everyone is in agreement when choosing to end a match early. I would like to see a future update that includes options for a single team to forfeit, however this commit does not include that. 

### Config Options
As mentioned above, added to the config are a few options to customize the forfeit functionality. Most are self explanatory, however I would like to give some insight into the `max-players` option.

On more popular PGM servers, it may hinder rather than help the experience when a certain amount of players are online. Rather than just disabling the whole system, I’ve included the `max-players` option in order to limit using the command when the participant count is larger. Of course it can also be disabled as described in the config. 

In its current form, I feel `/forfeit` will benefit most during smaller matches of players. However, it may also come in handy on larger matches that come to a stalemate. 

### Screenshot
Below you can see what the feedback is when using the /forfeit command two times. Performing the command a second time will withdraw your vote. 
Screen Shot 2020-01-10 at 1 05 31 AM

### Name Note
Personally I had a difficult time finding the proper name for this feature. I felt that in the end forfeit fit most naturally with PGM’s game modes. However some alternatives that came into my mind earlier include surrender and abandon. If these or similar names are wanted as aliases just let me know. 

### DIFF
--- a/src/main/i18n/templates/strings.properties
+++ b/src/main/i18n/templates/strings.properties
@@ -850,3 +850,25 @@ misc.seconds.pluralCompound = {0} seconds
 # {1} = the total pages
 pageHeader = Page {0} of {1}
 currentPage = Page {0}
+
+command.forfeit.notEnabled = Forfeit is not enabled at this time.
+command.forfeit.notPossible = It is not possible to forfeit this match.
+command.forfeit.observers = Only active participants can use /forfeit
+
+command.forfeit.disable = You have disabled /forfeit for this match.
+command.forfeit.enable = You have enabled /forfeit for this match.
+
+# {0} = forfeit status
+command.forfeit.voted = You have voted to end the match. ({0})
+command.forfeit.withdraw = You have withdrawn your forfeit vote.
+
+# {0} = player name
+# {1} = forfeit status
+forfeit.broadcast.vote = {0} has voted to end the match. ({1})
+forfeit.broadcast.withdraw = {0} has withdrawn their forfeit vote.
+
+# {0} = Total Forfeits
+# {1} = Total players
+forfeit.broadcast.stats = {0} / {1} votes
+
+forfeit.broadcast.hover = Click to vote in favor of ending the match.
--- a/src/main/java/tc/oc/pgm/commands/ForfeitCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/ForfeitCommands.java
@@ -1,0 +1,46 @@
+package tc.oc.pgm.commands;
+
+import app.ashcon.intake.Command;
+import app.ashcon.intake.parametric.annotation.Switch;
+import java.util.Optional;
+import net.md_5.bungee.api.ChatColor;
+import org.bukkit.command.CommandSender;
+import tc.oc.component.types.PersonalizedTranslatable;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.modules.ForfeitMatchModule;
+
+public class ForfeitCommands {
+
+  @Command(
+      aliases = {"forfeit"},
+      desc = "Vote to forfeit the match")
+  public static void forfeitMatch(
+      CommandSender sender, Match match, MatchPlayer player, @Switch('t') boolean toggle) {
+    Optional<ForfeitMatchModule> forfeit = match.getModule(ForfeitMatchModule.class);
+    if (forfeit.isPresent()) {
+      ForfeitMatchModule fmm = forfeit.get();
+
+      if (sender.hasPermission(Permissions.STAFF) && toggle) {
+        boolean newStatus = fmm.setEnabled(!fmm.isEnabled());
+
+        sender.sendMessage(
+            new PersonalizedTranslatable(
+                    newStatus ? "command.forfeit.enable" : "command.forfeit.disable")
+                .getPersonalizedText()
+                .color(ChatColor.GRAY));
+      } else {
+        if (fmm.isEnabled()) {
+          if (player.isObserving()) {
+            player.sendWarning(new PersonalizedTranslatable("command.forfeit.observers"));
+          } else {
+            fmm.forfeit(player, sender);
+          }
+        } else {
+          player.sendWarning(new PersonalizedTranslatable("command.forfeit.notEnabled"));
+        }
+      }
+    }
+  }
+}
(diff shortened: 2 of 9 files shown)


Pablete1234:
I had already thought of something similar, but always thought of it as `skip`, instead of forfeit, surrender or abanadon, and it being voted by everyone on the server, rather than just one team.

The biggest concern on adding such a command, is that it can be dangerous to have. If the minimum votes required isn't high enough, groups of people could end matches they don't like even if it messes with alot of people who did want to keep playing. If the votes required are too many, you may have a few people who still vote no on broken matches causing them to never end.

Regarding config options, i don't think there should be any for this. Decide on good defaults and ignore configuration.

src/main/i18n/templates/strings.properties
```diff
@@ -850,3 +850,25 @@ misc.seconds.pluralCompound = {0} seconds
...
+# {0} = player name
+# {1} = forfeit status
+forfeit.broadcast.vote = {0} has voted to end the match. ({1})
+forfeit.broadcast.withdraw = {0} has withdrawn their forfeit vote.
+
+# {0} = Total Forfeits
+# {1} = Total players
+forfeit.broadcast.stats = {0} / {1} votes
+
+forfeit.broadcast.hover = Click to vote in favor of ending the match.
```
TheMolkaPL: Missing new line

src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -62,6 +62,7 @@
   private final AtomicBoolean frozen;
   private final AtomicBoolean dead;
   private final AtomicBoolean visible;
+  private final AtomicBoolean forfeit;
```
TheMolkaPL: `true` value should expire after some time imo.

src/main/java/tc/oc/pgm/commands/ForfeitCommands.java
```diff
@@ -0,0 +1,46 @@
...
+import tc.oc.component.types.PersonalizedTranslatable;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.modules.ForfeitMatchModule;
+
+public class ForfeitCommands {
+
+  @Command(
+      aliases = {"forfeit"},
```
TheMolkaPL: Can we add additional `ff` and `surrender` aliases here?

src/main/java/tc/oc/pgm/modules/ForfeitMatchModule.java
```diff
@@ -0,0 +1,148 @@
...
+   * 2. If the match does not contain teams.
+   * 3. If player count is more than the max-player config.
+   */
+  private boolean isForfietPossible() {
+    Optional<TeamMatchModule> tmm = match.getModule(TeamMatchModule.class);
+    Optional<TimeLimitMatchModule> tlmm = match.getModule(TimeLimitMatchModule.class);
+
+    boolean hasTeams = tmm.isPresent();
+    boolean hasTimeLimit = tlmm.isPresent() && tlmm.get().hasTimeLimit();
+    boolean underPlayerLimit = maxPlayers != 0 ? getMatchPlayerCount() <= maxPlayers : true;
```
TheMolkaPL: Hard player count limit isn't the best way to go. Can we change `maxPlayers` to a percentage of players instead?

src/main/resources/config.yml
```diff
@@ -117,3 +117,20 @@ sidebar:
...
+  #
+  # Example: If max-players is at 25, matches which have 25 or less participants
+  # will be allowed to forfeit. Larger than 25 players and forfeit will disable.
+  #
+  # Note: set to 0 to disable this feature.
+  max-players: 10
+  
+  # When true, a broadcast will be sent to everyone after each /forfeit vote.
+  # If false feedback will only be sent to the player who used /forfeit.
+  broadcast-votes: true
```
TheMolkaPL: Missing new line

Electroid:
Thank you for your work on this, but I'm just not sure I want to bring this feature into PGM, config option or not. Either servers need to have more diligent moderation to prevent this or map makers needs to put more thought into better map design. This command just has the issue of creating unintended incentives that are hard to foresee.

TheMolkaPL:
> This command just has the issue of creating unintended incentives that are hard to foresee.

That's why I suggested a percentage of players (let's say 75% by default) who must vote to forfeit the match. Also, there should be a cooldown between match start and time to being able to vote.

applenick:
I understand where a `/forfeit` command would have unnecessary consequences when used on larger matches. Especially in situations such as players misusing the command for purposes other than intended. Which is why I designed it with the option to have a limit based on actual player count. As my intention was for this to be used when there are little to no players online. 

@Electroid If that does not fit the direction of PGM, I understand. My intention with submitting this pull request was to also receive feedback on whether I should continue. As this idea started simple, but soon there were multiple feature ideas I had thought of (single team forfeits, percentages, etc). Are there any changes you feel would make this acceptable for usage on PGM, such as what @TheMolkaPL presented? If not feel free to close this then. 

@TheMolkaPL & @Pablete1234 thank you both for the suggestions! `skip` is a really great alternative to forfeit. Also I originally thought of using percentages too, but I did identify the errors that could occur such as it becoming difficult for players to all vote `yes`. Which again, is why I felt having it for matches of say ~10 or less, would prove most useful.

Electroid:
Although it's possible this could be designed to mitigate those unintended consequences, sometimes simple is better.

PR #218, opened 2020-01-10
Map, MapModule, and MatchModule refactors

Electroid:
Major refactor to how `MapInfo`s and `Module`s are loaded.


src/main/java/tc/oc/pgm/api/map/MapInfo.java
```diff
@@ -0,0 +1,113 @@
...
+   * @return A collection of rules.
+   */
+  Collection<String> getRules();
+
+  /**
+   * Get the {@link org.bukkit.Difficulty#ordinal()} level for the map.
+   *
+   * @return The difficulty level.
+   */
+  int getDifficulty();
```
TheMolkaPL: Why is this an int instead of `Difficulty`?

src/main/java/tc/oc/pgm/api/map/MapInfo.java
```diff
@@ -0,0 +1,113 @@
...
+   * @return A name, alphanumeric with spaces are allowed.
+   */
+  String getName();
+
+  /**
+   * Get a short, human-readable description of the map's objective.
+   *
+   * @return A description.
+   */
+  String getDescription();
```
TheMolkaPL: What about a BaseComponent?

src/main/java/tc/oc/pgm/api/map/MapInfo.java
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Get the {@link org.bukkit.Difficulty#ordinal()} level for the map.
+   *
+   * @return The difficulty level.
+   */
+  int getDifficulty();
+
+  /**
+   * Get a collection of "hash tags" used to describe the map.
```
TheMolkaPL: The documentation should say if these are all maptags (with automatic maptags) or just XML defined.

src/main/java/tc/oc/pgm/api/map/MapInfo.java
```diff
@@ -0,0 +1,113 @@
...
+  MapInfo clone();
+
+  @Override
+  default int compareTo(MapInfo o) {
+    return getId().compareTo(o.getId());
+  }
+
+  static String normalizeName(@Nullable String idOrName) {
+    return idOrName == null
+        ? ""
```
TheMolkaPL: If it's null why should throw NPE to find bugs easier.

src/main/java/tc/oc/pgm/api/map/factory/MapFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import tc.oc.util.Version;
+
+/** A factory for creating {@link MapInfo}s and {@link MapContext}s. */
+public interface MapFactory extends ModuleContext<MapModule>, AutoCloseable {
+
+  RegionParser getRegions();
+
+  FilterParser getFilters();
+
+  KitParser getKits();
```
TheMolkaPL: We should deprecate these three parsers and explain why they shouldn't be here imho.

src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
```diff
@@ -52,9 +52,10 @@
   // die that do not fire an event e.g. the tick age limit, but this should be
   // rare and they will only leak until the end of the match.
   private final Set<FallingBlock> fallingBlocksThatWillNotLand = new HashSet<>();
+  private final Match match;
```
TheMolkaPL: This `match` field will now pop up in each `MatchModule` implementation. We should have `MatchModule.Impl` with this to keep it DRY.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -1,16 +1,15 @@
 package tc.oc.pgm.commands;
 
-import static com.google.common.base.Preconditions.*;
+import static com.google.common.base.Preconditions.checkNotNull;
```
TheMolkaPL: Keep that * import :P

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -27,43 +26,53 @@
...
+      desc = "Loads new maps and outputs any errors")
+  public static void loadNewMaps(MapLibrary library, @Switch('f') boolean force) {
+    library.loadNewMaps(force); // MapLibrary will handle sending output asynchronously
+  }
+
   @Command(
       aliases = {"maplist", "maps", "ml"},
       desc = "Shows the maps that are currently loaded",
-      usage = "[-a <author>] [-p <page>] [[!]#<maptag>...]",
+      usage = "[-a <author>] [-p <page>] [-# <tag1>,<tag2>]",
```
TheMolkaPL: The previous syntax was way easier imo. Just `/maps #many #tags` instead of commas, dashes and other difficult combinations you need to do to put multiple maptags.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -27,43 +26,53 @@
...
       help =
           "Shows all the maps that are currently loaded including ones that are not in the rotation.")
   public static void maplist(
       Audience audience,
       CommandSender sender,
       MapLibrary library,
-      MapTagsCondition mapTags,
+      @Fallback(Type.NULL) @Switch('t') String tags,
```
TheMolkaPL: This should fix the page number input I think? Also doing flags for maptags which always start with `#` isn't obvious to end-user.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -27,43 +26,53 @@
...
       @Fallback(Type.NULL) @Switch('p') Integer page)
       throws CommandException {
     if (page == null) page = 1;
 
-    Stream<PGMMap> search = library.getMaps().stream().filter(mapTags);
+    Stream<MapInfo> search = Sets.newHashSet(library.getMaps()).stream();
+    if (tags != null) {
+      final Set<String> tagSet =
+          Stream.of(tags.split(",")).map(String::trim).collect(Collectors.toSet());
+      search = search.filter(map -> matchesTags(map, tagSet));
```
TheMolkaPL: People can no longer negate maptags as @Pablete1234 suggested me to do. Let's if people would want to search for all DTM maps except the ones mixed with cores (SSB) - `/maps #monument !#core`

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -73,20 +82,30 @@ public static void maplist(
...
+        return (index + 1) + ". " + map.getStyledName(NameStyle.FANCY).toLegacyText();
       }
     }.display(audience, ImmutableSortedSet.copyOf(maps), page);
   }
 
-  private static boolean matchesAuthor(PGMMap map, String query) {
+  private static boolean matchesTags(MapInfo map, Set<String> tags) {
+    if (checkNotNull(tags).isEmpty()) return true;
+    for (MapTag tag : checkNotNull(map).getTags()) {
+      if (tags.contains(tag.getId())) {
```
TheMolkaPL: `tags` aren't always lower case, `getId()` is.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -97,144 +116,113 @@ private static boolean matchesAuthor(PGMMap map, String query) {
...
-          new PersonalizedText(mapInfoLabel("command.map.mapInfo.edition"), edition));
-    }
+  public void map(Audience audience, CommandSender sender, @Text MapInfo map) {
+    audience.sendMessage(
+        ComponentUtils.horizontalLineHeading(
+            ChatColor.DARK_AQUA
+                + map.getName()
+                + " "
+                + ChatColor.GRAY
+                + map.getVersion().toString(),
```
TheMolkaPL: I think this should be translatable? I guess in Arabic you should have `v1.0 Airship Battle` since they read right-to-left.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -97,144 +116,113 @@ private static boolean matchesAuthor(PGMMap map, String query) {
...
         result.extra(Components.space());
       }
 
-      MapTag mapTag = mapTags[i];
+      String mapTag = mapTags[i].getId();
       Component component =
-          mapTags[i]
-              .getComponentName()
-              .clickEvent(ClickEvent.Action.RUN_COMMAND, "/maplist " + mapTag.toString())
+          new PersonalizedText("#" + mapTag, ChatColor.GOLD)
```
TheMolkaPL: Instead of putting `#` here we should call `toString()` instead.

TheMolkaPL: BTW, maptags in voting books are yellow so I would keep them yellow here to for consistency.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -97,144 +116,113 @@ private static boolean matchesAuthor(PGMMap map, String query) {
...
+              .bold(false)
+              .clickEvent(ClickEvent.Action.RUN_COMMAND, "/maps -t " + mapTag)
```
TheMolkaPL: `/maps` is an alias to `/maplist` so I would keep `/maplist` here. :P

src/main/java/tc/oc/pgm/commands/provider/MapTagsConditionProvider.java
```diff
@@ -1,108 +0,0 @@
...
-                  if (mapTag.getName().startsWith(pair.first)) {
-                    mapTags.add(mapTagToString(mapTag, pair.second));
-                  }
-                }
-              }
-
-              return (List<String>) new ArrayList<>(mapTags);
-            })
-        .orElse(Collections.emptyList());
-  }
```
TheMolkaPL: We completly have lost tab suggestion for maptags :(

src/main/java/tc/oc/pgm/controlpoint/ControlPointModule.java
```diff
@@ -2,81 +2,92 @@
...
-@ModuleDescription(
-    name = "Control Points",
-    depends = {TeamModule.class, GoalModule.class, RegionModule.class, FilterModule.class})
-public class ControlPointModule extends MapModule<ControlPointMatchModule> {
-
-  private static final MapTag CONTROLPOINT_TAG = MapTag.forName("controlpoint");
+public class ControlPointModule implements MapModule<ControlPointMatchModule> {
 
+  private static final Collection<MapTag> TAGS =
+      ImmutableList.of(MapTag.create("controlpoint", "Control the Point", true, false));
```
TheMolkaPL: I feel that we should make the description a BaseComponent too. `Control The Point` should be translatable.

src/main/java/tc/oc/pgm/damage/DisableDamageModule.java
```diff
@@ -2,73 +2,63 @@
...
+public class DisableDamageModule implements MapModule {
   protected final SetMultimap<DamageCause, PlayerRelation> causes;
 
   public DisableDamageModule(SetMultimap<DamageCause, PlayerRelation> causes) {
     this.causes = causes;
   }
 
   @Override
-  public void loadTags(Set<MapTag> tags) {
-    if (causes.containsKey(DamageCause.FALL)) tags.add(NOFALLDAMAGE_TAG);
```
TheMolkaPL: Why complete removal?

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -1,14 +1,13 @@
 package tc.oc.pgm.listeners;
 
-import static com.google.common.base.Preconditions.*;
+import static com.google.common.base.Preconditions.checkNotNull;
```
TheMolkaPL: Fix the * import, please.

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
```diff
@@ -101,40 +101,34 @@ private void serializeMatch(Match match, JsonObject jsonObject) {
...
     checkNotNull(jsonObject);
 
-    MapInfo mapInfo = map.getInfo();
-    MapPersistentContext persistentContext = map.getPersistentContext();
-
-    jsonObject.addProperty("slug", mapInfo.slug());
-    jsonObject.addProperty("name", mapInfo.name);
-    jsonObject.addProperty("version", mapInfo.version.toString());
-    jsonObject.addProperty("objective", mapInfo.objective);
+    jsonObject.addProperty("id", map.getId());
```
TheMolkaPL: This is going to break at least OCC website. The slug is used to locate map icons afaik. Can we leave `slug` here, please?

src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+  public MapInfoImpl(
+      @Nullable String id,
+      Version proto,
+      Version version,
+      String name,
+      String description,
+      @Nullable Collection<Contributor> authors,
+      @Nullable Collection<Contributor> contributors,
+      @Nullable Collection<String> rules,
+      @Nullable Integer difficulty,
```
TheMolkaPL: Why is this an integer instead of a Difficulty?

src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+        info.getDescription(),
+        info.getAuthors(),
+        info.getContributors(),
+        info.getRules(),
+        info.getDifficulty(),
+        info.getTags(),
+        info.getMaxPlayers());
+  }
+
+  public MapInfoImpl(Element root) throws InvalidXMLException {
```
TheMolkaPL: I think we should make a public static factory method to parse the XML and return a new instance of this class.

src/main/java/tc/oc/pgm/map/MapTagImpl.java
```diff
@@ -0,0 +1,57 @@
+package tc.oc.pgm.map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import tc.oc.pgm.api.map.MapTag;
+
+public class MapTagImpl implements MapTag {
+
+  private static final String HASHTAG = "#";
```
TheMolkaPL: I called this `SYMBOL` originally because all these different special characters (maily on scoreboard) were called `SYMBOL`s.

src/main/java/tc/oc/pgm/map/MapTagImpl.java
```diff
@@ -0,0 +1,57 @@
...
+  private final String id;
+  private final String name;
+  private final boolean gamemode;
+  private final boolean auxiliary;
+
+  public MapTagImpl(String id, String name, boolean gamemode, boolean auxiliary) {
+    this.id = checkNotNull(id);
```
TheMolkaPL: We lost the regex pattern to validate if the maptag was in lower case, between a-z, 0-9, with - or _. This public and people from outside can create new instances of this class with illegal characters which will potentially lead to bugs.

src/main/java/tc/oc/pgm/api/map/MapTag.java
```diff
@@ -0,0 +1,44 @@
...
+   * @return A short, lowercase id without the "#".
+   */
+  String getId();
+
+  /**
+   * Get a full name for the tag.
+   *
+   * @return A full name.
+   */
+  String getName();
```
TheMolkaPL: `getName()` and `getDescription()` maybe?

src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+        Node.fromRequiredChildOrAttr(root, "name").getValueNormalize(),
+        Node.fromRequiredChildOrAttr(root, "objective", "description").getValueNormalize(),
+        parseContributors(root, "author"),
+        parseContributors(root, "contributor"),
+        parseRules(root),
+        XMLUtils.parseEnum(
+                Node.fromLastChildOrAttr(root, "difficulty"),
+                Difficulty.class,
+                "difficulty",
+                Difficulty.NORMAL)
```
TheMolkaPL: `Bukkit.getDifficulty()`?

src/main/java/tc/oc/pgm/api/map/MapInfo.java
Pablete1234: Since it's `@Nullable`, i *assume* if you use the `/map` without map argument it will get thru this and then if the result is an empty string will get replaced with the current map.

src/main/java/tc/oc/pgm/match/MatchImpl.java
```diff
@@ -131,7 +134,8 @@ protected MatchImpl(String id, PGMMap map, World world) {
     this.state = new AtomicReference<>(MatchPhase.IDLE);
     this.start = new AtomicLong(0);
     this.end = new AtomicLong(0);
-    this.capacity = new AtomicInteger(map.getPersistentContext().getTotalMaxPlayers());
+    this.capacity =
+        new AtomicInteger(map.getMaxPlayers().stream().findFirst().orElse(Bukkit.getMaxPlayers()));
```
TheMolkaPL: I think this should count `sum()`?

src/main/java/tc/oc/pgm/match/MatchManagerImpl.java
```diff
@@ -1,433 +1,206 @@
...
-    world.setAutoSave(false);
-
-    if (info.difficulty != null) {
-      world.setDifficulty(info.difficulty);
-    } else {
-      world.setDifficulty(server.getWorlds().get(0).getDifficulty());
-    }
-
-    return world;
+  public Match getMatch(@Nullable World world) {
```
TheMolkaPL: If the `world` is null throw NPE for finding bugs easier?

src/main/java/tc/oc/pgm/modes/ObjectiveModesModule.java
```diff
@@ -1,95 +1,102 @@
 package tc.oc.pgm.modes;
 
-import static tc.oc.pgm.map.ProtoVersions.*;
+import static tc.oc.pgm.api.map.MapProtos.MODES_IMPLEMENTATION_VERSION;
```
TheMolkaPL: I think we should keep the `*` import?

src/main/java/tc/oc/pgm/modules/EventFilterMatchModule.java
```diff
@@ -39,14 +55,16 @@
...
+  private final Match match;
 
   public EventFilterMatchModule(Match match) {
-    super(match);
+    this.match = match;
   }
 
   boolean cancel(Cancellable event, @Nullable MatchPlayer actor, @Nullable Component message) {
-    logger.fine("Cancel " + event + " actor=" + actor);
+    match.getLogger().fine("Cancel " + event + " actor=" + actor);
```
TheMolkaPL: Modules should have their own loggers if possible.

src/main/java/tc/oc/pgm/score/ScoreModule.java
```diff
@@ -80,69 +71,78 @@ public ScoreConfig getConfig() {
...
+      RegionParser regionParser = factory.getRegions();
+      ScoreConfig config = new ScoreConfig();
+      ImmutableSet.Builder<ScoreBoxFactory> scoreBoxFactories = ImmutableSet.builder();
+
+      for (Element scoreEl : scoreElements) {
+        config.scoreLimit = XMLUtils.parseNumber(scoreEl.getChild("limit"), Integer.class, -1);
+
+        // For backwards compatibility, default kill/death points to 1 if proto is old and <king/>
+        // tag
+        // is not present
```
TheMolkaPL: Unnecessary newline

src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -70,6 +88,37 @@
...
+            .filter(MapTag::isAuxiliary)
+            .map(MapTag::getName)
+            .collect(Collectors.toList());
+
+    String title = "";
+
+    if (gamemode.size() == 1) {
+      title = gamemode.get(0);
+    } else if (gamemode.size() >= 2) {
+      title = "Objectives";
```
TheMolkaPL: Should be translatable.

src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -70,6 +88,37 @@
...
+    }
+
+    if (auxiliary.size() == 1) {
+      title += (title.isEmpty() ? "" : " & ") + auxiliary.get(0);
+    } else if (gamemode.isEmpty() && auxiliary.size() == 2) {
+      title = auxiliary.get(0) + " & " + auxiliary.get(1);
```
TheMolkaPL: `&` should also be translatable.

src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -70,6 +88,37 @@
...
+    }
+
+    return title.isEmpty() ? "Match" : title;
```
TheMolkaPL: This too

src/main/java/tc/oc/pgm/commands/MapCommands.java
Electroid: ? It's just the google syntax cleaner.

src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
Electroid: There are some `MatchModule`s that don't even need a `Match` reference. They just use event listeners. Not really nessecary.

src/main/java/tc/oc/pgm/api/map/factory/MapFactory.java
Electroid: It's too widely used at this point. But I'll add some descriptions.

src/main/java/tc/oc/pgm/teams/TeamModule.java
```diff
@@ -40,15 +39,44 @@ public TeamModule(Set<TeamFactory> teams, @Nullable Boolean requireEven) {
   }
 
   @Override
-  public String toString() {
-    return getClass().getSimpleName() + "{teams=[" + Joiner.on(", ").join(teams) + "]}";
+  public Collection<MapTag> getTags() {
+    final int id = teams.size();
```
TheMolkaPL: `amount`?

src/main/java/tc/oc/pgm/api/map/MapInfo.java
Electroid: Moving away from having `Bukkit` literals in the api.

src/main/java/tc/oc/pgm/api/map/MapInfo.java
Electroid: Components are rendering level stuff, this interface is just descriptive.

src/main/java/tc/oc/pgm/terrain/TerrainModule.java
```diff
@@ -1,40 +1,28 @@
...
   private final TerrainOptions options;
 
   public TerrainModule(TerrainOptions options) {
     this.options = options;
   }
 
-  @Override
-  public void loadTags(Set<MapTag> tags) {
-    if (options.vanilla) tags.add(VANILLAWORLDGEN_TAG);
-  }
```
TheMolkaPL: Why complete removal?

src/main/java/tc/oc/pgm/commands/MapCommands.java
Electroid: Yea, will remove the `-p`, because everyone was just using `/maps N`

src/main/java/tc/oc/pgm/timelimit/TimeLimitModule.java
```diff
@@ -1,103 +1,89 @@
...
   private final @Nullable TimeLimit timeLimit;
 
   public TimeLimitModule(@Nullable TimeLimit limit) {
     this.timeLimit = limit;
   }
 
   @Override
-  public void loadTags(Set<MapTag> tags) {
-    if (timeLimit != null) tags.add(TIMELIMIT_TAG);
-  }
```
TheMolkaPL: Why complete removal?

src/main/java/tc/oc/pgm/commands/MapCommands.java
Electroid: Sure, but not a high priority.

src/main/java/tc/oc/pgm/commands/MapCommands.java
Electroid: I made them gold to keep the output consistent in `/map`. So might look at the voting book too.

src/main/java/tc/oc/pgm/commands/MapCommands.java
Electroid: `/maps` is the main command.

src/main/java/tc/oc/pgm/controlpoint/ControlPointModule.java
Electroid: In the future, not today.

src/main/java/tc/oc/pgm/map/MapInfoImpl.java
Electroid: See other comment.

src/main/java/tc/oc/pgm/worldborder/WorldBorderModule.java
```diff
@@ -32,50 +28,48 @@ public WorldBorderModule(List<WorldBorder> borders) {
   }
 
   @Override
-  public void loadTags(Set<MapTag> tags) {
-    tags.add(WORLDBORDER_TAG);
-  }
```
TheMolkaPL: Here too

src/main/java/tc/oc/pgm/match/MatchImpl.java
Electroid: Ah yes, nice catch.

src/main/java/tc/oc/pgm/timelimit/TimeLimitModule.java
Electroid: Too many tags.. you have to keep it simple.

src/main/java/tc/oc/pgm/api/map/MapInfo.java
TheMolkaPL: why?

src/main/java/tc/oc/pgm/commands/MapCommands.java
TheMolkaPL: Leave a todo there so we don't forget it when we migrate to text library.

src/main/java/tc/oc/pgm/timelimit/TimeLimitModule.java
TheMolkaPL: You could easily differentiate maps where you have to score n amount of points to win the match and maps where you have to score most points before the time ends. This one was pretty useful.

src/main/java/tc/oc/pgm/listeners/ServerPingDataListener.java
Electroid: `<slug>` is becoming `<id>` so I think we'll have to accept this breaking change.

src/main/java/tc/oc/pgm/listeners/MotdListener.java
```diff
@@ -31,7 +31,9 @@ public MotdListener() {
     this.format = Config.Motd.format();
 
     // Ensure there are always default keys
-    final MapInfo map = PGM.get().getMapOrder().getNextMap();
+    MapInfo map = PGM.get().getMapOrder().getNextMap();
+    if (map == null) map = PGM.get().getMapLibrary().getMaps().next();
```
Pablete1234: Is this assuming a map if a next map is missing? 
Next map not being defined is completely normal (for example, on voted pools) and shouldn't be avoided or treated like an error.

src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -62,12 +66,19 @@ public static void maplist(
...
-      search = search.filter(map -> matchesTags(map, tagSet));
+              .collect(Collectors.partitioningBy(s -> s.startsWith("!")));
+      search =
+          search.filter(
+              map ->
+                  matchesTags(
+                      map,
+                      tagSet.getOrDefault(false, new LinkedList<>()),
+                      Collections2.transform(
+                          tagSet.getOrDefault(true, new LinkedList<>()), s -> s.substring(1))));
```
Pablete1234: should probably just use:

```
        final Map<Boolean, Set<String>> tagSet = Stream.of(tags.split(","))
                .map(String::toLowerCase)
                .map(String::trim)
                .collect(Collectors.partitioningBy(s -> s.startsWith("!"),
                        Collectors.mapping(
                                (String s) ->  s.startsWith("!") ? s.substring(1) : s,
                                Collectors.toSet())));
```
that does both partitioning and removing the first char, that way you avoid the later transformation on the filter.

Also, i don't think linked lists are a good fit, should probably just use hash sets.

src/main/java/tc/oc/pgm/listeners/MotdListener.java
Electroid: If we don't know what the true next map will be, what should the motd say? `> PGM <` ?

Pablete1234: Either the current map, or something else like `Deciding next map`, `PGM`, or whatever other placeholder

Issue #225, opened 2020-01-12
Documentation on capture points bug


botinator:
So there is a bug with the CapturePoints module, and I reckon any module that relies on the `FiniteBlockRegion` implementation with maps below protocol versions `1.3.1`.

 The issue is that when these region definitions are generated, they are automatically wrapped into a `Union`. This is evading a check here: 

https://github.com/Electroid/PGM/blob/master/src/main/java/tc/oc/pgm/regions/FiniteBlockRegion.java#L145

In practice, this causes maps like `The Hill` to not work, and is a side effect of the changes made in the PR #96 which fixed another union bug (see: #82). 

There are multiple ways we can fix this. I'm considering importing the Region system from 1.12 PGM which fixes this bug and likely others, by allowing you to specify if you want a union wrap when you parse the element in the module file. For example see how it is o n 1.8:

```java
Region progressDisplayRegion =
        regionParser.parseRegionProperty(
            elControlPoint, BlockBoundedValidation.INSTANCE, "progress-display-region", "progress");
```

and on 1.12:

```java
Region progressDisplayRegion = regionParser.property(elControlPoint, "progress-display-region")
                                                   .alias("progress-display", "progress")
                                                   .validate(BlockBoundedValidation.INSTANCE)
                                                   .optionalUnion(null);
```

This is a thread to document what I've found so far, and discussion best options on fixing it.

Pablete1234:
on Loyalty DTM (and probably others) the monuments are broken because of this bug as well. The maps are on proto 1.3.0 and the monuments (destroyables) don't get old behaviour that went one block over, wich makes them detect 0 blocks and be completed from start.

PR #230, opened 2020-01-13
Patch union wrapped Cuboids in FiniteBlockRegion#fromWorld

botinator:

Temporary fix for #225, it just allows the points to function normally and prevents any issues where old maps will not detect monument breaks and therefore get stuck on the current map. This is a hot fix and not a long term solution.


botinator:
Pablo says he might have a more resilient fix for this in place, so it may be best to hold off merging that in to see what that looks like. This is really intended as a hot fix, and not a long term thing.

Pablete1234:
Closed in favor of #232

PR #243, opened 2020-01-17
Add a cooldown to /report, resolves Electroid#242

### LINKED ISSUE #242: Report cooldown
OCN had this. It's better to prevent report spamming with a short cooldown instead of having to kick or ban players who spam the command.

applenick:
Resolves #242 

The changes were inspired by [here](https://github.com/OvercastNetwork/ProjectAres/blob/b883ef579938c49076379fa306592da847fc672f/Commons/bukkit/src/main/java/tc/oc/commons/bukkit/report/ReportCommands.java#L105) 

I did not want to deviate too far from what has been done in the past. The cooldown is implemented with a hard-coded 15 second cooldown, which I feel will be appropriate for most situations.

 If it is desired I can add a config option or adjust the hard-coded limit, just let me know.

~~Edit: Forgot to mention, I did not add a bypass for players with a specific permission. As I can’t really imagine any situation where one would want to bypass this check. Anyway, if that is desired I can happily add it too.~~ (Outdated)

Screenshot of formatting (what would happen if user spams the command several times):
Screen Shot 2020-01-17 at 10 40 32 AM

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
@@ -3,10 +3,13 @@ package tc.oc.pgm.commands;
 import app.ashcon.intake.Command;
 import app.ashcon.intake.CommandException;
 import app.ashcon.intake.parametric.annotation.Text;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Map;
-import java.util.WeakHashMap;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import net.md_5.bungee.api.ChatColor;
 import org.bukkit.Bukkit;
 import org.bukkit.command.CommandSender;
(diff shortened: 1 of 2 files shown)


Meeples10:
There should probably be a config option for this that defaults to 15 seconds.

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -18,16 +24,40 @@
 
 public class ModerationCommands {
 
+  private static final int REPORT_COOLDOWN_SECONDS = 15;
+
+  private final Map<CommandSender, Instant> lastReportSent = new WeakHashMap<>();
```
Pablete1234: Should use a guava expiring cache, and maybe just use the UUID as key (if the command sender is the console just always allow)

Pablete1234:
I think the 15 second constant is fine, there's no need to over-complicate such a simple feature

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
Electroid: Also exempt players with the `Permissions.STAFF`

applenick: Thanks @Pablete1234 and @Electroid! 
I switched the  map over to use a guava loading cache. I haven’t used them before, so hopefully the way I implemented it is the proper way. Though after testing everything seemed to work, so it should be right 😄 

Also I added the staff bypass.

applenick:
Thanks @Pablete1234 and @Electroid!
I switched the map over to use a guava loading cache. I haven’t used them before, so hopefully the way I implemented it is the proper way. Though after testing everything seemed to work, so it should be right 😄

I also agree with @Pablete1234 that this is a very simple feature and really there is no reason to add an additional config option.

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -18,6 +26,14 @@
 
 public class ModerationCommands {
 
+  private static final int REPORT_COOLDOWN_SECONDS = 15;
+
+  private static final Cache<UUID, Instant> lastReportSent =
+      CacheBuilder.newBuilder()
+          .weakKeys()
```
Pablete1234: Using uuis, no need for weak keys

Pugzy:
Would it be better to make the translated message more generic?
> Please wait {0} before running that command again

Allowing this to be extracted to more commands in the future

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
applenick: Ah thanks, I was unsure! I’ll adjust that now

applenick:
> Would it be better to make the translated message more generic?
> 
> > Please wait {0} before running that command again
> 
> Allowing this to be extracted to more commands in the future

Great suggestion. I’ll adjust that as well.

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -27,7 +40,32 @@ public static void report(
...
+        if (secondsRemaining > 0) {
+          throw new CommandException(
+              AllTranslations.get()
+                  .translate(
+                      "command.cooldown",
+                      commandSender,
+                      ChatColor.AQUA
+                          + (secondsRemaining
+                              + " second"
+                              + (secondsRemaining != 1 ? "s" : "")
```
TheMolkaPL: This should be translatable.

applenick: Thanks Molka, I should have done this from the beginning!

TheMolkaPL:
I meant something like this: :P

https://github.com/Electroid/PGM/blob/31ebbcc74002981fcf05dc0af45e861a77d07aca/src/main/i18n/templates/strings.properties#L672-L674

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -18,6 +25,11 @@
 
 public class ModerationCommands {
 
+  private static final int REPORT_COOLDOWN_SECONDS = 15;
+
+  private static final Cache<UUID, Instant> lastReportSent =
+      CacheBuilder.newBuilder().expireAfterWrite(REPORT_COOLDOWN_SECONDS, TimeUnit.SECONDS).build();
```
TheMolkaPL: I think this should be called `LAST_REPORT_SENT`.

applenick: > I meant something like this: :P
> 
> https://github.com/Electroid/PGM/blob/31ebbcc74002981fcf05dc0af45e861a77d07aca/src/main/i18n/templates/strings.properties#L672-L674

Whoops! 😆 
I’ll make the adjustments now, thanks again haha.

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -27,7 +39,36 @@ public static void report(
...
+      @Text String reason)
+      throws CommandException {
+    if (!commandSender.hasPermission(Permissions.STAFF) && commandSender instanceof Player) {
+      // Check for cooldown
+      Instant lastReport = LAST_REPORT_SENT.getIfPresent(matchPlayer.getId());
+      if (lastReport != null) {
+        Duration timeSinceReport = Duration.between(lastReport, Instant.now());
+        long secondsRemaining = REPORT_COOLDOWN_SECONDS - timeSinceReport.getSeconds();
+        if (secondsRemaining > 0) {
+          Component secondsComponent = new PersonalizedText("" + secondsRemaining);
```
TheMolkaPL: I would use `Integer.toString(secondsRemaining)`, otherwise looks good to me.

applenick: Oh snap, I did not see your suggestion! I’m adjusting this right now. 👍

PR #245, opened 2020-01-18
Add soft dependency on ViaVersion

### LINKED ISSUE #16: Fix tab list for 1.7 players
`ViaRewind` allows `1.7` players to join the server as a `1.8` player. The `1.7` tab list is quite different, so we want to disable it, only problem is `PGM` has no way of determining if the player is a `1.8` player or a secret `1.7` player "disguised" as a `1.8` player. ~~Ideally, we do not import any new APIs to do this.~~ Any ideas?

Meeples10:

If a ViaVersion is installed on the server, the usual way of getting a player's client protocol version always returns `47`. This change ~~would allow an external plugin to set the player's protocol version to the correct value by retrieving it from ViaVersion (see [§Get the players protocol version](https://docs.viaversion.com/display/VIAVERSION/Basic+API+usage) in the ViaVersion docs), which in turn~~ would allow PGM to behave differently for 1.7, 1.8, 1.15, etc. players. This could be used to fix #16.

### DIFF
--- a/src/main/java/tc/oc/pgm/listeners/PGMListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/PGMListener.java
@@ -1,6 +1,6 @@
 package tc.oc.pgm.listeners;
 
-import static com.google.common.base.Preconditions.*;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.UUID;
 import net.md_5.bungee.api.ChatColor;
@@ -45,7 +45,6 @@ import tc.oc.pgm.events.PlayerPartyChangeEvent;
 import tc.oc.pgm.gamerules.GameRule;
 import tc.oc.pgm.gamerules.GameRulesModule;
 import tc.oc.pgm.modules.TimeLockModule;
-import tc.oc.util.ViaUtils;
 
 public class PGMListener implements Listener {
   private final Plugin parent;
@@ -72,7 +71,6 @@ public class PGMListener implements Listener {
 
   @EventHandler(priority = EventPriority.LOW)
   public void addPlayerOnJoin(final PlayerJoinEvent event) {
-    System.out.println(ViaUtils.getProtocolVersion(event.getPlayer())); // XXX REMOVE THIS
     if (this.mm.getMatch(event.getWorld()) == null) {
       event
           .getPlayer()
--- a/src/main/java/tc/oc/pgm/listeners/PGMListener.java
+++ b/src/main/java/tc/oc/pgm/listeners/PGMListener.java
@@ -45,6 +45,7 @@ import tc.oc.pgm.events.PlayerPartyChangeEvent;
 import tc.oc.pgm.gamerules.GameRule;
 import tc.oc.pgm.gamerules.GameRulesModule;
 import tc.oc.pgm.modules.TimeLockModule;
+import tc.oc.util.ViaUtils;
 
 public class PGMListener implements Listener {
   private final Plugin parent;
@@ -71,6 +72,7 @@ public class PGMListener implements Listener {
 
   @EventHandler(priority = EventPriority.LOW)
   public void addPlayerOnJoin(final PlayerJoinEvent event) {
+    System.out.println(ViaUtils.getProtocolVersion(event.getPlayer())); // XXX REMOVE THIS
     if (this.mm.getMatch(event.getWorld()) == null) {
       event
           .getPlayer()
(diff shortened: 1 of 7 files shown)


src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -75,6 +78,9 @@ public MatchPlayerImpl(Match match, Player player) {
     this.frozen = new AtomicBoolean(false);
     this.dead = new AtomicBoolean(false);
     this.visible = new AtomicBoolean(false);
+    this.protocolVersion =
+        new AtomicInteger(
+            ((CraftPlayer) player).getHandle().playerConnection.networkManager.protocolVersion);
```
Pablete1234: Put this in nmshacks

Electroid:
I would rather have a `SoftDepend` on ViaVersion, doesn't make since to make another plugin called PGMViaVersion just to set the version.

AustinLMayes:
Checking plugin names is unsafe, and could return some false positives. 

`Class.forName` is safer and is less prone to these false positives. 

Also, this would require that Via always be shaded in, since `ViaUtils` has imports from VV. A better solution would be to only load the utils class if Via is actually loaded, removing the need for it to be directly shaded in. You could register a special Via listener, for example, which sets the version on login only if the plugin is actually loaded.

src/main/java/tc/oc/util/ViaUtils.java
```diff
@@ -1,16 +1,48 @@
...
+   *     href="https://wiki.vg/Protocol_version_numbers">https://wiki.vg/Protocol_version_numbers</a>
+   */
   public static int getProtocolVersion(Player player) {
-    return enabled() ? Via.getAPI().getPlayerVersion(player) : NMSHacks.getProtocolVersion(player);
+    int version = NMSHacks.getProtocolVersion(player);
+    if (enabled()) {
+      try {
+        return (int) getPlayerVersion.invoke(viaAPI, player);
+      } catch (Exception e) {
+        e.printStackTrace();
```
TheMolkaPL: This should use a logger.

Pablete1234:
Why would using viaversion as dependency mean you need to shade it?

You can just add viaversion as dependency, check on runtime if it's present (with Class.forName), and use all the API without the need for reflection to call the api.

AustinLMayes:
@Pablete1234 because the enabled check is in the same class that uses the actual Via API, so loading the class requires that Via classes be loaded

Pablete1234:
@AustinLMayes i'm pretty sure that's not an issue whatsoever. As long as you don't call the class if not loaded, it's not an issue.

src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -71,6 +72,7 @@ public void onPlayerLogin(final PlayerLoginEvent event) {
 
   @EventHandler(priority = EventPriority.LOW)
   public void addPlayerOnJoin(final PlayerJoinEvent event) {
+    System.out.println(ViaUtils.getProtocolVersion(event.getPlayer())); // XXX REMOVE THIS
```
TheMolkaPL: XXX REMOVE THIS

src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -1,6 +1,6 @@
 package tc.oc.pgm.listeners;
 
-import static com.google.common.base.Preconditions.*;
+import static com.google.common.base.Preconditions.checkNotNull;
```
TheMolkaPL: Can you leave the * import?

AustinLMayes:
@Pablete1234 yes, but the `enabled()` check is also In the class, so checking if via was enabled using that method would fail if via isn't loaded (since the same class imports Via APIs directly)

Meeples10:
@AustinLMayes how would it fail? Based on my testing both `enabled()` and `getProtocolVersion(Player)` work as expected whether or not ViaVersion is loaded.

AustinLMayes:
Turns out on the iteration of the Bukkit plugin class loader the server version is song isn't prone to the issue (reloading breaks some static code blocks, leaves some phantom stuff etc)

Ignore. As long as it is never possible for ANY calls to the via code to be executed if it isn't loaded, it's fine. 

As for the actual imports, those aren't a factor since those are long gone before bytecode generation ever begins.

Pablete1234:
@AustinLMayes Indeed, imports mean nothing on bytecode. They're just a shorthand for not needing fully classified names on classes. 

The fact that you load a class that has calls to code that may not be loaded on the JVM, doesn't mean the code breaks just by loading the class.
It only breaks (by throwing exceptions when class/method/fields are not found) if the code that runs those actions can't be found when the invocation occurs, wich we're going to do if viaversion isn't loaded.

PR #248, opened 2020-01-20
Make commands using duration read config

KingOfSquares:
My earlier "fix" did not quite fix much. This should fix it.

Is this correct @TheMolkaPL ? :)

Signed-off-by: Simon simonmr60@gmail.com

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/provider/DurationProvider.java
+++ b/src/main/java/tc/oc/pgm/commands/provider/DurationProvider.java
@@ -27,6 +27,6 @@ public class DurationProvider implements BukkitProvider<Duration> {
           .toStandardDuration()
           .toDuration();
     }
-    return Duration.standardSeconds(30L);
+    return null;
   }
 }


AustinLMayes:
Just a friendly note that, in the future, you can simply force push to the branch this PR is based off of instead of remaking the entire pull request.

TheMolkaPL:
Also `git push <remote> --force`

KingOfSquares:
But my earlier commit didnt have the correct sign of right? I cant change that?

TheMolkaPL:
> I cant change that?

Sure you can. You can modify or even rewrite your history. If you do `git commit --amend` it will make a new commit, based on the previous one (removing the old one and attaching the changes to the new one). So if you do `git commit -s --amend` it will create a new signed-off commit based on the previous one. `-s` will add `Signed-off-by` thing to the description.

Now if you try to push this commit to remote, GitHub will reject the push because there is a difference in the history. The new commit you just created is child to commit older than the newest one on the remote branch. You can force that push by using the `--force` flag in the `push` command.

PR #254, opened 2020-01-23
Fix sounds whenever u kill yourself

VladyslavLukyanenko:
Signed-off-by: Vlad <me@vladyslav.xyz>


src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
```diff
@@ -54,7 +54,8 @@ private void playRaindrop(Competitor competitor) {
   @EventHandler(priority = EventPriority.MONITOR)
   public void onMatchPlayerDeath(MatchPlayerDeathEvent event) {
     ParticipantState killer = event.getKiller();
-    if (killer != null) {
+    MatchPlayer victim =  event.getVictim();
+    if (killer != null && killer.getId() != victim.getId()) {
```
TheMolkaPL: `getId` returns `UUID` so you should compare them with `equals`.

TheMolkaPL: Fixed in 5684ad082973d226785920257131af2d9186fea7

PR #260, opened 2020-01-25
Add Observer Tools

### LINKED ISSUE #220: Give observers night vision
This is useful for dark maps.

applenick:
# Observer Tools 

Resolves #220
Inspiration was taken mostly from @Pablete1234's comment on that issue. Thanks!

## Features:
This implementation of an observer tool set adds a new GUI menu for observers to interact with. While simple in functionality, this will most certainly prove useful for users who want more control over the features described below. I've designed this feature set with future improvements/additions in mind, so with minimal tweaking more tools/features can be added to the GUI if desired. 

overview

### Fly Speed
Adjust the flight speed between 4 set values. I performed a bit of testing between several values to see what made a difference to the user, and I believe the values chosen are good. 

Also I used the names `Regular`, `Fast`, `Faster`, and `Hyperspeed` as user friendly names to define the speeds. As I felt that using numbers would be more confusing. If anyone has better suggestions for names or speed values, I'm open to adjusting them.

fly-speed

### Night Vision
Allows the player to toggle night vision on/off. Potions are removed when joining matches, so this will not interfere in anyway. This also does not persist between matches, so users will be able to decide on a per match basis if they want night vision on or not.

night-vision

### Toggle Visibility
Simply toggles the Observer visibility setting. 

obs-vis

### Gamemode
This was one of the more tricky additions, as it toggles the observer between Creative and Spectator mode. Since the regular hotbar is not accessible while in Spectator, once the user clicks the tool, a message will be sent in the chat. Stating in order to access this menu they must use the `/tools` command.

 Also I made sure to include the `/!` command as suggested by @Pablete1234.

gamemode

## Future Improvements
As stated earlier, I feel this menu has the potential to house more useful tools. Hopefully with community feedback, we can gather new ideas and implement them when needed. 

There was one feature I feel could be implemented, though I chose not to for simplicity sake before receiving feedback. It would be allowing players the option to add one of the setting "tools" to their hotbar to easily activate the setting. For example, were the night vision tool added to the hotbar the user when clicking the item would be able to toggle night vision on/off without opening the inventory menu. 

## Final Note
Just wanted to state, I did not add any config options to this. As I feel nothing here would disrupt gameplay in a way server owners would want to disable it. However if that is requested it can be added.


src/main/java/tc/oc/pgm/gui/InventoryGUI.java
```diff
@@ -0,0 +1,132 @@
...
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.util.StringUtils;
+
+public abstract class InventoryGUI {
+
+  /*
+   * Code extracted from PickerMatchModule to allow for reuse
+   */
+
+  public Set<MatchPlayer> viewing = Sets.newHashSet();
```
TheMolkaPL: Keep it private or protected, please.

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public String title; // Title of the inventory
```
TheMolkaPL: Keep it private or protected, please.

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
```diff
@@ -0,0 +1,132 @@
...
+  private int size; // Size of inventory
+
+  public InventoryGUI(String title, int size) {
+    this.title = title;
```
TheMolkaPL: checkNotNull please

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
```diff
@@ -0,0 +1,132 @@
...
+    this.size = size;
```
TheMolkaPL: Check if the size is positive and can be divided by 9.

src/main/java/tc/oc/pgm/observers/tools/FlySpeedTool.java
```diff
@@ -0,0 +1,89 @@
...
+      this.value = value;
+    }
+
+    public float getValue() {
+      return value;
+    }
+
+    public Component getName() {
+      return new PersonalizedText(WordUtils.capitalize(this.name().toLowerCase())).color(color);
+    }
```
TheMolkaPL: This should be translatable.

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
applenick: Thanks for the suggestions @TheMolkaPL !
I have to head out for a few hours, but I will make your suggested changes as soon as I’m back. 😄

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
```diff
@@ -13,18 +16,27 @@
 
 public abstract class InventoryGUI {
 
-  /*
-   * Code extracted from PickerMatchModule to allow for reuse
-   */
-
-  public Set<MatchPlayer> viewing = Sets.newHashSet();
+  protected static int ROW_WIDTH = 9; // Number of columns per row
+  protected static int MAX_ROWS = 6; // Max allowed row size
```
TheMolkaPL: You forgot `final` :p

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
```diff
@@ -13,18 +16,27 @@
...
 
-  public String title; // Title of the inventory
-  private int size; // Size of inventory
+  private Set<MatchPlayer> viewing = Sets.newHashSet(); // Set of players who are viewing the gui
+  private String title; // Title of the inventory
+  private int rows; // The # of rows in the inventory
```
TheMolkaPL: I think all of these can be `final`?

src/main/java/tc/oc/pgm/gui/InventoryGUI.java
applenick: 🤦‍♂ Haha, whoops.

src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
```diff
@@ -0,0 +1,90 @@
...
+    toggleObserverGameMode(player);
+    menu.refreshWindow(player);
+  }
+
+  public void toggleObserverGameMode(MatchPlayer player) {
+    player.setGameMode(getOppositeMode(player.getGameMode()));
+    if (player.getGameMode() == GameMode.SPECTATOR) {
+      player.sendWarning(getToggleMessage(), true);
+    } else if (isCreative(player)) {
+      player.getBukkit().performCommand("!");
```
TheMolkaPL: We should keep a comment here saying that this command is from WorldEdit and it unstucks the player.

src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Override
+  public ChatColor getColor() {
+    return ChatColor.DARK_AQUA;
+  }
+
+  @Override
+  public List<String> getLore(MatchPlayer player) {
+    Component gamemode =
+        new PersonalizedText(WordUtils.capitalizeFully(player.getGameMode().name().toLowerCase()))
```
TheMolkaPL: I'm sure there is a translation for this in the client for thìs. Can we use it?

src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
```diff
@@ -0,0 +1,90 @@
...
+    return player.getGameMode().equals(GameMode.CREATIVE);
+  }
+
+  private Component getToggleMessage() {
+    Component command =
+        new PersonalizedText("/tools")
+            .color(ChatColor.AQUA)
+            .hoverEvent(
+                HoverEvent.Action.SHOW_TEXT,
+                new PersonalizedText("Click to open tool menu").color(ChatColor.GRAY).render())
```
TheMolkaPL: Should be translatable.

src/main/java/tc/oc/pgm/observers/tools/NightVisionTool.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public ChatColor getColor() {
+    return ChatColor.DARK_PURPLE;
+  }
+
+  @Override
+  public List<String> getLore(MatchPlayer player) {
+    Component status =
+        new PersonalizedText(hasNightVision(player) ? "On" : "Off")
```
TheMolkaPL: Should be translatable.

src/main/java/tc/oc/pgm/observers/tools/VisibilityTool.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Override
+  public ChatColor getColor() {
+    return ChatColor.YELLOW;
+  }
+
+  @Override
+  public List<String> getLore(MatchPlayer player) {
+    Component status =
+        new PersonalizedText(isVisible(player) ? "Shown" : "Hidden")
```
TheMolkaPL: Should be translatable.

src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
applenick: Curious. I’m not familiar with the client translations, could you explain?

Meeples10: These are the values in the 1.8.9 en_US.lang file:
```
gameMode.survival=Survival Mode
gameMode.creative=Creative Mode
gameMode.adventure=Adventure Mode
gameMode.spectator=Spectator Mode
```
Alternatively:
```
selectWorld.gameMode.survival=Survival
selectWorld.gameMode.creative=Creative
selectWorld.gameMode.adventure=Adventure
selectWorld.gameMode.spectator=Spectator
```

applenick: Thanks @Meeples10!
Ah I understand now, that’s very useful 👍

Meeples10: Have you confirmed that this actually works? I'm not sure if PGM translates vanilla messages.

src/main/java/tc/oc/pgm/observers/ObserverToolsMatchModule.java
```diff
@@ -0,0 +1,186 @@
...
+    menu.remove(getMatch().getPlayer((Player) event.getPlayer()));
+  }
+
+  public void openMenu(MatchPlayer player) {
+    if (hasUse(player)) {
+      menu.display(player);
+    }
+  }
+
+  private boolean hasUse(MatchPlayer player) {
```
Pablete1234: maybe rename to `canUse` instead ?

src/main/java/tc/oc/pgm/observers/tools/FlySpeedTool.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public FlySpeed getNext() {
+      return ordinal() < (values().length - 1) ? values()[ordinal() + 1] : NORMAL;
```
Pablete1234: have a:
`private static FlySpeed[] speeds = values();`

and have this method be:
`return speeds[(ordinal() + 1) % speeds.length]`

src/main/java/tc/oc/pgm/observers/ObserverTool.java
```diff
@@ -0,0 +1,41 @@
...
+import org.bukkit.Material;
+import org.bukkit.inventory.ItemFlag;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.ItemMeta;
+import tc.oc.component.Component;
+import tc.oc.component.render.ComponentRenderers;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.gui.InventoryGUI;
+
+public interface ObserverTool {
```
Pablete1234: maybe have this be a generic gui item thing?

src/main/java/tc/oc/pgm/commands/ObserverCommands.java
```diff
@@ -0,0 +1,33 @@
...
+import tc.oc.component.types.PersonalizedTranslatable;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.observers.ObserverToolsMatchModule;
+
+public class ObserverCommands {
+
+  @Command(
+      aliases = {"tools", "observertools", "ot"},
+      desc = "Open the observer tool menu")
+  public static void openObserverToolMenu(CommandSender sender, MatchPlayer player)
```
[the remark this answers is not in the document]
applenick: Hey @Pablete1234,
So what I encountered during testing was the lack of the ability to use any item while a player is a spectator. Unless there is some odd method to access items as a spectator which I’m unaware of, I’m unsure how exactly one can return to the menu without using the command. 

Here’s a look at what I mean:
gamemode-example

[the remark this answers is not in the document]
Pablete1234: @applenick the item you have selected is still the diamond, right-ckicking brings uo the menu again. IIRC, you must click againt block while in gm3 or there's no event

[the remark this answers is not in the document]
applenick: @Pablete1234 So I just did some more testing and I see what I missed the first time. You are correct, I was simply right clicking air in my previous tests. Right-clicking blocks are the only way to fire the event as a spectator. However I fear that if we relied solely on just the right-clicking of blocks we may run into players who are unfamiliar as I was.

 For now I believe it’s important to offer the command, the same way there is a command to join the match although we have the picker. 

Also in regards to your original comment, clicking any diamond does open the menu.

src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
```diff
@@ -0,0 +1,93 @@
...
+    menu.refreshWindow(player);
+  }
+
+  public void toggleObserverGameMode(MatchPlayer player) {
+    player.setGameMode(getOppositeMode(player.getGameMode()));
+    if (player.getGameMode() == GameMode.SPECTATOR) {
+      player.sendWarning(getToggleMessage(), true);
+    } else if (isCreative(player)) {
+      // Note: This performs the WorldEdit command to ensure the player is not stuck
+      player.getBukkit().performCommand("worldedit:!");
```
Electroid: What happens if `WorldEdit` is not loaded? Will it show a command not found error?

applenick: Good catch. I’ll add a check to see if WorldEdit is enabled or not.
Thanks @Electroid!

src/main/java/tc/oc/pgm/menu/InventoryMenu.java
```diff
@@ -0,0 +1,151 @@
...
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.util.StringUtils;
+
+public abstract class InventoryMenu {
+
+  protected static final int ROW_WIDTH = 9; // Number of columns per row
+  protected static final int MAX_ROWS = 6; // Max allowed row size
+
+  private final Set<MatchPlayer> viewing =
+      Sets.newHashSet(); // Set of players who are viewing the gui
```
Electroid: This doesn't seem to get cleaned up anywhere? How about using `WeakHashMap` ?

applenick: Hey @Electroid,
The viewing map does get cleaned up via the ObserverToolMatchModule, though that is only for that implementation. I switched it over to a `WeakHashMap` as suggested, to accommodate for when future implementations are added. 
Thanks!

PR #261, opened 2020-01-25
Disable Teammate kill sound

VladyslavLukyanenko:
Signed-off-by: Vlad <me@vladyslav.xyz>

### DIFF
--- a/src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
+++ b/src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
@@ -55,7 +55,9 @@ public class SoundsMatchModule extends MatchModule implements Listener {
   public void onMatchPlayerDeath(MatchPlayerDeathEvent event) {
     ParticipantState killer = event.getKiller();
     MatchPlayer victim = event.getVictim();
-    if (killer != null && !killer.getId().equals(victim.getId())) {
+    if (killer != null
+        && !killer.getId().equals(victim.getId())
+        && !killer.getParty().getName().equals(victim.getParty().getName())) {
       playRaindrop(killer);
     }
   }


src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
```diff
@@ -55,7 +55,9 @@ private void playRaindrop(Competitor competitor) {
   public void onMatchPlayerDeath(MatchPlayerDeathEvent event) {
     ParticipantState killer = event.getKiller();
     MatchPlayer victim = event.getVictim();
-    if (killer != null && !killer.getId().equals(victim.getId())) {
+    if (killer != null
+        && !killer.getId().equals(victim.getId())
```
Pablete1234: remove the own-kill check if we have a same-party check

VladyslavLukyanenko: Yeah thought about it

src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
```diff
@@ -32,51 +32,56 @@ public SoundsMatchModule createMatchModule(Match match) throws ModuleLoadExcepti
...
     ParticipantState killer = event.getKiller();
     MatchPlayer victim = event.getVictim();
-    if (killer != null
-        && !killer.getId().equals(victim.getId())
-        && !killer.getParty().getName().equals(victim.getParty().getName())) {
-      playRaindrop(killer);
+    if (killer != null) {
+      if (killer.getId().equals(victim.getId())) {
+
+      } else if (killer.getParty().getName().equals(victim.getParty().getName())) {
```
Meeples10: Any reason not to use `killer.getParty().equals(victim.getParty())`?

VladyslavLukyanenko: I can use that actually thanks for the heads up.

src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
```diff
@@ -32,49 +32,56 @@ public SoundsMatchModule createMatchModule(Match match) throws ModuleLoadExcepti
...
   public void onMatchPlayerDeath(MatchPlayerDeathEvent event) {
     ParticipantState killer = event.getKiller();
     MatchPlayer victim = event.getVictim();
-    if (killer != null && !killer.getId().equals(victim.getId())) {
-      playRaindrop(killer);
+    if (killer != null) {
+      if (killer.getId().equals(victim.getId())) {
+
+      } else if (killer.getParty().equals(victim.getParty())) {
+        playSound(killer, ENDERMAN_SCREAM);
```
Pablete1234: Let's not do this

VladyslavLukyanenko: Didnt know how to do it properly

src/main/i18n/templates/strings.properties
```diff
@@ -69,6 +69,7 @@ death.predictedSuffix = (predicted)
 # {3} = mob
 # {4} = distance in blocks
 
+death.friendlyfire.warning = Killing your teammates intentionally will get you punished
```
TheMolkaPL: I think we should keep the `death.` key for DeathMessageBuilder exclusively?

src/main/java/tc/oc/pgm/modules/SoundsMatchModule.java
```diff
@@ -32,49 +33,57 @@ public SoundsMatchModule createMatchModule(Match match) throws ModuleLoadExcepti
...
   public void onMatchPlayerDeath(MatchPlayerDeathEvent event) {
     ParticipantState killer = event.getKiller();
     MatchPlayer victim = event.getVictim();
-    if (killer != null && !killer.getId().equals(victim.getId())) {
-      playRaindrop(killer);
+    if (killer != null) {
+      if (!killer.getId().equals(victim.getId())) {
+        if (killer.getParty().equals(victim.getParty())) {
+          playSound(killer, ENDERMAN_SCREAM);
+          killer.sendWarning(new PersonalizedTranslatable("death.friendlyfire.warning"), false);
```
Pablete1234: It is not PGM's duty to moderate, and not every same-team kill is bad. Accidents can happen and there's no need to play enderman scream & warn them they're getting punished.

PR #265, opened 2020-01-26
Don't start a mapvote when restarting

applenick:
A simple fix.

I noticed both on OCC and testing locally, that when a restart is queued map voting will initiate a vote. I’m sure some players must have been confused before, as they’re voting for a map and then the server restarts. 


Brottweiler:
Yeah, I can confirm every time the server will restart, it still brings up a vote.

I think players might assume that map will play after restart and then just assume the random map that comes up after the restart won the vote.

Pablete1234:
This is really not the way to fix this issue. Map votes & scores should be persisted thru the restart instead of not voting at all

Issue #210, opened 2020-01-08
Commands do not expose help or usage instructions


Pablete1234:
I'm pretty sure this is an issue with intake, and not specifically PGM, but it's still relevant to have it noted as it heavily affects pgm's commands.

TheMolkaPL:
We should *maybe* migrate to another better supported command library? [EngineHub's Piston](https://github.com/EngineHub/Piston) may be an option. It is already been used by [WorldEdit 7](https://github.com/EngineHub/WorldEdit/releases/tag/7.0.0).

Sadly, it is not currently documented, but it is [being implemented to CommandBook](https://github.com/EngineHub/CommandBook/pull/117).

KingOfSquares:
@applenick  Suggested https://github.com/aikar/commands to me, thoughts? 
They have much better documentation, and seems like more recent activity.

KingOfSquares:
Solved in https://github.com/Electroid/intake/commit/59f112a279b44d1ee28a2f7361a10f75e4d0d6af

PR #266, opened 2020-01-27
WIP : Changing from Intake to Piston

KingOfSquares:
Hello. [Link to Piston](https://github.com/EngineHub/Piston)

I was inspired by #210 to start a migration to a newer more supported framework for in-game commands. 

While this is WIP formatting is probably going to be pretty ruined.

Things that needs to be done for this to work:

- [x] Change the annotation parts in the XXXcommands.java
- [x] Change the Exceptions in the XXXcommands.java
- [x] Add a replacement for @Default
- [ ] Add permissions (Probably with CommandConditionGenerator [link](https://github.com/EngineHub/Piston/blob/master/core-ap/runtime/src/main/java/org/enginehub/piston/gen/CommandConditionGenerator.java)
- [ ] Add the implementation part
- [ ] Remove all other dependencies on Intake (Not sure how much there is of this..)

Other things

- [ ] Look into moving translation of exception out of the commands.java

All feedback appreciated!! :)

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/AdminCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/AdminCommands.java
@@ -1,17 +1,12 @@
 package tc.oc.pgm.commands;
 
-import app.ashcon.intake.Command;
-import app.ashcon.intake.CommandException;
-import app.ashcon.intake.parametric.annotation.Default;
-import app.ashcon.intake.parametric.annotation.Switch;
-import app.ashcon.intake.parametric.annotation.Text;
-import java.util.AbstractMap;
-import java.util.Map;
-import java.util.stream.Collectors;
-import javax.annotation.Nullable;
+
+
 import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
+
 import org.joda.time.Duration;
+
 import tc.oc.pgm.AllTranslations;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.Permissions;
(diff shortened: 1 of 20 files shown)


Electroid:
I like where this is going, keep going. I'd rather not *also* maintain Intake.

AustinLMayes:
@KingOfSquares While converting, it would be nice if we could implement a simple translatable command backbone - which shifts all translation/rendering to the main `onCommand` entrypoint in `PGM`. 

See the [WorldEdit Implementation](https://github.com/EngineHub/WorldEdit/blob/master/worldedit-core/src/main/java/com/sk89q/worldedit/extension/platform/PlatformCommandManager.java) for an example. **Note:** That file is quite long and does a lot of extra work to support WE's multiple platforms, the main thing I"m referring to is the `catch` blocks in the `handle` method. 

There are a lot of instances in code currently where exceptions are either in plain English or translated in-method, which isn't great.

KingOfSquares:
Will try that @AustinLMayes !

KingOfSquares:
Im having issues with conflicting Guava code. So until i can resolve that this is halted.

TheMolkaPL:
> issues with conflicting Guava code

Can you explain more?

KingOfSquares:
> > issues with conflicting Guava code
> 
> Can you explain more?

The Minecraft Server uses Guava 17 for 1.8, but i need Guava > 20.00 for Piston to work. Im trying to get piston to relocate its own guava files so it wont conflict but.. i'm not able to do that yet. Posted about it in #debug on OCC Discord

KingOfSquares:
Right now this seems like a project a little bit out of reach for me. If someone manages to get Piston to work i would love to help, but i am going to try and work on other issues as of now.

PR #270, opened 2020-01-27
Fix tab list for players on 1.7.10, resolves #16

### LINKED ISSUE #16: Fix tab list for 1.7 players

Meeples10:
1.7.10 tab list


src/main/java/tc/oc/pgm/listeners/LongRangeTNTListener.java
```diff
@@ -271,7 +272,9 @@ private void removeTNT(Entity entity, boolean explode) {
 
   @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
   public void onPlayerJoin(final PlayerJoinEvent event) {
-    this.viewsByPlayer.put(event.getPlayer(), new View(event.getPlayer()));
+    if (ViaUtils.getProtocolVersion(event.getPlayer()) >= 47) {
```
Pablete1234: Create a util for protocol versions, `47` means nothing but `Protocol.VERSION_1_7` (or something along those lines) is easy to read

src/main/java/tc/oc/tablist/TabRender.java
```diff
@@ -53,6 +54,9 @@ private BaseComponent getContent(TabEntry entry, int index) {
   }
 
   private void appendAddition(TabEntry entry, int index) {
+    if (ViaUtils.getProtocolVersion(this.view.getViewer()) < 47) {
```
Pablete1234: This is a complete and absolute hack. Don't change the current tab view to avoid sending packets to 1.7 players, give them a new no-op view

src/main/java/tc/oc/tablist/TabRender.java
```diff
@@ -154,6 +158,9 @@ public void setHeaderFooter(TabEntry header, TabEntry footer) {
   }
 
   public void updateFakeEntity(TabEntry entry, boolean create) {
+    if (ViaUtils.getProtocolVersion(this.view.getViewer()) < 47) {
```
Pablete1234: See previous comment

src/main/java/tc/oc/pgm/PGMImpl.java
```diff
@@ -143,12 +160,13 @@
...
 import tc.oc.pgm.wool.WoolModule;
 import tc.oc.pgm.worldborder.WorldBorderModule;
+import tc.oc.tablist.TabManager;
 import tc.oc.util.SemanticVersion;
 
 public final class PGMImpl extends JavaPlugin implements PGM {
 
   private MatchManager matchManager;
-  private MatchTabManager matchTabManager;
+  private TabManager tabManager;
```
Pablete1234: I don't get why this was changed (and the subsequent changes) if you aren't even changing the MatchTabManager for 1.7/1.8+ players.

Meeples10: Sorry, I had meant to do something different than what I ended up doing and forgot to undo this.

src/main/java/tc/oc/tablist/TabRender.java
```diff
@@ -53,6 +54,9 @@ private BaseComponent getContent(TabEntry entry, int index) {
   }
 
   private void appendAddition(TabEntry entry, int index) {
+    if (ViaUtils.getProtocolVersion(this.view.getViewer()) < ViaUtils.VERSION_1_8) {
+      return;
+    }
```
Pablete1234: This check shouldn't be here, instead, this renderer shouldn't be called for 1.7 clients

src/main/java/tc/oc/tablist/TabRender.java
```diff
@@ -154,6 +158,9 @@ public void setHeaderFooter(TabEntry header, TabEntry footer) {
   }
 
   public void updateFakeEntity(TabEntry entry, boolean create) {
+    if (ViaUtils.getProtocolVersion(this.view.getViewer()) < ViaUtils.VERSION_1_8) {
+      return;
+    }
```
Pablete1234: Same here

src/main/java/tc/oc/tablist/TabView.java
```diff
@@ -365,4 +367,10 @@ protected void onRespawn(PlayerRespawnEvent event) {
   protected void onWorldChange(PlayerChangedWorldEvent event) {
     if (this.viewer == event.getPlayer()) this.respawnFakeEntities();
   }
+
+  public void onViewerJoinMatch(PlayerJoinMatchEvent event) {
```
Pablete1234: May i ask why this is now necessary?

src/main/java/tc/oc/tablist/TabView.java
```diff
@@ -16,7 +18,7 @@
  * bottom and cannot be seen. The fake team names all start with a '\u0001' character, so they will
  * always come before any real teams.
  */
-public class TabView {
+public class TabView implements Listener {
```
Pablete1234: I do not think this is or should be a listener

src/main/java/tc/oc/tablist/LegacyTabView.java
```diff
@@ -0,0 +1,85 @@
...
+
+  public void renderLayout(TabRender render) {}
+
+  public void renderContent(TabRender render) {}
+
+  public void markSlotsClean() {}
+
+  public void renderHeaderFooter(TabRender render, boolean force) {}
+
+  public void onViewerJoinMatch(PlayerJoinMatchEvent event) {}
```
Pablete1234: I think you only need to override the `render()` method to no-op here, everything else either isn't even used nor overriding (like `onViewerJoinMatch`) or you shouldn't be overriding anyways

src/main/java/tc/oc/pgm/tablist/MatchTabManager.java
```diff
@@ -94,8 +95,8 @@ public void run() {
   @Override
   public @Nullable TabView getView(Player viewer) {
     TabView view = super.getView(viewer);
-    if (view != null && !(view instanceof LegacyMatchTabView)) {
-      this.registerEvents((MatchTabView) view);
+    if (view instanceof ListeningTabView) {
```
Pablete1234:
```suggestion
    if (view != null && view instanceof ListeningTabView) {
```

src/main/java/tc/oc/pgm/tablist/MatchTabManager.java
```diff
@@ -151,8 +152,8 @@ public void onMatchUnload(MatchUnloadEvent event) {
   @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
   public void onJoinMatch(PlayerJoinMatchEvent event) {
     TabView view = this.getView(event.getPlayer().getBukkit());
-    if (view != null && !(view instanceof LegacyMatchTabView)) {
-      ((MatchTabView) view).onViewerJoinMatch(event);
+    if (view instanceof ListeningTabView) {
```
Pablete1234:
```suggestion
    if (view != null && view instanceof ListeningTabView) {
```

src/main/java/tc/oc/pgm/tablist/MatchTabManager.java
Meeples10: Is a null check really needed if there is an `instanceof` check? `instanceof` will return false if the object is null.

Pablete1234: true, why did it have it before?

PR #279, opened 2020-01-30
Map Vote Enhancement: Announce the winning map

### LINKED ISSUE #281: Currently playing doesnt show the authors of the map
https://i.badlion.net/LgmobAMxn9ECzdNUfxA6gR.png

applenick:
# Map Vote - Winner Announcement

Currently when a map poll comes to an end, the voting tallies are displayed then the server cycles to the next map. I believe it would make it easier for users, if the winning map were to be announced in a more prominent way.

I implemented two types of announcements via titles and the hotbar. Both of which use the same formatting so the messages are consistent. 

Edit:
Also resolves #281 

## Title:
title

## Hotbar:
hotbar

Players have the setting to toggle this feature between `title`, `hotbar` and `off`.

### Note: 
When naming the setting related to this feature, I was unsure whether to restrict it to map voting exclusively or keep it open for future improvements. One such idea was actually to make `announcements` relate to other already implemented title features, such as controlling for example when a match is about to start (3, 2, 1 countdown). So in that case, if the player has `announcements` toggled to hotbar, then instead of the countdown being sent via title, it would be in the hotbar instead. 

If anyone has feedback regarding the setting values or the formatting for announcements, it'd be appreciated 😄 

### DIFF
--- a/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -127,8 +127,7 @@ public class MapPoll {
         new PersonalizedText("" + countVotes(votes.get(map)), ChatColor.YELLOW),
         new PersonalizedText("] "),
         map.getStyledMapName(
-            winner ? MapNameStyle.HIGHLIGHT_WITH_AUTHORS : MapNameStyle.COLOR_WITH_AUTHORS,
-            winner ? ChatColor.GREEN : ChatColor.GOLD));
+            winner ? MapNameStyle.HIGHLIGHT_WITH_AUTHORS : MapNameStyle.COLOR_WITH_AUTHORS));
   }
 
   public void sendBook(MatchPlayer viewer) {
--- a/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -31,6 +31,7 @@ import tc.oc.pgm.api.map.MapInfo;
 import tc.oc.pgm.api.map.MapTag;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.util.components.Components;
 import tc.oc.world.NMSHacks;
 
 /** Represents a polling process, with a set of options. */
@@ -111,13 +112,9 @@ public class MapPoll {
     // Check if the winning map name's length suitable for the top title, otherwise subtitle
     boolean top = winner.getName().length() < TITLE_LENGTH_CUTOFF;
     Component mapName = new PersonalizedText(winner.getName()).bold(true).color(ChatColor.GREEN);
-    Component winText =
-        new PersonalizedTranslatable(top ? "poll.winner.sub" : "poll.winner.title")
-            .getPersonalizedText()
-            .bold(true)
-            .color(ChatColor.GOLD);
 
-    viewer.showTitle(top ? mapName : winText, top ? winText : mapName, 5, 60, 5);
+    viewer.showTitle(
+        top ? mapName : Components.blank(), top ? Components.blank() : mapName, 5, 60, 5);
   }
 
   private Component getMapChatComponent(MatchPlayer viewer, MapInfo map, boolean winner) {
--- a/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -111,7 +111,7 @@ public class MapPoll {
 
     // Check if the winning map name's length suitable for the top title, otherwise subtitle
     boolean top = winner.getName().length() < TITLE_LENGTH_CUTOFF;
-    Component mapName = new PersonalizedText(winner.getName()).bold(true).color(ChatColor.GREEN);
+    Component mapName = winner.getStyledMapName(MapNameStyle.COLOR).bold(true);
 
     viewer.showTitle(
         top ? mapName : Components.blank(), top ? Components.blank() : mapName, 5, 60, 5);
--- a/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -24,20 +24,21 @@ import org.bukkit.inventory.meta.BookMeta;
 import tc.oc.component.Component;
 import tc.oc.component.types.PersonalizedText;
 import tc.oc.component.types.PersonalizedTranslatable;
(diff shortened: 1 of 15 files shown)


AustinLMayes:
It just seems super quick to be in either place since it goes away on world change. Maybe if the vote end to cycle delay was longer?

Also titles are kinda risky because text size is widely affected by GUI scale and the text could be easily cut off - which isn't the best IMO.

Pablete1234:
@AustinLMayes Technically the cycle and the same time, i think since the recent refactors new match loading is a bit more async, but they happen at the same time.

As for ending the vote behorehand, that's kinda tricky. Right now a vote starts 5s after match end (match end schedules task for 5s later) and ends whenever pollNextMap is called (on cycle), or a  different map is set (like via /setnext). 
It doesn't have any way to know when the cycle is ending, or if you cancel the cycle, or if you have differing cycle times. It's agnostic to all of that, and i think it's better to stay that way.

@applenick 
Regarding this in particular, sure we could have the map name in the title or action bar, but i think the map message & the objective in chat are already good enough indicators of the map.

Either way if this is kept, the title has a long fade in and long fade out, but barely any display time, wich makes it a bit hard to read, should be extended a bit.

Also, like austin pointed out, displaying `Xi` or `Zoo` isn't the same as displaying `A journey through history`, `Bridge Over Troubled Water`, `Fairy Tales 2: A Tale or Two` or `Welcome To Wool Square`, they look fine on a "normal" gui scale, but on the automatic (default) scale, it looks like this:
image

What you may want to consider, is using the subtitle for the map name:
image

src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -225,4 +228,32 @@ private void updateScores() {
     if (voters == 0) return;
     votes.forEach((m, v) -> mapScores.put(m, Math.max(v.size() / voters, Double.MIN_VALUE)));
   }
+
+  private void announceMapWinner(Match match, MapInfo winner) {
```
Pablete1234: I would just merge this method & the `sendMessage` method, wich is already done per-player, just rename `sendMessage` to `announceMapWinner` with the player & picked map as params, and integrate this logic there, it's only called when the map has actually ended.

You could also make it so the picked map name is underlined or bolded in the chat msg

Electroid:
Great idea, there are some comments first, but let's not add another setting. This isn't a feature that needs a setting.

src/main/java/tc/oc/pgm/rotation/MapPoll.java
applenick: Awesome,
I was initially unsure whether it would be more appropriate to make this separate, however I will merge this into the sendMessage method. Thanks!

I’ll also make the formatting changes suggested 👍

applenick:
@AustinLMayes 
I understand how the time period may not be quick enough, however I’m going to play around with the title length a bit to get a period of time that feels long enough. 

@Pablete1234 
Thank you for all the suggestions. I’m going to experiment with the fade in/out on the title to get a look that feels longer for the player. I agree that having the map name in the sub-title would probably be best for now. We could also have a cut-off length for map names that appear in the title field before it’s in the sub-title. However, I’m unsure as they would introduce inconsistenty between each map. Though I’ll experiment and update everyone here on what the format looks like.

@Electroid I’ll remove the setting and most likely the hotbar announcement, as I feel the title is more prominent and cool 👍 

Thanks again to all of you for your feedback. I have limited time today, so an update will be a little later or tomorrow.

applenick:
Alright so back with an update. I took into account what was mentioned above and have updated the pull request with the following changes:

1. Removed the settings as @Electroid requested
2. Created a cutoff length (15) to decide whether a map name goes in the upper title or subtitle.
3. Made the text that goes along with the title shorter
4. Created a new `MapNameStyle` to be used separately from `NameStyle`, as it is solely for formatting map names.

The last change will most likely be the most controversial as I decided to move MapInfo away from using the `Named` interface. As discussed on Discord with @Pablete1234 , it doesn’t make sense for map names to use the same styling system as usernames. So hopefully my implementation is the type of idea they had in mind. However, feel free to correct me if my method was incorrect and I can gladly revert it back or implement a new method. However after testing, everything works accordingly.


Also last thing, since the styling of names were updated here this fixes #281 


Screenshots:

When a title length fits the requirement:
large

A title length that is over:
sub

New formatting of highlighted winner:
new-highlight

src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -116,7 +129,9 @@ private Component getMapChatComponent(MatchPlayer viewer, MapInfo map) {
         new PersonalizedText(" ").bold(!voted), // Fix 1px symbol diff
         new PersonalizedText("" + countVotes(votes.get(map)), ChatColor.YELLOW),
         new PersonalizedText("] "),
-        map.getStyledName(NameStyle.FANCY));
+        map.getStyledMapName(
+            winner ? MapNameStyle.HIGHLIGHT_WITH_AUTHORS : MapNameStyle.COLOR_WITH_AUTHORS,
+            winner ? ChatColor.GREEN : ChatColor.GOLD));
```
Pablete1234: Why not just append a `.underlined(winner)` (or bold)?

applenick: Originally I tried just modifying the map name as you suggested, however what I found was since the author list is included they are bolded/underlined as well. I feel that having just the map name highlighted is better visually. Also adjusting the color makes it more prominent too.

TheMolkaPL:
I think personally this might be confusing to new players to display map names in different positions. I think we should print it in subtitle, or depending on length, in title and subtitle but without `Next map` (keep it empty). So it should then show:

```
> Harb
>
```

and

```
>
> Race for Victory
```

applenick:
@TheMolkaPL Great suggestion! I’ve removed the additional text and when the titles are displayed it’s more consistent. 

If anyone wants a screenshot let me know, but it’s basically the same as my previous ones 🤷‍♂

Pablete1234:
Since map names are basically universally gold, i'd want to keep it consistent. /map, /maps, the announcer saying "Currently playing", all use gold color for the map name. That's also why i think making the winner map in chat appear bold or underlined is better, because you keep the consistency of gold coloring across the plugin

applenick:
@Pablete1234 
Ah I understand what you mean now. I’ll adjust the method to remove adjusting the chat color. You don’t mind if the title stays green though, or would you like that to be in gold too?

Pablete1234:
@applenick i think it makes more sense for title to be gold aswell

applenick:
Final screenshot of the formatting:
final-look

PR #283, opened 2020-01-31
Do not send 'You joined Observers' message when combat logging

Meeples10:
Resolves Electroid#282.

### DIFF
--- a/src/main/java/tc/oc/pgm/api/match/Match.java
+++ b/src/main/java/tc/oc/pgm/api/match/Match.java
@@ -393,4 +393,6 @@ public interface Match extends MatchPlayerResolver, MultiAudience, ModuleContext
    * @return If the {@link Match} was just ended.
    */
   boolean calculateVictory();
+
+  Listener getListener(Class<? extends Listener> listenerClass);
 }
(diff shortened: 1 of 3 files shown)


src/main/java/tc/oc/pgm/api/match/Match.java
```diff
@@ -393,4 +393,6 @@ default boolean finish() {
    * @return If the {@link Match} was just ended.
    */
   boolean calculateVictory();
+
+  Listener getListener(Class<? extends Listener> listenerClass);
```
Electroid: This doesn't seem like the right solution to this problem. Doesn't `CombatLogTracker` cancel the leave event or something?

Meeples10: It does, but the event is fired after `JoinMatchModule#leave(MatchPlayer)`.

Pablete1234: Code will need to be redesigned so that the event is triggered first, and if not cancelled, then the message is shown.

Doing this getListener logic is a complete hack

Meeples10: Should I close this PR and submit a new one for that or just add it to this one?

Pablete1234: You can just ammend the commit

Electroid: Yea, changing the event cancellation logic is the right way to go.

PR #284, opened 2020-01-31
Fix 'You joined Observers' message after event cancellation, resolves #282

### LINKED ISSUE #282: Typing /obs whenever in Danger, sends message that you joined obs
https://i.imgur.com/4UDM3k2.png

Meeples10:
Did not intend to close the previous pull request. @Electroid is this a suitable fix, or should I change how the event is fired/handled anyway?

### DIFF
--- a/src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
+++ b/src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
@@ -634,7 +634,7 @@ public class TeamMatchModule implements MatchModule, Listener, JoinHandler {
 
   @EventHandler(priority = EventPriority.MONITOR)
   public void onPartyChange(PlayerPartyChangeEvent event) {
-    if (event.getNewParty() instanceof Team) {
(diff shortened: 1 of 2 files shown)


src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
```diff
@@ -634,7 +634,7 @@ public boolean kickPlayerOffTeam(Team kickFrom, boolean forBalance) {
 
   @EventHandler(priority = EventPriority.MONITOR)
   public void onPartyChange(PlayerPartyChangeEvent event) {
-    if (event.getNewParty() instanceof Team) {
+    if (event.getOldParty() != null && event.getNewParty() != null) {
```
Electroid: Since this is `TeamMatchModule`, we will need to check if the `Party` is a `Team` and not a `Tribute`.

`event.getOldParty() instanceof Team && event.getNewParty() instanceof Team`

The reason this works is because `null instanceof <anything>` is always evaluated to be `false`.

PR #287, opened 2020-02-02
Implement player freezing, resolves #135

### LINKED ISSUE #135: Implement player freezing
Ubiquitous moderation feature. Non-PGM servers mostly don't need it (lobby) and PGM servers absolutely need it. Requires:

- [ ] Command: /f \<username> or /freeze
- [ ] Give ice block to players in observers with perms to pgm.freeze
- [ ] Freeze any players **right** clicked by ice block
- [ ] Don't freeze players with permission pgm.freeze.exempt
- [ ] When a player is frozen they can't move (like in the death screen)
- [ ] Display a title screen or chat message when frozen (depending on version) and play some dragon sound
- [ ] Send to adminchat: `[a] Eudaldca froze Cheater1234`
- [ ] Give freeze perms to pgm.mod
- [ ] Server configurable (enable/disable)

Would require to [port feature back from PGM](https://github.com/StratusNetwork/projectares/tree/master/PGM/src/main/java/tc/oc/pgm/freeze)

Meeples10:
Originally started by @ThatOneTqnk.

### DIFF
--- a/src/main/java/tc/oc/pgm/api/Permissions.java
+++ b/src/main/java/tc/oc/pgm/api/Permissions.java
@@ -31,7 +31,6 @@ public interface Permissions {
   String STAFF = ROOT + ".staff"; // Considered apart of the staff team
   String RELOAD = ROOT + ".reload"; // Reload the PGM configuration
   String FREEZE = ROOT + ".freeze"; // Can freeze players
-  String FREEZE_EXEMPT = ROOT + ".freeze.exempt"; // Cannot be frozen
 
   // Role-specific permission nodes
   Permission DEFAULT =
@@ -64,7 +63,6 @@ public interface Permissions {
               .put(JOIN_FORCE, true)
               .put(DEFUSE, true)
               .put(FREEZE, true)
-              .put(FREEZE_EXEMPT, true)
               .put(STAFF, true)
               .build());
 
--- a/src/main/java/tc/oc/pgm/api/Permissions.java
+++ b/src/main/java/tc/oc/pgm/api/Permissions.java
@@ -30,6 +30,8 @@ public interface Permissions {
   String DEBUG = ROOT + ".debug"; // Errors from map loading and debug commands
   String STAFF = ROOT + ".staff"; // Considered apart of the staff team
   String RELOAD = ROOT + ".reload"; // Reload the PGM configuration
+  String FREEZE = ROOT + ".freeze"; // Can freeze players
+  String FREEZE_EXEMPT = ROOT + ".freeze.exempt"; // Cannot be frozen
 
   // Role-specific permission nodes
   Permission DEFAULT =
@@ -61,6 +63,8 @@ public interface Permissions {
               .put(RESIZE, true)
               .put(JOIN_FORCE, true)
               .put(DEFUSE, true)
+              .put(FREEZE, true)
+              .put(FREEZE_EXEMPT, true)
               .put(STAFF, true)
               .build());
 
--- a/src/main/java/tc/oc/pgm/commands/AdminCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/AdminCommands.java
@@ -11,17 +11,14 @@ import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
-import org.bukkit.entity.Player;
 import org.joda.time.Duration;
 import tc.oc.pgm.AllTranslations;
-import tc.oc.pgm.Config;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.map.MapInfo;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchManager;
 import tc.oc.pgm.api.party.Competitor;
-import tc.oc.pgm.freeze.FreezeManager;
 import tc.oc.pgm.restart.CancelRestartEvent;
 import tc.oc.pgm.restart.RequestRestartEvent;
 import tc.oc.pgm.restart.RestartManager;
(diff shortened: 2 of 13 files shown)


src/main/java/tc/oc/pgm/api/Permissions.java
```diff
@@ -30,6 +30,8 @@
   String DEBUG = ROOT + ".debug"; // Errors from map loading and debug commands
   String STAFF = ROOT + ".staff"; // Considered apart of the staff team
   String RELOAD = ROOT + ".reload"; // Reload the PGM configuration
+  String FREEZE = ROOT + ".freeze"; // Can freeze players
+  String FREEZE_EXEMPT = ROOT + ".freeze.exempt"; // Cannot be frozen
```
Electroid: Don't really need this, just have the exempt be `STAFF`.

Electroid: If a player has the `FREEZE` permission, they can't be frozen.

src/main/java/tc/oc/pgm/freeze/FreezeManager.java
```diff
@@ -0,0 +1,124 @@
...
+import tc.oc.pgm.AllTranslations;
+import tc.oc.pgm.Config;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.chat.Sound;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.match.ObservingParty;
+
+/** Controls player freezing and those in freeze mode */
+public class FreezeManager {
```
Electroid: No need for a `FreezeManager`. I would rather implement a `MatchPlayer#setFrozen(boolean)` method.

src/main/java/tc/oc/pgm/freeze/FreezeManager.java
```diff
@@ -0,0 +1,124 @@
...
+
+  public static FreezeManager get() {
+    if (freezeManager == null) freezeManager = new FreezeManager();
+    return freezeManager;
+  }
+
+  private Map<Player, Boolean> freezeModeMap = new HashMap<>();
+  private Set<Player> frozenPlayers = new HashSet<>();
+
+  public void assignFreezeItem(Player freezer) {
```
Electroid: There's some class called ObserverToolFactory or something, this should go there.

src/main/java/tc/oc/pgm/api/Permissions.java
EudaldC: We might not want this, e.g. referees cannot be frozen but might not want them showing up in /staff or otherwise having the pgm.staff permissions

Electroid: So then it would make sense that if a player has the `FREEZE` permission, they can't be frozen? seems reasonable to me.

EudaldC: not necessarily, we wouldn't want map developers having permission to freeze but we definitely want them to be excempt

src/main/java/tc/oc/pgm/freeze/FreezeManager.java
Pablete1234: setFrozen already exists, but it's for player movement prevention (like when you die), not for punishment-wise freezing

Electroid: Wait, then why can't we use that code?

src/main/java/tc/oc/pgm/api/Permissions.java
Meeples10: So what is the final decision on permissions?

src/main/java/tc/oc/pgm/freeze/FreezeManager.java
Pablete1234: @Electroid well, they should be using it. That is the "low-level" way to prevent the player from moving, then everyting else that goes ontop knowing who's frozen, being able to unfreeze, that stuff, needs to be handled outisde. Just because someone died doesn't mean you can run /unfreeze on that player

src/main/java/tc/oc/pgm/api/Permissions.java
Electroid: It's hard to justify having a special permission *just* to be exempt. If you can freeze someone, makes sense you can't be frozen yourself (since you could just unfreeze yourself).

So let's just exempt those with the `FREEZE` permission.

Electroid: Keep in mind `Permissions` was designed to be really simple, that's why there isn't a permission node for every nook and cranny.

src/main/java/tc/oc/pgm/freeze/FreezeManager.java
Electroid: What I'm trying to understand is why is `FreezeListener` required if we already have a freezing mechanism with `isFrozen`? (Aside from the title and UI stuff that says you're frozen).

src/main/java/tc/oc/pgm/spawns/ObserverToolFactory.java
```diff
@@ -15,6 +15,7 @@
 public class ObserverToolFactory {
 
   public static final String EDIT_WAND_PERMISSION = "worldedit.wand";
+  public static final String PLAYER_FREEZER_PERMISSION = "pgm.freeze";
```
Electroid: No need for another static field. Just reference `Permissions.FREEZE`

src/main/java/tc/oc/pgm/freeze/FreezeManager.java
Pablete1234: `isFrozen` is just the player state of not being able to move, that doesn't know why that player was put on frozen mode.

If you would use that as the punishment flag, you could run /unfreeze on dead people and pgm would be like "Unfrozen player", when in reality, that player wasn't frozen (the punishment-freeze).

As to why the listener exists, is because it still has to prevent you from throwing items, or placing blocks, that `isFrozen` doesn't handle.
 When you're dead what handles no-block-placing is the fact that you are on the `Spawning` state, not the fact that you're frozen. When you're frozen in-game, it has to prevent block placing even if your player state is `Alive`.

src/main/java/tc/oc/pgm/commands/AdminCommands.java
```diff
@@ -148,6 +151,31 @@ public void pgm(CommandSender sender) {
...
+        sender.sendMessage(
+            ChatColor.RED + AllTranslations.get().translate("command.freeze.mustBePlayer", sender));
+        return;
+      }
+      freezeManager.setFreezeMode(playerSender, !freezeManager.getFreezeMode(playerSender));
+    } else {
+      freezeManager.toggleFreeze(playerSender, target);
+    }
+  }
+
```
Pablete1234: This should be in `ModerationCommands`

Electroid:
Going to close this due to scope creep, but feel free to implement this on a fork.

KingOfSquares:
Hey so since this is a part of the milestone API v1, maybe reopen it?

applenick:
@KingOfSquares I have a new PR with several moderation improvements, including re-adding of the /freeze command/tool, that I’ll be submitting tonight or tomorrow.

KingOfSquares:
> @Pablete1234 & @KingOfSquares I have a new PR with several moderation improvements, including re-adding of the /freeze command/tool, that I’ll be submitting in a day or so.

@Meeples10

Meeples10:
Closing this in favor of @applenick's PR.

PR #297, opened 2020-02-08
Prevent wearing the team selector helmet

### LINKED ISSUE #286: Clicking Team Selector, wears the helmet
https://i.gyazo.com/44f8fe6018503fc670d8457f1cfb4d54.mp4

Meeples10:

Resolves #286.

### DIFF
--- a/src/main/java/tc/oc/pgm/spawns/states/Observing.java
+++ b/src/main/java/tc/oc/pgm/spawns/states/Observing.java
@@ -2,9 +2,11 @@ package tc.oc.pgm.spawns.states;
 
 import java.util.List;
 import org.bukkit.Location;
+import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.event.Event;
 import org.bukkit.event.entity.EntityDamageEvent;
+import org.bukkit.event.player.PlayerInteractEvent;
 import org.bukkit.event.player.PlayerRespawnEvent;
 import org.bukkit.permissions.PermissionAttachment;
 import tc.oc.block.BlockVectors;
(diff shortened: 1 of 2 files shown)


Pablete1234:
I wonder, why are you adding a new listener to handle this, instead of cancelling the event in the same place where event gets handled?

Meeples10:
> 
> 
> I wonder, why are you adding a new listener to handle this, instead of cancelling the event in the same place where event gets handled?

If it were handled there, it would also prevent players from wearing leather helmets if they are currently participating in a match. The current way only prevents it if the player is an observer.

src/main/java/tc/oc/pgm/spawns/states/Observing.java
```diff
@@ -109,4 +111,13 @@ public void onEvent(EntityDamageEvent event) {
     super.onEvent(event);
     event.setCancelled(true);
   }
+
+  @Override
+  public void onEvent(PlayerInteractEvent event) {
+    super.onEvent(event);
+    if (event.getItem().getType() == Material.LEATHER_HELMET) {
```
Electroid: Is there a static variable for the helmet join hat? Because this will prevent *any* leather hat.

src/main/java/tc/oc/pgm/spawns/states/Observing.java
```diff
@@ -109,4 +111,13 @@ public void onEvent(EntityDamageEvent event) {
     super.onEvent(event);
     event.setCancelled(true);
   }
+
+  @Override
+  public void onEvent(PlayerInteractEvent event) {
+    super.onEvent(event);
+    if (PickerMatchModule.validatePickerItem(event.getPlayer().getItemInHand())) {
```
Electroid: Seems weird for `PickerMatchModule` to expose this implementation detail. How about add a `ObserverInteractEvent` listener to `PickerMatchModule` and cancel the event.

Electroid:
Closing due to inactivity, welcome to reconsider

PR #299, opened 2020-02-08
Sort map vote ending message by vote count

AustinLMayes:
Reduces the number of (potentially) intensive vote count checks to 1, and sorts the totals message by number of votes.

### DIFF
--- a/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.Set;
 import java.util.TreeMap;
@@ -46,6 +47,9 @@ public class MapPoll {
   private final int voteSize;
 
   private final Map<MapInfo, Set<UUID>> votes = new HashMap<>();
+  // Maps populated during finishVote()
+  private final Map<MapInfo, Integer> finalTotals = new HashMap<>();
+  private final List<MapInfo> finalOrder = new ArrayList<>();
 
   MapPoll(Match match, Map<MapInfo, Double> mapScores, int voteSize) {
     this.match = new WeakReference<>(match);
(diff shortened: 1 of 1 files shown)


Pablete1234:
i don't think re-ordering maps is a good idea. Keeping the order consistent thru all the voti g process is good

src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -222,7 +218,16 @@ private int countVotes(Set<UUID> uuids) {
...
   MapInfo finishVote() {
-    MapInfo picked = getMostVotedMap();
+    // Cache vote totals, so we don't have to count for each compare
+    for (Entry<MapInfo, Set<UUID>> map : votes.entrySet()) {
+      finalTotals.put(map.getKey(), countVotes(map.getValue()));
+    }
+    // Populate final order
+    finalTotals.entrySet().stream()
+        .sorted(Comparator.comparingInt((e) -> ((Entry<MapInfo, Integer>) e).getValue()).reversed())
+        .forEachOrdered(e -> finalOrder.add(e.getKey()));
```
Pablete1234: I'd rather not change the order of maps
Also, if you're using a stream, don't use a terminal operation that mutates a pre-made collection, just collect it to a list and assign it

src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -222,7 +218,16 @@ private int countVotes(Set<UUID> uuids) {
    * @return The picked map to play after the vote
    */
   MapInfo finishVote() {
-    MapInfo picked = getMostVotedMap();
+    // Cache vote totals, so we don't have to count for each compare
+    for (Entry<MapInfo, Set<UUID>> map : votes.entrySet()) {
+      finalTotals.put(map.getKey(), countVotes(map.getValue()));
```
Pablete1234: Instead of making a new map for the cache, change the Set<UUID> to a specific collection that keeps track of double-votes and caches the vote count.

src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -193,14 +197,6 @@ public boolean toggleVote(MapInfo vote, UUID player) throws CommandException {
     return false;
   }
 
-  /** @return The map currently winning the vote, null if no vote is running. */
-  private MapInfo getMostVotedMap() {
```
Pablete1234: Don't get rid of this method, see comment about specific collection that caches vote count

Electroid:
What's the status here?

Pablete1234:
@Electroid It's trying to fix something that works as intended, and making an optimization that shouldn't be done in that way but rather a more solid approach if he wants to.
IMO can be closed and make the optimization in another PR if relevant.

AustinLMayes:
Will leave improvements to a later pull request - the whole class needs a bit of a rewrite.

PR #300, opened 2020-02-08
WIP: Add moderation commands

### LINKED ISSUE #156: Basic moderation commands
I think to give a better out-of-the-box experience, it is nice that PGM implements some basic moderation commands that all use similar & consistent formatting. 
Moderation should use bukkit's methods and also provide a pgm moderation event for APIs to be able to integrate.
A config value that enables/disables mod commands is also relevant, since if you want to use a completely different moderating system having overriding commands would be an annoyance.

- [x] Ban
- [x] Temp ban
- [x] Kick
- [x] Warn
- [x] Mute
- [x] Freeze - See #135 
- [x] Report
- [x] Staff
- [x] Vanish
 - Hide from tab list
 - Force you on obs (and force you to stay on obs)
 - Make player not visible to any other obs
 - Disable your name being tabbed-in on commands
 - Hide from /staff
 - Don't allow others to /msg to you (unless bypass permission)
 - Don't allow you to type in team or global chat
 - Do allow admin chat & /msg to others with bypass
 - Don't count towards /match observer count
 - Display join/leave messages when you vanish/unvanish

applenick:
# Add Moderation Commands

I'm back with another feature set, this time tackling moderation commands 🎉 

This resolves #156 when complete.

## Commands:
What has been completed and status of other things:

**/staff: Fully implemented**
When no staff are online:
staff-offline

When staff are online the count will be shown, and a list of names next to it:
staff-online

**/warn: Fully implemented**
new-warning

**/kick: Fully implemented**
kick-example

**/mute: Fully implemented**
muted

**/ban: Fully implemented**
ban-example

**/tempban: Fully implemented**
temp

Extra Screenshots:
Example of when a punishment is broadcasted, in this case a warning.
Though the general format is [Punishment Type] {staff} > {target player} > {reason}
broadcast

NOTE: The screenshots here display the server title as OCC’s, however the config offers an option to specify the server name as well as whether to display a link to a rules like or not.

## Updated Note:
I believe there is currently an ongoing debate on whether this feature should be added directly into PGM or not. As the author of this PR it’s easy to see my personal position on the subject. However, if it’s decided against I can always see about making a separate plugin specifically for OCC to implement these 🤷‍♂ 

### DIFF
--- a/src/main/java/tc/oc/pgm/api/player/MatchPlayer.java
+++ b/src/main/java/tc/oc/pgm/api/player/MatchPlayer.java
@@ -139,14 +139,6 @@ public interface MatchPlayer extends Audience, Named, Tickable, InventoryHolder 
   boolean isFrozen();
 
   /**
-   * Get whether the {@link MatchPlayer} us currently muted. Muted players can not send commands to
-   * global or team chat.
-   *
-   * @return Whether the {@link MatchPlayer} is muted.
-   */
-  boolean isMuted();
-
-  /**
    * Get whether the {@link MatchPlayer} can interact with things in the {@link Match}.
    *
    * @return Whether the {@link MatchPlayer} can interact.
@@ -209,13 +201,6 @@ public interface MatchPlayer extends Audience, Named, Tickable, InventoryHolder 
    * @param gameMode - The gamemode to set
    */
   void setGameMode(GameMode gameMode);
-
-  /**
-   * Set whether the {@link MatchPlayer} is muted or not.
-   *
-   * @param mute Whether to mute.
-   */
-  void setMuted(boolean mute);
 
   /**
    * Get the protocol version of the {@link MatchPlayer}'s client
--- a/src/main/java/tc/oc/pgm/api/player/MatchPlayer.java
+++ b/src/main/java/tc/oc/pgm/api/player/MatchPlayer.java
@@ -139,6 +139,14 @@ public interface MatchPlayer extends Audience, Named, Tickable, InventoryHolder 
   boolean isFrozen();
 
   /**
+   * Get whether the {@link MatchPlayer} us currently muted. Muted players can not send commands to
+   * global or team chat.
+   *
+   * @return Whether the {@link MatchPlayer} is muted.
+   */
+  boolean isMuted();
+
+  /**
    * Get whether the {@link MatchPlayer} can interact with things in the {@link Match}.
    *
    * @return Whether the {@link MatchPlayer} can interact.
@@ -201,6 +209,13 @@ public interface MatchPlayer extends Audience, Named, Tickable, InventoryHolder 
    * @param gameMode - The gamemode to set
    */
   void setGameMode(GameMode gameMode);
+
+  /**
+   * Set whether the {@link MatchPlayer} is muted or not.
+   *
+   * @param mute Whether to mute.
+   */
+  void setMuted(boolean mute);
 
   /**
    * Get the protocol version of the {@link MatchPlayer}'s client
--- a/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
@@ -2,28 +2,50 @@ package tc.oc.pgm.commands;
 
 import app.ashcon.intake.Command;
 import app.ashcon.intake.CommandException;
+import app.ashcon.intake.parametric.annotation.Switch;
 import app.ashcon.intake.parametric.annotation.Text;
+import com.google.common.base.Strings;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-import java.time.Duration;
+import com.google.common.collect.Lists;
 import java.time.Instant;
+import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import net.md_5.bungee.api.ChatColor;
+import net.md_5.bungee.api.chat.BaseComponent;
 import org.bukkit.Bukkit;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
+import org.joda.time.Duration;
 import tc.oc.component.Component;
 import tc.oc.component.types.PersonalizedText;
 import tc.oc.component.types.PersonalizedTranslatable;
 import tc.oc.named.NameStyle;
+import tc.oc.pgm.AllTranslations;
+import tc.oc.pgm.Config;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.chat.Audience;
+import tc.oc.pgm.api.chat.Sound;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.events.PlayerReportEvent;
+import tc.oc.util.components.ComponentUtils;
+import tc.oc.util.components.Components;
+import tc.oc.util.components.PeriodFormats;
 
 public class ModerationCommands {
+
+  private static final Sound WARN_SOUND = new Sound("mob.enderdragon.growl", 1f, 1f);
+  private static final Component BROADCAST_DIV = new PersonalizedText(" » ").color(ChatColor.GOLD);
+
+  private static final Component CONSOLE_NAME =
+      new PersonalizedTranslatable("console.displayName")
+          .getPersonalizedText()
+          .color(ChatColor.YELLOW)
+          .italic(true);
 
   private static final int REPORT_COOLDOWN_SECONDS = 15;
 
@@ -45,7 +67,7 @@ public class ModerationCommands {
       // Check for cooldown
       Instant lastReport = LAST_REPORT_SENT.getIfPresent(matchPlayer.getId());
       if (lastReport != null) {
-        Duration timeSinceReport = Duration.between(lastReport, Instant.now());
+        java.time.Duration timeSinceReport = java.time.Duration.between(lastReport, Instant.now());
         long secondsRemaining = REPORT_COOLDOWN_SECONDS - timeSinceReport.getSeconds();
         if (secondsRemaining > 0) {
           Component secondsComponent = new PersonalizedText(Long.toString(secondsRemaining));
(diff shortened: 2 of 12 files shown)


Pablete1234:
The format for that warn looks pretty out of place, could you copy back the one from ocn?

PGM doesn't need to (generally) perssist the punishments. It has to throw moderation events and if another plugin wishes to perssist them they should be able to.
For bans and tempbans use bukkit/vanilla ways to ban them, for kicks/warns no need to store them, for mutes we will probably have to perssist if a player is muted in the datastore (but not the mute itself)

Lookup or /l IMO, is out of scope, since we're not storing punishments.
I don't think each command needs a config. If you don'g want it used don't give the permission to use it to people. (pgm.moderation.*)

applenick:
Hey @Pablete1234,
Thanks for the feedback, I really appreciate it!

I’ll adjust the /warn formatting, I thought it would be a good idea to make warnings stand out more. However yeah it is not really consistent with the other punishment formatting styles. Is there anything particular about my style that you dislike? Such as is it the magic headings or the text in the middle that’s troubling you? Say if the heading were to use the regular horizontalLineHeading method and the text inside was just replaced with the reason, would that be better? Or should the format be more simple like `⚠ WARNING ⚠ {Reason}`

I understand the decision to not persist punishments and agree with it. I’ll add the moderation events next. However I’m curious how `/tempban` should be handled, as there is no bukkit implementation of bans that expire as far as I’m aware. Should I simply remove the temp ban command and leave it for other plugins to handle removing bans after a period of time? 

Also thanks for the feedback on the two additional questions, of course `/l` makes sense to not implement, and great idea regarding the permissions. 

Thanks for all your help!

src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -243,15 +245,24 @@ public void send(
...
-                new PersonalizedText(message.trim())));
-    match.getPlayers().stream().filter(filter).forEach(player -> player.sendMessage(component));
-    Audience.get(Bukkit.getConsoleSender()).sendMessage(component);
+    if (!sender.isMuted()) {
+      final Component component =
+          new PersonalizedText(
+              Components.format(
+                  format,
+                  sender == null
+                      ? new PersonalizedText("Console", ChatColor.AQUA, ChatColor.ITALIC)
```
TheMolkaPL: You can use this: https://github.com/Electroid/PGM/blob/6bc2e673cbb1d180adaa4b4e25d32d6f2717fe6c/src/main/i18n/templates/strings.properties#L33

Brottweiler:
Here's OCN's warn message in chat for reference. I don't know how the Title looked when getting warned, though. I wonder if the Title also had obfuscated characters...

image

src/main/java/tc/oc/pgm/api/player/MatchPlayer.java
```diff
@@ -138,6 +138,14 @@
    */
   boolean isFrozen();
 
+  /**
+   * Get whether the {@link MatchPlayer} us currently muted. Muted players can not send commands to
+   * global or team chat.
+   *
+   * @return Whether the {@link MatchPlayer} is muted.
+   */
+  boolean isMuted();
```
Electroid: Not sure I want to bring this state into `MatchPlayer`. I would rather `ChatDispatcher` store this information in a `Set<UUID> muted`.

src/main/resources/config.yml
```diff
@@ -109,3 +109,17 @@ sidebar:
...
+
+moderation:
+  # If the moderation commands/enforcement should be used
+  enabled: true
+  
+  # Link to a rules website to display on kick screen
+  rules-link: ''
+  
+  # If the rules link is visible on kick screen
+  rules-link-visible: true
```
Electroid: No need for this, just check if `rules-link` is defined.

src/main/java/tc/oc/pgm/api/player/MatchPlayer.java
applenick: Alrighty! Will make that change asap.

src/main/resources/config.yml
applenick: Ah good point. Will adjust

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -2,29 +2,51 @@
...
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import net.md_5.bungee.api.ChatColor;
+import net.md_5.bungee.api.chat.BaseComponent;
 import org.bukkit.Bukkit;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
+import org.joda.time.Duration;
```
Pablete1234: Don't use joda time, use java time

applenick: Question regarding joda. It seems that [PeriodFormats](https://github.com/Electroid/PGM/blob/master/src/main/java/tc/oc/util/components/PeriodFormats.java) is using it and I’d like to utilize one of its methods for displaying how much time until a temp ban expires. Any advice?

Pablete1234: May be out of scope for this PR, but on 1.9+ pgm all the times had been migrated away from joda to java 8  time

applenick:
Quick update regarding the warning format. After taking time to go through the ProjectAres source, I discovered that the later versions of PGM did not actually send a chat message when a warning was given. Rather it was only a title. So I found this to be the ideal solution as it’s simple and more noticeable. If anyone feels that including a chat message too would be beneficial, let me know. Though with both the title and broadcast I believe it’s easy to understand.

Also I started with the initial work on moderation events. I wasn’t sure whether to make individual sub-events for each type or not. So I went with two types, `PlayerPunishmentEvent` and `PlayerTimedPunishmentEvent`, the first is mainly used for broadcasts atm but an outside plugin would be able to easily utilize it and would be able to retrieve all the necessary data about the punishment. The Timed event extends the prior event and adds logic related to getting when a punishment expires.

Fun fact I was unaware of, so I was able to implement the TempBan without having to store any thing in the Datastore. As the Bukkit method of adding a ban to the server allows for an expiry date to be specified. Who knew xD

Screenshot of how it’s now implemented:
new-warning

Pablete1234:
> As the Bukkit method of adding a ban to the server allows for an expiry date to be specified. Who knew xD

I was like certain bukkit had implemented such a thing

Anyways, nice work, the warning is fine with that formatting & on the title.
The only thing we would have to perssist then, is if a player has been muted.

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
Pablete1234: You should still make the import be java time in new code tho, if a specific place needs joda time make that the fully qualified name. Currently you're just using joda all arround (even on the event) wich you really shouldn't

applenick:
@Pablete1234 Thanks!
In regards to `/mute` I don’t really think it’s 100% necessary to persist mutes any more than they are now. Say a troublesome player is muted, will there be any reason to keep them muted for a long terms of time between server restarts? If a player continues to spam or etc, then other punishments could always be issued. Though there could be scenarios I’m not thinking of. Any thoughts on not persisting mutes long term?

Pablete1234:
Tbh, that may be a good enough solution, just have PGM do mutes until server restart unless another plugin perssists them longer

applenick:
Awesome, and of course another plugin can always listen for the moderation event to persist them longer. 

@Pablete1234 Do you have any other recommendations for this PR? I understand there has been some debate on whether this should be added or not on Discord. Being the author of this I’ve refrained from commenting due simply to bias. Until that has been settled anything else that can be improved upon? I’ve decided against implementing `/vanish` for now, in order to keep this PR simple.

src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -162,6 +186,35 @@ public void sendReply(Match match, Audience audience, MatchPlayer sender, @Text
     sendDirect(match, sender, receiver.getBukkit(), message);
   }
 
+  @Command(
+      aliases = {"unmute", "um"},
+      usage = "<player>",
+      desc = "Unmute a player",
+      perms = Permissions.MUTE)
```
Pablete1234: command to unmute in chat dispatcher? Should be in ModerationCommands

src/main/java/tc/oc/pgm/commands/ModerationCommands.java
```diff
@@ -108,4 +130,298 @@ public static void report(
...
+      @Switch('w') boolean warn) {
+    MatchPlayer targetMatchPlayer = match.getPlayer(target);
+
+    // if -w, also warn the player but don't broadcast warning
+    if (warn) {
+      warn(sender, target, match, reason, true);
+    }
+
+    // Call Event for general punishments
+    punish(PunishmentType.MUTE, targetMatchPlayer, sender, reason, silent);
```
Pablete1234: you should, like the kick command, check if event is cancelled and tell the mod it was cancelled.

src/main/java/tc/oc/pgm/listeners/ModerationListener.java
```diff
@@ -0,0 +1,97 @@
...
+          event.getType(),
+          event.getPlayer().getMatch(),
+          event.getSender(),
+          event.getPlayer(),
+          event.getReason(),
+          event.isSilent());
+    }
+  }
+
+  @EventHandler(priority = EventPriority.HIGHEST)
```
Pablete1234:
```suggestion
  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
```

src/main/java/tc/oc/pgm/listeners/ModerationListener.java
```diff
@@ -0,0 +1,97 @@
...
+  public void onBan(PlayerPunishmentEvent event) {
+    if (!event.isCancelled()) {
+      if (event.getType().equals(PunishmentType.BAN)) {
+        banPlayer(event, null);
+      }
+    }
+  }
+
+  @EventHandler(priority = EventPriority.HIGHEST)
```
Pablete1234: Same as above

src/main/java/tc/oc/pgm/listeners/ModerationListener.java
```diff
@@ -0,0 +1,97 @@
...
+import tc.oc.named.NameStyle;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.commands.ModerationCommands;
+import tc.oc.pgm.commands.ModerationCommands.PunishmentType;
+import tc.oc.pgm.events.PlayerPunishmentEvent;
+import tc.oc.pgm.events.PlayerTimedPunishmentEvent;
+import tc.oc.util.components.Components;
+
+public class ModerationListener implements Listener {
```
Pablete1234: I am generally not sure if this should be handled in a listener, or if it should just be directly handled in the ModerationCommands, but you have to pick one and commit to it.

 - If you want to kill this class, just have the commands send the event and if not cancelled, apply the punishment & broadcast (All on ModerationCommands).

 - If you want to keep this class, don't perform the player kick on the moderation commands, just launch the event there, and if cancelled return a message to the user. Apply the punishment & broadcast it here, if event isn't cancelled.

Electroid:
Hey @applenick, thanks for putting this PR together. After some reflection, I do think this is out-of-scope for PGM.

Is it nice? Yes. Is it relevant to PGM's core purpose? No.

Moderation plugins are basically *the* most popular type of plugins out there, so server owners have a plethora of options. Only muting needs to be fixed, which can be done with #304. 

Apologies for the lag in decision time, as the project grows, decisions like these help educate future decisions down the road.

applenick:
@Electroid I understand your decision and how you want to keep new features within the scope of PGM. Perhaps the contributor community, with your guidance, can determine a more concrete roadmap as to what should be worked on for PGM. Personally I enjoy working on “new” features for PGM, and would rather leave refactoring/bug fixes/larger projects to those more familiar with the codebase. 

@Pablete1234 Thank you for all the suggestions! I do appreciate the help, and thank you for your guidance. 😄

PR #301, opened 2020-02-08
Fire more transform events

AustinLMayes:
Event called by structure grows (Sapling -> Tree), (Mushroom -> Huge Mushroom), naturally or using bonemeal.

https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/world/StructureGrowEvent.html

Also fires transform events for two-high blocka

### DIFF
--- a/src/main/java/tc/oc/pgm/modules/MaxBuildHeightMatchModule.java
+++ b/src/main/java/tc/oc/pgm/modules/MaxBuildHeightMatchModule.java
@@ -23,8 +23,7 @@ public class MaxBuildHeightMatchModule implements MatchModule, Listener {
   @EventHandler(ignoreCancelled = true)
   public void checkBuildHeight(BlockTransformEvent event) {
     if (event.getNewState().getType() != Material.AIR) {
-      if (event.getNewState().getY() + getExtraHeight(event.getNewState().getMaterial())
-          >= this.buildHeight) {
+      if (event.getNewState().getY() >= this.buildHeight) {
         event.setCancelled(
             true,
             new PersonalizedTranslatable(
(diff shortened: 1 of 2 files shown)


src/main/java/tc/oc/pgm/modules/MaxBuildHeightMatchModule.java
```diff
@@ -33,4 +34,10 @@ public void checkBuildHeight(BlockTransformEvent event) {
       }
     }
   }
+
+  private int getExtraHeight(Material material) {
+    if (material.name().contains("DOOR") && !material.name().contains("TRAP")) return 1;
```
Electroid: Can we put this in a util? Also, `BlockTransformEvent` is a very, *very* hot part of the code: Can we optimize this with enums and switches? 

I would even check have two checks, so this will be short-circuited.
```java
if (event.getNewState().getY() < this.buildHeight - 1) return;
if (event.getNewState().getY()
    + getExtraHeight(event.getNewState().getMaterial()) < this.buildHeight) return;
// setCancelled

Comment (Electroid) [reply src/main/java/tc/oc/pgm/modules/MaxBuildHeightMatchModule.java]: Also, this isn't just an issue with `MaxBuildHeightMatchModule`, right? The bug will still happen for `RegionMatchModule`, can we fix it for both?

Comment (AustinLMayes) [reply src/main/java/tc/oc/pgm/modules/MaxBuildHeightMatchModule.java]: Code changed in latest commit, re-review

File: src/main/java/tc/oc/pgm/modules/MaxBuildHeightMatchModule.java [hunk 2]
```diff
@@ -33,4 +34,23 @@ public void checkBuildHeight(BlockTransformEvent event) {
       }
     }
   }
+
+  private int getExtraHeight(Material material) {
```
TheMolkaPL: `getMaterialHeight` maybe? Should return 1 or 2.

AustinLMayes: This is old code, review the latest commit only

PR #305, opened 2020-02-10
Add admin chat symbol and more sounds

applenick:
# Add admin chat symbol and more sounds

### Admin Chat Symbol
Earlier today there was a suggestion by @Brottweiler to add `$` as a symbol shortcut for sending a message in the admin chat. I don’t believe there is any harm in having this as a shortcut, and with some testing I found it really is easy to use 😄 

screenshot

### Sounds
Also added are two new sound effects (if sounds are enabled), both staff specific but should be a nice QOL feature to improve the experience. First, utilizing the same sound as private messages, will be when an admin chat message is received. Personally I am a fan of chat sounds, and feel they help to make identifying important information easier. Adding onto that is a sound that will play when a report is received, basically the same reason as the admin chat one.

### Notes:
I understand not all staff member may enjoy having every single admin chat message play a sound, however there is the sound setting so that should suffice. In addition to that, I'm now thinking it may be better to have  a different sound than the DM one, so if there is feedback regarding that let me know. 

### DIFF
--- a/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
+++ b/src/main/java/tc/oc/pgm/commands/ModerationCommands.java
@@ -19,8 +19,11 @@ import tc.oc.component.types.PersonalizedTranslatable;
 import tc.oc.named.NameStyle;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.chat.Audience;
+import tc.oc.pgm.api.chat.Sound;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.setting.SettingKey;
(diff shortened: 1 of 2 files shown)


Pablete1234:
IMO, we should have pgm support multi-option settings. The sound setting shouldn't be on/off, it should be a set of flags for enabled/disabled sounds, each able to be triggered individually. But that's going to require quite a refactor.

Electroid:
Then we're going to end up with way to many settings.. I'm ok with the current changes.

Pablete1234:
@Electroid Well the idea wouldn't be to have more settings, but rather for the Sound setting to be "flags".

There's just one sound setting, but the value for it can be: `ADMIN_CHAT,OBJECTIVES,PRIVATE_MESSAGES`, wich makes it so you hear those 3 sounds but not any others.
Regardless, that's a refactor that should go elsewhere

PR #315, opened 2020-02-12
Add setting to determine if map vote book is displayed

AustinLMayes:
Some people just don't want to vote :)

### DIFF
--- a/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -31,7 +31,8 @@ public enum SettingKey {
     }
   }, // Changes if observers are visible
   SOUNDS("sounds", SOUNDS_ON, SOUNDS_OFF), // Changes if sounds are played
-  ;
+  VOTE_BOOK("vote-book", VOTES_ON, VOTES_OFF) // Changes if the vote book is shown on cycle
+;
 
   private final List<String> aliases;
   private final SettingValue[] values;
--- a/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -31,8 +31,8 @@ public enum SettingKey {
     }
   }, // Changes if observers are visible
   SOUNDS("sounds", SOUNDS_ON, SOUNDS_OFF), // Changes if sounds are played
-  VOTE_BOOK("vote-book", VOTES_ON, VOTES_OFF) // Changes if the vote book is shown on cycle
-;
+  VOTE("vote", VOTE_ON, VOTE_OFF), // Changes if the vote book is shown on cycle
+  ;
 
   private final List<String> aliases;
   private final SettingValue[] values;
--- a/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
+++ b/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
@@ -37,8 +37,8 @@ public enum SettingValue {
   SOUNDS_ON("sounds", "on"), // Always play sounds
   SOUNDS_OFF("sounds", "off"), // Never play sounds
 
-  VOTES_ON("vote_book", "on"), // Show the vote book on cycle
-  VOTES_OFF("vote_book", "off"), // Don't show the vote book on cycle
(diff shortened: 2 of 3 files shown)


Pablete1234:
The book should always be sent to the player. What you can make a setting, is if opens or not.
Move the setting-check a few lines lower

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -31,7 +31,8 @@ public void update(MatchPlayer player) {
     }
   }, // Changes if observers are visible
   SOUNDS("sounds", SOUNDS_ON, SOUNDS_OFF), // Changes if sounds are played
-  ;
+  VOTE_BOOK("vote-book", VOTES_ON, VOTES_OFF) // Changes if the vote book is shown on cycle
+;
```
Pablete1234: Should probably leave this line unmodified

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -36,6 +36,9 @@
 
   SOUNDS_ON("sounds", "on"), // Always play sounds
   SOUNDS_OFF("sounds", "off"), // Never play sounds
+
+  VOTES_ON("vote-book", "on"), // Show the vote book on cycle
+  VOTES_OFF("vote-book", "off"), // Don't show the vote book on cycle
```
TheMolkaPL: Must be `vote_book`, look on line 58 in this class.

TheMolkaPL: https://github.com/Electroid/PGM/pull/315/files/d1cf7a0a2ff55e979f6ed0edfaec034eb368bbe3#diff-36e7b49e5ff5568811f0ef2f27fef7c1R58

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
AustinLMayes: Auto-formatter did this, checks won't pass unless it stays (I think)

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
AustinLMayes: Good catch

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
TheMolkaPL: That's because you forgot about `,` at the end of the line.

src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -36,6 +36,9 @@
 
   SOUNDS_ON("sounds", "on"), // Always play sounds
   SOUNDS_OFF("sounds", "off"), // Never play sounds
+
+  VOTES_ON("vote_book", "on"), // Show the vote book on cycle
+  VOTES_OFF("vote_book", "off"), // Don't show the vote book on cycle
```
TheMolkaPL: I would call them `VOTE_BOOK_ON/OFF` for consistency.

src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -31,7 +31,8 @@ public void update(MatchPlayer player) {
     }
   }, // Changes if observers are visible
   SOUNDS("sounds", SOUNDS_ON, SOUNDS_OFF), // Changes if sounds are played
-  ;
+  VOTE_BOOK("vote-book", VOTES_ON, VOTES_OFF) // Changes if the vote book is shown on cycle
```
Electroid: `VOTE("vote")` -- similar to `PICKER`, when we have a GUI for settings we'll explain it.

AustinLMayes: I decided to use `vote_book` since it was a bit more descriptive than just `vote`. Like toggling the "picker" is a bit more obvious than toggling "vote". 

I can still change it if you wish

Electroid: I've been trying to keep setting keys a single word :)

Electroid:
@AustinLMayes LGTM, squash then I'll merge it up.

PR #321, opened 2020-02-15
Fix double jump for testers

AustinLMayes:
If anyone has a better solution, I'm happy to hear it.

### DIFF
--- a/src/main/java/tc/oc/pgm/doublejump/DoubleJumpMatchModule.java
+++ b/src/main/java/tc/oc/pgm/doublejump/DoubleJumpMatchModule.java
@@ -13,6 +13,7 @@ import org.bukkit.event.entity.PlayerDeathEvent;
 import org.bukkit.event.player.PlayerToggleFlightEvent;
 import org.bukkit.util.Vector;
 import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.match.Match;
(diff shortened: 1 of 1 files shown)


Electroid:
I don't get it, what's the issue?

AustinLMayes:
@Electroid if you're testing a map with this module enabled, and go into creative, you can't fly around.

src/main/java/tc/oc/pgm/doublejump/DoubleJumpMatchModule.java
```diff
@@ -97,6 +98,8 @@ public void setKit(Player player, @Nullable DoubleJumpKit kit) {
   public void onPlayerToggleFlight(final PlayerToggleFlightEvent event) {
     Player player = event.getPlayer();
     Jumper jumper = this.jumpers.get(player);
+    if (player.getGameMode() == GameMode.CREATIVE && player.hasPermission(Permissions.DEBUG))
+      return; // HACK: Allow players who are testing to fly
```
Pablete1234: If a map has creative & double-jump for everyone, it could be working differently because of the perms, wich isn't desirable.

Either you consider people on creative just shouldn't double-jump at all since they can fly, or we don't add this at all.

Electroid:
I think the issue is that this introduces non-obvious behavior, that you would only be able to debug if you looked at the code. Going to pass on this one.

PR #325, opened 2020-02-16
Drop inv on quit

KingOfSquares:
This should drop the players inventory with the PlayerQuitEvent.
Having two problems atm i need to fix before merge:

Empty armor slots contains AIR and throws an error when trying to drop it.

Make it so observers wont trigger this.

Just posting this in case someone sees any errors..

Signed-off-by: Simon Mørland <simonmr60@gmail.com>


KingOfSquares:
This was a mistake posting lol, will open again when i do some more..

PR #330, opened 2020-02-18
Allow map objectives to be empty

AustinLMayes:
Re-adds the behavior in PGM which allowed maps to have no defined objective. 

This does, however, yell at the map developers to add one.


AustinLMayes:
Actually nvm

PR #332, opened 2020-02-22
Refactor code into separate modules

Electroid:
Now:
* `util`
* `util-bukkit`
* `core`
* `server`

Soon:
* `api`
* `community`


Electroid:
Will delay extracting `api` into its own module, it has too many direct dependencies in `core`.

Electroid:
This is essentially, other than some embedded improvements, just a code move. If you've developed with PGM before, nothing much has changed.

PR #335, opened 2020-02-25
Add initial support for net.kyori.text library

Electroid:
PGM relies too heavily on custom `Component` code and its created a lot of technical debt. Text formatting is a well solved problem and we do not need to reinvent the wheel. The `net.kyori.text` library is a well supported, platform agnostic library that gives us all the functionality we need.

Since this is a significant part of the codebase, we will try out something we haven't done before, which is a rolling refactor. Both `net.kyroi` and `tc.oc` versions of `Component` will be used until a full refactor is achieved.

This process will also allow us to take a second look at translation strings, which have a haphazard set of standards over the years: `legacy.properties` will contain all the old strings, and as code is migrated to the new library, it will be put in the `en-US.properties` file.

We should be able to remove a significant amount of boilerplate `Component` code and improve our translation strings at the same time. Sometime during the process, I will also setup `Crowdin` so we can start adding internationalization support.


util-bukkit/src/main/i18n/templates/en-US.properties
```diff
@@ -0,0 +1 @@
+admin.reload = Configuration reloaded
```
TheMolkaPL: Missing new line here.

PR #338, opened 2020-02-27
Match Stats

KingOfSquares:
### Match Stats

Hello!

This PR includes a feature of tracking some stats temporarily, and some different ways to display them:

On a MatchPlayerDeathEvent the players involved will both get a temporary message pop up over the hotbar:

With the command `/stats` these same numbers appear in chat:

And at the end of the match, the best stats from all players participating will appear(The stat with the bowshot will only appear if a kill with a bow has been done in the match):

Does this go into the community module?
All feedback appriciated! (Colors, layout, method names..)

05.03.2020: Updated formatting on end of match stats, and redid the changes into a MatchModule


Pablete1234:
IMO you're overformatting stuff. Don't use so many parenthesis and command, keep it simpler

BGMP:
@KingOfSquares do you really need me to format the code again? ;-; Not sure if I'm added to your fork as a collaborator <,<

KingOfSquares:
No im kidding :p

KingOfSquares:
Thanks Electroid! Working on it now, just wondering about something you wrote, what is the function of this part?
` private final Map<String, Entry> entries;

    public StatsMatchModule() {
        entries = new HashMap<>();
    }`

Why make that constructor? And why not just define "entires" as a HashMap from the start?

KingOfSquares:
Ready for review 👼 
@Electroid @Pablete1234 @TheMolkaPL

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -40,23 +40,7 @@
...
-import tc.oc.pgm.commands.MapCommands;
-import tc.oc.pgm.commands.MapPoolCommands;
-import tc.oc.pgm.commands.MatchCommands;
-import tc.oc.pgm.commands.ModeCommands;
-import tc.oc.pgm.commands.ObserverCommands;
-import tc.oc.pgm.commands.SettingCommands;
-import tc.oc.pgm.commands.StartCommands;
-import tc.oc.pgm.commands.TeamCommands;
-import tc.oc.pgm.commands.TimeLimitCommands;
+import tc.oc.pgm.commands.*;
```
TheMolkaPL: Why this change?

core/src/main/java/tc/oc/pgm/api/Modules.java
```diff
@@ -56,33 +56,7 @@
...
-import tc.oc.pgm.modules.ModifyBowProjectileMatchModule;
-import tc.oc.pgm.modules.ModifyBowProjectileModule;
-import tc.oc.pgm.modules.MultiTradeMatchModule;
-import tc.oc.pgm.modules.PlayableRegionMatchModule;
-import tc.oc.pgm.modules.PlayableRegionModule;
-import tc.oc.pgm.modules.SoundsMatchModule;
-import tc.oc.pgm.modules.TimeLockModule;
-import tc.oc.pgm.modules.ToolRepairMatchModule;
-import tc.oc.pgm.modules.ToolRepairModule;
+import tc.oc.pgm.modules.*;
```
TheMolkaPL: Why this change?

core/src/main/java/tc/oc/pgm/commands/StatCommands.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class StatCommands {
+
+  @Command(
+      aliases = {"stats"},
+      desc = "Shows your stats for this match")
+  public static void checkStats(CommandSender sender) {
+    sender.sendMessage(
+        ComponentUtils.horizontalLineHeading(
+            ChatColor.DARK_GREEN + "Current stats",
```
TheMolkaPL: Should be placed inside the translations file.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+
+  public static class Factory implements MatchModuleFactory<StatsMatchModule> {
+
+    @Override
+    public StatsMatchModule createMatchModule(Match match) throws ModuleLoadException {
+      return new StatsMatchModule();
+    }
+  }
+
+  public static class playerStats {
```
TheMolkaPL: `PlayerStats`

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+    private int kills = 0;
+    private int deaths = 0;
+    private int killstreak = 0;
+    private int killstreakMax = 0;
+    private int longestBowKill = 0;
```
TheMolkaPL: You don't have to define zeros in each :P Keep it simple :P

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    public void setLongestBowKill(double distance) {
+      if (distance > longestBowKill) {
+        longestBowKill = Integer.parseInt(new DecimalFormat("#").format(Math.round(distance)));
+      }
+    }
+
+    public Component getBasicStatsMessage() {
+      String KD;
```
TheMolkaPL:
```suggestion
      String kd;
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+      if (deaths == 0) {
+        KD = "0";
+      } else {
+        DecimalFormat df = new DecimalFormat("#.##");
```
TheMolkaPL: Can we change it to a constant in this class? Also, call it `decimalFormat` because `df` is very cryptic.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+                  "stats.basic",
+                  new PersonalizedText(Integer.toString(kills), ChatColor.GREEN),
+                  new PersonalizedText(Integer.toString(killstreak), ChatColor.GREEN),
+                  new PersonalizedText(Integer.toString(deaths), ChatColor.RED),
+                  new PersonalizedText(KD, ChatColor.GREEN))
+              .render());
+    }
+  }
+
+  public static Map<String, playerStats> allPlayerStats;
```
TheMolkaPL: Please make it private.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  @EventHandler
+  public void onPlayerDeath(MatchPlayerDeathEvent event) {
+    MatchPlayer victim = event.getVictim();
+    MatchPlayer murderer = null;
+    try {
+      murderer = event.getKiller().getParty().getPlayer(event.getKiller().getId());
+    } catch (NullPointerException ignored) {
+    }
```
TheMolkaPL: You shouldn't be catching `NPE`s :D If something returns `null` please check it with the `if` statements

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
TheMolkaPL: Also, I would persist player `UUID`s, not player names since `UUID`s are unique by definition.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+            });
+
+    if (murderer != null
+        && PlayerRelation.get(victim.getParticipantState(), murderer.getState())
+            != PlayerRelation.ALLY
+        && PlayerRelation.get(victim.getParticipantState(), murderer) != PlayerRelation.SELF) {
+      playerStats murdererStats = allPlayerStats.get(murderer.getBukkit().getName());
+      if (murdererStats == null) {
+        allPlayerStats.put(murderer.getBukkit().getName(), new playerStats());
+        murdererStats = allPlayerStats.get(murderer.getBukkit().getName());
```
TheMolkaPL: I see your are using `murderer.getBukkit().getName()` and `event.getVictim().getBukkit().getName()` in this method. Can you make local variables in this method to make the code cleaner?

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+          allPlayerStats.get(player.getBukkit().getName()).kills);
+      allKillstreaks.put(
+          player.getBukkit().getDisplayName(),
+          allPlayerStats.get(player.getBukkit().getName()).killstreakMax);
+      allDeaths.put(
+          player.getBukkit().getDisplayName(),
+          allPlayerStats.get(player.getBukkit().getName()).deaths);
+      allBowshots.put(
+          player.getBukkit().getDisplayName(),
+          allPlayerStats.get(player.getBukkit().getName()).longestBowKill);
```
TheMolkaPL: You should change the name and display name to local variables here too.

util-bukkit/src/main/i18n/templates/strings.properties
```diff
@@ -944,3 +944,13 @@ moderation.reports.hover = Reported by {0}
...
+
+# {X} = The number of the relevant stat
+stats.basic = Kills: {0}(Killstreak: {1}), Deaths: {2}, K/D: {3}
+
+# {0} = Username
+# {1} = The number of the relevant stat
+stats.kills = Kills: {0} : {1}
+stats.killstreak = Killstreak: {0} : {1}
+stats.deaths = Deaths: {0} : {1}
+stats.bowshot = Longest bowshot: {0} : {1} block(s)
```
TheMolkaPL: Missing newline character.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+  Component getDeathsMessage(Map.Entry<String, Integer> topResult) {
+    return new Component(
+        new PersonalizedTranslatable(
+                "stats.deaths",
+                new PersonalizedText(topResult.getKey()),
+                new PersonalizedText(Integer.toString(topResult.getValue()), ChatColor.RED))
+            .render());
+  }
+
+  Component getBowshotMessage(Map.Entry<String, Integer> topResult) {
```
TheMolkaPL: I think we should make a class representing `Map.Entry<String, Integer> topResult`.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
Pablete1234: when you have `df.format(number);` it being called df isn't very cryptic tbh

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+    // override the stats message.
+    event
+        .getMatch()
+        .getScheduler(MatchScope.LOADED)
+        .runTaskLater(
+            0,
+            () -> {
+              NMSHacks.sendHotbarMessage(
+                  victim.getBukkit(), finalVictimStats.getBasicStatsMessage());
+            });
```
Pablete1234: Make this a repeating task, send the message for a second or so or else it fades away too quick

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  @EventHandler
+  public void onMatchEnd(MatchFinishEvent event) {
+    Match match = event.getMatch();
+    HashMap<String, Integer> allKills = new HashMap<>();
+    HashMap<String, Integer> allKillstreaks = new HashMap<>();
+    HashMap<String, Integer> allDeaths = new HashMap<>();
+    HashMap<String, Integer> allBowshots = new HashMap<>();
```
Pablete1234:
```suggestion
    Map<String, Integer> allKills = new HashMap<>();
    Map<String, Integer> allKillstreaks = new HashMap<>();
    Map<String, Integer> allDeaths = new HashMap<>();
    Map<String, Integer> allBowshots = new HashMap<>();
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+      viewer.sendMessage(deathMessage);
+      if (bestBowshot.getValue()
+          != 0) { // Prevent this from showing if bows are not used on the map played
+        viewer.sendMessage(bowshotMessage);
+      }
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private Map.Entry<String, Integer> sortStats(HashMap<String, Integer> map) {
```
Pablete1234:
```suggestion
  private Map.Entry<String, Integer> sortStats(Map<String, Integer> map) {
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,233 @@
...
+    Object[] a = map.entrySet().toArray();
+    Arrays.sort(
+        a,
+        (Comparator)
+            (o1, o2) ->
+                ((Map.Entry<String, Integer>) o2)
+                    .getValue()
+                    .compareTo(((Map.Entry<String, Integer>) o1).getValue()));
+    return (Map.Entry<String, Integer>) a[0];
```
Pablete1234: Something along the lines of:
```
return map.entrySet().stream()
  .max(Comparator.comparingInt(Map.Entry::getValue))
  .orElse(null);
````

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 13 above)
Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: A `Record` here would be amazing, too bad we're on java 8 and not 14 :cry:

File: core/src/main/java/tc/oc/pgm/PGMPlugin.java (hunk 1 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/PGMPlugin.java]: It was either my IDE or Maven formatting, any issues with doing this?

File: core/src/main/java/tc/oc/pgm/commands/StatCommands.java (hunk 3 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/commands/StatCommands.java]: Of course! Great catch

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 7 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: > Can we change it to a constant in this class? Also, call it `decimalFormat` because `df` is very cryptic.

What do you mean change it to a constant, the K/D variable?

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 8 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: But when i enter in `player#getBukkit#getDisplayName` it keeps the colors and everything when displaying the best, does this also work with UUID? I only see a potential error if players use /nick, but it should be prevented to disguise as someone existing anyways.

Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: And why private? It is accessed by `StatCommands`

File: core/src/main/java/tc/oc/pgm/PGMPlugin.java (hunk 1 above)
Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/PGMPlugin.java]: It's just necessary :P

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 7 above)
Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Make `DecimalFormat` a constant so it is not created each time it is used. Look that this is going to parse your input string (`#.##`) each time to create a new `DecimalFormat` instance.

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 8 above)
Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: > But when i enter in `player#getBukkit#getDisplayName` it keeps the colors and everything when displaying the best, does this also work with UUID? I only see a potential error if players use /nick, but it should be prevented to disguise as someone existing anyways.

Player names are not unique so you shouldn't use them as a key anywhere.

> And why private? It is accessed by StatCommands

https://dzone.com/articles/following-oop-principles-hermetization-in-java

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 7 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Yes of course!

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 13 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: What do you mean Molka? Represent `topResult` as class in which way?

File: core/src/main/java/tc/oc/pgm/PGMPlugin.java (hunk 1 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/PGMPlugin.java]: It looks nicer? I can undo it

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 14 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: How do i make it repeat? I tried `#runTaskTimer` with a delay of 0 and interval of 5, but that looked like a infinite thing, right?

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 17 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Why this in favour of what i wrote? Is there a standard i'm doing wrong? :)
Or does it just look prettier (Your does)

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 8 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Will this: `match.getPlayer(UUID).getBukkit().getDisplayName()` return null if the player with the given UUID is offline?

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 17 above)
Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: It's easier to read ^^

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 13 above)
Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Make a class with two fields, `String` and `int`. A dedicated class with named fields is easier to understand than `Map.Entry`.

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 8 above)
Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Yes

Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Then how would i get the playername of a player that has the best results but is offline?

Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: You need to store that `PlayerStats`.

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 13 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: This should be resolved in coming update, thanks :)

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 8 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: I got it!!

File: core/src/main/java/tc/oc/pgm/PGMPlugin.java (hunk 1 above)
Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/PGMPlugin.java]: Do i really have to? I dont see the point of filling up the import list when you are going to register all commands and match modules anyways :p

Comment (TheMolkaPL) [reply core/src/main/java/tc/oc/pgm/PGMPlugin.java]: You don't have to. I'm always trying keep local code standards and change only what's necessary. :P

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 8 above)
Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Authors are not online but their name is displayed. The player names can be resolved from their UUID, and should be stored in a cache anyways. It's not stats' responsability to know playernames

File: core/src/main/java/tc/oc/pgm/PGMPlugin.java (hunk 1 above)
Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/PGMPlugin.java]: You should avoid wildcard imports, they're bad

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 14 above)
Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: i'm not sure with the scheduler. I know bukkit has it's own runRepeatingTask but idk why the match scheduler would not expose it

Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Yeah that is exposed with the scheduler through `runTaskTimer` But that is neverending! How do i stop it o_o

Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: It returns an int for the task id. you can cancel task by it's id

Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: Yeah, but i have to cancel the task through `Bukkit#getScheduler#cancelTask`
`Match#getScheduler(MatchScope.LOADED)#cancelTask` does not exist.

Comment (KingOfSquares) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: And the normal `runTaskTimer` does not return an int, you have to do `#getTaskId` it seems

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java (hunk 17 above)
Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java]: More concise (less code, same outcome)
Better readability
Avoids extra allocation of an object array.
Doesn't do weird casting from object to entry.
The complexity of finding a max or min is O(n) instead of O(n log n) that good sorting algs have
Isn't painful to read (yes, i'm mentioning readability twice, because it is double as relevant).

File: core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java [hunk 18]
```diff
@@ -0,0 +1,299 @@
...
+            event
+                .getMatch()
+                .getScheduler(MatchScope.LOADED)
+                .runTaskTimer(
+                    0,
+                    5,
+                    () -> {
+                      NMSHacks.sendHotbarMessage(
+                          finalMurderer.getBukkit(), finalMurdererStats.getBasicStatsMessage());
+                    })
```
applenick: I’m curious why you chose to call this directly instead of 
`finalMurderer.sendHotbarMessage(finalMurdererStats.getBasicStatsMessage())`?

The [MatchPlayer#sendHotbarMessage](https://github.com/Electroid/PGM/blob/master/core/src/main/java/tc/oc/pgm/api/chat/Audience.java#L61) method calls NMSHacks anyway, so it may be a little more consistent if you like.

Besides that, great job! This will be great once merged 😄

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,299 @@
...
+    int victimTaskId =
+        event
+            .getMatch()
+            .getScheduler(MatchScope.LOADED)
+            .runTaskTimer(
+                0,
+                5,
+                () -> {
+                  NMSHacks.sendHotbarMessage(
+                      victim.getBukkit(), finalVictimStats.getBasicStatsMessage());
```
applenick: See other comment about this.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,299 @@
...
+    private void onDeath() {
+      deaths++;
+      killstreak = 0;
+    }
+
+    private DecimalFormat decimalFormatBowshot = new DecimalFormat("#");
+
+    private void setLongestBowKill(double distance) {
+      if (distance > longestBowKill) {
+        longestBowKill = Integer.parseInt(decimalFormatBowshot.format(Math.round(distance)));
```
TheMolkaPL: Why are you doing this that complicated way? :P You can cast numbers between datatypes.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,299 @@
...
+      }
+    }
+
+    private DecimalFormat decimalFormatKd = new DecimalFormat("#.##");
```
TheMolkaPL: Can be `final`

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,299 @@
...
+                  new PersonalizedText(Integer.toString(deaths), ChatColor.RED),
+                  new PersonalizedText(kd, ChatColor.GREEN))
+              .render());
+    }
+  }
+
+  private static Map<UUID, PlayerStats> allPlayerStats;
+
+  public StatsMatchModule() {
+    allPlayerStats = new HashMap<>();
```
TheMolkaPL: This is defining static context from the instance context.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,299 @@
...
+            .getMatch()
+            .getScheduler(MatchScope.LOADED)
+            .runTaskLater(
+                70,
+                () -> {
+                  Bukkit.getScheduler().cancelTask(murdererTaskId);
+                });
+      }
+    }
+  }
```
TheMolkaPL: This method seems to be very long. Could you try to slice it into smaller ones? Would make it easier to read.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,299 @@
...
+
+  Component getBowshotMessage(TopResult topResult, Match match) {
+    return new Component(
+        new PersonalizedTranslatable(
+                "stats.bowshot",
+                playerName(match, topResult.uuid),
+                new PersonalizedText(Integer.toString(topResult.stat), ChatColor.YELLOW))
+            .render());
+  }
```
TheMolkaPL: These methods are actually them same. The only difference seems to be translation keys and colors. To keep this try could try make a generic method with these as parameters, and these methods would call it.

```java
Component getBowshotMessage(TopResult topResult, Match match) {
  return getMessage("stats.bowshot", topResult, match, ChatColor.YELLOW);
}

Component getMessage(String messageKey, TopResult topResult, Match match, CharColor color) {
  return new Component(
      new PersonalizedTranslatable(
              messageKey,
              playerName(match, topResult.uuid),
              new PersonalizedText(Integer.toString(topResult.stat), color))
          .render());
}
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: Ah! I never knew that, thanks! :)

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: Like this?
` private void setLongestBowKill(double distance) {
      if (distance > longestBowKill) {
        longestBowKill = (int) distance;
      }
    }`

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: What does that mean? haha

TheMolkaPL: `allPlayerStats` is a static variable. This means that the variable is global and not object depended. If you create a new instance of `StatsMatchModule` the contents of `allPlayerStats` should still be the same. The problem is that you are modifying static (global) variable from the object (its constructor). You should either make `allPlayerStats` non-static and make it object depended, or initialize it just once and make it a constant. If so, there will only be one instance of this `allPlayerStats` map.

```java
private final Map<UUID, PlayerStats> allPlayerStats = new HashMap<>();
```

or

```java
private static final Map<UUID, PlayerStats> ALL_PLAYER_STATS = new HashMap<>();
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
TheMolkaPL: Exactly, but don't compare doubles with `<` `>` `==`

https://howtodoinjava.com/java/basics/correctly-compare-float-double/

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: Do we even need a Component constructor for each message then? Why not just call `getMessage` from a method directly.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: What do i even need the constructor for? I only added it because Electroids example has it. The Module does not need to get passed any objects on creation anyways.., i will make i private static final,

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: So like this?
`private void setLongestBowKill(double distance) {
      if (new BigDecimal(distance).compareTo(new BigDecimal(longestBowKill)) > 0) {
        longestBowKill = (int) distance;
      }
    }`

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: Yeah i'm splitting it up, you will see next commit.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+@ListenerScope(MatchScope.RUNNING)
+public class StatsMatchModule implements MatchModule, Listener {
+
+  public static class Factory implements MatchModuleFactory<StatsMatchModule> {
+
+    @Override
+    public StatsMatchModule createMatchModule(Match match) throws ModuleLoadException {
+      return new StatsMatchModule(match);
+    }
+  }
```
Electroid: You don't need this. In `Modules.java` you can put instead:
```java
register(StatsMatchModule.class, StatsMatchModule::new);
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    private void onDeath() {
+      deaths++;
+      killstreak = 0;
+    }
+
+    private void setLongestBowKill(double distance) {
+      if (distance > longestBowKill) {
+        longestBowKill = (int) distance;
```
Electroid: To avoid the possibility of a `0` bow kill:

```suggestion
        longestBowKill = (int) Math.ceil(distance);
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+  private TopResult sortStats(Map<UUID, Integer> map) {
+    Map.Entry<UUID, Integer> mapEntry =
+        map.entrySet().stream().max(Comparator.comparingInt(Map.Entry::getValue)).orElse(null);
+    TopResult topResult = new TopResult();
+
+    if (mapEntry == null) { // Should never happen, but acts as a failsafe
+      topResult.uuid = UUID.fromString("3c7db14d-ac4b-4e35-b2c6-3b2237f382be");
+      topResult.stat = 0;
+      return topResult;
+    }
```
Electroid: You don't need this `null` check, these maps will never be null.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+      viewer.sendMessage(killstreakMessage);
+      viewer.sendMessage(deathMessage);
+      if (bestBowshot.stat != 0) viewer.sendMessage(bowshotMessage);
+    }
+  }
+
+  private class TopResult {
+    UUID uuid;
+    int stat;
+  }
```
Electroid: You don't need this class.

Just use `Map.Entry<UUID, Integer>`

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+  private void sendLongHotbarMessage(MatchPlayer player, Component message) {
+    int taskId =
+        match
+            .getScheduler(MatchScope.LOADED)
+            .runTaskTimer(
+                0,
+                5,
+                () -> {
+                  player.sendHotbarMessage(message);
+                })
```
Electroid:
```suggestion
                () -> player.sendHotbarMessage(message));
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+            .getTaskId();
+
+    match
+        .getScheduler(MatchScope.LOADED)
+        .runTaskLater(
+            20 * 4,
+            () -> {
+              Bukkit.getScheduler().cancelTask(taskId);
+            });
```
Electroid:
```suggestion
            () -> Bukkit.getScheduler().cancelTask(taskId));
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,250 @@
...
+                messageKey,
+                playerName(topResult.uuid),
+                new PersonalizedText(Integer.toString(topResult.stat), color).render())
+            .render());
+  }
+
+  private PersonalizedText playerName(UUID playerUUID) {
+    if (Bukkit.getPlayer(playerUUID) == null) {
+      if (Bukkit.getOfflinePlayer(playerUUID).getName() == null) {
+        return new PersonalizedText("Noone", ChatColor.MAGIC, ChatColor.BLACK);
```
Electroid:
```suggestion
        return new PersonalizedText("Unknown", ChatColor.MAGIC, ChatColor.BLACK);
```

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,247 @@
...
+                  "stats.basic",
+                  new PersonalizedText(Integer.toString(kills), ChatColor.GREEN),
+                  new PersonalizedText(Integer.toString(killstreak), ChatColor.GREEN),
+                  new PersonalizedText(Integer.toString(deaths), ChatColor.RED),
+                  new PersonalizedText(kd, ChatColor.GREEN))
+              .render());
+    }
+  }
+
+  private static final Map<UUID, PlayerStats> allPlayerStats = new HashMap<>();
```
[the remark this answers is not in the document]
KingOfSquares: Already fixed.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -0,0 +1,247 @@
...
+    MatchPlayer victim = event.getVictim();
+    MatchPlayer murderer = null;
+
+    if (event.getKiller() != null)
+      murderer = event.getKiller().getParty().getPlayer(event.getKiller().getId());
+
+    UUID victimUUID = victim.getId();
+    PlayerStats victimStats = allPlayerStats.get(victimUUID);
+
+    if (PlayerStatsDoesNotExist(victimUUID)) victimStats = putNewPlayer(victimUUID);
```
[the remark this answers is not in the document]
KingOfSquares: Already fixed.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: Ok!

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: This was done because of @TheMolkaPL s suggestion earlier. You can see his reasoning https://github.com/Electroid/PGM/pull/338#discussion_r388542333
I can remove it if you still disagree

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: I am not sure why that would be needed, i already prevent any message containing bow kills from showing if they are `0`.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
Electroid: I would remove it.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
Electroid: It's good form.

core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
KingOfSquares: Sure, i've removed it now.

Electroid:
@KingOfSquares Can you add a `stats.enabled` in the PGM config to optionally disable this? Then should be good to go.

KingOfSquares:
> @KingOfSquares Can you add a `stats.enabled` in the PGM config to optionally disable this? Then should be good to go.

Yes of course, we talked about that i remember.

Electroid:
@KingOfSquares Can you rebase so it works?

KingOfSquares:
> @KingOfSquares Can you rebase so it works?

Yes, works now.

PR #339, opened 2020-02-27
Fix observer item drops in match start phase

Pugzy:
Prevents "match players" dropping items on disconnect outside of match running state.

This fixes "players" who had joined during the start countdown dropping all of the items once leaving.


Electroid:
Thanks!

PR #341, opened 2020-02-28
Improve moderation commands

applenick:
# Improve Moderation commands
After some feedback regarding the now merged moderation commands, I’ve worked on a quick fix that addresses the most concerning issues. Everything tested find and should be ready to merge if no additional suggestions are made.

## Changes made:

Additions
- Added `/mutes` - Sends a list of usernames who are currently muted
- Added an offline ban command: 
`/offlineban (username) (reason) -t (optional time | don't include for perm ban)`
- Added `/alts (target) (page)` for viewing online alt-accounts
- Added `/baninfo (player|uuid)` for looking up punishment info for a player 

---
Removed
- Removed -s and -w flags from `/mute (player) (reason)`
- Removed -s flag from /warn
---
Etc.
- Merged temp-ban and ban into `/ban`
    - New format:
 `/ban (player) (reason) -t (optional time | don't include for perm ban)`
- Improved formatting for punishment broadcasts
- IP ban now bans usernames in addition to the IP 
  - (Player can't rejoin with different IP to ban evade)
- Mutes and Warns are now only broadcasted to online staff
- Fixed bug where sounds play when muted player sends a /msg

## Screenshots 
Please see following comments to see updated screenshots


Brottweiler:
On OCN warns were never broadcasted to regular players. Maybe warns should be always silent? Maybe mutes as well?

And I gotta say, [Ban - 7 days] doesn't look great 😕 
Could (7 days) maybe be added at the end of the ban message? Just an idea, not sure how it will look like.

applenick:
New format for punishments:
Screen Shot 2020-02-28 at 1 07 52 AM

Also as @Brottweiler suggested, made warns/mutes only send a broadcast to adminchat.

Electroid:
Why not something simpler like `N day ban`?

applenick:
@Electroid 
The punishment length only appears for temp-bans, however I get what you’re saying. 

Though the current format retains consistency for all punishments. Say we make it so it was
`{staff} > {length} {punishment} > {target} > {reason}` as you suggested. 

While more simple, it will only show a length for temp-bans. Which I personally believe will be more confusing for players. Since it could vary depending on the lengths, such as N hours to N days. Also the `PeriodFormats` seems to give durations in plural, so it would be `N days ban`

Personally I’d say we keep this format, but let me know what you would like me to do.

applenick:
Update:
After talking with @Brottweiler, it seems that also banning the username may be more troublesome for server administrators when mistakes are made. So I’ve reverted that change so this is as close to the vanilla ban system as possible.

Brottweiler:
## IP ban

> banning the username may be more troublesome for server administrators

Just to add onto this, if I would ever unban an IP, I have to unban the IP *and* the username. Instead, it's better if ipban *only* bans the IP and not the player (but it will kick the player). 
If I want to ban the username as well, I can just issue a normal (and silent) `ban` afterwards.

## Formatting

I like the `staff » Ban » 7 days » player » REASON`, it looks good, but OCN always showed the ban time, even when permanent, and it didn't say `Ban » 7 day`.

You already know the formatting on OCN, but I don't remember how the temp bans looked like. But permanent ban would look like `staff » Permanent Ban » player » REASON`. 7day ban probably looked like `staff » 7 day ban » player » REASON`.

applenick:
So another update:
Was able to combine `/ban` and `/tempban` into one, using a `-t (time)` flag in order to define the length. I’ve discussed this with @Brottweiler, and we believe it would be easier to have them as one command instead of two. 

Also due to the way Intake handles things, I’ve added a `/offlineban` to handle players who are offline, which was requested by staff on OCC. Regarding the problem with intake, say you do `/ban <some offline name>` currently, it will throw an exception saying the player could not be found. This new command solves that problem. The reason I simply did not change the input to a string and check if the player is online from the `/ban` command, is because intake will not tab-complete for strings. 🤷‍♂ 

Last change is regarding the punishment formatting, which I’ve once again updated. The final format is `{staff} » (length) {punishment} » {target}» {reason}`

Let me know if there is anything else that can be updated, otherwise I believe is fine for now.

Pablete1234:
I am not sure if this has already been changed, but the current colors used in chat & formatting are pretty violent (lots of red & dark colors). On OCN they used to be less violent colors

applenick:
Hey @Pablete1234,
Thanks for the feedback. What do you think of this?
Screen Shot 2020-03-01 at 10 08 31 AM
Screen Shot 2020-03-01 at 10 09 03 AM

applenick:
Thanks everyone, glad the new format is better 😄 

I was working on an additional `/alts` command that looks at online players and allows staff to determine who’s on the same IP. Would it be alright to hold off merging this PR until that’s complete or would you rather I make a new PR?

Brottweiler:
Could the mute warn by default? I don't see any reason why you would ever need to mute someone without wanting to issue a warning as well?

applenick:
Another update

Added the following commands:
`/mutes` - Lists muted players who are online
Screen Shot 2020-03-06 at 12 52 43 AM


`/alts (target)` Allows staff to check which players are on the same address (useful before using `/ipban`
Screen Shot 2020-03-06 at 12 52 13 AM
(Screenshot is of using example `/alts applenick3`, which shows a single entry of alts if any. Using `/alts` with no target give paginated results of all alts on the server. 


`/baninfo (target)` - Displays information regarding a player’s active punishment (ban / temp-ban)
Screen Shot 2020-03-06 at 1 08 05 AM

Screen Shot 2020-03-06 at 1 07 26 AM

Other changes:
Made `/mute` issue a warning as per @Brottweiler’s suggestion


Sorry if I altered the comment history on this PR due to the rebase. Let me know if you want to have the commits merged into one or something.

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -187,48 +215,46 @@ public void kick(
...
-      @Switch('s') boolean silent) {
-    MatchPlayer targetMatchPlayer = match.getPlayer(target);
-    Component senderName = formatPunisherName(sender, match);
-    if (punish(PunishmentType.BAN, targetMatchPlayer, sender, reason, silent)) {
-      banPlayer(target, reason, senderName, null);
-      target.kickPlayer(formatPunishmentScreen(PunishmentType.BAN, senderName, reason, null));
-    }
-  }
+      @Switch('s') boolean silent,
+      @Switch('t') String length)
```
Pablete1234: I'm pretty sure we can use a Duration here, and have a duration parser

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+      usage = "<player> <reason> -t (length)",
+      desc = "Ban an offline player from the server",
+      flags = "t",
+      perms = Permissions.BAN)
+  public void offlineBan(
+      CommandSender sender,
+      MatchManager manager,
+      String target,
+      @Text String reason,
+      @Switch('t') String length)
```
Pablete1234: same here

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+      perms = Permissions.STAFF)
+  public void alts(
+      Audience audience,
+      CommandSender sender,
+      MatchManager manager,
+      @Fallback(Type.NULL) Player target,
+      @Default("1") int page)
+      throws CommandException {
+    if (target != null) {
+      MatchPlayer targetMP = manager.getPlayer(target);
```
Pablete1234: Personally i'd rename `target` to `targetPl`, and keep this as `target`

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+        sender,
+        new PersonalizedText(target).color(ChatColor.DARK_AQUA),
+        reason,
+        true,
+        duration);
+  }
+
+  @Command(
+      aliases = {"alts", "listalts"},
+      usage = "[target] (page)",
```
Pablete1234:
```suggestion
      usage = "[target] [page]",
```

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+        public Component formatEmpty() throws CommandException {
+          throw new CommandException("No alternate accounts found!");
+        }
+      }.display(audience, altAccounts, page);
+    }
+  }
+
+  @Command(
+      aliases = {"baninfo", "lookup", "l"},
+      usage = "[player] -u (if input is UUID)",
```
Pablete1234:
```suggestion
      usage = "[player]",
```

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+      desc = "Lookup baninfo about a player",
+      perms = Permissions.STAFF)
+  public void banInfo(CommandSender sender, String target, @Switch('u') boolean uuid)
```
Pablete1234:
```suggestion
  public void banInfo(CommandSender sender, String target)
```

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+      throws CommandException {
+    BanEntry ban = Bukkit.getBanList(BanList.Type.NAME).getBanEntry(target);
```
Pablete1234: Make name or UUID toggle based on the format, not based on a flag.
Minecraft usernames must be between 3 and 16 characters, if it's over 16, it's guaranteed to be a UUID (possibly with, possibly without dashes)

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
```diff
@@ -281,6 +307,290 @@ public void ipBan(
...
+
+    if (ban != null) {
```
Pablete1234: Switch this one arround to:
 ```
if (ban == null ||
     (ban.getExpiration() != null && ban.getExpiration.toInstant().isBefore(Instant.now())) { 
    throw CommandEx...;
}
```
It will make the rest of the command not be indented by one and probably easier to read

core/src/main/java/tc/oc/pgm/community/commands/ModerationCommands.java
TheMolkaPL: The command should automatically detect if the input is a player name or UUID

applenick:
@Pablete1234 and @TheMolkaPL 
Thanks for the feedback! I made all the suggested changes and should be good to go now 👍 

Also after speaking with @Brottweiler again, we decided it's best to have usernames banned in the `/ipban` after all. Due to a scenario where a player who is ipbanned one day, simply rejoins the next with a different IP.

applenick:
@Electroid I can’t request you for a re-review, but just letting you know the changes here should be good to go.  I’m sure OCC would like to get these changes implemented soon, so just let me know if there are any final adjustments and I can take care of them.

applenick:
So not sure how I missed this, but the old ModerationCommands class was never deleted 😱 
Removed it and also added a darker header color to `/reports` just to make it prettier. 

@Electroid this should be ready now 👍

PR #351, opened 2020-03-02
Reintroduce fireworks 

applenick:
# Reintroduce Fireworks 🎆
I noticed recently when testing that the _iconic_ post match fireworks were not apart of this version of PGM. So after taking a look through the ProjectAres repo, I discovered that we’re missing all of the firework features. Personally I really enjoy fireworks being launched at the end of a match or when an objective is completed. So this PR uses a slightly modified version of the code found [here](https://github.com/OvercastNetwork/ProjectAres/tree/master/PGM/src/main/java/tc/oc/pgm/fireworks) to reintroduce those features back. 

### Changes:
* Adds post-match victory fireworks 
  * (Fireworks shoot off from location of winning team at the match end)
* Adds objective fireworks 
  * (Fireworks shoot off for various objectives, such as placing a wool or capturing a control point)

I found it difficult to get a screenshot of these changes in action. If someone really wants to see it I’ll see what I can do. 

If anyone has suggestions on additional places where fireworks would be useful let me know.


core/src/main/java/tc/oc/pgm/Config.java
```diff
@@ -478,4 +478,39 @@ public static boolean isAppealVisible() {
...
+      }
+
+      public static int iterations() {
+        return Math.max(1, getConfiguration().getInt("fireworks.post-match.iterations", 15));
+      }
+
+      public static int power() {
+        return Math.max(0, getConfiguration().getInt("fireworks.post-match.power", 2));
+      }
+    }
```
Electroid: It never made any sense to make these configurable, just make these static variables.

core/src/main/java/tc/oc/pgm/Config.java
```diff
@@ -478,4 +478,39 @@ public static boolean isAppealVisible() {
...
+
+    public static class Goals {
+      public static boolean enabled() {
+        return getConfiguration().getBoolean("fireworks.goals.enabled", false);
+      }
+    }
```
Electroid: Simple: `fireworks.enabled`

core/src/main/java/tc/oc/pgm/fireworks/FireworkUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+import org.bukkit.entity.EntityType;
+import org.bukkit.entity.Firework;
+import org.bukkit.inventory.meta.FireworkMeta;
+
+public class FireworkUtil {
+
+  public static @Nonnull Firework spawnFirework(
```
Electroid: Can be moved to the `MatchModule`

core/src/main/java/tc/oc/pgm/fireworks/FireworkUtil.java
```diff
@@ -0,0 +1,47 @@
...
+    meta.setPower(power);
+    meta.addEffect(effect);
+
+    Firework firework = (Firework) location.getWorld().spawnEntity(location, EntityType.FIREWORK);
+    firework.setFireworkMeta(meta);
+
+    return firework;
+  }
+
+  public static @Nonnull Location getOpenSpaceAbove(@Nonnull Location location) {
```
Electroid: Don't we have some `Locations` or `LocationUtil` class?

core/src/main/java/tc/oc/pgm/fireworks/ObjectivesFireworkListener.java
```diff
@@ -0,0 +1,117 @@
...
+import tc.oc.pgm.core.CoreLeakEvent;
+import tc.oc.pgm.destroyable.DestroyableDestroyedEvent;
+import tc.oc.pgm.flag.event.FlagCaptureEvent;
+import tc.oc.pgm.regions.Bounds;
+import tc.oc.pgm.regions.Region;
+import tc.oc.pgm.wool.PlayerWoolPlaceEvent;
+import tc.oc.util.bukkit.BukkitUtils;
+import tc.oc.util.bukkit.block.BlockVectors;
+
+public class ObjectivesFireworkListener implements Listener {
```
Electroid: Integrate this into a `FireworksMatchModule implements MatchModule, Listener`

core/src/main/java/tc/oc/pgm/fireworks/PostMatchFireworkMatchModule.java
```diff
@@ -0,0 +1,109 @@
...
+import tc.oc.pgm.api.match.MatchModule;
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.match.event.MatchFinishEvent;
+import tc.oc.pgm.api.party.Competitor;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.ListenerScope;
+import tc.oc.util.bukkit.BukkitUtils;
+
+@ListenerScope(MatchScope.LOADED)
+public class PostMatchFireworkMatchModule implements MatchModule, Listener {
```
Electroid: There can just be 1 fireworks `MatchModule`, handle both event and post-match

core/src/main/resources/config.yml
```diff
@@ -120,3 +120,16 @@ moderation:
...
+  post-match:
+    enabled: true
+    number: 5
+    delay: 40 # ticks
+    frequency: 40 # ticks
+    iterations: 15
+    power: 2
+  # Goals spawn fireworks when objectives are completed
+  goals:
+    enabled: true
```
Electroid:
```yml
fireworks:
     enabled: true
```

core/src/main/java/tc/oc/pgm/fireworks/PostMatchFireworkMatchModule.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private BukkitTask task;
```
Electroid: Don't use `BukkitTask`, instead `Match#getScheduler(MatchScope.LOADED)`

Electroid: Seems like you do already, so no need to store the `BukkitTask`, it will get canceled automatically.

core/src/main/java/tc/oc/pgm/Config.java
Pablete1234: post-match fireworks are barely used tho. Even pgm 1.9+ never used it except for arcade, ctw/dtc servers never used them

core/src/main/resources/config.yml
Pablete1234: I would keep it as:
```suggestion
fireworks:
  post-match: false
  goals: true
```
(post-match default being false)

applenick:
@Electroid & @Pablete1234 thanks for all the suggestions!
I was unsure how much of this to modify as it came straight from the ProjectAres repo. Now that you both approve I’ll clean up this PR with the changes that were suggested. I’ll update this once the changes are made.

core/src/main/java/tc/oc/pgm/fireworks/FireworkUtil.java
applenick: @Electroid I was unable to find a `Locations` or `LocationUtil` class. Perhaps you have some guidance on this? I placed both of the FireworkUtil methods into the `FireworkMatchModule` for now.

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,237 @@
...
+
+  private final Match match;
+
+  private final boolean postEnabled = Config.Fireworks.postMatchEnabled();
+  private final boolean goalsEnabled = Config.Fireworks.goalsEnabled();
+  private final int number = 5;
+  private final int delay = 40;
+  private final int frequency = 40;
+  private final int iterationCount = 15;
+  private final int power = 2;
```
Electroid: Should be static, and follow the static variable naming rules: `ITERATION_COUNT`

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,237 @@
...
+  @EventHandler(priority = EventPriority.MONITOR)
+  public void onCoreLeak(final CoreLeakEvent event) {
+    if (goalsEnabled && event.getCore().isVisible()) {
+      this.spawnFireworkDisplay(
+          event.getMatch().getWorld(),
+          event.getCore().getCasingRegion(),
+          event.getCore().getColor(),
+          8,
+          1.5,
+          2);
```
Electroid: What are these magic numbers? They should probably be static variables too.

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,237 @@
...
+  @EventHandler(priority = EventPriority.MONITOR)
+  public void onDestroyableBreak(final DestroyableDestroyedEvent event) {
+    if (goalsEnabled && event.getDestroyable().isVisible()) {
+      this.spawnFireworkDisplay(
+          event.getMatch().getWorld(),
+          event.getDestroyable().getBlockRegion(),
+          event.getDestroyable().getColor(),
+          4,
+          3,
+          2);
```
Electroid: Magic numbers.

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,237 @@
...
+    meta.setPower(power);
+    meta.addEffect(effect);
+
+    Firework firework = (Firework) location.getWorld().spawnEntity(location, EntityType.FIREWORK);
+    firework.setFireworkMeta(meta);
+
+    return firework;
+  }
+
+  public static @Nonnull Location getOpenSpaceAbove(@Nonnull Location location) {
```
Electroid: We don't use `@Nonnull`, only `@Nullable`, so they can be removed here.

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
Electroid: Also, why do we need static numbers? Can these variables not be derived by the size, location, etc.. of the objective?

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,237 @@
...
+
+  public static List<FireworkEffect.Type> AVAILABLE_TYPES =
+      ImmutableList.<FireworkEffect.Type>builder()
+          .add(Type.BALL)
+          .add(Type.BALL_LARGE)
+          .add(Type.BURST)
+          .add(Type.STAR)
+          .build();
+
+  public class FireworkRunner implements Runnable {
```
Electroid: `private static class`

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,237 @@
...
+
+  @EventHandler(priority = EventPriority.MONITOR)
+  public void onMatchEnd(final MatchFinishEvent event) {
+    if (!postEnabled) return;
+    match
+        .getScheduler(MatchScope.LOADED)
+        .runTaskTimer(delay, frequency, new FireworkRunner(event.getWinners()));
+  }
+
+  public static List<FireworkEffect.Type> AVAILABLE_TYPES =
```
Electroid: `FIREWORK_TYPES`

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
applenick: @Electroid To be honest I haven’t looked into why these numbers were set. As I stated in the original PR, most of this came right from the ProjectAres repo. So my guess is that these numbers are most likely finely tuned to whatever OCN used to have them as. 

When I performed testing of these changes everything worked really nicely, so I don’t really see a need to change them/risk having them look bad. If you want I can make these static variables as you suggested, but anything custom may take some time (I don’t have much free time for the next week).  

Let me know what you would like me to do. Thanks!

Electroid: Hmm, that's okay for now, we can skip over the magic numbers

core/src/main/java/tc/oc/pgm/observers/ObserverToolsMatchModule.java
```diff
@@ -36,13 +34,6 @@
...
 public class ObserverToolsMatchModule implements MatchModule, Listener {
 
-  public static class Factory implements MatchModuleFactory<ObserverToolsMatchModule> {
-    @Override
-    public ObserverToolsMatchModule createMatchModule(Match match) throws ModuleLoadException {
-      return new ObserverToolsMatchModule(match);
-    }
-  }
-
   // Slot where tool item is placed
```
Pablete1234: This doesn't seem like it belongs in this commit

applenick: Oh whoops 🤦‍♂

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,259 @@
...
+  private final Match match;
+
+  private static final boolean POST_ENABLED = Config.Fireworks.postMatchEnabled();
+  private static final boolean GOALS_ENABLED = Config.Fireworks.goalsEnabled();
+
+  private static final int NUMBER = 5;
+  private static final int DELAY = 40;
+  private static final int FREQUENCY = 40;
+  private static final int ITERATION_COUNT = 15;
+  private static final int POWER = 2;
```
Pablete1234: Could use some better naming or descriptions (i'm not 100% sure if i got them all right or i mixed any of them up):
```suggestion
  private static final int MAX_ROCKET_COUNT = 5; // Maximum rockets to throw at once, one per player
  private static final int INITIAL_DELAY = 40; // Ticks before starting to throw rockets
  private static final int FREQUENCY = 40; // Ticks between rocket throws
  private static final int ITERATION_COUNT = 15; // Amount of times rockets are thrown
  private static final int ROCKET_POWER = 2; // Power of the rockets, 1 = low, 2 = medium, 3 = high
```

applenick: Great suggestion @Pablete1234! I’ll take care of this as soon as I can, sorry for the delay.

applenick:
@Pablete1234 thanks again for the help! Everything should be cleaned up now and up to date with master. 

@Electroid unless you have any additional suggestions, this should be ready to go when you want 👍

core/src/main/java/tc/oc/pgm/fireworks/FireworkMatchModule.java
```diff
@@ -0,0 +1,258 @@
...
+  private static final double CORE_LEAK_RADIUS = 1.5;
+
+  private static final int DESTROYABLE_BREAK_COUNT = 4;
+  private static final double DESTROYABLE_BREAK_RADIUS = 3;
+
+  private static final int CONTROL_POINT_COUNT = 8;
+  private static final double CONTROL_POINT_RADIUS = 1;
+
+  private static final int FLAG_CAPTURE_COUNT = 6;
+  private static final double FLAG_CAPTURE_RADIUS = 1;
```
Electroid: TBH, if there are all just "magic" numbers, then let's skip the static variables.
Nice in theory, but there's way too many vars here.

Pablete1234: I'd much rather have them as variables than magic numbers. If we ever feel like changing them it's much easier to have it this way

applenick: @Electroid I agree with @Pablete1234. It is rather excessive for magic numbers, but personally I feel it’s more organized and understandable for new contributors who may want to adjust them later on. 

If you dislike them all at the top of the file, I can move each respective pair above the event they’re related to like:

```java
...
private static final int CORE_LEAK_COUNT = 8;
private static final double CORE_LEAK_RADIUS = 1.5;

   public void onCoreLeak(final CoreLeakEvent event) {
     if (GOALS_ENABLED && event.getCore().isVisible()) {
       this.spawnFireworkDisplay(
           event.getMatch().getWorld(),
           event.getCore().getCasingRegion(),
           event.getCore().getColor(),
           CORE_LEAK_COUNT,
           CORE_LEAK_RADIUS,
           ROCKET_POWER);
     }
   }

```

Electroid: @applenick Go ahead and make that change, then ready to go

core/src/main/java/tc/oc/pgm/Config.java
```diff
@@ -477,4 +477,15 @@ public static boolean isAppealVisible() {
       return getAppealMessage().length() > 0;
     }
   }
+
+  public static class Fireworks {
+
+    public static boolean postMatchEnabled() {
+      return getConfiguration().getBoolean("fireworks.post-match", false);
```
Electroid: Post match default is false? Why not true?

applenick: It was suggested above by @Pablete1234 to have it default false. Personally I would have it as true. Want me to adjust it?

applenick:
@Electroid I went ahead and adjusted it, let me know if there is anything else.

PR #356, opened 2020-03-13
Chat and Sound improvements

### LINKED ISSUE #352: Disable admin chat sound
Direct messages leaving a ping-sound makes total sense.  That's because someone is specifically asking YOU and YOU alone.  But for admin chat, lots of times there's active discussions (at least within stratus event servers) and if you're playing in the match, or anything where you don't care about the ongoing discussion, it's super annoying to hear PING PING PING PING, when it doesn't relate to you, unlike messages where it always does concern you.  I know you can toggle sounds, but afaik that does both of them, and it would make sense for most people to have admin chat w/o sound and DMs with sound.

applenick:
# Chat and Sound improvements

This resolves #352 by allowing users to toggle their sound settings to disallow extra sound effects such as admin chat and reports. 

### All Changes Made:
- Adjusted setting values for sounds to `ALL`, `MESSAGES`, `NONE`.
  - `ALL` receive all sounds
  - `MESSAGES` only `/msg` sounds are played
  - `NONE` no sound effects play
- ~Color change for Admin chat to differentiate it from direct messages.~
- Changed DM formatting (see below)
- Lowered pitch for admin chat sound 
  -  New sound is just a tad bit different so should help tell the difference between different alerts
- Added a new `/say (team) (message)` command to allow staff to speak in team chats while observing. 

**Update: screenshot outdated, see below for new info.**

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
@@ -34,6 +34,7 @@ import tc.oc.util.StringUtils;
 import tc.oc.util.bukkit.BukkitUtils;
 import tc.oc.util.bukkit.OnlinePlayerMapAdapter;
 import tc.oc.util.bukkit.component.Component;
+import tc.oc.util.bukkit.component.ComponentRenderers;
 import tc.oc.util.bukkit.component.Components;
 import tc.oc.util.bukkit.component.types.PersonalizedText;
 import tc.oc.util.bukkit.component.types.PersonalizedTranslatable;
@@ -53,8 +54,7 @@ public class ChatDispatcher implements Listener {
   private static final String DM_SYMBOL = "@";
   private static final String ADMIN_CHAT_SYMBOL = "$";
 
-  public static final String ADMIN_CHAT_PREFIX = "[" + ChatColor.RED + "A" + ChatColor.WHITE + "]";
-  private static final String DM_CHAT_PREFIX = "[" + ChatColor.GOLD + "DM" + ChatColor.WHITE + "]";
+  public static final String ADMIN_CHAT_PREFIX = "[" + ChatColor.GOLD + "A" + ChatColor.WHITE + "]";
 
   public ChatDispatcher(MatchManager manager) {
     this.manager = manager;
@@ -189,7 +189,7 @@ public class ChatDispatcher implements Listener {
         match,
         sender,
         message,
-        DM_CHAT_PREFIX + " {0}: {1}",
+        ComponentRenderers.toLegacyText(formatPrivateMessage("command.message.from"), receiver),
         viewer -> viewer.getBukkit().equals(receiver),
         null);
 
@@ -197,7 +197,8 @@ public class ChatDispatcher implements Listener {
         match,
         manager.getPlayer(receiver), // Allow for cross-match messages
         message,
-        DM_CHAT_PREFIX + " -> {0}: {1}",
+        ComponentRenderers.toLegacyText(
+            formatPrivateMessage("command.message.to"), manager.getPlayer(receiver).getBukkit()),
         viewer -> viewer.getBukkit().equals(sender.getBukkit()),
         null);
   }
@@ -215,6 +216,12 @@ public class ChatDispatcher implements Listener {
     }
 
     sendDirect(match, sender, receiver.getBukkit(), message);
+  }
+
+  private static PersonalizedText formatPrivateMessage(String key) {
+    return new PersonalizedText(
+        new PersonalizedTranslatable(key).getPersonalizedText().color(ChatColor.GRAY).italic(true),
+        new PersonalizedText(" {0}: {1}"));
   }
 
   private static MatchPlayer getApproximatePlayer(Match match, String query, CommandSender sender) {
--- a/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
@@ -53,8 +53,7 @@ public class ChatDispatcher implements Listener {
   private static final String DM_SYMBOL = "@";
   private static final String ADMIN_CHAT_SYMBOL = "$";
 
-  public static final String ADMIN_CHAT_PREFIX =
-      "[" + ChatColor.DARK_RED + "A" + ChatColor.WHITE + "]";
+  public static final String ADMIN_CHAT_PREFIX = "[" + ChatColor.RED + "A" + ChatColor.WHITE + "]";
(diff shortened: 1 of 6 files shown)


Xerocoles:
You should use the lighter-red instead of the dark red, cause that's the same colour as the red on most maps' xmls, so that would blend in with chat more unless you do the lighter-red (which looks totally fine anyways)

applenick:
@Xerocoles good idea.

Updated format:
Screen Shot 2020-03-14 at 10 28 12 AM

Brottweiler:
Not sure I want `[A]` to be red lol, I don't see how it was a problem for it for be orange like before.

Xerocoles:
I'm realizing light red could also potentially be not ideal due to mod flairs (which is a lot of the people using /a) being the same colour. That being said, I think it still works fine, and it's defo better than dark red.

But do NOT go back to orange please, it was definitely a problem.  You couldn't instantly tell which it was -- sometimes I'd have two conversations going on, one in admin chat, one in PMs, and there were _multiple_ times where me or someone else would be responding in the wrong format (ie, another staff member pms you, but you see its a staff and you respond in /a, which includes the rest of staff who didnt see the first part of that convo.  That's super easily fixed, you just have two diff colours for A or PM.  I think light red is fine, but you could do a ton of different colours)

applenick:
@Brottweiler I have to agree with @Xerocoles. Seeing what it looks like with a separate formatting, it’s a wonder why it wasn’t like this from the beginning. Plus, it’s only people with permission who will ever see this formatting. I suggested that too since if it were say the DM color being altered, then it would cause more confusion since people are already used to the gold for DM colors. 

If you dislike the light red color, I’d be happy to alter it to whatever color you prefer. Though personally I like the light red 🤷‍♂

Brottweiler:
What about aqua? Or would that not stand out as much?

applenick:
Screen Shot 2020-03-15 at 9 47 46 AM
Let me know which you like best.

applenick:
Screen Shot 2020-03-15 at 10 43 24 AM
Updated the private message format, based off of the OCN private message format provided by @Pablete1234. And now the admin chat can stay gold! 🎉 

I think this looks good, however I wonder if the `( )` should be formatted with a dark gray?

@Brottweiler & @Xerocoles any thoughts?

Brottweiler:
You might be able to skip the brackets entirely and also not make it italics?

applenick:
I like the idea of no brackets, ever cleaner! I kinda like the italics but I’ll try it without and see how it looks.

applenick:
@Brottweiler here is the suggested format.

Screen Shot 2020-03-15 at 10 52 39 AM

Brottweiler:
That looks good. I am indifferent if you choose to use italics or not, though, but it definitely looks better without brackets.

applenick:
Alrighty I decided to add back the italics, personally feel it looks better. 
Screen Shot 2020-03-15 at 12 45 55 PM

Let me know if there are any other changes 😄

Pablete1234:
Italics look better

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -217,6 +218,12 @@ public void sendReply(Match match, Audience audience, MatchPlayer sender, @Text
     sendDirect(match, sender, receiver.getBukkit(), message);
   }
 
+  private static PersonalizedText formatPrivateMessage(String key) {
+    return new PersonalizedText(
+        new PersonalizedTranslatable(key).getPersonalizedText().color(ChatColor.GRAY).italic(true),
+        new PersonalizedText(" {0}: {1}"));
```
TheMolkaPL: I think this should be translatable too.

applenick: @TheMolkaPL In a previous PR(not sure which of mine) I’m pretty sure someone stated they would rather keep formatting “styles" outside of the translation files. I’m been trying to adhere to that. Do you think this situation would be acceptable to be translatable?

TheMolkaPL: In Arabic this should be `TEST :applenick3 [A]` since they read right-to-left.

Pablete1234: Excuse me, but i'm not sure this makes much sense. Minecraft doesn't support right-to-left to start with, and it's not like changing order of parameters in formatting is the way that you apply RTL

applenick:
I worked on cleaning up some things in ChatDispatcher, such as adding some missing translations. 

The new command I’ve added, `/say (team) (message)` allows for staff members to send messages (while observing) to other teams. 

Screenshot of a few formatting things:
Screen Shot 2020-03-19 at 4 59 31 PM
As you can see, observing staff will be able to speak in individual team chats to notify teams of certain things. This should prove useful say if a player is spamming in `/t` and instead of a staff having to say a message in global, they can tell that player in their team chat. 
Also pictured are the final formats for [A] and /msg from previous discussion, in case anyone wanted to see.

applenick:
@Electroid I’m moving this out of Draft as I can’t think of anything else to change with this. Let me know if there are any changes you would like and I’ll take care of them as soon as I can.

applenick:
@Pablete1234 I agree, a new setting system would really make things easier 😆 

It’s tricky, as I’m trying to have it so users (not staff) are unable to view any setting that does not relate to them. Otherwise a simple `/toggle adminsounds on/off` would have worked very well to resolve this xD So atm, all extra sounds such as raindrops/report sounds are disabled when set to `messages`. I feel this is acceptable for the short-term, as I believe most people will either want all sounds or simply none. 

Let me know if there are any changes regarding this you feel would be wise to adjust, otherwise thanks for the feedback!

Brottweiler:
>as I’m trying to have it so users (not staff) are unable to view any setting that does not relate to them

FWIW, when a player `/toggle chat` they see can switch to admin chat setting, which obviously doesn't do anything for them, but they still see it as an option currently.

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -172,54 +188,91 @@ public void sendDirect(Match match, MatchPlayer sender, Player receiver, @Text S
...
+        match,
+        player,
+        message,
+        party.getChatPrefix().toLegacyText() + PREFIX_FORMAT,
+        viewer ->
+            party.equals(viewer.getParty())
+                || (viewer.isObserving()
+                    && viewer.getBukkit().hasPermission(Permissions.ADMINCHAT)),
+        SettingValue.CHAT_TEAM);
   }
```
Electroid: I think this is neat, but we don't need it. If staff want to say something, they can just use global, rarely do you need to address a single team.

PR #366, opened 2020-03-21
Add Minecraft server files to gitignore

rmsy:
These files are generated when you run the embedded Minecraft server in the project directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,17 @@ Commons/core/src/main/i18n/translations/*
 
 # Minecraft generated files
 maps.zip
+banned-ips.json
+banned-players.json
+ops.json
+help.yml
+server.properties
+sportpaper.yml
+update
+usercache.json
+whitelist.json
+logs/
+plugins/
+world/
+world_nether/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -34,3 +34,17 @@ Commons/core/src/main/i18n/translations/*
...
+server.properties
+sportpaper.yml
+update
+usercache.json
+whitelist.json
+logs/
+plugins/
+world/
+world_nether/
+world_the_end/
```
TheMolkaPL: Aren't these inside the `matches/` directory?

rmsy:
For matches, maybe (it’s been a while) - but﻿ I believe these are the directories that are generated when the server starts initially.

.gitignore
rmsy: For matches, maybe (it’s been a while) - but﻿ I believe these are the directories that are generated when the server starts initially

Pablete1234:
Should just have a server folder added to the gitignore, and change the working directory to there

PR #367, opened 2020-03-22
Allow external plugins to manage tablist player order

Meeples10:
Signed-off-by: Meeples10 <mc.meeples10@gmail.com>

I realize that reflection is bad, and I would be happy to change it if anyone has a better idea.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/tablist/MatchTabManager.java
+++ b/core/src/main/java/tc/oc/pgm/tablist/MatchTabManager.java
@@ -43,9 +43,11 @@ public class MatchTabManager extends TabManager implements Listener {
       new DefaultMapAdapter<>(new FreeForAllTabEntry.Factory(), true);
 
   private BukkitTask renderTask;
+  private PlayerOrder.Factory playerOrderFactory;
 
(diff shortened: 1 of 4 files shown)


Pablete1234:
Use a factory stype instead of reflection.

Also i don't see why you need a player order wrapping every player, just make it a Comparator<MatchPlayer>

Electroid:
```java
public void setPlayerOrder(Comparator<MatchPlayer> order) {}
```

Pablete1234:
This is not going to work as expected. The ordering has to be viewer dependant (you see yourself first).

The DefaultPlayerComparator you added, is redundant.
Add something along the lines of:

```java
public class PlayerOrder implements Comparator<MatchPlayer> {
    public static class Factory implements PlayerOrderFactory {
        @Override
        public Comparator<MatchPlayer> getOrder(MatchPlayer viewer) {
            return new PlayerOrder(viewer);
        }
    }

    // ... The rest of the current player order
}
```

then, the default PlayerOrderFactory would be `this.playerOrderFactory = new PlayerOrder.Factory();`
if via the api, the factory is changed, it will provide other type of comparators for player

PR #370, opened 2020-03-23
Translation improvements

rmsy:
This adds a CrowdIn configuration file for [our CrowdIn project](https://crowdin.com/project/pgm). I removed the integration with `Electroid/PGM` and linked it to `rmsy/PGM`, so that I could test that the integration was working smoothly. It just needs to be relinked to the correct repository, which can be done [here](https://crowdin.com/project/pgm/settings#integration), **after** this PR is merged. Once that's done, CrowdIn should automatically create a pull request to merge in the current approved translations from the various languages. As soon as that's merged, translations will work out-of-the-box.

Additionally, I fixed various rendering issues throughout PGM where translations were not being displayed. There are still a few sticklers (some error messages, `/map`, etc.), but this gets us probably 85% of the way there.

One issue I do need to figure out -- Spanish accents are not being rendered correctly, for some reason:
image

They are stored and encoded correctly in the output files, as far as I can tell. Not sure what's going on there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-PGM ![build](https://github.com/Electroid/PGM/workflows/build/badge.svg)
+PGM ![build](https://github.com/Electroid/PGM/workflows/build/badge.svg) [![Crowdin](https://badges.crowdin.net/pgm/localized.svg)](https://crowdin.com/project/pgm)
 ===
 
 The original PvP Game Manager for Minecraft.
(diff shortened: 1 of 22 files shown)


KingOfSquares:
Does this change depending on which version you are on?

rmsy:
> Does this change depending on which version you are on? (Spanish characters bugging out)

Good question. I believe I was on a 1.8.x version; I have not tested with any other versions.

EudaldC:
@rmsy it should work in 1.8, Spanish translations have been around way earlier than 1.8. It's an encoding problem, at some point the text is being converted from UTF8 to ASCII.

rmsy:
@EudaldC Thank you for the tip! [By default, `ResourceBundles` are loaded as ISO-8859-1](https://stackoverflow.com/a/4660195/559330), apparently. I submitted #375 to address.

PR #376, opened 2020-03-25
Small fixes for match stats

KingOfSquares:
Fix list as of last commit:
-Fix formatting so KD includes up to two decimals.
-Change "stats.best" into "Top stats this match"
-Fix an issue with offline players having the best results resulting in an error for invalid UUIDs
-Cover for NPE when match is ended without any playerstats

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
@@ -37,7 +37,7 @@ public class StatsMatchModule implements MatchModule, Listener {
   private final Map<UUID, PlayerStats> allPlayerStats = new HashMap<>();
   // Since Bukkit#getOfflinePlayer reads the cached user files, and those files have an expire date
   // + will be wiped if X amount of players join, we need a seperate cache for players with stats
-  private final Map<UUID, Player> cachedPlayers = new HashMap<>();
+  private final Map<UUID, String> cachedUsernames = new HashMap<>();
 
   public StatsMatchModule(Match match) {
     this.match = match;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/modules/StatsMatchModule.java
```diff
@@ -167,6 +173,19 @@ public void onMatchEnd(MatchFinishEvent event) {
...
+  public void onPlayerLeave(PlayerQuitEvent event) {
+    Player player = event.getPlayer();
+    if (allPlayerStats.containsKey(player.getUniqueId()))
+      cachedPlayers.put(player.getUniqueId(), player);
+  }
+
+  @EventHandler
+  public void onPlayerJoin(PlayerJoinEvent event) {
+    UUID playerUUID = event.getPlayer().getUniqueId();
+    cachedPlayers.remove(playerUUID);
```
Electroid: Do you have these two methods mixed up? You're adding to cached on leave and removing on join?

KingOfSquares: I may have used "cahce" wrong. I need to store every player that has stats for each match that logs off in case they get a top stat, because Bukkit might lose track of them if too many new players join.
If they log back on i dont need to store them anymore.
Do you have any suggestions for something else i could call this?

Electroid: We can't keep `Player` objects in-memory after someone leaves, this will cause memory leaks. Instead, you would need to store their username, as a string.

KingOfSquares: Ok!

KingOfSquares: But is there another more suitable name for the storage?

applenick: @KingOfSquares could do 
`private final Map<UUID, String> cachedUsernames = new HashMap<>();`

KingOfSquares: Ok! I was just wondering if not to use "cache", thanks :)

KingOfSquares:
Does this need anything more? I've added all i wanted.
Currently investigating a bug with kills in the void not registering, but that's nothing serious.

PR #378, opened 2020-03-26
Add sound to custom projectile impact

applenick:
# Custom projectile sounds 🏹 
I was playing a match of Glacial Impact earlier and noticed that custom projectiles (snowballs) did not play the very helpful _ding_ when impacting a target. 

This PR makes it so the same _ding_ played for arrows, also play for custom projectiles.

I don’t believe there is much to improve upon, but if there is let me know. Otherwise should be good to go 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
@@ -19,6 +19,7 @@ import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.events.ListenerScope;
 import tc.oc.pgm.projectile.EntityLaunchEvent;
 import tc.oc.util.bukkit.chat.Sound;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
```diff
@@ -123,6 +127,16 @@ public void fixEntityDamage(EntityDamageByEntityEvent event) {
...
           event.getEntity().setVelocity(velocity);
         }
+
+        // If the projectile is not an arrow, play an impact sound.
+        if (event.getEntity() instanceof Player
+            && (projectile instanceof Projectile && !(projectile instanceof Arrow))) {
+          Projectile customProjectile = (Projectile) projectile;
+          if (customProjectile.getShooter() instanceof Player) {
+            Player bukkitShooter = (Player) customProjectile.getShooter();
+            SoundsMatchModule.playSound(match.getPlayer(bukkitShooter), PROJECTILE_SOUND);
```
Pablete1234: I am not convinced that this is an optional sound. Vanilla minecraft does the ding no matter your sound setting, and this also should

Electroid: I would agree, let's keep it default.

applenick: @Pablete1234 and @Electroid thanks for clarifying this. I was unsure about whether to make it default or not and I agree, it’s been adjusted!

PR #379, opened 2020-03-26
Add a line in /map showing which pools a map is in, if any.

### LINKED ISSUE #257: Add map pool the map is present in to /map
Make the `/map` command display what map pool(s) the map is in, if any

KingOfSquares:
Closes #257 
Title. Looks like this:
2 or more pools will display like this

1 pool will display like this

How it looks if map is in no pools

### DIFF
--- a/core/src/main/java/tc/oc/pgm/commands/MapCommands.java
+++ b/core/src/main/java/tc/oc/pgm/commands/MapCommands.java
@@ -14,6 +14,7 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
@@ -21,6 +22,7 @@ import net.md_5.bungee.api.ChatColor;
 import net.md_5.bungee.api.chat.ClickEvent;
 import net.md_5.bungee.api.chat.HoverEvent;
 import org.bukkit.command.CommandSender;
+import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.map.Contributor;
 import tc.oc.pgm.api.map.MapInfo;
@@ -28,6 +30,8 @@ import tc.oc.pgm.api.map.MapLibrary;
 import tc.oc.pgm.api.map.MapOrder;
 import tc.oc.pgm.api.map.MapTag;
 import tc.oc.pgm.commands.annotations.Text;
+import tc.oc.pgm.rotation.MapPool;
+import tc.oc.pgm.rotation.MapPoolManager;
 import tc.oc.pgm.util.PrettyPaginatedResult;
 import tc.oc.util.bukkit.chat.Audience;
 import tc.oc.util.bukkit.component.Component;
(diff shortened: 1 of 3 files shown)


util-bukkit/src/main/i18n/templates/strings.properties
```diff
@@ -547,6 +547,7 @@ command.map.mapInfo.sourceCode.tip = View the XML code that controls this map
 command.map.mapInfo.proto = Proto
 command.map.mapInfo.source = Source
 command.map.mapInfo.folder = Folder
+command.map.mapInfo.containedInMapPools = Pools
```
Pablete1234:
```suggestion
command.map.mapInfo.pools = Pools
```

core/src/main/java/tc/oc/pgm/rotation/MapPoolManager.java
```diff
@@ -152,4 +152,16 @@ public void matchEnded(Match match) {
...
+   * Checks all loaded {@link MapPool}s for a given {@link MapInfo} and returns all {@link MapPool}s
+   * that contains it
+   */
+  public List<MapPool> contains(MapInfo map) {
+    List<MapPool> containedIn = new ArrayList<>();
+    for (MapPool mapPool : this.mapPools) {
+      if (mapPool.getMaps().contains(map)) containedIn.add(mapPool);
+    }
+    return containedIn;
+  }
```
Pablete1234: This feels so specific that it looks out of place.

Just from wherever you call, do:
```java
poolManager.getPools().stream()
    .filter(p -> p.getMaps().contains(map)).collect(Collectors.toList());
```

core/src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -192,6 +196,16 @@ public void map(Audience audience, CommandSender sender, @Text MapInfo map) {
     }
 
     audience.sendMessage(createTagsComponent(map.getTags()));
+
+    if (PGM.get().getMapOrder() instanceof MapPoolManager) {
+      List<MapPool> mapPools = ((MapPoolManager) PGM.get().getMapOrder()).contains(map);
```
Pablete1234:
```
String pools = ((MapPoolManager) ..).getPools().stream()
    .filter(p -> p.getMaps().contains(map)).collect(Collectors.joining(", "));
```

core/src/main/java/tc/oc/pgm/commands/MapCommands.java
```diff
@@ -246,6 +260,17 @@ private static Component createPlayerLimitComponent(CommandSender sender, MapInf
...
+    PersonalizedText pools = new PersonalizedText();
+    int iteration = 1;
+    for (MapPool mapPool : mapPools) {
+      pools.extra(mapPool.getName());
+      if (mapPools.size() > iteration) pools.extra(", ");
+      iteration++;
+    }
+    return new PersonalizedText(pools, ChatColor.GOLD);
+  }
+
```
Pablete1234: this can go away
```suggestion
```

core/src/main/java/tc/oc/pgm/commands/MapCommands.java
KingOfSquares: .joining gives an error

KingOfSquares: With the filter it will give me all `MapPool`s i want right. But how do i make a Collector that collects all those `MapPool`s names.

KingOfSquares: Nvm, i think i figured it out!

PR #385, opened 2020-03-29
Make ChatDispatcher compatible with plugins that listen for AsyncPlayerChatEvent

### LINKED ISSUE #304: Make ChatDispatcher compatible with plugins that listen to AsyncPlayerChatEvent
This will likely involve firing an `AsyncPlayerChatEvent` before sending the message in `ChatDispatcher`, and checking if its cancelled.

KingOfSquares:
Closes #304 

This will fire an AsyncPlayerChatEvent for every message sent with ChatDispatcher. If that event is canceled, the message will not be sent. 

This will also work with any feature that reads:
-The player sending the message
-The message sent
-The recipients of the message

This was tested with: 
https://www.spigotmc.org/resources/mute-system-1-8-x-1-14-x.15570/
https://www.spigotmc.org/resources/mute-chat-1-8-1-12.27156/

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
@@ -239,6 +239,9 @@ public class ChatDispatcher implements Listener {
 
   @EventHandler(ignoreCancelled = true, priority = EventPriority.MONITOR)
   public void onChat(AsyncPlayerChatEvent event) {
+    // Ignore test messages from method "send"
+    if (event.getMessage().contains("8af5859a-1d6e-4fbd-ae5f-4c63ff630fa4")) return;
+
     event.setCancelled(true);
 
     final MatchPlayer player = manager.getPlayer(event.getPlayer());
@@ -302,6 +305,25 @@ public class ChatDispatcher implements Listener {
       String format,
       Predicate<MatchPlayer> filter,
       @Nullable SettingValue type) {
+
+    // Cancel the message if some other plugin cancels chat events for this player
+    if (sender != null) {
+      AsyncPlayerChatEvent event =
+          new AsyncPlayerChatEvent(
+              false,
+              sender.getBukkit(),
+              // This message prevents PGM from sending the test as a message
+              message + " 8af5859a-1d6e-4fbd-ae5f-4c63ff630fa4",
+              match.getPlayers().stream()
+                  .filter(filter)
+                  .map(MatchPlayer::getBukkit)
+                  .collect(Collectors.toSet()));
+
+      match.callEvent(event);
+
+      if (event.isCancelled()) return;
+    }
+
     // When a message is empty, this indicates the player wants to change their default chat channel
     if (message == null) {
       try {
--- a/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
@@ -323,23 +323,25 @@ public class ChatDispatcher implements Listener {
       Predicate<MatchPlayer> filter,
       @Nullable SettingValue type) {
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -238,6 +238,9 @@ public void sendReply(Match match, Audience audience, MatchPlayer sender, @Text
 
   @EventHandler(ignoreCancelled = true, priority = EventPriority.MONITOR)
   public void onChat(AsyncPlayerChatEvent event) {
+    // Ignore test messages from method "send"
+    if (event.getMessage().contains("8af5859a-1d6e-4fbd-ae5f-4c63ff630fa4")) return;
```
Electroid: This is too hacky, you'll need to develop another way.

Electroid: You could keep a `Cache` of recently sent `AsyncPlayerChatEvent`s.

KingOfSquares: I did a version of that now. Thanks for feedback.

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -301,6 +321,27 @@ public void send(
...
+          new AsyncPlayerChatEvent(
+              false,
+              sender.getBukkit(),
+              message,
+              match.getPlayers().stream()
+                  .filter(filter)
+                  .map(MatchPlayer::getBukkit)
+                  .collect(Collectors.toSet()));
+
+      messageCache.put(event, "boring message");
```
Electroid, line 338: Just use a `""` string and explain why it's empty.

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -301,6 +321,27 @@ public void send(
...
+
+      match.callEvent(event);
+
+      if (event.isCancelled()) return;
```
Electroid, line 342: We need to respect any changes to the event, say if the message is changed or the viewers are changed. So more than just check if cancelled.

KingOfSquares: Of course. I totally forgot that haha. Thanks!

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -302,6 +322,29 @@ public void send(
...
+      AsyncPlayerChatEvent event =
+          new AsyncPlayerChatEvent(false, sender.getBukkit(), message, recipients);
+
+      messageCache.put(event, ""); // The value in the Cache is not used for anything
+
+      match.callEvent(event);
+
+      if (!event.getMessage().equals(message)) message = event.getMessage();
+      if (!event.getRecipients().equals(recipients)) recipients = event.getRecipients();
+      if (!event.getFormat().equals(format)) format = event.getFormat();
```
applenick, line 344: Figured out the issue.

You’re replacing `format` with the format provided by `AsyncPlayerChatEvent`, which in return is following the String [format syntax](https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html#syntax).

I would suggest simply omitting this check, as `send()` is provided a specific format to PGM (team/global/admin) and I don’t believe we need to care about changes made for that. 

Tested this change locally and all works well.

KingOfSquares: Ah, great catch. I will remove the check. Thanks!

Electroid:
See 7a55789c034d3de1e85c5626762a55071e57b02b

PR #387, opened 2020-03-29
Freeze world after match finish

### LINKED ISSUE #92: Lava keeps flowing when match is not running
1. break a core
2. wait until match ends
3. lava keeps flowing

EricZeiberg:
The world should freeze after completion of a match, so liquids don't leak further and TNT that has been ignited but not yet exploded does not cause any more damage. Fixes #92. 

My first PR to this codebase so let me know if I implemented this the right way, feedback is welcome.


Electroid:
Thanks for contributing! Welcome :)

PR #389, opened 2020-03-30
Players now stick to their originally picked teams

### LINKED ISSUE #328: Make players stick to their team
If a player auto joins a match, they should be stuck with the team they first joined. If they join observers, it shouldn't matter if the other team has less players, they should auto join to the last team they were on.

This would also be helpful for premium players. If they join observers, they don't have to remember what team they were on. Just auto join and you end up on the same team you previously left.

EricZeiberg:
This changes the behavior of joining teams slightly as now, by default, all players will be placed back on their originally picked teams if they leave the game or join observers. Donors can switch teams, if they would like, by explicitly selecting a certain team, but if no specific team is selected, they too will be placed back on their original team. 

All other joining behavior (full team / teams) should stay the same. Fixes #328.


core/src/main/java/tc/oc/pgm/modules/TeamRestrictMatchModule.java
```diff
@@ -0,0 +1,53 @@
...
+    return playerTeamMap;
+  }
+
+  public Team getLastTeam(UUID id) {
+    Party lastParty = playerTeamMap.get(id);
+    if (lastParty instanceof Team) {
+      return (Team) lastParty;
+    } else {
+      return null;
+    }
```
Pablete1234: If you won't return the party unless it's a team, why don't you just store teams to start with?

EricZeiberg: PlayerJoinPartyEvent returns a specific party, plus for FFA matches, you don't want to restrict someone joining. This way, only if its a team based match, will someone be restricted.

Electroid:
Although, `TeamMatchModule` is big, this should go in there instead of a separate module.

core/src/main/java/tc/oc/pgm/modules/TeamRestrictMatchModule.java
Pablete1234: That wasn't my point. The idea is, instead of storing a `Map<UUID, Party>`, you should just store teams in a `Map<UUID, Team>` and ignore all of the adding logic if it isn't a team. Instead of testing for the type of party afterwards, test for it beforehand.

EricZeiberg: Fixed, thanks.

PR #403, opened 2020-04-14
Check for null when right clicking items in observer mode

Booquefius:
First this checks to see if the clicked item from the event is null and then if so just stops right there. If not it makes sure that the event is a right click, if that item is not null and if the type of the item is equal to the Team Join button's material. If so the code continues.


core/src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
```diff
@@ -312,8 +312,11 @@ public void run() {
 
   @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
   public void cancelArmorEquip(ObserverInteractEvent event) {
-    if (event.getClickType() == ClickType.RIGHT
-        && event.getClickedItem().getType() == Button.TEAM_JOIN.material) {
+    if (event.getClickedItem() == null) {
+      return;
+    }
+
```
Electroid, line 318: You don't need this. You check below does that.

Electroid: Also run `mvn clean install` so the code formatter runs.

KingOfSquares:
fixed in f08ce4d

PR #411, opened 2020-04-18
Change visibility of PlayerOrder constructor and add getter for viewer field

Meeples10:
Signed-off-by: Meeples10 <mc.meeples10@gmail.com>


EricZeiberg:
What's the reasoning for this change?

KingOfSquares:
Im guessing it's for rendering the tablist based on who is watching?

Meeples10:
>What's the reasoning for this change?

@EricZeiberg External plugins can't extend `PlayerOrder` without a public constructor and a getter for the viewer.

Electroid:
This is now on master. Be cautious though, PGM doesn't have an official API, this might break in the future.

Issue #326, opened 2020-02-17
Community edition


Electroid:
Figuring out what belongs in PGM and what doesn't is a tricky balancing act, and it's been difficult in the past. On one had, PGM works great "out-of-the-box" with features such as `/a` or settings. On the other hand, recent proposals and issues for #300, #305, #317, #280 highlight that putting them all in one package might be problematic in the future.

**For developers:**
I've already started breaking up PGM into "maven modules" such as `api`, `core`, `util`, and now I am considering adding a new module, `community`, which is essentially a "catch-all" for things that are useful for "scrimmage servers" but do not need to be in `core`. 

Existing features like `/report` would be moved to `community` and recently rejected proposals for `/punish`, `/freeze`, and other moderation tools would be allowed back under `community`.

The `community` module would be slightly special, because it *optionally* depends on PGM. You can run the `Community` plugin without PGM, say on a map development server, but if PGM is present, then extra functionality that requires `Match` access would also be present. This module could also depend on `util` so we don't have to reinvent the wheel with translations and other things.

**For users:**
Commands like `/report`, `/punish`, `/freeze` and other continue to have support and default functionality in PGM. Commands like `/vanish` would be reconsidered under the new community scheme.

**For server owners:**
The PGM jar has the `community` module included by default, with an optional config option to remove it. The `community` jar would also be available for a separate download, for other use cases such as map development servers.

User and developer feedback is important, as well as compromise for a project like this one. It may take some time to get the module separation onto master, so I would be willing to fast-track some of the rejected moderation PRs and refactor them out later. Comments and feedback would be great!

cc @applenick @Pablete1234 @Brottweiler @KingOfSquares @AustinLMayes

PR #425, opened 2020-04-22
Introduce non-dynamic map pools

applenick:
# Map Pool Improvements 🗺️ 
Changes to map pools, including a command to set specific map pools and the introduction of non-dynamic pools. 

## Non-Dynamic Pools
Running events can be difficult atm due to map pools dynamically adjusting to accommodate player counts. The solution is to introduce non-dynamic pools. When a non-dynamic pool is set, PGM will not adapt to different player counts as long as a staff member is online (this can be disabled via config if desired) and the non-dynamic pool is set. If the match ends and no staff are online, it will revert to an appropriate dynamic pool or this can be done manually.

Changes to `map-pools.yml`:
The major change that has been made to the map-pool file is the addition of a `dynamic:` field. By default no changes are required for current servers, as all map-pools will be set as dynamic unless specified false. The following config would be used if for example an all-blitz map pool was desired:
```yml
pools:
  blitz:
    type: ordered
    enabled: true
    dynamic: false
    players: 1
    maps: ...
```
When dynamic is set to false. This pool will never be set automatically by player counts, so `players:` is irrelevant. 

If a staff member wants this pool to be set, they would simply execute `/setpool blitz` 

Also added is a new event, `MapPoolAdjustEvent` which is called whenever the map pool is switched (both dynamically and non, including a isForced to determine if it was set by command). I'm hoping this will allow other plugins to listen for the event in order to add cool features, for example broadcasting a special message when a specific map pool is loaded.

## Other Changes
- Add a `random` option to map pools 
  - When `type: random` is set in the map-pools.yml, it will follow a rotation style but at complete random.
- Add a staff broadcast to `/setnext` 
- Adjust `/pools` format to include # of maps (non-dynamic pools do not show player counts)
- Fix so blitz maps don't bug dynamic pools (instead of counting obs at `0.5`, upped to `0.85` for blitz. Thanks @Pablete1234 )
- Renamed a method from ModerationCommands to better fit its purpose. 

## Screenshots
Additional information shown under `/pools` includes the map count per pool. Also non-dynamic pools can be identified by not having a players field.
Screen Shot 2020-04-21 at 10 41 19 PM

When `/setpool <pool>` is used, a useful message will be broadcast to inform staff.
Screen Shot 2020-04-21 at 10 44 16 PM

Similar to the previous screenshot, now staff will be alerted when a map is `/setnext`
Screen Shot 2020-04-21 at 10 41 54 PM

## Feedback
After performing testing for a few days, I've yet to find any major problems so everything should be good :thumbsup:

As always I am open to feedback of any kind and will make changes where needed.

Only design question I had:
1. Was it a good idea to add the `map-pools.require-staff` to config, or was that unnecessary? 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
+++ b/core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
@@ -20,7 +20,6 @@ import tc.oc.pgm.api.map.MapInfo;
 import tc.oc.pgm.api.map.MapOrder;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.party.Competitor;
-import tc.oc.pgm.community.commands.ModerationCommands;
 import tc.oc.pgm.cycle.CycleCountdown;
 import tc.oc.pgm.listeners.ChatDispatcher;
 import tc.oc.pgm.restart.CancelRestartEvent;
@@ -32,6 +31,7 @@ import tc.oc.pgm.start.StartMatchModule;
 import tc.oc.pgm.timelimit.TimeLimitCountdown;
 import tc.oc.pgm.timelimit.TimeLimitMatchModule;
 import tc.oc.pgm.util.StringUtils;
+import tc.oc.pgm.util.UsernameFormatUtils;
 import tc.oc.pgm.util.chat.Audience;
 import tc.oc.pgm.util.component.Component;
 import tc.oc.pgm.util.component.types.PersonalizedText;
@@ -128,7 +128,7 @@ public class AdminCommands {
     Component successful =
         new PersonalizedTranslatable(
                 "command.admin.set.success",
-                ModerationCommands.formatStaffName(sender, match),
+                UsernameFormatUtils.formatStaffName(sender, match),
                 mapName)
             .getPersonalizedText()
             .color(ChatColor.GRAY);
@@ -138,7 +138,7 @@ public class AdminCommands {
   @Command(
       aliases = {"setpool", "setrot"},
       desc = "Set a custom map pool or revert to dynamic",
-      usage = "[pool name] -r (revert to dynamic)",
+      usage = "[pool name] -r (revert to dynamic) -t (time limit for map pool)",
       flags = "r",
       perms = Permissions.SETNEXT)
   public static void setPool(
@@ -146,7 +146,8 @@ public class AdminCommands {
       Match match,
       MapOrder mapOrder,
       @Nullable String poolName,
-      @Switch('r') boolean reset)
+      @Switch('r') boolean reset,
+      @Switch('t') Duration timeLimit)
       throws CommandException {
     if (!match.getCountdown().getAll(CycleCountdown.class).isEmpty()) {
       sender.sendMessage(
@@ -162,11 +163,12 @@ public class AdminCommands {
             : mapPoolManager.getMapPoolByName(poolName);
 
     if (newPool == null) {
-      Component noPoolError =
-          new PersonalizedTranslatable("command.pools.noPoolMatch")
+      Component error =
+          new PersonalizedTranslatable(
+                  reset ? "command.pools.noDynamic " : "command.pools.noPoolMatch")
               .getPersonalizedText()
               .color(ChatColor.RED);
-      sender.sendMessage(noPoolError);
+      sender.sendMessage(error);
     } else {
       if (newPool.equals(mapPoolManager.getActiveMapPool())) {
         sender.sendMessage(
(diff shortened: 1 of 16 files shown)


Pugzy:
Something I had suggested in Discord was naming the pool type to be `shuffled` to better reflect that it's the ordering thats mixed up rather than it randomly picking maps each cycle. 

Also an idea for the setpool command, what if you could specify either a number or duration after the pool name so `/setpool blitz 1h` would stick to the pool until the time is exceed and another `/setpool gear 2` would do two maps from the pool.

applenick:
@Pugzy A few seconds before your comment I made the change from random to `shuffled` haha.

Specifying a duration seems like it would be a very useful feature, I’ll look into that as well as the # of maps idea. I’ll most likely include it in this PR or perhaps a followup, depending on how much changes. 

Thanks for the feedback!

core/src/main/java/tc/oc/pgm/events/MapPoolAdjustEvent.java
```diff
@@ -0,0 +1,63 @@
...
+
+import javax.annotation.Nullable;
+import org.bukkit.command.CommandSender;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.rotation.MapPool;
+
+/** MapPoolAdjustEvent is called when the active {@link MapPool} is set to another * */
+public class MapPoolAdjustEvent extends Event {
```
Electroid, line 11: Not sure this event is necessary. Since you're just printing to chat, I would skip the event and just print.

EricZeiberg:
Another cool idea would be to add a string argument to `/pool` where you can specify the specific pool and see all the maps in it. I believe currently you can only see the maps for the pool that's being played.

core/src/main/java/tc/oc/pgm/events/MapPoolAdjustEvent.java
applenick: So personally I plan on having a plugin which is going to listen to this event, to be used in order to perform other cool stuff. As I wanted to keep specific features out of PGM for the sake of not evading the scope. Is that alright or would you still like it removed?

Electroid: Hmm, I'm okay with keeping this for now, but there's no guarantees it will still exist when the API is published and 1.0 is released. So you can keep it for now.

core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
```diff
@@ -12,23 +12,30 @@
...
+import net.md_5.bungee.api.ChatColor;
 import org.bukkit.command.CommandSender;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.map.MapInfo;
 import tc.oc.pgm.api.map.MapOrder;
 import tc.oc.pgm.api.match.Match;
-import tc.oc.pgm.api.match.MatchManager;
 import tc.oc.pgm.api.party.Competitor;
+import tc.oc.pgm.community.commands.ModerationCommands;
```
KingOfSquares, line 23: Based on the wording in #326 i do not think PGM should depend on the `community` module being present.

core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
```diff
@@ -115,13 +124,61 @@ public static void setNext(
...
+    }
+    MapPoolManager mapPoolManager = MapPoolCommands.getMapPoolManager(sender, mapOrder);
+    MapPool newPool =
+        reset
+            ? mapPoolManager.getAppropriateDynamicPool(match).orElse(null)
+            : mapPoolManager.getMapPoolByName(poolName);
+
+    if (newPool == null) {
+      Component noPoolError =
+          new PersonalizedTranslatable("command.pools.noPoolMatch")
```
KingOfSquares, line 166: Im guessing this will be displayed if someone does `/setpool -r` and there is no dynamic pools present? I think there should be a more intuitive response if so (e.g "No dynamic pools loaded") 

This could maybe be located in `getAppropriateDynamicPool`?

core/src/main/java/tc/oc/pgm/rotation/MapPoolManager.java
```diff
@@ -170,15 +172,30 @@ public void setNextMap(MapInfo map) {
...
-  @Override
-  public void matchEnded(Match match) {
-    int activePlayers = match.getPlayers().size() - (match.getObservers().size() / 2);
-
-    mapPools.stream()
+  public Optional<MapPool> getAppropriateDynamicPool(Match match) {
+    int obs =
+        match.getModule(BlitzMatchModule.class) != null
+            ? (int) (match.getObservers().size() * 0.85)
+            : (match.getObservers().size() / 2);
```
KingOfSquares, line 179: What about `*0.5` so it stays consistent with the line above?

applenick:
I realized I never uploaded a screenshot I took of the new time limit for map pools.
Screen Shot 2020-04-22 at 5 48 39 PM

PR #426, opened 2020-04-22
Add support for multi-version NBT tags

Electroid:
Allows PGM to add arbitrary item tags without depending on 1.8 NMS.


Electroid:
Going to break up this PR into smaller stuff. I'll start by merging the NBT, because that's good to go.

PR #429, opened 2020-04-25
Split datastore implementation into a driver and add async APIs

rafibaum:
This PR is the first of a larger refactor to add in MySQL support to PGM. It separates out the SQLite specific semantics into a separate class which can then be swapped out for a different database backend if desired. Calling classes have been refactored to use async API's where it makes sense.


rafibaum:
Closing. Electroid implemented it on his branch.

PR #431, opened 2020-04-26
Begin migration away from SportPaper APIs

Electroid:
A lot of this PR is also moving away from Minecraft 1.8 dependencies, to allow for multi-version.

PR #434, opened 2020-04-29
Vanish

applenick:
# Vanish 👻 
This PR implements a vanish mode for staff members, allowing them to appear offline to regular players. 

When a player is vanished the following occurs:
1. A leave message is broadcasted
2. Forced to observing team, and will be unable to join using picker or command
3. Visibility is reset so only other staff can see you
4. Tab name is completely removed for regular players (staff remain visible)

**Additional notes**
* Vanish status persists until the server restarts
* A broadcast is sent to admin chat when a vanished staff member actually joins/quits
* Attempting to /msg <vanished player> will send an user not found response
* All global/team/dms are disabled for vanished players (only admin chat)
* Vanished players are excluded from `/list`

## Commands
`/vanish` - Will vanish you from the server, making it appear like you are offline. 

`/unvanish` - Will make you visible again to players, will seem as if you join rejoined.

`/list` - Displays a list of online players (now grouped by teams or participants)

## Screenshots

First is an example of using `/vanish` from the perspective both staff member and regular.

Staff:
staff-view

Regular:
player-view

As you can see the staff member is sent a message along with refreshing their observer kit to remove the picker. From the regular view, it appears as if the staff member has quit the server.

Below is the broadcast message sent to admin chat when a vanished staff member joins/quits. 
Screen Shot 2020-04-28 at 11 27 47 PM

The final set are of the new `/list` command, as you will see it is neat and organized by teams. Also included an example view of when a staff member views the list, as staff will be able to see players who are vanished as identified with a strikethrough (same as the tab list). 
Screen Shot 2020-04-28 at 11 20 46 PM

Screen Shot 2020-04-28 at 5 15 36 PM

Screen Shot 2020-04-28 at 5 15 26 PM
## Final Notes
I believe this PR handles most of the features related to vanish, however there are a few accommodations that will have to be made for some setups. 

For example on OCC, observers have access to the commandbook `/tp` command. Even if a player is vanished, if a non-staff performs `/tp <vanished player>` they will still be teleported to them. 

I plan on forking CommandBook and adding in checks that will accommodate this, however it's always possible other servers may use different plugins for various things. I’ll link my CommandBook fork here once it is ready, for future reference. 

That is something that will have to be taken into account when running PGM and using the vanished feature. I did include both a `PlayerVanishEvent` and `isVanished` metadata that is attached to the player. So there are multiple ways one could check for vanished players via separate plugins. 

As always feedback is appreciated, and I will make adjustments where necessary :thumbsup:


core/src/main/java/tc/oc/pgm/commands/UserCommands.java
```diff
@@ -0,0 +1,112 @@
...
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.community.features.VanishManager;
+import tc.oc.pgm.teams.TeamMatchModule;
+import tc.oc.pgm.util.component.Component;
+import tc.oc.pgm.util.component.Components;
+import tc.oc.pgm.util.component.types.PersonalizedText;
+import tc.oc.pgm.util.component.types.PersonalizedTranslatable;
+import tc.oc.pgm.util.named.NameStyle;
+
+public class UserCommands {
```
KingOfSquares, line 20: `UserCommands` feels so generic many other commands could fit in here, what do you think about `ListCommands`?

core/src/main/java/tc/oc/pgm/community/events/PlayerVanishEvent.java
```diff
@@ -0,0 +1,35 @@
+package tc.oc.pgm.community.events;
+
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.event.ExtendedCancellable;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+public class PlayerVanishEvent extends ExtendedCancellable {
+
+  private MatchPlayer vanisher;
+  private boolean vanished;
```
KingOfSquares, line 10: Both of these can be `final`?

core/src/main/java/tc/oc/pgm/community/features/VanishManager.java
```diff
@@ -0,0 +1,176 @@
...
+import tc.oc.pgm.util.component.ComponentRenderers;
+import tc.oc.pgm.util.component.types.PersonalizedTranslatable;
+import tc.oc.pgm.util.named.NameStyle;
+
+public class VanishManager implements Listener {
+
+  private static final String VANISH_KEY = "isVanished";
+  private static final MetadataValue VANISH_VALUE = new FixedMetadataValue(PGM.get(), true);
+
+  private List<UUID> vanishedPlayers;
```
KingOfSquares, line 41: `private final List<UUID>` ?

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -167,12 +170,27 @@ public void sendAdmin(Match match, MatchPlayer sender, @Nullable @Text String me
...
     if (matchReceiver != null) {
+
+      // Vanish Check - Don't allow messages to vanished
+      if (vanish.isVanished(matchReceiver.getId())) {
+        sender.sendMessage(
+            new PersonalizedTranslatable("command.playerNotFound")
+                .getPersonalizedText()
+                .color(ChatColor.RED));
+        return;
+      }
```
KingOfSquares, line 192: Exclude staff from this? As in staff can send vanished staff messages? Maybe its enough with adminchat, what do you think?

core/src/main/java/tc/oc/pgm/api/PGM.java
```diff
@@ -35,6 +36,8 @@
 
   ScheduledExecutorService getAsyncExecutor();
 
+  VanishManager getVanishManager();
+
```
KingOfSquares, line 40: This makes PGM depend on the vanish feature being present, which goes against #326 since it currently is a community feature. Maybe check for the vanish value in the same way you intended for the util package? (See other comment for where i'm thinking about)

core/src/main/java/tc/oc/pgm/community/features/VanishManager.java
```diff
@@ -0,0 +1,176 @@
...
+    // Keep track of the UUID and apply/remove META data, so we can detect vanish status from other
+    // projects (i.e utils)
+    if (vanish) {
+      addVanished(player.getId());
+      player.getBukkit().setMetadata(VANISH_KEY, VANISH_VALUE);
+      player.getSettings().setValue(SettingKey.CHAT, SettingValue.CHAT_ADMIN);
+    } else {
+      removeVanished(player.getId());
+      player.getBukkit().removeMetadata(VANISH_KEY, VANISH_VALUE.getOwningPlugin());
+    }
```
KingOfSquares, line 89: See other comment about #326

core/src/main/java/tc/oc/pgm/community/features/VanishManager.java
```diff
@@ -0,0 +1,176 @@
...
+
+    match.callEvent(new PlayerVanishEvent(player, vanish));
+  }
+
+  /* Commands */
+
+  @Command(
+      aliases = {"vanish", "disappear", "v"},
+      desc = "Vanish from the server",
+      perms = Permissions.STAFF)
```
EudaldC, line 118: I really dislike the approach of pgm.staff having all the perms, as it makes the plugin permissions a lot less customizable. I may want some staff groups in Stratus to be able to vanish but not freeze or lookup alts, especially with permission realms in mind, imagine a tournament setting for example.

I know it's down to @Electroid to decide in which parts of the project complexity is to be avoided, but I strongly believe that simplifying permission nodes is not worth the result.

Pablete1234:
You could probably display in the action bar a message that reminds you of your vanished status

core/src/main/java/tc/oc/pgm/community/features/VanishManager.java
Electroid: I think `pgm.vanish` should be ok.

core/src/main/java/tc/oc/pgm/api/PGM.java
Electroid: Hmm, I'd say make `VanishManager` an interface in `tc.oc.pgm.api`. With 2 implementation, a `NoopVanishManager` in `pgm` and a `VanishManagerImpl` in `community`.

core/src/main/java/tc/oc/pgm/community/events/PlayerVanishEvent.java
```diff
@@ -0,0 +1,35 @@
+package tc.oc.pgm.community.events;
+
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.event.ExtendedCancellable;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+public class PlayerVanishEvent extends ExtendedCancellable {
```
Electroid, line 7: Make sure to extend existing events, seems like this would fall under `MatchPlayerEvent`

core/src/main/java/tc/oc/pgm/community/events/PlayerVanishEvent.java
Electroid: Yes

core/src/main/java/tc/oc/pgm/community/features/VanishManager.java
Pablete1234: The issue if you make it `pgm.vanish` is that to setup permission for your staff you start getting a long list that becomes hard to manage. You end up doing permission node hunting to find out what to give and what not to give your mods access to, happened on 1.12 pgm and it became kinda hard to manage

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
applenick: Will there be any reason to message other staff members? Using `/a` should suffice in my opinion and most staff usually know each other outside of the server, so any private conversations can be held via other platforms.

applenick:
Addressed most of the issues brought up from the feedback.

I decided to give separate permissions for both vanish and freeze (`pgm.vanish` `pgm.freeze`), though I do agree with @Pablete1234’s point. It may be better longterm to switch moderation and similar commands to a `pgm.staff.<type>` format. 

Let me know if there is anything else I can adjust😄

core/src/main/java/tc/oc/pgm/api/player/VanishManager.java
```diff
@@ -0,0 +1,37 @@
...
+
+/** A manager that holds information related to {@link MatchPlayer}s who are vanished. */
+public interface VanishManager {
+
+  /**
+   * A list of {@link MatchPlayer} who are online and vanished.
+   *
+   * @return A list of {@link MatchPlayer}
+   */
+  List<MatchPlayer> getOnlineVanished();
```
Electroid, line 15: Unless it needs to be a list. `Collection<MatchPlayer> getVanished()`

core/src/main/java/tc/oc/pgm/api/player/VanishManager.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * Set the {@link MatchPlayer}'s vanish status
+   *
+   * @param player - The target {@link MatchPlayer}
+   * @param vanish - Whether the player is vanished or not
+   * @param quiet - Whether to broadcast related join/quit messages
+   * @throws CommandException When player is already vanished/unvanished
+   */
+  void setVanished(MatchPlayer player, boolean vanish, boolean quiet) throws CommandException;
```
Electroid, line 33: We shouldn't use `CommandException` here, that's a concern of the commands not the interface.

How about this method return a `boolean`?

core/src/main/java/tc/oc/pgm/commands/JoinCommands.java
```diff
@@ -47,7 +47,7 @@ public static void join(
       }
     }
 
-    if (sender.hasPermission(Permissions.JOIN_FORCE) && force) {
+    if (sender.hasPermission(Permissions.JOIN_FORCE) && force && !player.isVanished()) {
```
Electroid, line 50: Does this check need to be inside `JoinMatchModule`? I think so, because it should also guard against the picker.

core/src/main/java/tc/oc/pgm/commands/ListCommands.java
```diff
@@ -0,0 +1,112 @@
...
+              .color(ChatColor.GRAY));
+      tmm.getTeams()
+          .forEach(
+              team ->
+                  sendTeamInfo(
+                      sender, team.getComponentName(), team.getPlayers(), team.getMaxPlayers()));
+    } else {
+      sendTeamInfo(
+          sender,
+          new PersonalizedTranslatable("commands.users.participants"),
```
Electroid, line 46: It would be great if we can start using `TextComponent` and `TranslatableComponent` with the new text library. Soon I won't be accepting any more changes with the old text. Or else we'll never migrate.

core/src/main/java/tc/oc/pgm/community/events/PlayerVanishEvent.java
```diff
@@ -0,0 +1,30 @@
+package tc.oc.pgm.community.events;
+
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.player.event.MatchPlayerEvent;
+
+public class PlayerVanishEvent extends MatchPlayerEvent {
+
+  private final boolean vanished;
```
Electroid, line 9: Since this is sorted in the `MatchPlayer` object, you don't need to store it in the event as well.

core/src/main/java/tc/oc/pgm/community/events/PlayerVanishEvent.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public PlayerVanishEvent(MatchPlayer vanisher, boolean vanished) {
+    super(vanisher);
+    this.vanished = vanished;
+  }
+
+  public boolean isVanished() {
+    return vanished;
```
Electroid, line 17: `vanisher.isVanished()`

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -0,0 +1,209 @@
...
+                  vanish ? "vanish.activate.already" : "vanish.deactivate.already"),
+              player.getBukkit()));
+    }
+
+    // Keep track of the UUID and apply/remove META data, so we can detect vanish status from other
+    // projects (i.e utils)
+    if (vanish) {
+      addVanished(player.getId());
+      player.getBukkit().setMetadata(VANISH_KEY, VANISH_VALUE);
+      player.getSettings().setValue(SettingKey.CHAT, SettingValue.CHAT_ADMIN);
```
Electroid, line 102: Seems like `addVanished` should do this?

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -0,0 +1,209 @@
...
+    } else {
+      removeVanished(player.getId());
+      player.getBukkit().removeMetadata(VANISH_KEY, VANISH_VALUE.getOwningPlugin());
```
Electroid, line 105: Same here

util/src/main/i18n/templates/strings.properties
```diff
@@ -1041,6 +1042,25 @@ stats.current = Current stats
 
 stats.disabled = You have disabled stat tracking
 
+vanish.activate = You have vanished!
+vanish.deactivate = You are now visible!
```
Electroid, line 1046: I think you want to use the same vanish terminology, not visible.

core/src/main/java/tc/oc/pgm/commands/ListCommands.java
applenick: I was wondering about this, as I seem to have run into a few issues. First using `MatchPlayer#showHotbar` did not actually show the hotbar. After searching the repo, I was unable to find any calls to the new hotbar, so not sure if that’s been tested. 

Also in `ListCommands` I’m utilizing `Components#join` to neatly format the player names. Since that’s still using the old component system, what should be done there? I can migrate the method if you like, however there are quite a few calls to that method from `ModerationCommands`.

 I’m planning on a PR soon that will have some changes to `ModerationCommands`, maybe it would be more appropriate to migrate all the moderations commands to the new text library then? 

I did however use `TranslatableComponent` for the main vanish commands. 👍

Electroid: Great, if you have some blockers don’t worry about it for now.

Electroid:
Merged into #431

PR #439, opened 2020-05-02
[DOCS] Rework paragraph

Picajoluna:
I think I got this PR thing right...

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -22,10 +22,10 @@ const features = [
     imageUrl: "img/open-hands.png",
     description: (
       <>
-        Not only it is easy to use, but contributing is also more accessible
-        thanks to having no dependency injection. PGM is supposed to be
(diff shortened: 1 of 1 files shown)


EricZeiberg:
I think you should have kept the word injection and not installation. It still requires dependency installation, but the difference between PGM 1.8 and OCN's version is that dependencies weren't injected during runtime.

Picajoluna:
Yeah Indicado let me know afterwards, I hadn't caught the nuance...  We can keep dependency injections, maybe like so?

>  Not only is PGM easy, but the plugin runs without dependency injections: contributing has never been more accessible.

Indicardo:
From Electroid:
> Most people don't know what dependency injection is so I'm okay leaving it out

Given that the website is now also a "gateway" (sort of speaking) for PGM and its community, we should aim towards using simpler terms, without technicisms. Hence I think it's ok as it is right now

PR #451, opened 2020-05-03
Add inventory menu API and settings GUI

rafibaum:
This PR contains two major changes. The first being the addition of an inventory menu API which makes the creation of inventory GUI menus straightforward. I refactored observer tools to use this new API and then made the second change, the addition of a settings menu. The "Observer Tools" item is now a "Settings" item which combines the functionality of observer tools with a settings GUI. You can access this menu either by right clicking the tool like previously or by running `/settings`. If you run `/settings` in a match, observer tools are removed from the menu since they aren't relevant in a match.


core/src/main/java/tc/oc/pgm/menu/InventoryMenuMatchModule.java
```diff
@@ -0,0 +1,156 @@
...
+   */
+  @EventHandler
+  public void onMove(InventoryMoveItemEvent event) {
+    if (hasInventory(event.getDestination()) || hasInventory(event.getSource())) {
+      event.setCancelled(true);
+    }
+  }
+
+  /**
+   * Removes a {@link Inventory} from the manager
```
KingOfSquares, line 144: This should be rephrased to reflect that it's not the method to use to remove inventories from the manager, but rather that it removes inventories from the manager whenever players stop viewing them.

core/src/main/java/tc/oc/pgm/menu/InventoryMenuMatchModule.java
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * @param event the close inventory event
+   */
+  @EventHandler
+  public void onClose(InventoryCloseEvent event) {
+    if (event.getInventory() == null) {
+      return;
+    }
```
KingOfSquares, line 152: How can this ever return true?

core/src/main/java/tc/oc/pgm/menu/InventoryMenuMatchModule.java
```diff
@@ -0,0 +1,156 @@
...
+   */
+  InventoryMenu getInventory(Inventory inventory) {
+    return inventoryMap.get(inventory);
+  }
+
+  /**
+   * Removes {@link Inventory} and it's link from the linked mappings
+   *
+   * @param inventory the inventory to remove
+   * @return the walrus inventory this inventory is linked to
```
KingOfSquares, line 92: `walrus inventory` 😅

core/src/main/java/tc/oc/pgm/menu/items/ItemBuilder.java
```diff
@@ -0,0 +1,177 @@
...
+   */
+  public ItemBuilder addEnchantment(Enchantment enchantment, int level) {
+    itemStack.addUnsafeEnchantment(enchantment, level);
+    return this;
+  }
+
+  /**
+   * Set's the owner of skull, if the item is a skull
+   *
+   * @param owner the owner of the skull
```
KingOfSquares, line 102: Should probably specify that it takes the username of the owner and not some other reference.

core/src/main/java/tc/oc/pgm/menu/items/ItemHolder.java
```diff
@@ -0,0 +1,58 @@
+package tc.oc.pgm.menu.items;
+
+import org.bukkit.inventory.Inventory;
+import org.bukkit.inventory.ItemStack;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.menu.InventoryMenu;
+
+/**
+ * A holder for {@link InventoryItem} inside of a {@link InventoryMenu}. Keeps track of what slot
```
KingOfSquares, line 9: `A holder for an {@link InventoryItem}...`

core/src/main/java/tc/oc/pgm/menu/items/ItemHolder.java
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Puts the item into the requested {@link Inventory} that has been generated for a specific
+   * player. If the item returned is null it will not be put into the bukkit inventory. This is not
+   * an error but an intended feature to allow for items to not generate under certain cicrumstances
+   * (For example, the back button not generating if there is no page to go back to)
+   *
+   * @param player the player
+   * @param bukkitInventory the bukkit inventory to put the item in
+   * @param inventory the walrus inventory the item is from
```
KingOfSquares, line 48: `walrus inventory` :sweat_smile:

core/src/main/java/tc/oc/pgm/settings/ObserverToolsMatchModule.java
```diff
@@ -0,0 +1 @@
+
```
KingOfSquares, line 1: Forgot to delete?

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
```diff
@@ -0,0 +1,166 @@
...
+
+@ListenerScope(MatchScope.LOADED)
+public class SettingsMatchModule implements MatchModule, Listener {
+
+  public static class Factory implements MatchModuleFactory<SettingsMatchModule> {
+    @Override
+    public SettingsMatchModule createMatchModule(Match match) throws ModuleLoadException {
+      return new SettingsMatchModule(match);
+    }
+  }
```
KingOfSquares, line 43: Why not just do `SettingsMatchModule::new` in `Modules`?

util/src/main/i18n/templates/ui.properties
```diff
@@ -43,6 +43,54 @@ setting.visibility.hidden = Hidden
...
+setting.vote = Vote menu
+setting.vote.lore = Status: {0}
+setting.vote.value.on = On
+setting.vote.value.off = Off
+
+setting.stats = Stats
+setting.stats.lore = Tracking: {0}
+setting.stats.value.on = On
+setting.stats.value.off = Off
+
```
KingOfSquares, line 93: Duplicate strings should be merged

All strings with "off" and "on" should be replaced with `misc.on` and `misc.off`

`setting.stats` should be removed in favour of `match.stats.overall` (might want to change the key name)

Strings for "Shown and Hidden" already exists (might want to change key name maybe)

core/src/main/java/tc/oc/pgm/menu/InventoryMenu.java
```diff
@@ -1,151 +1,79 @@
...
 
-  protected static final int ROW_WIDTH = 9; // Number of columns per row
-  protected static final int MAX_ROWS = 6; // Max allowed row size
-
-  private final WeakHashMap<MatchPlayer, InventoryMenu> viewing =
-      new WeakHashMap<>(); // Map of players who are viewing the gui, along with the menu
-  private final String title; // Title of the inventory
-  private final int rows; // The # of rows in the inventory
+/** Interface that describes a programmable inventory menu. */
+public interface InventoryMenu {
```
Electroid, line 8: I like the interface, I think methods need a little tweaking. Right now there's 3 `open`s and a `purge` and `refresh`,  it's not clear what each of these do from a bird's eye view.

First, I'd say move this to `tc.oc.pgm.util.menu` and expand the scope from `MatchPlayer` to `Player`. There's no reason, that I can think of, that we should scope this to a `Match`. For instance, when we support multiple matches we might have some menu that allows you to switch between them. For settings, in the `onClick` callback, you can use `Datastore#getSettings(Player#getId)` to get the player's settings.

Next, I'd say you only need 1 open. Any cyclical back checks can be an implementation detail.
```java
/**
 * Opens the menu for a player.
 *
 * @param viewer A player.
 * @param referrer A previous menu that referred this menu, or null if the first menu.
 */
void open(Player viewer, @Nullable InventoryMenu referrer);
```

Finally, you really only need 1 type of invalidate, which basically just re-draws the items.
```java
/**
 * Refreshes the menu for a player, on next tick.
 *
 * @param viewer A player.
 */
void invalidate(Player viewer);
```

On the implementation side, you could use metadata for back checks or history. e.g. (a `Stack<InventoryMenu>` stored as metadata?)

core/src/main/java/tc/oc/pgm/menu/items/InventoryItem.java
```diff
@@ -0,0 +1,51 @@
...
+  void purgeAll();
+
+  /**
+   * This clears the item cache for a specific player, meaning the next time this item is generated
+   * for that player it will not used the cached version but will call {@link #item(InventoryMenu
+   * inventory, MatchPlayer player)}
+   *
+   * @param player the player the cache is being cleared for
+   */
+  void purge(MatchPlayer player);
```
Electroid, line 50: Not sure this belongs at the item level. See my suggested comment for `item`.

core/src/main/java/tc/oc/pgm/menu/items/InventoryItem.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * <p>This method can return null, in which case the item should not be rendered and is treated as
+   * the player not having permission/not meeting requirements to see the item
+   *
+   * @param inventory the inventory that this item is inside
+   * @param player the player this item is being generated for
+   * @return the generated itemstack, usually used to put in normal bukkit inventories or null if
+   *     the item stack should not be rendered
+   */
+  ItemStack item(InventoryMenu inventory, MatchPlayer player);
```
Electroid, line 24: To allow for an item to be "re-rendered", I might suggest the following:
```java
// return false = no render
// return true = item was modified and is ready to go
boolean renderItem(InventoryMenu menu, Player viewer, ItemStack item);
```

core/src/main/java/tc/oc/pgm/menu/InventoryMenuMatchModule.java
```diff
@@ -0,0 +1,156 @@
...
+
+    Inventory inventory = event.getInventory();
+
+    if (hasInventory(inventory)) {
+      int x = event.getSlot() % 9;
+      int y = event.getSlot() / 9;
+      event.setCancelled(true);
+      getInventory(inventory).clickItem(x, y, match.getPlayer(event.getActor()), event.getClick());
+    }
+  }
```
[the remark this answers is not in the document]
rafibaum: I believe that's true when you make a click while you're in an inventory screen, but you don't actually click on an inventory (ie, you drop an item outside the window).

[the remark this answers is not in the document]
rafibaum: Also not convinced reordering would make a difference. The difference between hashing and checking a map entry vs a couple comparisons and pointer dereferences is likely negligible.

core/src/main/java/tc/oc/pgm/menu/InventoryMenuMatchModule.java
rafibaum: No idea. Matt put this in and I'll leave it in unless I know it's safe to remove.

core/src/main/java/tc/oc/pgm/menu/InventoryMenuMatchModule.java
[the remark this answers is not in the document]
KingOfSquares: You can also use 121 on line 107

core/src/main/java/tc/oc/pgm/menu/InventoryMenu.java
```diff
@@ -1,151 +1,79 @@
...
+   * Acts as a {@link MatchPlayer} clicking a {@link InventoryItem} in a specific {@link
+   * InventoryMenu}
    *
-   * <p>If the window is open but should be closed, close it and return null.
+   * @param x the x coordinate
+   * @param y the y coordinate
+   * @param player the player clicking the item
+   * @param clickType the type of click which is occurring
+   */
+  void clickItem(int x, int y, MatchPlayer player, ClickType clickType);
```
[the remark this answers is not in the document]
rafibaum: I gave this a go but I don't think Slot is a good fit here. Slot is used by kits to map intuitive human slot locations you'd find in XML to the actual underlying slot and a lot of the API is built around that. Here we're building custom inventories where the slots aren't expected to line up with an actual inventory layout you'd find in Minecraft so I think the (x, y) system is a better fit.

core/src/main/java/tc/oc/pgm/menu/items/ItemHolder.java
```diff
@@ -0,0 +1,58 @@
...
+import tc.oc.pgm.menu.InventoryMenu;
+
+/**
+ * A holder for {@link InventoryItem} inside of a {@link InventoryMenu}. Keeps track of what slot
+ * the item is in and the translation between bukkit raw slots and a x,y coordinate system
+ */
+public class ItemHolder {
+
+  public final int y;
+  public final int x;
```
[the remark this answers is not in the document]
rafibaum: https://github.com/Electroid/PGM/pull/451#discussion_r420808199 Tried this refactor. See here as to why I don't think it's a good idea.

core/src/main/java/tc/oc/pgm/menu/InventoryMenu.java
rafibaum: Doing history as an implementation detail makes it a bit messy. I'd rather build it into the API or not have it at all and I think for this case I'll just chop it out since, while relevant for Walrus, isn't necessary here. Everything else is good though, and in the future we can build a different implementation which adds pagination/history.

core/src/main/java/tc/oc/pgm/menu/items/InventoryItem.java
rafibaum: Based on the current API it's impossible to know if the item has changed without rendering it, which is why invalidation is manual.

core/src/main/java/tc/oc/pgm/menu/items/InventoryItem.java
rafibaum: https://github.com/Electroid/PGM/pull/451#discussion_r420948605

core/src/main/java/tc/oc/pgm/settings/ObserverTool.java
```diff
@@ -0,0 +1,38 @@
...
+
+  public List<String> getLore(Player player);
+
+  public Material getMaterial(Player player);
+
+  default ItemStack createItem(Player player) {
+    ItemStack stack = new ItemStack(getMaterial(player));
+    ItemMeta meta = stack.getItemMeta();
+
+    // TODO: Check translations
```
Electroid, line 28: ?

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenuManager.java
```diff
@@ -0,0 +1,151 @@
...
+import org.bukkit.plugin.Plugin;
+import tc.oc.pgm.util.inventory.InventoryUtils;
+
+public class InventoryMenuManager implements Listener {
+
+  private final Map<Inventory, InventoryMenu> inventoryMap;
+  private final Plugin plugin;
+
+  /** Creates a new inventory manager. */
+  public InventoryMenuManager(Plugin plugin) {
```
Electroid, line 23: Instead of making one of these for each `Match`, register it as a global singleton in `PGMPlugin`

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenuManager.java
```diff
@@ -0,0 +1,151 @@
...
+import org.bukkit.event.Listener;
+import org.bukkit.event.inventory.InventoryClickEvent;
+import org.bukkit.event.inventory.InventoryCloseEvent;
+import org.bukkit.event.inventory.InventoryMoveItemEvent;
+import org.bukkit.event.inventory.InventoryType;
+import org.bukkit.inventory.Inventory;
+import org.bukkit.plugin.Plugin;
+import tc.oc.pgm.util.inventory.InventoryUtils;
+
+public class InventoryMenuManager implements Listener {
```
Electroid, line 17: `InventoryMenuListener`

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenuManager.java
```diff
@@ -0,0 +1,151 @@
...
+
+  private final Map<Inventory, InventoryMenu> inventoryMap;
+  private final Plugin plugin;
+
+  /** Creates a new inventory manager. */
+  public InventoryMenuManager(Plugin plugin) {
+    this.plugin = plugin;
+    inventoryMap = Maps.newHashMap();
+    plugin.getServer().getPluginManager().registerEvents(this, plugin);
```
Electroid, line 26: Usually don't self-register.

util/src/main/java/tc/oc/pgm/util/menu/items/InventoryClickAction.java
```diff
@@ -0,0 +1,10 @@
+package tc.oc.pgm.util.menu.items;
```
Electroid, line 1: Our package naming notation has typically stayed singular, so `menu.item`

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenuBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Creates a new inventory builder
+   *
+   * @param manager
+   * @param rows the number of rows the inventory should have
+   */
+  public InventoryMenuBuilder(InventoryMenuManager manager, int rows) {
+    this.manager = manager;
+    this.rows = rows;
+    this.name = TextComponent.of("");
```
Electroid, line 27: `TextComponent.empty()`

util/src/main/java/tc/oc/pgm/util/menu/items/InventoryItemBuilder.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.function.BiFunction;
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import tc.oc.pgm.util.menu.InventoryMenu;
+import tc.oc.pgm.util.menu.InventoryMenuManager;
+
+public class InventoryItemBuilder {
+  // TODO: Add back button
+  private final InventoryMenuManager manager;
+  private final BiFunction<InventoryMenu, Player, ItemStack> createItem;
```
Electroid, line 13: Anytime you have a builder with a complex function or manager passed around.. it's probably better to just make a static class in the settings module: `SettingItem`. I'd say you don't need the builder.

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenuBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+  public InventoryMenuBuilder addItem(InventoryItem item) {
+    for (int y = 0; y < rows; y++) {
+      for (int x = 0; x < 9; x++) {
+        boolean matched = false;
+        for (ItemHolder itemHolder : items) {
+          if (itemHolder.x == x && itemHolder.y == y) {
+            matched = true;
+            break;
+          }
+        }
```
Electroid, line 56: This is too inefficient

I would also skip the `InventoryMenuBuilder` and just have a static `SettingsInventoryMenu` class. By sub-classing instead of tons of builders we remove a ton of boilerplate code, and it makes the menu's more extendable.

util/src/main/i18n/translations/ui_ca_ES.properties
```diff
@@ -37,9 +37,9 @@ setting.visibility = Visibilitat
 # {0} = status (e.g. "Shown" or "Hidden")
 setting.visibility.lore = Visibilitat dels observadors\:{0}
 
-setting.visibility.shown = Mostrar
+setting.observers.value.on = Mostrar
```
Electroid, line 40: I don't think we typically fix non-English language translations until we merge Crowdin, but I think this should be ok. Leave it in. (Crowdin has "memory" and does this for us later)

core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -57,6 +129,37 @@ public String getName() {
...
+   *
+   * @param value the formatted value the setting is currently set to
+   * @return the complete lore text for the setting
+   */
+  public Component getLore(Component value) {
+    return TranslatableComponent.of(
+            SETTING_TRANSLATION_KEY + this.getName().toLowerCase() + ".description")
+        .args(value)
+        .color(TextColor.GRAY);
+  }
```
Electroid, line 152: Can we pre-compute the name and lore in the enum constructor? These won't change.

core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -57,6 +129,37 @@ public String getName() {
     return aliases.get(0);
   }
 
+  /**
+   * Get the display name of the {@link SettingKey}.
+   *
+   * @return The formatted display name.
+   */
+  public Component getDisplayName() {
```
Electroid, line 137: `getName`

core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -57,6 +129,37 @@ public String getName() {
...
+    return TranslatableComponent.of(displayNameKey, color, TextDecoration.BOLD);
+  }
+
+  /**
+   * Gets the lore of the setting.
+   *
+   * @param value the formatted value the setting is currently set to
+   * @return the complete lore text for the setting
+   */
+  public Component getLore(Component value) {
```
Electroid, line 147: `getDescription(SettingValue)`

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -71,6 +105,25 @@ public String getName() {
...
   }
 
+  /**
+   * Gets the formatted name of the {@link SettingValue}.
+   *
+   * @return formatted name of the {@link SettingValue}.
+   */
+  public Component getDisplayName() {
+    return TranslatableComponent.of(displayNameKey).color(color);
+  }
```
Electroid, line 115: Pre-compute in the enum constructor.

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -71,6 +105,25 @@ public String getName() {
     return name;
   }
 
+  /**
+   * Gets the formatted name of the {@link SettingValue}.
+   *
+   * @return formatted name of the {@link SettingValue}.
+   */
+  public Component getDisplayName() {
```
Electroid, line 113: `getName`. "display name" and "lore" and just implementation details of the menu.

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -71,6 +105,25 @@ public String getName() {
...
+    return TranslatableComponent.of(displayNameKey).color(color);
+  }
+
+  /**
+   * Gets the material that should be used to represent the {@link SettingValue}.
+   *
+   * @return the material that should be used to represent the {@link SettingValue}.
+   */
+  @Nullable
+  public Material getMaterial() {
```
Electroid, line 123: `getIcon` might be a more useful name here. The "material" of a setting is weird.

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
```diff
@@ -0,0 +1,173 @@
...
+    manager.disable();
+  }
+
+  @EventHandler
+  public void onObserverKitApply(ObserverKitApplyEvent event) {
+    refreshKit(event.getPlayer());
+  }
+
+  @EventHandler
+  public void onToolClick(PlayerInteractEvent event) {
```
Electroid, line 113: `ObserverInteractEvent`?

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
```diff
@@ -0,0 +1,173 @@
...
+
+  private ItemStack createToolItem(Player player) {
+    Component displayName =
+        TranslatableComponent.of(INVENTORY_TITLE, TextColor.AQUA, TextDecoration.BOLD);
+    Component lore = TranslatableComponent.of("setting.lore", TextColor.GRAY);
+
+    return new ItemBuilder()
+        .material(TOOL_MATERIAL)
+        .name(TextTranslations.translateLegacy(displayName, player))
+        .lore(TextTranslations.translateLegacy(lore, player))
```
Electroid, line 150: Edit `ItemBuilder` to also accept in `Component`s. In the future, I'm thinking it can do some magic to update translations in the background.

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
```diff
@@ -0,0 +1,173 @@
...
+  // Material of tool item item
+  public static final Material TOOL_MATERIAL = Material.DIAMOND;
+
+  public static final String INVENTORY_TITLE = "setting.title";
+  public static final int SETTINGS_ROWS = (SettingKey.values().length + 9 - 1) / 9;
+
+  private final Match match;
+  private InventoryMenuManager manager;
+  private InventoryMenu observerMenu;
+  private InventoryMenu otherMenu;
```
Electroid, line 52: Why 2 menus?

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
```diff
@@ -0,0 +1,173 @@
...
+      material = setting.getMaterial();
+    }
+    ItemStack stack = new ItemStack(material);
+
+    ItemMeta meta = stack.getItemMeta();
+    meta.setDisplayName(
+        TextTranslations.translateLegacy(setting.getDisplayName(), player.getBukkit()));
+    Component lore = setting.getLore(value.getDisplayName());
+    meta.setLore(Lists.newArrayList(TextTranslations.translateLegacy(lore, player.getBukkit())));
+    meta.addItemFlags(ItemFlag.values());
```
Electroid, line 168: `ItemBuilder`?

core/src/main/java/tc/oc/pgm/settings/tools/GamemodeTool.java
```diff
@@ -0,0 +1,102 @@
...
+    menu.invalidate(player);
+  }
+
+  public void toggleObserverGameMode(Player player) {
+    player.setGameMode(getOppositeMode(player.getGameMode()));
+    if (player.getGameMode() == GameMode.SPECTATOR) {
+      match.getPlayer(player).sendWarning(getToggleMessage());
+    } else if (isCreative(player)) {
+      // Note: When WorldEdit is present, this executes a command to ensure the player is not stuck
+      if (Bukkit.getPluginManager().isPluginEnabled("WorldEdit")) {
```
Electroid, line 70: We should save this as a static variable.

util/src/main/i18n/templates/ui.properties
```diff
@@ -32,16 +34,60 @@ setting.nightvision = Night Vision
...
+setting.observers.value.off = Hidden
+
+setting.sounds = Sounds
+
+# {0} = sounds (e.g. "All", "DMs", "None")
+setting.sounds.description = Status: {0}
+
+setting.sounds.value.messages = DMs
+
+setting.vote = Vote Menu
```
Electroid, line 84: Vote Book?

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
Pablete1234: In-game settings menu doesn't include changing gamemode, flight speed, obs visibility, or night vision, while the observer menu will have all the settings, at least i assume that's the reason

rafibaum: ^ that's the reason. The inventory item objects themselves are reused across both menus so the memory impact is minimal.

util/src/main/java/tc/oc/pgm/util/menu/items/InventoryItemBuilder.java
rafibaum: What do you mean by that? Should I create some sort of "creator" interface which serves the same purpose as the builder?

core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
rafibaum: The original `getName` is still needed in a bunch of places so the distinction needs to be made.

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenuBuilder.java
rafibaum: See my comment above. Not sure what pattern you're suggesting I replace the builders with.

core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
rafibaum: Actually I think you want a larger refactor? See my comment below.

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
rafibaum: Are you suggesting I move the formatted name and description out of the Settings objects and into the menu implementation?

core/src/main/java/tc/oc/pgm/api/PGM.java
```diff
@@ -40,6 +41,8 @@
 
   VanishManager getVanishManager();
 
+  InventoryMenuListener getInventoryMenuListener();
```
Electroid, line 44: We shouldn't need to expose this. See my comment in `SettingsMatchModule`.

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
Electroid: It's alright, just keep it as is.

core/src/main/java/tc/oc/pgm/settings/ObserverTool.java
```diff
@@ -0,0 +1,37 @@
...
+    ItemMeta meta = stack.getItemMeta();
+
+    meta.setDisplayName(
+        TextTranslations.translateLegacy(
+            getName().color(getColor()).decoration(TextDecoration.BOLD, true), player));
+    meta.setLore(getLore(player));
+    meta.addItemFlags(ItemFlag.values());
+
+    stack.setItemMeta(meta);
+    return stack;
```
Electroid, line 35: `ItemBuilder`?

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
```diff
@@ -0,0 +1,155 @@
...
+import tc.oc.pgm.spawns.events.ObserverKitApplyEvent;
+import tc.oc.pgm.util.inventory.ItemBuilder;
+import tc.oc.pgm.util.menu.InventoryMenu;
+import tc.oc.pgm.util.menu.InventoryMenuBuilder;
+import tc.oc.pgm.util.menu.InventoryMenuListener;
+import tc.oc.pgm.util.menu.item.InventoryItem;
+import tc.oc.pgm.util.menu.item.InventoryItemBuilder;
+
+@ListenerScope(MatchScope.LOADED)
+public class SettingsMatchModule implements MatchModule, Listener {
```
Electroid, line 35: `Settings` are independent of a `Match` so it doesn't make much sense to use a `MatchModule` here. I would also make it a `Listener`, that way we do not need to expose `PGM#getInventoryMenuListener`.

core/src/main/java/tc/oc/pgm/settings/tools/FlySpeedTool.java
```diff
@@ -0,0 +1,96 @@
...
+import org.bukkit.Material;
+import org.bukkit.entity.Player;
+import org.bukkit.event.inventory.ClickType;
+import tc.oc.pgm.settings.ObserverTool;
+import tc.oc.pgm.util.menu.InventoryMenu;
+import tc.oc.pgm.util.text.TextTranslations;
+
+public class FlySpeedTool implements ObserverTool {
+
+  private static final String TRANSLATION_KEY = "setting.flyspeed.";
```
Electroid, line 17: It's weird to define this but not use it in `getName` or `getLore`

core/src/main/java/tc/oc/pgm/settings/tools/GamemodeTool.java
```diff
@@ -0,0 +1,104 @@
...
+   * @param match the match this tool is being used in
+   */
+  public GamemodeTool(Match match) {
+    this.match = match;
+    worldEditEnabled = Bukkit.getPluginManager().isPluginEnabled("WorldEdit");
+  }
+
+  @Override
+  public Component getName() {
+    return TranslatableComponent.of("setting.gamemode");
```
Electroid, line 37: Same here (as well as the other tools) they keys are all over the place.

util/src/main/java/tc/oc/pgm/util/menu/items/InventoryItemBuilder.java
Electroid: My suggestion is instead of having `InventoryItemBuilder` and `InventoryMenuBuilder`, just define your own `SettingsItem` and `SettingsMenu`. Likely that will simplify constructors, also each menu has its own item rendering and slot distribution strategy.

Electroid:
Any progress here? Otherwise I think this will be deferred into `v0.9`.

rafibaum:
Sorry! Been away the past couple weeks. I'll start work on this again.

core/src/main/java/tc/oc/pgm/settings/SettingsMatchModule.java
rafibaum: I do need to pass around an instance of the listener either way since the implementation needs to query shared data structures that know which player is viewing which inventory menu etc. How should I keep track of that singleton instance outside of PGM?

PR #456, opened 2020-05-05
Fix localization for observer items on login

### LINKED ISSUE #443: Toolbar items don't get properly translated on login
Steps to reproduce:
1. Log in with a language other than English. Your toolbar items will be in English.
2. Join a match and then go back to spectating.
3. Your items will be in your local language.
I suspect locale information isn't available when we give an observer their tools.

KingOfSquares:
Fixes #443

### DIFF
--- a/core/src/main/java/tc/oc/pgm/spawns/states/Observing.java
+++ b/core/src/main/java/tc/oc/pgm/spawns/states/Observing.java
@@ -3,6 +3,7 @@ package tc.oc.pgm.spawns.states;
 import com.google.common.collect.Sets;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
@@ -16,6 +17,7 @@ import org.bukkit.material.Door;
 import org.bukkit.permissions.PermissionAttachment;
 import tc.oc.pgm.api.Config;
 import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.MatchScope;
 import tc.oc.pgm.api.match.event.MatchStartEvent;
 import tc.oc.pgm.api.party.Competitor;
 import tc.oc.pgm.api.player.MatchPlayer;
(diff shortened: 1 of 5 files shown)


core/src/main/java/tc/oc/pgm/spawns/states/Observing.java
```diff
@@ -72,22 +74,34 @@ public void enterState() {
...
+
+                // Let other modules give observer items
+                player.getMatch().callEvent(new ObserverKitApplyEvent(player));
+
+                // Apply observer spawn kit, if there is one
+                spawn.applyKit(player);
+
+                player.getBukkit().updateInventory();
+              },
+              500,
```
Pablete1234, line 101: This just adds a race condition based on time. Should figure out when the settings are actually sent and apply after that. Possibly just give the tools at the start and on locale change re-send them

cswhite2000: SportPaper does have a PlayerLocaleChangeEvent, but that doesn't work with trying to move away from using SportPaper. The only other way I can think of for doing something on locale change would be to add a ProtocolLib packet listener for PacketPlayInSettings, but that has its own problems.

Pablete1234: Checking for the locale being different than previous locale i guess is always an option, but it's not the best one

rafibaum: Since we're moving away from SportPaper, we need some alternative. @cswhite2000 what's the issue with using a ProtocolLib packet listener?

cswhite2000: We would have to add ProtocolLib as a dependency for pgm, and it doesn't seem like its worth doing that for this. But, if we planning on using protocollib for supporting multiple versions for the tab list, then it probably would be worth it.

rafibaum: Could do this as a soft-dependency then, where the functionality is disabled if ProtocolLib isn't on the server (similar to JGit handling).

KingOfSquares:
I will finish this after #469

KingOfSquares:
> We can't rely on the time delay for everything. Here's my idea:
> 
> Define a `PlayerLocaleChangeEvent` in `tc.oc.pgm.util`.
> Listen to SportPaper's `PlayerLocaleChangeEvent` and fire our own `PlayerLocaleChangeEvent` event.
> When we migrate to 1.8, we'll figure out a fix that doesn't use SportPaper's event.
> 
> I'd highly recommend looking at the 1.12 PGM codebase to figure out where else we need to refresh on `PlayerLocaleChangeEvent`.

Unsure if this is relevant https://github.com/OvercastNetwork/ProjectAres/blob/b883ef579938c49076379fa306592da847fc672f/PGM/src/main/java/tc/oc/pgm/picker/PickerMatchModule.java#L278-L282

PR #462, opened 2020-05-07
Vanish Improvements

applenick:
# Vanish Subdomain login

This PR adds a requested feature for the `/vanish` command. The detection of players who login using a subdomain starting with `vanish` for example `vanish.oc.tc`. This will allow staff members who wish to forcefully join vanished to do so.

Thanks and if there are any other suggestions for `/vanish` leave your feedback 😄 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
+++ b/core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
@@ -166,10 +166,6 @@ public class VanishManagerImpl implements VanishManager, Listener {
         loginSubdomains.invalidate(player.getId());
         player.setVanished(true);
         addVanished(player);
-
-        player.sendWarning(
-            TranslatableComponent.of("vanish.login.domain", TextColor.GRAY)
-                .args(TextComponent.of(domain, TextColor.AQUA)));
       }
     }
   }
--- a/core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
+++ b/core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
@@ -17,7 +17,6 @@ import net.kyori.text.TextComponent;
 import net.kyori.text.TranslatableComponent;
 import net.kyori.text.format.TextColor;
 import net.kyori.text.format.TextDecoration;
-import org.bukkit.Bukkit;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
@@ -160,21 +159,15 @@ public class VanishManagerImpl implements VanishManager, Listener {
         .getBukkit()
         .hasPermission(Permissions.VANISH)) { // Player is not vanished, but has permission to
 
-      // If majority of staff are vanished, auto vanish to keep hidden. Or if player logs in via a
-      // "vanish" subdomain.
+      // Automatic vanish if player logs in via a "vanish" subdomain.
       String domain = loginSubdomains.getIfPresent(player.getId());
-      boolean staff = checkStaffCount();
-      if (staff || domain != null) {
+      if (domain != null) {
         player.setVanished(true);
         addVanished(player);
 
-        if (staff) {
-          player.sendWarning(TranslatableComponent.of("vanish.login.auto", TextColor.GRAY));
-        } else if (domain != null) {
-          player.sendWarning(
-              TranslatableComponent.of("vanish.login.domain", TextColor.GRAY)
-                  .args(TextComponent.of(domain, TextColor.AQUA)));
-        }
+        player.sendWarning(
+            TranslatableComponent.of("vanish.login.domain", TextColor.GRAY)
+                .args(TextComponent.of(domain, TextColor.AQUA)));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


util/src/main/i18n/templates/moderation.properties
```diff
@@ -175,3 +175,5 @@ vanish.chat.deny = You can only use the admin chat while vanished
 
 vanish.hotbar = You are currently vanished
 
+vanish.login.auto = Majority of staff are vanished. Due to this we've vanished you as well
```
Electroid, line 178: I don't understand this feature

applenick: Hey Electroid,
So this idea was suggested by @Pablete1234 yesterday. Basically if more than half of the online staff are vanished, any joining staff will be auto-vanished to prevent players from knowing there are staff online (example for watching a suspected hacker).

applenick:
@Electroid, as requested the staff count feature has been removed 👍

util/src/main/i18n/templates/moderation.properties
Pablete1234: This is a nice feature, there may be 2 or 3 staff members online who are watching a hacker vanished, and suddently you join unvanished, broadcasting a join-message to all players. This has already happened and can ruin entire operations

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -133,10 +137,38 @@ public void vanish(MatchPlayer sender, @Switch('s') boolean silent) throws Comma
   }
 
   /* Events */
+  private final Cache<UUID, String> loginSubdomains =
+      CacheBuilder.newBuilder().expireAfterAccess(1, TimeUnit.SECONDS).build();
```
Electroid, line 141: Make this `30` seconds, which is the login timeout.

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -133,10 +137,38 @@ public void vanish(MatchPlayer sender, @Switch('s') boolean silent) throws Comma
...
+
+  @EventHandler(priority = EventPriority.MONITOR)
+  public void onPreJoin(PlayerLoginEvent event) {
```
Electroid, line 144: Maybe you want to do this check in `AsyncPreLoginEvent`?

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -133,10 +137,38 @@ public void vanish(MatchPlayer sender, @Switch('s') boolean silent) throws Comma
...
-      matchManager.getPlayer(event.getPlayer()).setVanished(true);
+    MatchPlayer player = matchManager.getPlayer(event.getPlayer());
+    if (isVanished(player.getId())) { // Player is already vanished
+      player.setVanished(true);
+    } else if (player
+        .getBukkit()
+        .hasPermission(Permissions.VANISH)) { // Player is not vanished, but has permission to
+
+      // Automatic vanish if player logs in via a "vanish" subdomain.
+      String domain = loginSubdomains.getIfPresent(player.getId());
```
Electroid, line 163: You'll want to remove from the `Cache`

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
Electroid: Also, before the if statement, always invalidate the `Cache` for the player.

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -145,6 +177,10 @@ public void checkMatchPlayer(MatchPlayerAddEvent event) {
     event.getPlayer().setVanished(isVanished(event.getPlayer().getId()));
   }
 
+  private boolean isVanishSubdomain(String address) {
+    return address.startsWith("vanish"); // TODO Maybe allow config value?
```
Electroid, line 181: No need for config, I think this is ok.

Pablete1234: I'd make it `vanish.`, otherwise it may match more than you want, i can only imagine some `vanished.network` having trouble with it

core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
applenick: As discussed `AsyncPreLoginEvent` does not have a way to determine the address used to login to the server with. Sticking with `PlayerLoginEvent` for now.

util/src/main/i18n/templates/moderation.properties
```diff
@@ -175,3 +175,4 @@ vanish.chat.deny = You can only use the admin chat while vanished
 
 vanish.hotbar = You are currently vanished
 
+vanish.login.domain = You logged in with {0} so we auto-vanished you
```
Electroid, line 178: No need for special message, the hot bar is fine enough.

applenick: Ok! I’ll make this change right away

PR #469, opened 2020-05-08
Add git repos as a possible remote source for maps

### LINKED ISSUE #4: Remote map sources
Right now a `MapSource` is required to be a `File` on the local filesystem. We want to allow PGM to load maps from `Github` (more important) or `Dropbox`. 

Two important considerations in the design.
1. Map loading and XML parsing should be done asynchronously on server startup. Meaning, that server startup should block until at least 1 map is loaded, the rest are parsed on a different thread(s).

2. PGM should **not** download the *entire* `Github` or `Dropbox` repository, it should use their APIs to only load maps when requested (eg. load all the maps on rotation, search and load for a map when it sets next)

KingOfSquares:
Closes #4 

After this PR you will be able to have any amount of online git repos as a map source.

Example of config can be seen in the changes to the config.yml

86264ca refactored this implementation, see edit history on this post for old functionality
With latest commit its now possible to choose which branch you want to copy as well.


Pablete1234:
Using:
`  - https://github.com/KingOfSquares/EpicMaps`
gives no hint nor clue about this being a *git* repository, and this will not work if you prop-in a random download link on some website that isn't a git repository.

In cardinal they were prefixed with `git:` so they looked like:
`- git:https://github.com/Pablete1234/PPTM`
 maybe consider something along those lines or some other way to specify this is git, and not just a path

KingOfSquares:
> Using:
> ` - https://github.com/KingOfSquares/EpicMaps`
> gives no hint nor clue about this being a _git_ repository, and this will not work if you prop-in a random download link on some website that isn't a git repository.
> 
> In cardinal they were prefixed with `git:` so they looked like:
> `- git:https://github.com/Pablete1234/PPTM`
> maybe consider something along those lines or some other way to specify this is git, and not just a path

I agree with this, just need to make sure for the user that its required. I think i recall elec saying he wanted to move the "default" maps to git? That would solve this issue

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -274,17 +274,40 @@ public void reloadConfig() {
...
+      MapSourceFactory factory = null;
       try {
-        factory =
-            source.equalsIgnoreCase("default")
-                ? DefaultMapSourceFactory.INSTANCE
-                : new SystemMapSourceFactory(source);
+        if (source.equalsIgnoreCase("default"))
+          factory =
+              new GitMapSourceFactory(
+                  "https://github.com/KingOfSquares/DefaultMaps", logger, false);
```
EricZeiberg, line 282: I don't think having hardcoded values pointing to a personal repo is a good idea. This should be in the config somewhere?

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -274,17 +274,40 @@ public void reloadConfig() {
...
+        if (source.startsWith("git;")) {
```
Electroid, line 283: Let's change the format to be the following:
```yml
map:
  folders:
    - "maps/"
  repositories:
    - uri: "https://github.com/Electroid/maps.git"
      branch: "tournament"
      folders:
        - "ctw/"
        - "dtc/"
```

You'll need to do some refactoring in `Config` and `PGMConfig`

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -274,17 +274,40 @@ public void reloadConfig() {
...
+          String gitSource = source.substring(4);
+          factory = new GitMapSourceFactory(gitSource, logger, true);
+        }
+        if (new File(source).exists()) factory = new SystemMapSourceFactory(source);
+      } catch (StringIndexOutOfBoundsException e) { // Git source parsing exception
```
Electroid, line 288: Add a method to `TextParser` to parse a git repository. `TextParser#parseGitRepository`

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -274,17 +274,40 @@ public void reloadConfig() {
...
+        logger.log(
+            Level.INFO,
+            "To add a git source you have to format your source like this: git;URL;USERNAME;PASSWORD");
+        logger.log(
+            Level.INFO, "Username and password only required if the git repository is private");
+        logger.log(Level.WARNING, source + " will be ignored until restart");
+        logger.log(Level.WARNING, "");
+        continue;
+      }
+
```
Electroid, line 310: Too verbose, don't need it.

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
```diff
@@ -0,0 +1,152 @@
...
+    int uniqueIdentifier =
+        Math.abs(source.hashCode()); // Prevents errors on similar repository names
+
+    String dirName;
+    if (credentialsProvided)
+      dirName =
+          (source.substring(source.lastIndexOf("/"), source.indexOf(";")) + "_" + uniqueIdentifier);
+    else dirName = (source.substring(source.lastIndexOf("/") + 1) + "_" + uniqueIdentifier);
+
+    return new File("./gitMaps/" + dirName);
```
Electroid, line 43: Make the default folder `maps/url`

If my repository is `github.com/Electroid/maps`, then the folder will be lowercased to: `maps/github.com/electroid/maps`

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
```diff
@@ -0,0 +1,152 @@
...
+
+    if (credentialsProvided) this.gitURI = source.substring(0, source.indexOf(";"));
+    else this.gitURI = source;
+
+    this.logger = logger;
+    this.dir = getFile(source);
+    this.updateOnRestart = updateOnRestart;
+
+    // Disable if JGit is not present
+    if (Class.forName("org.eclipse.jgit.api.Git") == null) {
```
Electroid, line 58: You need to do this check before calling the `GitMapSourceFactory` constructor, or else it won't work.

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
```diff
@@ -0,0 +1,152 @@
...
+    try {
+      paths = loadAllPaths();
+    } catch (IOException e) {
+      throw new MapMissingException(dir.getPath(), "Unable to list files in" + dir.getName(), e);
+    }
+
+    return paths.parallel().flatMap(this::loadNewSource).iterator();
+  }
+
+  private void refreshRepo() throws GitAPIException, MapMissingException {
```
Electroid, line 110: This needs to be async.

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
Pablete1234: that both creates a mess to navigate thru and adds severe inconsistencies from what characters are allowed in URLs and what is allowed on folder names. 
Trying to remove invalid characters will lead to possible collissions.

I think it's best to give them an id. in Cardinal we used the hash of the url for the folder name. Not very human readable, but it's the plugin handling those folders anyways, and it can't have these issues.

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Pablete1234: While i agree with this, special care should be taken not to over-complicate it for the user.
Just the URI should be enough to configure the git repo, all other fields should be optional for fine-tuning that the average user may not do

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
Electroid: A hash is not really user-friendly. If too many folders is a concern, maybe `github-com-electroid-maps` would be better? `.`, `/`, and `_` become `-`.

Pablete1234: It would be easier to read like that, but you still have collision issues.
github.com/Electroid/2-maps
github.com/Electroid-2/maps

Those would both end up with the same folder, and i don't think there's any good way to resolve it. Would have to start designing some "folder-name-escaping" technique where `-` becomes `--` and so on, wich feels like a bad solution in the long run

core/src/main/java/tc/oc/pgm/PGMPlugin.java
KingOfSquares: > I don't think having hardcoded values pointing to a personal repo is a good idea. This should be in the config somewhere?

Of course, just had to put something there for review.



> Don't define it here. Define it in the default `config.yml`
> We'll change the actual repo before merging.

ok!

core/src/main/java/tc/oc/pgm/PGMPlugin.java
KingOfSquares: I agree with Pablo. I can make it work with just a git link, but also work with branches/folders. Maybe also with .gitignore?

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
KingOfSquares: Right now i name folders the last word in url + a hash like
`Tournament_651324`

I could make something like electroids
`github-com-electroid-maps`
but then i would have to check for duplicate folders and do some escaping technique like Pablo mentioned.
I could also do Electroids format and add the unique hash at the end.

I could also support custom names in the config, so i could just prompt the user that if they are trying to download some repo that will crash name wise they just need to change the name of either one? I think maybe custom folder names would be nice anyways.

I will move repos to `maps` folder.

core/src/main/java/tc/oc/pgm/PGMPlugin.java
KingOfSquares: Can i leave some in? I agree i went overboard with this one, but maybe just a
`Map source someMapSource is invalid and will be ignored until restart`
is fine?

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Electroid: Sure, you make make a single git link work as well.

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
Electroid: I'd rather do the `-` and accept the risk of collision (which should be extremely low).

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
KingOfSquares: But sync on server startup?

core/src/main/java/tc/oc/pgm/PGMPlugin.java
KingOfSquares: Cant i just make an `GitRepository` object instead? Unsure how i would make a text parser for that since it has multiple strings as part of it `URIish`, `branch` and `folders`

KingOfSquares: With the URIish from JGit it parses the url, password and username for me.

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
KingOfSquares: Another question, its already called async in `MapLibraryImpl` so why would it not be async already?

core/src/main/java/tc/oc/pgm/PGMPlugin.java
KingOfSquares: Your proposed config does not work with multiple repos. You can not have multiple ConfigurationSections with the same key. You would have to increment it with "repo", "repo1" "repo2" etc..

Electroid: You can, that's why the `-` is in front of the uri.

```yml
  repositories:
    - "https://github.com/Electroid/old-maps.git"
    - uri: "https://github.com/Electroid/other-maps.git"
    - uri: "https://github.com/Electroid/maps.git"
      branch: "tournament"
      folders:
        - "ctw/"
        - "dtc/"
```
Any of the following above should work.

core/src/main/java/tc/oc/pgm/map/source/GitMapSourceFactory.java
Electroid: See the documentation for `MapSourceFactory`. The `Iterator` needs to be "lazy", which means it returns immediately, but will start blocking on the first `next()`

core/src/main/java/tc/oc/pgm/PGMPlugin.java
KingOfSquares: Whats the point of "folders"? You mentioned that it was not necessary to do the partial clone thing in this PR.

KingOfSquares:
I added a parser for the git repositories, but since it does not take in `Strings` directly i hesitated putting it in `TextParser`. Still waiting for a response on the folder thing, but since this code has been laying around for a while i just thought i could push it while waiting.

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -146,6 +155,41 @@
...
+            Level.WARNING,
+            "Config ERROR: A git repository was formatted incorrectly and will be ignored until restart");
+      } catch (URISyntaxException e) {
+        logger.log(
+            Level.WARNING,
+            "Config ERROR: A URI("
+                + e.getInput()
+                + ") was formatted incorrectly and will be ignored until restart",
+            e);
+      }
```
Electroid, line 188: Instead of logging, throw the error, but wrapped as a `TextException`

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -21,20 +22,25 @@
...
 import tc.oc.pgm.api.Config;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.map.source.gitSource.GitRepository;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.text.TextException;
 
 public final class PGMConfig implements Config {
 
+  private static final Logger logger = Logger.getLogger("PGMConfig");
```
Electroid, line 42: See my other comment about not logging, so this would not be needed.

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -146,6 +155,41 @@
...
+    for (String repository : simpleRepos) {
+      gits.add(new GitRepository(repository));
+    }
+
+    for (Map<?, ?> repository : thoroughRepos) {
+      try {
+
+        URIish uri = new URIish((String) repository.get("uri"));
+
+        String branch = (String) repository.get("branch");
```
Electroid, line 171: Will this default to `master` if not defined?

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -265,6 +267,8 @@ public void reloadConfig() {
     } catch (TextException e) {
       getGameLogger().log(Level.SEVERE, e.getLocalizedMessage(), e);
       return;
+    } catch (URISyntaxException e) {
+      getLogger().log(Level.WARNING, e.getReason());
```
Electroid, line 271: See other comment, won't be needed if wrapped in `TextException`

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -273,19 +277,33 @@ public void reloadConfig() {
...
+          } catch (Throwable e) {
+            e.printStackTrace();
+            logger.log(
+                Level.WARNING,
+                "Map source " + source.getDir().getName() + " is invalid and will be ignored");
+          }
+        }
+      }
+    } catch (ClassNotFoundException e) {
+      logger.log(Level.SEVERE, "Unable to load JGit(was it excluded when compiling the jar?)");
```
Electroid, line 306: Don't log an error here, just `// No-op`

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
```diff
@@ -0,0 +1,149 @@
...
+      @Override
+      public MapSource next() {
+        return delegate.next();
+      }
+    };
+  }
+
+  private void refreshRepo() throws GitAPIException, MapMissingException {
+    Git git;
+    logger.log(Level.INFO, "Refreshing local git repository " + gitURI.getPath() + "...");
```
Electroid, line 104: No need for logging statements here.

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
```diff
@@ -0,0 +1,149 @@
...
+    try {
+      git = Git.open(dir);
+    } catch (IOException e) {
+      logger.log(
+          Level.INFO,
+          "Unable to open local git repository "
+              + gitURI.getPath()
+              + ", trying to clone remote git repository...");
+      logger.log(
+          Level.INFO, "This will happen the first time PGM tries to read a new git repository");
```
Electroid, line 115: Or here.

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
```diff
@@ -0,0 +1,149 @@
...
+      try {
+
+        CloneCommand clone = Git.cloneRepository().setDirectory(dir).setURI(gitURI.toString());
+        clone.setCredentialsProvider(provider);
+        git = clone.call();
+        changeBranch(git);
+
+        logger.log(Level.INFO, "Successfully cloned git repository " + gitURI.getPath() + "!");
```
Electroid, line 123: Or here

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
```diff
@@ -0,0 +1,149 @@
...
+        return;
+      } catch (GitAPIException f) {
+        f.printStackTrace();
```
Electroid, line 126: Don't print the stack trace, since the exception is thrown.

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
```diff
@@ -0,0 +1,149 @@
...
+                Level.WARNING,
+                "Unable to fetch repo " + gitURI.getPath() + ", caused by: " + e.getMessage());
+            e.printStackTrace();
+            if (e instanceof TransportException) {
+              logger.log(
+                  Level.INFO, "Wrong username and password provided for " + gitURI.getPath());
+            }
+          } catch (MapMissingException e) {
+            e.printStackTrace();
+          }
```
Electroid, line 80: Although I said originally to make it lazy, this is probably too weird. So just git pull when initializing the `GitMapSourceFactory` and re-pull it before calling `loadNewSources`

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
```diff
@@ -0,0 +1,149 @@
...
+import org.eclipse.jgit.transport.URIish;
+import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider;
+import tc.oc.pgm.api.map.MapSource;
+import tc.oc.pgm.api.map.exception.MapMissingException;
+import tc.oc.pgm.map.source.SystemMapSourceFactory;
+
+public class GitMapSourceFactory extends SystemMapSourceFactory {
+
+  private final URIish gitURI;
+  private final Logger logger;
```
Electroid, line 25: This factory should not need a logger, any errors should be thrown.

core/src/main/java/tc/oc/pgm/api/Config.java
```diff
@@ -29,11 +30,19 @@
...
-  List<String> getMapSources();
+  List<String> getFolderMapSources();
+
+  /**
+   * Gets a list of @link{GitRepository} which are used as remote map sources
+   *
+   * @see GitRepository
+   * @return A list of git map sources
+   */
+  List<GitRepository> getGitMapSources();
```
Electroid, line 45: I'd change this to be `List<? extends MapSourceFactory> getMapSources()`

Electroid: And only have 1 method.

core/src/main/java/tc/oc/pgm/PGMConfig.java
Pablete1234: Shouldn't default to `master`, should default to whatever upstream's default is (could be named something else)

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Pablete1234: Maybe not severe, but leaving in a warning that you're trying to load a git repo on a PGM build that does not include JGit seems pretty important, otherwise it can be hard to figure out why your repo isn't loading

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
KingOfSquares: What do you mean "before"? On `/loadnewmaps` it only calls the `loadNewSources()`.

KingOfSquares: Do you mean do something like this?:
```
@Override
public Iterator loadNewSources(){
refreshRepo();
super.loadNewSources();
}

```

core/src/main/java/tc/oc/pgm/PGMConfig.java
KingOfSquares: Yes

core/src/main/java/tc/oc/pgm/api/Config.java
KingOfSquares: That makes no sense, do you want to move the parsing of `MapSourceFactory`s into `PGMConfig`?

KingOfSquares: Because neither `GitRepository` or `String` extend `MapSourceFactory` :p

Electroid: Yea, move parsing into `PGMConfig`

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
Electroid: Yes

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Electroid: `Git is not loaded (are you running a custom version of PGM without JGit?)`

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Electroid: If folders are specified, only those folders in the git repo are to be loaded. If folders is not specified, all folders in the git repo are loaded.

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
KingOfSquares: Then i see no reason to also pull when initializing the `GitMapSourceFactory`, why twice?

core/src/main/java/tc/oc/pgm/api/Config.java
KingOfSquares: Problem with this: since you said not to include any logging in the Config it removes to the ability to display any helpful messages on any errors with the map sources. ( I can only print the stack traces, which i dont think is very helpful for a normal user).

Electroid: You can throw a `TextException`, with a useful message.

core/src/main/java/tc/oc/pgm/map/source/gitSource/GitMapSourceFactory.java
Electroid: You only need to initialize or pull when `loadnewsources` is first loaded

util/src/main/i18n/templates/error.properties
```diff
@@ -7,4 +7,14 @@ error.invalidFormat = Invalid format: '{0}' is not a {1}
 error.outOfRange = Invalid value: {0} is not between {1}
 error.suggestionSuffix = (did you mean '{0}'?)
 
+# {0} - the error message
+error.config = Config error: {0}
```
Electroid, line 11: This is too generic of a message. Error messages should be friendly and understandable by users that don't know anything about the codebase.

util/src/main/i18n/templates/error.properties
```diff
@@ -7,4 +7,14 @@ error.invalidFormat = Invalid format: '{0}' is not a {1}
...
+
+error.gitNotPresent = Git is not loaded (are you running a custom version og PGM without JGit?)
```
Electroid, line 13: I would say it's okay for this message to not be translated.

util/src/main/i18n/templates/error.properties
```diff
@@ -7,4 +7,14 @@ error.invalidFormat = Invalid format: '{0}' is not a {1}
...
+error.invalidGitRepository = A git repository was formatted incorrectly and will be ignored until reload
```
Electroid, line 14: See `TextException.invalidFormat` use that instead of this message.

util/src/main/i18n/templates/error.properties
```diff
@@ -7,4 +7,14 @@ error.invalidFormat = Invalid format: '{0}' is not a {1}
...
+# {0} - some URI(text)
+error.invalidUri = A URI({0}) was formatted incorrectly and will be ignored until reload
```
Electroid, line 16: Use `TextException.invalidFormat`

util/src/main/i18n/templates/error.properties
```diff
@@ -7,4 +7,14 @@ error.invalidFormat = Invalid format: '{0}' is not a {1}
...
+# {0} - name of git directory
+error.wrongCredentials = Unable to pull from {0}, password or username incorrect
```
Electroid, line 18: `error.badCredentials = Bad credentials: {0} username or password is incorrect`

util/src/main/i18n/templates/error.properties
KingOfSquares: The {0} in Config error is the actual error.

KingOfSquares: So, it would be `Config Error: Bad credentials: {0} username or password is incorrect`

Electroid:
FYI, I'm making changes to this branch to get it up to speed and fixing various issues.

KingOfSquares:
> FYI, I'm making changes to this branch to get it up to speed and fixing various issues.


Ok! So are you fixing the errors yourself?

Electroid:
This has now been rolled up into #503

PR #471, opened 2020-05-08
Added a changelog

rafibaum:
Now that PGM is a public plugin seeking to make a release to the wider Minecraft community, it's important we start implementing a release system. A changelog is part of that and I've written up one for the next release with all the changes since 0.7. Contributors will be expected to update this document as part of their PR with their changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,36 @@
+## 0.8 (Unreleased)
+
+### Features
+- Rotations continue to work across server restarts
+- /prox is now an alias for /proximity
+- /setpool has been added, allows you to modify map pools on-the-fly
+- Support for MySQL databases
+- Can now customize permissions for match observers and participants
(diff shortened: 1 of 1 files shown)


EricZeiberg:
I would sort the feature list, so all new commands, then new XML modules, etc...

rafibaum:
Good idea, I'll do that.

KingOfSquares:
Don't think users will care about refactors?

Electroid:
I think we should keep this very user centric. New commands, new xml features. In terms of refactors, if we do our job right, we shouldn't need to keep track of it :)

Pablete1234:
Refactors should be mentioned when they mean something to the user. Refactored X to support X and X new features, is something that you can mention, but saying that you refactored some part of the code without any user-facing counterpart isn't needed

Electroid:
This is now on the release notes.

PR #477, opened 2020-05-10
Vanish & Map Pools quick fix

applenick:
# Vanish and Map Pools quick fix
This is another quick PR to fix a few bugs with vanish and map pools.
## Vanish Subdomain fix
There was a reported issue where logging in via a `vanish` subdomain would lead to the server announcing the join message. This fixes that, plus adds a small enhancement where if a player logged in using a `vanish` subdomain they will have their vanish status removed upon quit. (Only if they did not toggle their vanish while online).

## Map Pools fix
So apparently somewhere between my last several PRs, I may have stashed some changes that were crucial to non-dynamic pools working😓. Anyway, this resolves that issue by actually checking when to revert back to dynamic pools. The second bug fixed relates to map activity and fetching the proper dynamic pool after the server restarts.

I don’t believe there’s much else to add here, so unless there is any feedback this should be good to go 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
+++ b/core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
@@ -227,25 +227,7 @@ public class AdminCommands {
       perms = Permissions.RELOAD)
   public void pgm() {
     PGM.get().reloadConfig();
-  }
-
-  @Command(
-      aliases = {"reloadpools"},
-      desc = "Reload map pools",
-      perms = Permissions.RELOAD)
-  public void reloadMapPools(Audience viewer) {
-    if (PGM.get().getMapOrder() instanceof MapPoolManager) {
-      int poolCount = ((MapPoolManager) PGM.get().getMapOrder()).reload();
-      viewer.sendWarning(
-          TranslatableComponent.of("command.admin.reloadPools", TextColor.GREEN)
-              .args(TextComponent.of(poolCount, TextColor.GOLD))
-              .append(TextComponent.space())
-              .append(
-                  TranslatableComponent.of(
-                      poolCount != 1 ? "pool.title" : "pool.name", TextColor.GREEN)));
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
```diff
@@ -229,6 +229,25 @@ public void pgm() {
     PGM.get().reloadConfig();
   }
 
+  @Command(
+      aliases = {"reloadpools"},
```
Electroid, line 233: I would rather add this functionality to `/pgm`, and no need for any command output. It's just like reloading any other part of the config.

util/src/main/i18n/templates/command.properties
```diff
@@ -77,4 +77,9 @@ command.list.participants = Participants
 command.message.noReply = Did not find a message to reply to, use {0}
 
 # {0} = player name
-command.message.blocked = {0} is not accepting messages at this time.
\ No newline at end of file
+command.message.blocked = {0} is not accepting messages at this time.
+
+# {0} = # number of map pools found
+command.admin.reloadPools = Successfully reloaded {0}
```
KingOfSquares, line 83: `Successfully reloaded {0} pools` ?

core/src/main/java/tc/oc/pgm/commands/AdminCommands.java
applenick: Alrighty! I mainly split it up to provide the output, but if it’s not necessary I’ll add this to `/pgm` 👍

util/src/main/i18n/templates/command.properties
applenick: I actually had it so `mapPools.title` & `mapPools.name` were appended to the end in order to translate the words. But I’m removing it anyway 🤷‍♂️

Issue #479, opened 2020-05-11
Proximity doesn't track safeties placed in water


Picajoluna:
It seems that wool safeties which are placed in water do not get tracked by `/proximity`.  The image below is a screenshot of us testing this, and me running `/proximity` three times in a row, after each lime wool safety is placed. As shown, the proximity distance value for Lime wool does not update.

![wool_in_water.png](https://i.imgur.com/502NVrY.png)

I believe this is an old bug.

PR #481, opened 2020-05-11
Remove previous block check on wool proximity

Pugzy:
Fixes issue described in #479. 

This bug was resolved on 1.12 Project Ares in the same way by [removing the block check](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/wool/MonumentWool.java#L93).

Not sure what the block check was needed for but assume it wasn't needed as it was removed.


Pugzy:
Let me fix that..

PR #482, opened 2020-05-11
Add stats to tab list footer

EricZeiberg:
Adds a player's stats to their tab list footer. This will only show up if the player has the stats setting turned on. 

It's a bit hacky to place this together with the time footer, and a separate `DynamicTabEntry` would be more appropriate, but the tablist rendering code does not support multiline headers / footers, and I've been told to avoid modifying that code for the time being. 

Signed-off-by: Eric Zeiberg <ejzeiberg@gmail.com>


Palmidence:
> Use the same light `GRAY` color, like the "Time: " indicator. Also instead of using `|`, just use a space.

Hard disagree, I think a separator like `|` works better than a space in this case for UX. There's no real way to "divide" the text without making it look strange and jumbled together, a `|` is a good compromise. I ran a tellraw and title for It, and it's not too nice. As far as I know, we also use `|` in other places regardless.

Electroid:
@EricZeiberg Can we get a screenshot of just the gray with the separators?

EricZeiberg:
vs


I personally like the the separators more

Indicardo:
I think I already suggested this somewhere, but how hard would it be to have the stats row *before* time?

EricZeiberg:
Not hard at all. I can make a mockup if people are interested in that.

Electroid:
Grey with separators with time and the bottom would be good.

EricZeiberg:
Here's how that looks

Indicardo:
That is the best looking so far 👍

PR #487, opened 2020-05-14
Ensure dispensers do not equip observers with armor

### LINKED ISSUE #421: Observers can steal dispensed armor
This is a weird one, not sure if PGM or SportPaper

If an observer is close to a dispenser with armor when it is activated, they can steal the armor. Look at [video](https://youtu.be/r-2cehgMUTw) for explanation.

cswhite2000:
Resolves #421 

This will check BlockDispenseEvents and see if they are equipping a player with armor. If an observer is nearby this cancels the event and simulates the dispenser firing. It has to be done this way, since dispensers equipping armor only fire BlockDispenseEvent, and that event doesn't have information on who is getting the armor.

Not sure if EventFilterMatchModule is the best place to put this.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/ArmorType.java
+++ b/core/src/main/java/tc/oc/pgm/kits/ArmorType.java
@@ -23,6 +23,37 @@ public enum ArmorType {
     return inventorySlot >= FIRST_ARMOR_SLOT && inventorySlot < FIRST_ARMOR_SLOT + values().length;
   }
 
+  public static ArmorType getArmorType(ItemStack itemStack) {
+    switch (itemStack.getType()) {
+      case DIAMOND_BOOTS:
+      case IRON_BOOTS:
+      case GOLD_BOOTS:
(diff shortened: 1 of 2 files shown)


EricZeiberg:
This seems very hacky, especially the coping of NMS code. There's likely a better solution. Maybe if an obs is detected within range, fill their armor slots temporarily so that they can't equip?

cswhite2000:
It is very hacky, but I don't see a better way to do it.
Resource wise keeping track of where observers and dispensers are doesn't seem worth it to fix this bug.
This method has the benefit of only being called very rarely when dispensers are actually used and then ignoring events where the bug won't happen.
I'd love a better way to do this, but with spigot 1.8 this seems to me like the best way.

Electroid:
I'm afraid this is just to hacky to accept. This bug is quite rare to encounter, but we may not have a good alternative at this point.

Pablete1234:
Filling up armor slots for obs with something may be a better solution, just barrier blocks or something

PR #489, opened 2020-05-16
Update proto in example xml

CrazyisCreeps:
Probably shouldn't use a proto that isn't supported in this version of PGM


CrazyisCreeps:
Closed in favor of other PR

PR #491, opened 2020-05-16
Update ProximityMetric.java to allow for "none" as an option, allowing to disable it on a per-map basis

guillevn:
It allows for none as a proximity metric
(wool-proximity-metric="none", monument-proximity-metric="none" or any other proximity metric in any gamemode) which lets mapmakers disable proximity on a per map basis, allowing for more customability. 

From what Matt has told me, it works because it doesn't have any listener, so proximity never updates for any of the teams, so it remains as "infinity" when you do /proximity.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/goals/ProximityMetric.java
+++ b/core/src/main/java/tc/oc/pgm/goals/ProximityMetric.java
@@ -9,7 +9,8 @@ public class ProximityMetric {
   public static enum Type {
     CLOSEST_PLAYER("closest player"),
     CLOSEST_BLOCK("closest block"),
-    CLOSEST_KILL("closest kill");
+    CLOSEST_KILL("closest kill"),
+    NONE("no proximity")
 
     public final String description;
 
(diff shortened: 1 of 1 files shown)


KingOfSquares:
Did you test this? Don't you have to edit how PGM reacts to the type `NONE`? Or does PGM just ignore prox if it gets passed an unknown type

guillevn:
@King yes it’s been tested, last stratus tournament used it, and it worked with the expected functionality.

Pugzy:
Spoke to @mattarnold98 who said this was ripped straight from some other changes (but works).

It would be nice to get a description of what your pull request does and how it does it. The pull request, title, change and commit message make no sense on their own. Usually, the one making the pull request is the one who did the modification and testing so can provide this information.

guillevn:
@Pugzy I'm sorry about not crediting where the change come from (https://github.com/mattarnold98/PGM/commit/a17b9f98f44b8221a20a2bab78632592bd0aeb91).

As for what it does, it allows for `none` as a proximity metric
(`wool-proximity-metric="none"`, `monument-proximity-metric="none"` or any other proximity metric in any gamemode) which lets mapmakers disable proximity on a per map basis, allowing for more customability. We used it to allow CTW matches to tie instead of depending on pre-touch proximity, because we felt that was unfair in a competitive setting. At the same time, we didn't want to disable it server-wide because our tiebreaker (it needing to break ties) needed pre-touch proximity to work as intended.

From what Matt has told me, it works because it doesn't have any listener, so proximity never updates for any of the teams, so it remains as "infinity" when you do /proximity.

PR #492, opened 2020-05-16
Show player ping in tab list

dentmaged:
Shows the player's actual ping in the tab list instead of the default 9999ms.

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/tablist/PlayerTabEntry.java
+++ b/util/src/main/java/tc/oc/pgm/util/tablist/PlayerTabEntry.java
@@ -4,6 +4,7 @@ import java.util.UUID;
 import javax.annotation.Nullable;
 import net.md_5.bungee.api.chat.BaseComponent;
 import org.bukkit.Skin;
+import org.bukkit.craftbukkit.v1_8_R3.entity.CraftPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.event.player.PlayerSkinPartsChangeEvent;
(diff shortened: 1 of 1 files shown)


Pugzy:
I did something like this a while ago and ran into two issues.

1) The tab entry is unrelated to the player at all so the ping remains constant and will not change as the player's actual ping fluctuates.

2) It didn't look too great, the consistency of having as little visual clutter on the tab list imo looks better than a mess of different coloured ping indicators (that have no real use due to their accuracy, 1-5 bars). The only exception to this is when using BLC where the `9999` red tab is a little unsightly (maybe a 0 would look better) but again what does knowing someone's ping actually achieve?

util/src/main/java/tc/oc/pgm/util/tablist/PlayerTabEntry.java
```diff
@@ -78,4 +79,9 @@ protected void onSkinPartsChange(PlayerSkinPartsChangeEvent event) {
   public String toString() {
     return this.getClass().getSimpleName() + "{" + this.player.getName() + "}";
   }
+
+  @Override
+  public int getPing() {
+    return ((CraftPlayer) this.player).getHandle().ping;
+  }
```
KingOfSquares, line 86: `this.player` is only updated on each `PlayerSkinPartsChangeEvent`. To keep it consistent you would have to request the ping form the active player object at a set interval.

Electroid:
I think for the reasons already mentioned above, we should keep the ping the same.

PR #495, opened 2020-05-18
Hide vanished players from server ping and other fixes

### LINKED ISSUE #493: Misleading setnext message
So the message in admin chat that it sends when somebody setnexts a map, is quite misleading.
It makes it look like the person who set the map, is the *author* of the map being set.  It just shouldn't use the passive voice tbh

Screen Shot 2020-05-16 at 7 08 08 PM

Doesn't it look like as though Zaquez made the map called Forge :p

It should just be "Zaquez set the next map to Forge"

applenick:
# Hide vanished players from server ping

This is just a quick PR to address a few minor bugs 🐛

## Fixes:
* Hide vanished players from server ping
* Disallow `/report` being used on vanished players
* Possibly fix issue with vanished players being seen by observers
* Fix a NPE with AntiGriefListener
* Adjust `/setnext` staff broadcast (Resolves #493)

## Screenshot:
Screen Shot 2020-05-17 at 5 07 33 PM

Let me know if there are any issues, but I did test all the changes and everything should be fine 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/community/commands/ReportCommands.java
+++ b/core/src/main/java/tc/oc/pgm/community/commands/ReportCommands.java
@@ -84,6 +84,15 @@ public class ReportCommands {
     }
 
     MatchPlayer accused = match.getPlayer(player);
+
+    // Don't allow reports of vanished players
+    if (accused.isVanished()) {
+      sender.sendWarning(
+          TextComponent.of(
+              "Could not find player named '" + player.getName() + "'", TextColor.RED));
+      return;
+    }
+
     PlayerReportEvent event = new PlayerReportEvent(sender, accused, reason);
     match.callEvent(event);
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/tc/oc/pgm/community/commands/ReportCommands.java
```diff
@@ -84,6 +84,15 @@ public void report(MatchPlayer sender, Match match, Player player, @Text String
     }
 
     MatchPlayer accused = match.getPlayer(player);
+
+    // Don't allow reports of vanished players
+    if (accused.isVanished()) {
+      sender.sendWarning(
+          TextComponent.of(
+              "Could not find player named '" + player.getName() + "'", TextColor.RED));
```
applenick, line 92: Let me know if you want this translated. I was unsure since the matching text error comes from Intake, I did not believe it would matter since other languages most likely see the English warning no matter what. 

See [here](https://github.com/Electroid/intake/blob/e34dfe29eb0f79bde70dc8678c55ea9aafd568e0/bukkit/src/main/java/app/ashcon/intake/bukkit/parametric/provider/DynamicPlayerProvider.java#L44) for details.

Electroid: Yes, add a note.

PR #499, opened 2020-05-23
Scoreboard can show more CTW teams and wools per team.

alexsosnovsky:
This pull request comprises three changes to how the scoreboard displays CTW data.

1. If a CTW map has more than five teams, there are no empty scoreboard rows. This increases the amount of teams that can be shown.
2. If a team has more than five wools, only a single space is inserted between each wool. This increases the amount of wools that can be shown. This only applies when the scoreboard is displaying compact wools.
3. Wools now have an optional "inline" attribute. If at least one of a team's wools is set to inline="true", that team's wools are shown on the same line as the team name. If this exceeds the maximum row length, spaces are inserted only between every two wools. The scoreboard is still reasonably readable in this mode.

The following shows a before and after with an 8 team CTW; each team needs to capture 7 wools. The only XML change is that the red team's wools are set as inline="true".
Click the pictures to see larger versions.

Before:

After (match start):

After (mid match):

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointMatchModule.java
@@ -36,8 +36,4 @@ public class ControlPointMatchModule implements MatchModule {
     }
     HandlerList.unregisterAll(this.announcer);
   }
-
-  public List<ControlPoint> getControlPoints() {
-    return controlPoints;
-  }
 }
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointMatchModule.java
@@ -36,4 +36,8 @@ public class ControlPointMatchModule implements MatchModule {
     }
     HandlerList.unregisterAll(this.announcer);
   }
+
+  public List<ControlPoint> getControlPoints() {
+    return controlPoints;
+  }
 }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -152,4 +152,8 @@ public class CoreMatchModule implements MatchModule, Listener {
       }
     }
   }
+
+  public List<Core> getCores() {
+    return cores;
+  }
 }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -152,8 +152,4 @@ public class CoreMatchModule implements MatchModule, Listener {
       }
     }
   }
-
-  public List<Core> getCores() {
-    return cores;
-  }
 }
--- a/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
@@ -476,6 +476,9 @@ public class SidebarMatchModule implements MatchModule, Listener {
         }
         firstTeam = false;
 
+        // Add a row for the team name
+        rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
+
         if (isCompactWool()) {
           String woolText = " ";
           boolean firstWool = true;
@@ -490,22 +493,12 @@ public class SidebarMatchModule implements MatchModule, Listener {
                 }
               });
 
-          boolean inline = false;
-          String altText = " ";
-          int index = 1;
           for (Goal goal : sortedWools) {
             if (goal instanceof MonumentWool && goal.isVisible()) {
               MonumentWool wool = (MonumentWool) goal;
-              if (wool.getDefinition().isInline() == true) {
-                inline = true;
-              }
               if (!firstWool) {
                 if (sortedWools.size() > 5) {
                   woolText += " ";
-                  if (index % 2 == 0) {
-                    altText += " ";
-                  }
-                  index += 1;
                 } else {
                   woolText += "   ";
                 }
@@ -513,38 +506,13 @@ public class SidebarMatchModule implements MatchModule, Listener {
               firstWool = false;
               woolText += wool.renderSidebarStatusColor(competitor, viewingParty);
               woolText += wool.renderSidebarStatusText(competitor, viewingParty);
-              altText += wool.renderSidebarStatusColor(competitor, viewingParty);
-              altText += wool.renderSidebarStatusText(competitor, viewingParty);
             }
           }
-
-          if (!inline) {
-            // This team's wools are not inline; add a row for the team name
-            rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
-
-            // Add a row for the compact wools
-            rows.add(woolText);
-          } else {
-            // Add team name & compact wools on a single line.
-            String teamName = competitor.getStyledName(NameStyle.FANCY).render().toLegacyText();
-            String inlineString = teamName + woolText;
-            if (inlineString.length() <= 32) {
-              rows.add(teamName + woolText);
-            } else {
-              inlineString = teamName + altText;
-              if (inlineString.length() <= 32) {
-                rows.add(teamName + altText);
-              } else {
-                rows.add(teamName + " " + altText.replaceAll("\\s", ""));
-              }
-            }
-          }
+          // Add a row for the compact wools
+          rows.add(woolText);
 
         } else {
-          // Wools are not compact; add a row for the team name
-          rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
-
-          // Add a row for each of this team's goals
+          // Not compact; add a row for each of this team's goals
           for (Goal goal : gmm.getGoals()) {
             if (!goal.isShared() && goal.canComplete(competitor) && goal.isVisible()) {
               rows.add(this.renderGoal(goal, competitor, viewingParty));
--- a/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
@@ -497,7 +497,11 @@ public class SidebarMatchModule implements MatchModule, Listener {
             if (goal instanceof MonumentWool && goal.isVisible()) {
               MonumentWool wool = (MonumentWool) goal;
               if (!firstWool) {
-                woolText += "   ";
+                if (sortedWools.size() > 5) {
+                  woolText += " ";
+                } else {
+                  woolText += "   ";
+                }
               }
               firstWool = false;
               woolText += wool.renderSidebarStatusColor(competitor, viewingParty);
--- a/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
@@ -476,9 +476,6 @@ public class SidebarMatchModule implements MatchModule, Listener {
         }
         firstTeam = false;
 
-        // Add a row for the team name
-        rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
-
         if (isCompactWool()) {
           String woolText = " ";
           boolean firstWool = true;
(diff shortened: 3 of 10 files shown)


core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -221,6 +221,10 @@ private boolean isBlitz() {
     return match.getModule(BlitzMatchModule.class) != null;
   }
 
+  private boolean isWool() {
+    return match.getModule(WoolMatchModule.class) != null;
+  }
+
```
Pablete1234, line 227: Being wool or not, shouldn't be the reason used to compact, See the `isCompactWool` method, what it defines is that wools should be compacted if they don't fit the scoreboard.

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,15 +470,12 @@ private void renderSidebar() {
       }
 
       for (Competitor competitor : sortedCompetitors) {
-        if (!firstTeam) {
+        if (!(firstTeam || (isWool() && competitorsWithGoals.size() > 5))) {
```
Pablete1234, line 473: Make the "isWool()" into a "isHighlyCompact()" or something along the lines, don't use "competitors with goals" as your reason, make it the total rows being > MAX_ROWS even using the normally compact format

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -489,21 +490,60 @@ public int compare(Goal a, Goal b) {
...
+          int index = 1;
           for (Goal goal : sortedWools) {
             if (goal instanceof MonumentWool && goal.isVisible()) {
               MonumentWool wool = (MonumentWool) goal;
+              if (wool.getDefinition().isInline() == true) {
+                inline = true;
+              }
               if (!firstWool) {
-                woolText += "   ";
+                if (sortedWools.size() > 5) {
```
Pablete1234, line 503: Again, instead of basing this on amount of wools, base it off of if the characters will fit

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -489,21 +490,60 @@ public int compare(Goal a, Goal b) {
                 }
               });
 
+          boolean inline = false;
+          String altText = " ";
+          int index = 1;
           for (Goal goal : sortedWools) {
             if (goal instanceof MonumentWool && goal.isVisible()) {
               MonumentWool wool = (MonumentWool) goal;
+              if (wool.getDefinition().isInline() == true) {
```
Pablete1234, line 499:
```suggestion
              if (wool.getDefinition().isInline()) {
```

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,15 +470,12 @@ private void renderSidebar() {
...
       for (Competitor competitor : sortedCompetitors) {
-        if (!firstTeam) {
+        if (!(firstTeam || (isWool() && competitorsWithGoals.size() > 5))) {
           // Add a blank row between teams
           rows.add("");
         }
         firstTeam = false;
 
-        // Add a row for the team name
-        rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
```
Pablete1234, line 476: I really do not think that this should be changed.
That not all teams & their wools will fit is a shame, but joining team name & objectives is a worse solution than the problem itself, i'd just accept that the last teams' wools (wich, because they should be sorted, should be the losing one) won't be visible in scoreboard and only visible in /match

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -489,21 +490,60 @@ public int compare(Goal a, Goal b) {
...
+              } else {
+                rows.add(teamName + " " + altText.replaceAll("\\s", ""));
+              }
+            }
+          }
 
         } else {
+          // Wools are not compact; add a row for the team name
+          rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
+
```
Pablete1234, line 546: See above comment for all of this inline functionality

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -489,21 +490,60 @@ public int compare(Goal a, Goal b) {
...
             if (goal instanceof MonumentWool && goal.isVisible()) {
               MonumentWool wool = (MonumentWool) goal;
+              if (wool.getDefinition().isInline() == true) {
+                inline = true;
+              }
               if (!firstWool) {
-                woolText += "   ";
+                if (sortedWools.size() > 5) {
+                  woolText += " ";
+                  if (index % 2 == 0) {
```
Pablete1234, line 505: Adding a space every other wool looks extremely weird in the scoreboard, do either one or none, but don't toggle

Pugzy:
Having some team and objectives in line and others not it makes it looks messy.
Compacting and removing lines doesn't seem like the correct solution as it only serves the use-case where you have more than usual teams/objectives but does not make it expandable beyond that.

The removal of the IP address branding and the extra line is also a tad weird as when adding this feature it was understood that it would remove scoreboard real estate in exchange for server identity and distinction, for consistency stake I believe this information should remain.

I think a pagination approach or some sort of scrolling design would be more appealing and appropriate where the scoreboard cycles between pages of objectives, maybe even keeping the players own objectives visible at all times.

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
Electroid, line 499: Inlining looks really weird, I don't agree that we should allow that.

alexsosnovsky:
I have removed inline objectives from this pull request. The merits and implementation of this feature can be discussed in the future.

Additionally, as requested, I have changed how the plugin calculates spacing between rows and wools.

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,48 +473,59 @@ private void renderSidebar() {
...
+
+            List<Goal> sortedWools = new ArrayList<>(gmm.getGoals(competitor));
+            Collections.sort(
+                sortedWools,
+                new Comparator<Goal>() {
+                  @Override
+                  public int compare(Goal a, Goal b) {
+                    return a.getName().compareToIgnoreCase(b.getName());
+                  }
+                });
```
KingOfSquares, line 500: You should use lambda here. Something like:
`(Goal a, Goal b) -> a.getName().compareToIgnoreCase(b.getName());`

Pablete1234: Well this was already like that, so it can stay, the better way would be to use a:
`Comparators.comparing(Goal::getName, String.CASE_INSENSITIVE_ORDER)`

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -470,86 +473,61 @@ private void renderSidebar() {
...
+
+            boolean horizontalCompact =
+                !((MAX_PREFIX + MAX_SUFFIX) < sortedWools.size() + 3 * (sortedWools.size() - 1));
```
Pablete1234, line 503: Leaving a small comment here mentioning why this is done could be helpful for future mantainers

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -227,6 +227,11 @@ private boolean isCompactWool() {
         && MAX_ROWS < wmm.getWools().keySet().size() * 2 - 1 + wmm.getWools().values().size();
   }
 
+  private boolean isSuperCompactWool() {
```
Pablete1234, line 230:
```suggestion
  private boolean isSuperCompact() {
```
Maybe this shouldn't be restricted to just compacting by wools, if other objectives need to show maybe it's worth removing the in-between-team lines in the scoreboard. 
Obviously the method should change accordingly to add up non-wool objectives, one line for compacted wools, and one line for the space between teams

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,48 +473,59 @@ private void renderSidebar() {
       }
 
       for (Competitor competitor : sortedCompetitors) {
```
Pablete1234, line 475:
```suggestion
      for (Competitor competitor : sortedCompetitors) {
          if (rows.size() + 2 >= MAX_ROWS) break; // No point in rendering the team at all, no space
```

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,48 +473,59 @@ private void renderSidebar() {
...
-
-          List<Goal> sortedWools = new ArrayList<>(gmm.getGoals(competitor));
-          Collections.sort(
-              sortedWools,
-              new Comparator<Goal>() {
-                @Override
-                public int compare(Goal a, Goal b) {
-                  return a.getName().compareToIgnoreCase(b.getName());
+        // Prevent team name from showing up if there is no space for its objectives
+        if (!isSuperCompactWool || rows.size() + 2 < MAX_ROWS) {
```
Pablete1234, line 483: This condition is better handled above (before potentially adding the whitespace line), also prevents the whole code getting indented inwards by 1

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,48 +473,59 @@ private void renderSidebar() {
...
+
+            for (Goal goal : sortedWools) {
+              if (goal instanceof MonumentWool && goal.isVisible()) {
+                MonumentWool wool = (MonumentWool) goal;
+                if (!firstWool) {
+                  if (horizontalCompact) {
+                    woolText += " ";
+                  } else {
+                    woolText += "   ";
+                  }
```
Pablete1234, line 513:
```suggestion
                woolText += " ";
                if (!firstWool && !horizontalCompact) woolText += " ";
```

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -466,48 +473,59 @@ private void renderSidebar() {
...
-                public int compare(Goal a, Goal b) {
-                  return a.getName().compareToIgnoreCase(b.getName());
+        // Prevent team name from showing up if there is no space for its objectives
+        if (!isSuperCompactWool || rows.size() + 2 < MAX_ROWS) {
+
+          // Add a row for the team name
+          rows.add(competitor.getStyledName(NameStyle.FANCY).render().toLegacyText());
+
+          if (isCompactWool) {
+            String woolText = " ";
```
Pablete1234, line 489:
```suggestion
            String woolText = "";
```

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -221,10 +225,39 @@ private boolean isBlitz() {
...
+    if (cpmm != null) rowsUsed += cpmm.getControlPoints().size();
+
+    FlagMatchModule fmm = match.getModule(FlagMatchModule.class);
+    if (fmm != null) rowsUsed += fmm.getFlags().size();
+
+    CoreMatchModule cmm = match.getModule(CoreMatchModule.class);
+    if (cmm != null) rowsUsed += cmm.getCores().size();
+
+    DestroyableMatchModule dmm = match.getModule(DestroyableMatchModule.class);
+    if (dmm != null) rowsUsed += dmm.getDestroyables().size();
```
Electroid, line 258: I think you could use `GoalMatchModule` to get all the goals without checking each module.

alexsosnovsky: Yeah, I see that would be the easy way. Sorry, I misunderstood how goals work.

core/src/main/java/tc/oc/pgm/controlpoint/ControlPointMatchModule.java
```diff
@@ -36,4 +36,8 @@ public void unload() {
     }
     HandlerList.unregisterAll(this.announcer);
   }
+
+  public List<ControlPoint> getControlPoints() {
+    return controlPoints;
+  }
```
Electroid, line 42: Remove this.

core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
```diff
@@ -152,4 +152,8 @@ public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
       }
     }
   }
+
+  public List<Core> getCores() {
+    return cores;
+  }
```
Electroid, line 158: And remove this.

PR #501, opened 2020-05-23
Add option to rage to allow all projectiles to be instant-kill

AustinLMayes:
```xml
<rage all-projectiles="true"/>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rage/RageModule.java
+++ b/core/src/main/java/tc/oc/pgm/rage/RageModule.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableList;
 import java.util.Collection;
 import java.util.logging.Logger;
 import org.jdom2.Document;
+import org.jdom2.Element;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/rage/RageModule.java
```diff
@@ -30,8 +38,11 @@ public MatchModule createMatchModule(Match match) {
     @Override
     public RageModule parse(MapFactory factory, Logger logger, Document doc)
         throws InvalidXMLException {
-      if (doc.getRootElement().getChild("rage") != null) {
-        return new RageModule();
+      Element rageEle = doc.getRootElement().getChild("rage");
+      if (rageEle != null) {
+        boolean allProjectiles =
+            XMLUtils.parseBoolean(rageEle.getAttribute("all-projectiles"), false);
```
Electroid, line 44: `projectile=<entity class> or <all>`, defaults to `arrow`?

PR #505, opened 2020-05-26
Add map author prefix/permission

### LINKED ISSUE #152: Map Author Ranks
Sorry if this doesn't go here, I was told like a week ago that it's most likely a PGM bug, but it used to be that when a map made by yourself is playing, you get a blue asterisk beside your name, and you get donator perms.  That would be excellent to have back.

KingOfSquares:
Adds a special permission for map authors!

Requires any users of the current version of PGM to add the section in the config manually

Thanks to: @Pablete1234 
Pair programming partners: @BGMP and @Pugzy 

Closes #152

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/MapmakerMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/MapmakerMatchModule.java
@@ -4,6 +4,7 @@ import java.util.Collection;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.map.Contributor;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
@@ -27,7 +28,7 @@ public class MapmakerMatchModule implements MatchModule, Listener {
         .getBukkit()
         .addAttachment(
             PGM.get(),
-            "pgm.group.mapmaker",
+            Permissions.MAPMAKER,
             authors.stream().anyMatch(c -> c.isPlayer(player.getId())));
     PGM.get().getPrefixRegistry().removePlayer(player.getId()); // Refresh prefixes for the player
   }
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/modules/MapmakerMatchModule.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @EventHandler(ignoreCancelled = true)
+  public void onPlayerAddEvent(final MatchPlayerAddEvent event) {
+    MatchPlayer player = event.getPlayer();
+    player
+        .getBukkit()
+        .addAttachment(
+            PGM.get(),
+            "pgm.group.mapmaker",
```
Electroid, line 30: Should define this in `Permissions`

KingOfSquares: Define the permission group? So move it from `config` -> `Permissions`?

KingOfSquares: Then i would also have to code in a prefix provider for mapmakers?

Pablete1234: All that's being asked is for the constant string `pgm.group.mapmaker` to be defined in the `Permissions.java` file instead of here

core/src/main/resources/config.yml
```diff
@@ -136,6 +136,12 @@ groups:
       - "-worldedit.navigation.jumpto.tool"
       - "-commandbook.teleport"
 
+  # A special group for authors of the playing map
+  mapmaker:
+    prefix: "&9*"
+    permissions:
+      - "pgm.premium"
```
Electroid, line 143: Also add a `+` before, so its consistent with the rest of the config.

core/src/main/java/tc/oc/pgm/modules/MapmakerMatchModule.java
KingOfSquares: Aaaah, i see. Thanks!

PR #513, opened 2020-05-28
make extra voting an actual permission node

jacobbearden:
Signed-off-by: Jacob Bearden <jacob@bearden.io>


Electroid:
Thanks for your first contribution! 🎉 Welcome :)

PR #515, opened 2020-05-28
add maps-default/ to gitignore

jacobbearden:
Signed-off-by: Jacob Bearden <jacob@bearden.io>


jacobbearden:
you merged the other one just as i noticed this so here's a second PR i guess

PR #517, opened 2020-05-29
Changes to /setpool and mutes

applenick:
# Changes to /setpool and mutes 😱 
# Add match limit to /setpool
This was suggested originally by @Pugzy and more recently @Diamyx. Basically staff can use `/setpool <pool> -m (count)` to set a limit on the number of matches to play in a non-dynamic rot before auto reverting back to the regular dynamic pools. 

In addition to adding the match limit, I also updated the broadcast message to be more aligned with the recent changed to `/setnext`

### Screenshots
New format for `/setpool` broadcast
Screen Shot 2020-05-28 at 5 49 01 PM

New format with time option
Screen Shot 2020-05-28 at 5 47 10 PM

New match limit broadcast
Screen Shot 2020-05-28 at 5 48 48 PM

# Provide mute reason to muted players
The second change is a minor one, but should provide a better user experience to those who find themselves mutes. Currently when a player is muted and they attempt to send a message, they receive no information regarding what their mute reason was. I’ve altered the storage of muted players along with the message format to allow for muted players to view the reason they were muted. This will be shown when a muted player attempts to send a message while muted.

### Screenshot
Example: player is muted for the reason `Don’t spam` and then attempts to send a message 3 times.
Screen Shot 2020-05-28 at 5 34 45 PM

These are minor changes, but as always I have tested to ensure these all work and should be good to go 👍 Please let me know if there are any modifications I should make, and I’ll get to them asap.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/CancelCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/CancelCommand.java
@@ -3,7 +3,6 @@ package tc.oc.pgm.command;
 import app.ashcon.intake.Command;
 import net.kyori.text.TranslatableComponent;
 import net.kyori.text.format.TextColor;
-import org.bukkit.command.CommandSender;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.restart.CancelRestartEvent;
@@ -19,7 +18,7 @@ public final class CancelCommand {
       aliases = {"cancel", "cancelrestart", "cr"},
       desc = "Cancels all countdowns",
       perms = Permissions.STOP)
-  public void cancel(Audience audience, CommandSender sender, Match match) {
+  public void cancel(Audience audience, Match match) {
     if (RestartManager.isQueued()) {
       match.callEvent(new CancelRestartEvent());
       audience.sendMessage(
--- a/core/src/main/java/tc/oc/pgm/command/CancelCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/CancelCommand.java
@@ -3,13 +3,16 @@ package tc.oc.pgm.command;
 import app.ashcon.intake.Command;
 import net.kyori.text.TranslatableComponent;
 import net.kyori.text.format.TextColor;
+import org.bukkit.command.CommandSender;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.listeners.ChatDispatcher;
 import tc.oc.pgm.restart.CancelRestartEvent;
 import tc.oc.pgm.restart.RestartManager;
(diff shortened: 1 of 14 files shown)


eudaldca:
Does /cancelrestart also send a message? I suppose it should for consistency. Nice work!

applenick:
@eudaldca Thanks for reminding me!
The restart command and cancel restart are each in a different class, so slipped my mind 😆 

Anyway this is what the `/cancelrestart` broadcast looks like:
Screen Shot 2020-05-29 at 10 55 39 AM

Palmidence:
What about saying `applenick has canceled a queued restart` instead?

applenick:
Thanks for the feedback @Palmidence!  

Here’s how it looks now:
Screen Shot 2020-05-29 at 11 25 14 PM

core/src/main/java/tc/oc/pgm/command/CancelCommand.java
```diff
@@ -18,11 +21,15 @@
...
-  public void cancel(Audience audience, Match match) {
+  public void cancel(Audience audience, CommandSender sender, Match match) {
     if (RestartManager.isQueued()) {
       match.callEvent(new CancelRestartEvent());
       audience.sendMessage(
           TranslatableComponent.of("admin.cancelRestart.restartUnqueued", TextColor.RED));
+      ChatDispatcher.broadcastAdminChatMessage(
+          TranslatableComponent.of("admin.cancelRestart.broadcast", TextColor.GRAY)
+              .args(UsernameFormatUtils.formatStaffName(sender, match)),
+          match);
```
Electroid, line 32: I think this is over-verbosity on messages, we already have restart message.

applenick: Just to clarify, are you referring to the `/queuerestart` broadcast as well, or should both be removed?

util/src/main/i18n/templates/map.properties
```diff
@@ -73,14 +73,14 @@ map.protectPortal = You may not obstruct this area
...
-# {1} = player name
-pool.change.force = The map pool has been set to {0} by {1}
-
-# {0} = map pool name (e.g. "Mega")
-# {1} = a duration (e.g. "2 hours")
-# {2} = player name
-pool.change.forceTimed = The map pool has been set to {0} for {1} by {2}
+# {0} = player name
+# {1} = map pool name (e.g. "Mega")
+pool.change.force = {0} set the map pool to {1}
```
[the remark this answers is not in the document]
applenick: I sent you a message regarding clarification on this one. Thanks

applenick:
@Electroid 
Sorry these changes took longer than expected, as you know I was occupied with the large text refactor 😆 

I’ve removed the broadcast messages and returned the pool broadcast parameters to their original positions. I don’t believe there are any more changes to make so it should be good to go 👍

PR #528, opened 2020-06-09
XML option to show where a flag will respawn at next.

mrcookieunderscore13:
This change is intended to help king of the flag play better (ie desert sanctuary). With this change, as soon as the flag is picked up, if show-respawn-on-pickup="true", it will say in chat where the flag will respawn at next.

I added the attribute show-respawn-on-pickup to <flag>. It defaults to false. Also, the chosen post must have a name in order for the chat message to display.

Signed-off-by: mrcookie

### DIFF
--- a/util/src/main/i18n/templates/gamemode.properties
+++ b/util/src/main/i18n/templates/gamemode.properties
@@ -217,6 +217,10 @@ flag.willRespawn = {0} will respawn in {1}
 flag.willRespawn.named = {0} will respawn at {1} in {2}
 
 # {0} = flag name
+# {1} = post name
+flag.willRespawn.next = {0} will respawn next at {1}
+
+# {0} = flag name
 flag.respawn = {0} has respawned
 
 # {0} = flag name
--- a/core/src/main/java/tc/oc/pgm/flag/Flag.java
+++ b/core/src/main/java/tc/oc/pgm/flag/Flag.java
@@ -88,6 +88,7 @@ public class Flag extends TouchableGoal<FlagDefinition> implements Listener {
   private final Set<Team> completers;
   private BaseState state;
   private boolean transitioning;
+  private @Nullable Post predeterminedPost;
 
   protected Flag(Match match, FlagDefinition definition, ImmutableSet<Net> nets)
       throws ModuleLoadException {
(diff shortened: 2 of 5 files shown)


KingOfSquares:
Nice! 👍 

I think `flag.announceNextRespawn` has a rather strange phrasing, but i'm not native so i'm not entirely sure about that..

Rest is great!

applenick:
> I think `flag.announceNextRespawn` has a rather strange phrasing, but i'm not native so i'm not entirely sure about that..

Perhaps something like `flag.willRespawn.next` would be more suited for this?

KingOfSquares:
Didnt think of that actually, was thinking about the phrase.

applenick:
@KingOfSquares OH I read your reply a bit fast, you were referring to  `{0} will respawn at {1} next` not `flag.announceNextRespawn`? xD Whoops

KingOfSquares:
Yes

mrcookieunderscore13:
> Nice! 👍
> 
> I think `flag.announceNextRespawn` has a rather strange phrasing, but i'm not native so i'm not entirely sure about that..
> 
> Rest is great!

I think the wording is fine, but I am happy to change it--I just don't know what I would change it to.

mrcookieunderscore13:
> Here's my thought: if named posts are used, PGM should always announce where the flag will spawn next. Anything wrong with that?

The one issue that comes to mind is if a map uses a net to change the respawn post, then the message could be wrong

mrcookieunderscore13:
> Here's my thought: if named posts are used, PGM should always announce where the flag will spawn next. Anything wrong with that?

So I made it so `show-respawn-on-pickup` default to true. I left it as an xml option incase a map author wants his/her map to not show the message. If a net defines a respawn post, `show-respawn-on-pickup` is forced to false for flags that can be captured at the net. This is because if a net changes the respawn post, it is impossible to know what post the flag will respawn to when it is picked up.

So, the message will display for a given flag when the following are true:
1) The post is named
2) No nets that the flag can be captured in define a respawn post
3) The map author has not defined `show-respawn-on-pickup="false"` in the `<flag>` element

util/src/main/i18n/templates/gamemode.properties
```diff
@@ -216,6 +216,10 @@ flag.willRespawn = {0} will respawn in {1}
 # {2} = duration
 flag.willRespawn.named = {0} will respawn at {1} in {2}
 
+# {0} = flag name
+# {1} = post name
+flag.announceNextRespawn = {0} will respawn at {1} next
```
Electroid, line 221: `flag.willRespawn.next = {0} will respawn next at {1}`

util/src/main/i18n/templates/gamemode.properties
```diff
@@ -216,6 +216,10 @@ flag.willRespawn = {0} will respawn in {1}
 # {2} = duration
 flag.willRespawn.named = {0} will respawn at {1} in {2}
 
+# {0} = flag name
+# {1} = post name
+flag.announceNextRespawn = {0} will respawn next at {1}
```
Electroid, line 221: `flag.willRespawn.next`

mrcookieunderscore13:
I accidentally messed up the rebase, so the commit history is a little bit scuffed. Everything should be good to go though.

PR #536, opened 2020-06-14
Fix a few text related issues

### LINKED ISSUE #534: Renaming teams does not update team chat prefix
After a team is aliased as something else, talking in chat still uses the old prefix, like `(Red)` instead of `(New name)`

applenick:
# Fix a few text related issues
Most of these issues come from the recent text refactor, all of which are resolved by this PR 🎉 

* Vanished mapmakers are no longer exposed via various map info locations
* Fixed issue related to names not rendering decorations (death message bug)
* Resolve team chat prefix not using proper aliased name (resolves #534)

## Screenshot
Screen Shot 2020-06-14 at 1 11 40 AM

Screen Shot 2020-06-14 at 12 34 22 AM

### DIFF
--- a/core/src/main/java/tc/oc/pgm/teams/Team.java
+++ b/core/src/main/java/tc/oc/pgm/teams/Team.java
@@ -37,8 +37,6 @@ public class Team extends SimpleParty implements Competitor, Feature<TeamFactory
   private TeamMatchModule tmm;
   private JoinMatchModule jmm;
   protected @Nullable String name = null;
-  protected @Nullable Component componentName;
-  protected Component chatPrefix;
   protected Integer minPlayers, maxPlayers, maxOverfill;
 
   // Recorded in the match document, Tourney plugin sets this
@@ -128,10 +126,7 @@ public class Team extends SimpleParty implements Competitor, Feature<TeamFactory
 
   @Override
   public Component getName(NameStyle style) {
-    if (componentName == null) {
-      this.componentName = TextComponent.of(getNameLegacy(), TextFormatter.convert(getColor()));
-    }
-    return componentName;
+    return TextComponent.of(getNameLegacy(), TextFormatter.convert(getColor()));
   }
 
   /**
@@ -166,7 +161,6 @@ public class Team extends SimpleParty implements Competitor, Feature<TeamFactory
     if (Objects.equals(this.name, newName) || this.getNameLegacy().equals(newName)) return;
     String oldName = this.getNameLegacy();
     this.name = newName;
-    this.componentName = null;
     this.match.callEvent(new PartyRenameEvent(this, oldName, this.getNameLegacy()));
   }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/teams/Team.java
```diff
@@ -37,8 +37,6 @@
   private TeamMatchModule tmm;
   private JoinMatchModule jmm;
   protected @Nullable String name = null;
-  protected @Nullable Component componentName;
```
Electroid, line 40: We should be caching the `Componet` name. Instead, listen to the team rename event and change it.

util/src/main/java/tc/oc/pgm/util/text/types/PlayerComponent.java
```diff
@@ -38,32 +38,49 @@ static Component of(@Nullable Player player, String defName, NameStyle style) {
       return component.build();
     }
 
-    Component usernameComponent;
+    // For name styles that don't allow vanished, make vanished appear offline
+    if (!style.showVanish && isVanished(player)) {
```
Electroid, line 42: I think we need to clean up this code. Maybe separate methods for each `NameStyle`?

applenick: I agree, with the previous refactor I tried to keep it similar to the old `PersonalizedPlayer` class, but imo would be better to separate into different methods as you suggested 👍 

I’ll take care of this in a bit.

applenick:
@Electroid 
I reverted the change related to Team components caching. I decided against using the team rename event and instead opted for a reset to the chatPrefix which should fix the issue with team names updating.

As for `PlayerComponent`, I went ahead and cleaned up `NameStyle` to be more aligned with recent changes and what’s currently supported. Also as requested I split up the rendering for each style into different methods. 

Let me know if you have any notes or suggestions! Tested and everything works so should be fine 👍

util/src/main/java/tc/oc/pgm/util/text/types/PlayerComponent.java
```diff
@@ -81,6 +68,109 @@ static Component of(CommandSender sender, NameStyle style) {
         : TranslatableComponent.of("misc.console", TextColor.DARK_AQUA);
   }
 
+  // What an offline or vanished username renders as
+  static Component formatOffline(String name, boolean plain) {
```
Electroid, line 72: Make all of these functions return `Component.Builder` to avoid copying `Component`s among the methods. In `PlayerComponent.of` you can build it before returning it.

applenick: Ah better idea! I’ll fix this up 👍

applenick: I believe you meant `TextComponent.Builder`, but it’s all cleaned up now 👍

PR #537, opened 2020-06-14
Allow voting book to be forced open

Pugzy:
When the setting `vote` is set to `off` the voting book does not popup when given to the player, great.

BUT I believe that once you interact with the book (opening and clicking a map) the book should reopen for you to vote for another map until you close the book manually. ATM it's a bit of a faff to have the `vote` `off` as you need to manually open the book each time you select more than one map.

This adds a forced argument to the `sendBook` method which if set to true will force the book open. If false it will use the player's settings. This is then used via a command flag `-f` when a book is voted for by clicking its name in the book.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
@@ -275,7 +275,11 @@ public final class MapPoolCommand {
 
   @Command(aliases = "votenext", desc = "Vote for the next map", usage = "map")
   public static void voteNext(
-      MatchPlayer player, CommandSender sender, MapOrder mapOrder, @Text MapInfo map)
+      MatchPlayer player,
+      CommandSender sender,
+      MapOrder mapOrder,
+      @Switch('o') boolean forceOpen,
+      @Text MapInfo map)
       throws CommandException {
     MapPool pool = getMapPoolManager(sender, mapOrder).getActiveMapPool();
     MapPoll poll = pool instanceof VotingPool ? ((VotingPool) pool).getCurrentPoll() : null;
@@ -291,7 +295,7 @@ public final class MapPoolCommand {
             voteResult ? TextColor.GREEN : TextColor.RED,
             map.getStyledName(MapNameStyle.COLOR));
     player.sendMessage(voteAction);
-    poll.sendBook(player);
+    poll.sendBook(player, forceOpen);
   }
 
   public static MapPoolManager getMapPoolManager(CommandSender sender, MapOrder mapOrder)
--- a/core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -136,7 +136,7 @@ public class MapPoll {
         .build();
   }
 
-  public void sendBook(MatchPlayer viewer) {
+  public void sendBook(MatchPlayer viewer, boolean forceOpen) {
     String title = ChatColor.GOLD + "" + ChatColor.BOLD;
     title += TextTranslations.translate("vote.title.map", viewer.getBukkit());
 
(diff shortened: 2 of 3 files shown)


core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
```diff
@@ -275,7 +275,11 @@ public static void skip(
 
   @Command(aliases = "votenext", desc = "Vote for the next map", usage = "map")
   public static void voteNext(
-      MatchPlayer player, CommandSender sender, MapOrder mapOrder, @Text MapInfo map)
+      MatchPlayer player,
+      CommandSender sender,
+      MapOrder mapOrder,
+      @Switch('f') boolean force,
```
Pablete1234, line 281: I don't think this needs to be a parameter for the command
```suggestion
```

core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
```diff
@@ -291,7 +295,7 @@ public static void voteNext(
             voteResult ? TextColor.GREEN : TextColor.RED,
             map.getStyledName(MapNameStyle.COLOR));
     player.sendMessage(voteAction);
-    poll.sendBook(player);
+    poll.sendBook(player, force);
```
Pablete1234, line 298:
```suggestion
    poll.sendBook(player, true);
```

core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -180,7 +180,7 @@ private Component getMapBookComponent(MatchPlayer viewer, MapInfo map) {
                 TextComponent.of(
                     map.getTags().stream().map(MapTag::toString).collect(Collectors.joining(" ")),
                     TextColor.YELLOW)))
-        .clickEvent(ClickEvent.runCommand("/votenext " + map.getName()))
+        .clickEvent(ClickEvent.runCommand("/votenext -f " + map.getName()))
```
Pablete1234, line 183:
```suggestion
        .clickEvent(ClickEvent.runCommand("/votenext " + map.getName()))
```

core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -136,7 +136,7 @@ private Component getMapChatComponent(MatchPlayer viewer, MapInfo map, boolean w
         .build();
   }
 
-  public void sendBook(MatchPlayer viewer) {
+  public void sendBook(MatchPlayer viewer, boolean force) {
```
Pablete1234, line 139:
```suggestion
  public void sendBook(MatchPlayer viewer, boolean forceOpen) {
```
force in this context could mean if the book is forced, i think making it more explicit can help

core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
Pugzy: If I do `votenext The Nile` I don't want the book to popup.

Pablete1234: Without seeing the book, you can't even know `The Nile` is an option, neither will you know if it's been voted for or against. The command is an implementation detail for how the book works, never intended to be used standalone. If you wish to make it a command make it so that skipping the -f flag will display voting messages in chat that can replace the functionality of the book.

core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
Electroid: How about `-o` for open. I think we may want to keep the flag, to allow running the command without opening the book.

Pugzy:
Should I update the usage of `forceOpen` to just be `open` to match the `-o` flag?

PR #539, opened 2020-06-18
Reintroduce bold tab name

applenick:
# Reintroduce bold tab name 🎉 

This is a quick PR to reintroduce bolded names to the tab, as well as fix dead players not having their names grayed out. (This was just discovered due to #536)

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/tablist/PlayerTabEntry.java
+++ b/util/src/main/java/tc/oc/pgm/util/tablist/PlayerTabEntry.java
@@ -2,7 +2,6 @@ package tc.oc.pgm.util.tablist;
 
 import java.util.UUID;
 import javax.annotation.Nullable;
-import net.kyori.text.format.TextDecoration;
 import net.md_5.bungee.api.chat.BaseComponent;
 import org.bukkit.Skin;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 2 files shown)


util/src/main/java/tc/oc/pgm/util/tablist/PlayerTabEntry.java
```diff
@@ -47,7 +48,9 @@ public PlayerTabEntry(Player player) {
   @Override
   public BaseComponent getContent(TabView view) {
     return TextTranslations.toBaseComponent(
-        PlayerComponent.of(player, NameStyle.TAB), view.getViewer());
+        PlayerComponent.of(player, NameStyle.TAB)
+            .decoration(TextDecoration.BOLD, view.getViewer().equals(player)),
```
Electroid, line 52: Should this be done inside the `PlayerComponent` renderer?

applenick: Gotcha 👍 
I moved the rendering to the `PlayerComponent` though it did introduce a `viewer` parameter. Also I opted to not add back the `showSelf` to NameStyle, as the bolding of names will only be done in the tab. Does not make sense to define it if other styles will not utilize it. Let me know if this was the right approach or if you have an alternative.

KingOfSquares:
Bold names have dissapeared from other places as well, like death/kill messages.

applenick:
@KingOfSquares 
 #536 resolved the bolded name issues for death messages 😄

PR #540, opened 2020-06-22
Add custom vote management 

applenick:
# Add custom vote management 

This PR adds a new command (`/vote`) which allows for up to 5 custom maps to be added by staff to override the next map vote. 

## Command
`/vote add [map name]` Adds the specified map to the selection
`/vote remove [map name]`Removes specified map from the selection
`/vote clear` Removes all maps that were added
` /vote list` Displays a list of selected maps

## Screenshots:

Command feedback when a map is added or removed from the selection
Screen Shot 2020-06-22 at 12 16 44 AM

List format when using `/vote list` ( the first number’s color changes depending on how many maps are in the selection)
Screen Shot 2020-06-22 at 12 17 40 AM

Command feedback when using `/vote clear` to clear entire selection
Screen Shot 2020-06-22 at 12 17 00 AM

Error when trying to add maps over the limit (5)
Screen Shot 2020-06-22 at 12 23 36 AM

Example of voting book when custom selection is active and was provided with 3 maps:
Screen Shot 2020-06-22 at 9 34 44 AM

## Feedback
I rewrote this command from originally being `/setvote` into `/vote`, allowing for a more organized and standard command structure. If anyone had additional features that would fit the scope of adjusting votes, please let me know!

As always this was tested and should be good to go if no changes are requested 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
@@ -2,6 +2,8 @@ package tc.oc.pgm.command;
 
 import app.ashcon.intake.Command;
 import app.ashcon.intake.CommandException;
+import app.ashcon.intake.bukkit.parametric.Type;
+import app.ashcon.intake.bukkit.parametric.annotation.Fallback;
 import app.ashcon.intake.parametric.annotation.Default;
 import app.ashcon.intake.parametric.annotation.Switch;
 import app.ashcon.intake.parametric.annotation.Text;
@@ -24,12 +26,14 @@ import tc.oc.pgm.api.map.MapOrder;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.cycle.CycleCountdown;
+import tc.oc.pgm.listeners.ChatDispatcher;
 import tc.oc.pgm.rotation.MapPoll;
 import tc.oc.pgm.rotation.MapPool;
 import tc.oc.pgm.rotation.MapPoolManager;
 import tc.oc.pgm.rotation.Rotation;
 import tc.oc.pgm.rotation.VotingPool;
 import tc.oc.pgm.util.PrettyPaginatedComponentResults;
+import tc.oc.pgm.util.UsernameFormatUtils;
 import tc.oc.pgm.util.chat.Audience;
 import tc.oc.pgm.util.named.MapNameStyle;
 import tc.oc.pgm.util.text.TextFormatter;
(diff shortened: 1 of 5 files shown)


core/src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -88,10 +95,47 @@ public void matchEnded(Match match) {
...
               if (RestartManager.isQueued()) return;
-              currentPoll = new MapPoll(match, mapScores, VOTE_SIZE);
+              currentPoll =
+                  shouldCustomVote()
+                      ? new MapPoll(
+                          match,
+                          getCustomVoteMapWeighted(),
+                          Math.min(MAX_VOTE_OPTIONS, customVoteMaps.size()))
+                      : new MapPoll(match, mapScores, VOTE_SIZE);
+
```
Pablete1234, line 105: I assume that custom votes should be added to the poll alongside the other maps, instead of completely overwrite

applenick: @Pablete1234 When implementing, my idea was for a complete override. I feel this gives staff more control, so say if they only wanted players to select from 2 or 3 maps. Perhaps an additional flag could be added that will “fill” the remaining selections with random maps from the current pool?

Pablete1234: Yeah, how i understood it is if someone is nagging you about *HEY SN THIS MAP* you can just add it to the vote and that way players can decide if they wanna play it, alongside other 4 "normal" maps

core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
```diff
@@ -276,6 +280,118 @@ public static void skip(
...
+      MapOrder mapOrder,
+      Match match,
+      @Switch('c') boolean clear,
+      @Switch('r') boolean remove,
+      @Switch('l') boolean list)
+      throws CommandException {
+    MapPool pool = getMapPoolManager(sender, mapOrder).getActiveMapPool();
+    VotingPool vote = (pool instanceof VotingPool ? ((VotingPool) pool) : null);
+
+    if (vote == null) {
```
AustinLMayes, line 302: Probably shouldn't allow the pool to be modified during a voting period.

applenick:
Thanks for all the great feedback. I see now this will require some reworking, so I converted this to a draft while I work on that. I’ll push an update later today 👍

applenick:
I rewrote a majority of this PR to address most of the feedback that was initially received. If there are any other suggestions let me know and I’d be glad to make accommodations 👍

core/src/main/java/tc/oc/pgm/command/VotingCommand.java
```diff
@@ -0,0 +1,170 @@
...
+import tc.oc.pgm.rotation.VotingPool;
+import tc.oc.pgm.util.UsernameFormatUtils;
+import tc.oc.pgm.util.chat.Audience;
+import tc.oc.pgm.util.named.MapNameStyle;
+import tc.oc.pgm.util.text.TextTranslations;
+
+public class VotingCommand {
+
+  @Command(
+      aliases = {"add", "set"},
```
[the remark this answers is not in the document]
applenick: @Pablete1234 
I agree, makes more sense for it to just be add. However for the tab-complete issue, apparently only the first alias is shown. 
Screen Recording 2020-06-25 at 11 15 AM

What I would like to adjust however is the feedback that is given when executing `/vote` which displays this:
Screen Shot 2020-06-25 at 11 16 10 AM

Imo, only the first alias should be shown in messages like this. When I have time, I’m going to figure out what exactly handles this (not sure atm if it’s intake of PGM, most likely intake).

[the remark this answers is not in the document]
Pablete1234: Aliases make sense when it's "the same word", like list & ls or remove & rm, but add/set is just misleading

util/src/main/i18n/templates/map.properties
```diff
@@ -54,6 +54,29 @@ map.setNext.confirm = You may not set the next map when a restart is queued. Use
 # {1} = name of map that was unset
 map.setNext.revert = {0} has removed {1} as the next map
 
+# {0} = player name (who added the map)
+# {1} = map name 
+map.setVote.addMap = {0} added {1} to the custom vote
```
Electroid, line 59: Let's change these keys to `vote.custom.add` etc (make them generic away from maps)

core/src/main/java/tc/oc/pgm/command/VotingCommand.java
```diff
@@ -0,0 +1,170 @@
...
+  @Command(
+      aliases = {"clear"},
+      desc = "Clear all custom map selections from the next vote",
+      perms = Permissions.SETNEXT)
+  public void clearMaps(CommandSender sender, Match match, MapOrder mapOrder)
+      throws CommandException {
+    VotingPool vote = getVotingPool(sender, mapOrder);
+
+    Component current =
+        TextComponent.of(Integer.toString(vote.getCustomVoteMaps().size()), TextColor.RED);
```
KingOfSquares, line 98: You don't need to wrap the integer in `Integer.toString` when using `TextComponent.of`, it has a constructor `TextComponent.of(int value)`

core/src/main/java/tc/oc/pgm/command/VotingCommand.java
```diff
@@ -0,0 +1,170 @@
...
+  @Command(
+      aliases = {"list", "ls"},
+      desc = "View a list of maps that have been selected for the next vote")
+  public void listMaps(CommandSender sender, Audience viewer, MapOrder mapOrder)
+      throws CommandException {
+    VotingPool vote = getVotingPool(sender, mapOrder);
+
+    int currentMaps = vote.getCustomVoteMaps().size();
+    TextColor listNumColor =
+        currentMaps > 1 ? currentMaps < 5 ? TextColor.GREEN : TextColor.YELLOW : TextColor.RED;
```
KingOfSquares, line 125: `5` -> `vote.MAX_VOTE_OPTIONS`? Maybe also define a `MIN_VOTE_OPTIONS`

applenick: Thanks @KingOfSquares ! Great suggestions 👍

util/src/main/i18n/templates/map.properties
applenick: Very good idea, looks much better now 😄

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
 
 # {0} = option name (e.g. "Airship Battle")
 vote.abstain = Removed your vote for {0}
+
+vote.disabled = Voting is disabled
+
+# {0} = player name (who added the map)
+# {1} = map name 
+vote.custom.add = {0} added {1} to the custom vote
```
Electroid, line 127: `{0} added {1} to the vote`

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+# {0} = player name (who removed the map)
+# {1} = map name 
+vote.custom.remove = {0} removed {1} from the custom vote
```
Electroid, line 131: `{0} removed {1} from the vote`

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+# {0} = map name
+vote.custom.existing = {0} has already been added to the custom vote
```
Electroid, line 134: Remove this message, in your implementation, just show he `vote.custom.add` message.

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+# {0} = player name (who cleared the map selection)
+vote.custom.clear = {0} has cleared the custom vote selection
```
Electroid, line 137: Remove this message, instead use the `vote.custom.remove` string with a list for `{1}`. It would look like:
`Electroid removed cos(tnt) and Boom Box from the vote`. Use the translations list utility.

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+# {0} = map name
+vote.custom.notFound = {0} could not be removed since it was never added
```
Electroid, line 140: Is there a generic not found message? Use that instead

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+vote.custom.inProgress = Voting selection can not be modified at this time
```
Electroid, line 142: Just say voting is disabled, and reuse that message.

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+vote.custom.limit = Map limit for the custom vote has been reached
```
Electroid, line 144: Make this generic not just for maps.

util/src/main/i18n/templates/map.properties
Electroid: Also drop the `custom` from all the keys, so just `vote.add`

util/src/main/i18n/templates/map.properties
```diff
@@ -119,3 +119,29 @@ vote.against = Voted against {0}
...
+
+vote.custom.list = Selected Maps
```
Electroid, line 146: `vote.title.map = Selected Maps`

In the future we could have `vote.title.gamemode`

Electroid:
Also, reopen this PR on branch `v0.9`. We're no longer using the word `master`

PR #545, opened 2020-06-24
Start sequential posts from 1

iangbb:
As the default flag post is post 0, we should start from 1 when iterating sequential posts, as otherwise post 0 is duplicated in the sequence.

For three posts A, B, C, the current situation: A -> A -> B -> C -> A -> B -> C -> ...

Starting from 1: A -> B -> C -> A -> B -> C -> ...

Signed-off-by: Ian <46601096+iangbb@users.noreply.github.com>

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/Flag.java
+++ b/core/src/main/java/tc/oc/pgm/flag/Flag.java
@@ -242,9 +242,7 @@ public class Flag extends TouchableGoal<FlagDefinition> implements Listener {
       return post;
     }
     if (definition.isSequential()) {
-      if (sequentialPostCounter == definition.getPosts().size()) {
-        sequentialPostCounter = 0;
-      }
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/flag/Flag.java
```diff
@@ -234,19 +234,18 @@ public boolean canDropAt(Location location) {
...
 
   public Post getReturnPost(Post post) {
     if (post.isSpecifiedPost()) {
       return post;
     }
     if (definition.isSequential()) {
-      Post returnPost = definition.getPosts().get(sequentialPostCounter++);
       if (sequentialPostCounter == definition.getPosts().size()) {
         sequentialPostCounter = 0;
       }
```
Pablete1234, line 247:
```suggestion
      sequentialPostCounter %= definition.getPosts().size();
```

PR #550, opened 2020-07-06
Add Wool Flag Beams

AustinLMayes:
Adds the `LegacyFlagBeamMatchModule` which renders wool blocks (using fake zombies) over flags so that all players can see the beams. Currently renders wool beams for all player versions for continuity. 

Code originally written by @Electroid [here](https://github.com/StratusNetwork/projectares/commit/52c82a81e7c12722e9c644d526b227dcce373f06#diff-59b90ff90f481d22f799020fdd02c882) and modified by @jasoryeh and @ShinyDialga

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Modules.java
+++ b/core/src/main/java/tc/oc/pgm/api/Modules.java
@@ -43,6 +43,7 @@ import tc.oc.pgm.filters.FilterModule;
 import tc.oc.pgm.fireworks.FireworkMatchModule;
 import tc.oc.pgm.flag.FlagMatchModule;
 import tc.oc.pgm.flag.FlagModule;
+import tc.oc.pgm.flag.LegacyFlagBeamMatchModule;
 import tc.oc.pgm.gamerules.GameRulesMatchModule;
 import tc.oc.pgm.gamerules.GameRulesModule;
 import tc.oc.pgm.goals.GoalMatchModule;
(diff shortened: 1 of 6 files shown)


Pablete1234 (approved):
Thanks for taking the time to upstream changes

Electroid:
Going to put this on hold for a bit, this should only be applied to 1.7 players.

jasoryeh:
> Going to put this on hold for a bit, this should only be applied to 1.7 players.

The change to apply to all players was mainly because we saw the wool flag beams as being too advantageous for 1.7 players because the wool blocks are way more visible (a solid block vs. smaller particles) than particles, if anything the fair solution for all players should be to either figure out particles for 1.7 or use flag beams only since 1.8+ players can currently see particles + the wool beam.

Also to further review this a change could be to further extend the wool height to > 32 blocks since the particles appear to go up farther than that.

Pablete1234:
The reason this feature exists is so that 1.7 clients can see *something* instead of nothing about a flag, since currently, the particles don't show from very far for those players.
Maybe a config option (or maybe a player setting?) can be added to enable this behaviour for everyone, but i do not think it should be a default for 1.8+ clients since the particles are way better representation than the zombies, wich have issues like arrows bouncing off of them client-side

jasoryeh:
That could be an option. 

Looking at it from 1.7 perspectives I think one more change can be made to extend the wool beam higher so it matches the particle's height as it appears the particles extend far taller than what is present currently.

Electroid:
I'm willing to try this out in production, see how people like it. We can potentially make it optional later if need be.

PR #554, opened 2020-07-07
Rework /vote logic

applenick:
# Rework /vote logic

This PR started out as a new command (`/vote fill`) but has since removed that command in favor of implementing both a `Replacement` and `Override` selection mode.

## `Replacement`
When vote selection is in the replacement mode, maps added to the selection will replace maps that are provided by the current pool.

Example:
Let’s say we wanted to add three maps to the vote `The Nile`, `Ozone`, and `Sand Wars`. Since three maps are added, two will be provided based on the pool. 

Using `/vote list` (As you can see, there is now an indicator of what mode is active, in this case Replace)
Screen Shot 2020-07-07 at 11 16 45 AM

When the vote is ready, you’ll notice that 2 additional maps from the current pool were added. 
Screen Shot 2020-07-07 at 11 17 06 AM

## `Create`
Create preserves the current functionality, where the selected maps will be the **only**  maps provided in the poll.

Using `/vote list` again, we can see the three selected maps. Also pictured is the hover message that is displayed over the mode indicator. Note: you can click the status to run `/vote mode` to toggle the mode. 
Screen Shot 2020-07-07 at 11 20 59 AM

When the vote is ready, only the selected maps will be visible. 
Screen Shot 2020-07-07 at 11 18 30 AM

Also here is the formatting used when executing `/vote mode` twice.
Screen Shot 2020-07-07 at 11 20 15 AM

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
@@ -45,13 +45,15 @@ public class MapPoll {
 
   private final WeakReference<Match> match;
   private final Map<MapInfo, Double> mapScores;
+  private final Set<MapInfo> overrides;
   private final int voteSize;
 
   private final Map<MapInfo, Set<UUID>> votes = new HashMap<>();
 
-  MapPoll(Match match, Map<MapInfo, Double> mapScores, int voteSize) {
+  MapPoll(Match match, Map<MapInfo, Double> mapScores, Set<MapInfo> overrides, int voteSize) {
     this.match = new WeakReference<>(match);
     this.mapScores = mapScores;
+    this.overrides = overrides;
     this.voteSize = voteSize;
 
     selectMaps();
@@ -68,7 +70,10 @@ public class MapPoll {
     NavigableMap<Double, MapInfo> cumulativeScores = new TreeMap<>();
     double maxWeight = cummulativeMap(0, sortedDist, cumulativeScores);
 
-    for (int i = 0; i < voteSize; i++) {
+    // Add all override maps before selecting random
+    overrides.forEach(map -> votes.put(map, new HashSet<>()));
+
+    for (int i = overrides.size(); i < voteSize; i++) {
       NavigableMap<Double, MapInfo> subMap =
           cumulativeScores.tailMap(Math.random() * maxWeight, true);
       Map.Entry<Double, MapInfo> selected = subMap.pollFirstEntry();
--- a/core/src/main/java/tc/oc/pgm/rotation/VotingPool.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/VotingPool.java
@@ -2,6 +2,7 @@ package tc.oc.pgm.rotation;
 
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
@@ -30,6 +31,9 @@ public class VotingPool extends MapPool {
 
   /** Override pool is used when custom maps are defined, does not count for regular vote score * */
   private Set<MapInfo> customVoteMaps = Sets.newHashSet();
+
+  /** Whether custom vote should replace existing maps or override the poll * */
+  private boolean replaceMaps = true;
 
   public VotingPool(MapPoolManager manager, ConfigurationSection section, String name) {
     super(manager, section, name);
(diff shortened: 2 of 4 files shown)


core/src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -119,6 +119,24 @@ public boolean addCustomVoteMap(MapInfo info) {
...
+        MapInfo random = getRandom();
+        if (!isCustomMapSelected(random)) {
+          selected.add(random);
+          addCustomVoteMap(random);
+        } else {
+          maxRetrys--;
+        }
+      }
+    }
+    return selected;
```
Pablete1234: This is adding maps randomly, it'd be nice if it added maps based on the vote-selection logic of scores and weights

applenick: That’s a really good idea. I’ll look into switching this logic around later today 👍

Electroid: Can we just auto-fill for the user? Why require them to fill?

Pablete1234: It's just an issue of how vote modifying commands were approached, it was discussed in #540 [here](https://github.com/PGMDev/PGM/pull/540#discussion_r443683772). It would be expected that if you just add a map to the vote it would "force" one map and keep the other 4 random. Instead, what vote managing commands do, is expect you to select 2-5 options for the user to pick from. Some form of a "setnext between this options".

I think vote commands should always replace options of whatever vote will be held next, and maybe optionally make it so it makes the vote *only* the maps added via command, but i heavily doubt that will ever be used (wich is what we currently have)

applenick: I converted this PR to a draft and will work on adjusting the logic to handle vote management as you suggested @Pablete1234 👍

applenick:
@Electroid & @Pablete1234 
I updated the logic, please see the new PR overview for how everything works. Let me know if you have any feedback and thanks for the suggestions 😄

core/src/main/java/tc/oc/pgm/rotation/MapPoll.java
```diff
@@ -68,6 +70,9 @@ private void selectMaps() {
     NavigableMap<Double, MapInfo> cumulativeScores = new TreeMap<>();
     double maxWeight = cummulativeMap(0, sortedDist, cumulativeScores);
 
+    // Add all override maps before selecting random
+    overrides.forEach(map -> votes.put(map, new HashSet<>()));
+
     for (int i = 0; i < voteSize; i++) {
```
Pablete1234:
```suggestion
    for (int i = overrides.size(); i < voteSize; i++) {
```

core/src/main/java/tc/oc/pgm/rotation/VotingPool.java
```diff
@@ -98,12 +101,13 @@ public void matchEnded(Match match) {
               // If there is a restart queued, don't start a vote
               if (RestartManager.isQueued()) return;
               currentPoll =
-                  shouldCustomVote()
+                  shouldOverride()
                       ? new MapPoll(
                           match,
                           getCustomVoteMapWeighted(),
+                          Sets.newHashSet(),
```
Pablete1234:
```suggestion
                          Collections.emptySet(),
```

Pablete1234 (dismissed):

util/src/main/i18n/templates/map.properties
```diff
@@ -134,3 +134,13 @@ vote.limit = Selection limit for the vote has been reached
 
 vote.title.map = Selected Maps
 
+# {0} = player name (who toggled mode)
+# {1} = mode name
+vote.toggle = {0} toggled vote mode to {1}
+
+vote.mode.replace = Replace
+
+vote.mode.override = Override
```
Electroid: Replace and override basically mean the same thing, can we use different terms that make it more clear?

Electroid (changes requested):

util/src/main/i18n/templates/map.properties
applenick: @AustinLMayes suggested `create` as an alternative for override 🎉 
If you have different terms you wanted just let me know and I’ll be glad to adjust it.

Otherwise should be good to go now 👍

Electroid (approved):
Rock on

PR #555, opened 2020-07-09
Make changes needed for Ingame & fix cycling bugs

dentmaged:
Ingame is a plugin that manages tournaments and ranked matches. It automatically cycles to matches (even when no players are online hence the changes to CountdownRunner), assigns players to teams and supports customisable vetoes.

Ingame is still a WIP and will be open-sourced in the coming days.

Signed-off-by: William Jeffcock <jeffcockw@gmail.com>

### DIFF
--- a/core/src/main/java/InitiateMatchStartCountdown/InitiateMatchStartCountdown.java
+++ b/core/src/main/java/InitiateMatchStartCountdown/InitiateMatchStartCountdown.java
@@ -1,0 +1,31 @@
+package InitiateMatchStartCountdown;
+
+import java.time.Duration;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.event.MatchEvent;
+
+public class InitiateMatchStartCountdown extends MatchEvent {
+
+  private final Duration duration;
+
+  public InitiateMatchStartCountdown(Match match, Duration duration) {
+    super(match);
+    this.duration = duration;
+  }
+
+  public Duration duration() {
+    return duration;
+  }
+
+  private static final HandlerList handlers = new HandlerList();
+
+  public static HandlerList getHandlerList() {
+    return handlers;
+  }
+
+  @Override
+  public HandlerList getHandlers() {
+    return handlers;
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/command/graph/MatchPlayerProvider.java
+++ b/core/src/main/java/tc/oc/pgm/command/graph/MatchPlayerProvider.java
@@ -11,7 +11,7 @@ import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.util.text.TextException;
 
-final class MatchPlayerProvider implements BukkitProvider<MatchPlayer> {
+public final class MatchPlayerProvider implements BukkitProvider<MatchPlayer> {
 
   @Override
   public boolean isProvided() {
--- a/core/src/main/java/tc/oc/pgm/start/StartCountdown.java
+++ b/core/src/main/java/tc/oc/pgm/start/StartCountdown.java
@@ -2,12 +2,15 @@ package tc.oc.pgm.start;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
+import InitiateMatchStartCountdown.InitiateMatchStartCountdown;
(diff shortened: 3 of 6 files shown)


core/src/main/java/tc/oc/pgm/start/StartCountdown.java
```diff
@@ -49,11 +52,11 @@ protected boolean showTitle() {
   @Override
   public void onStart(Duration remaining, Duration total) {
     super.onStart(remaining, total);
+    Bukkit.getPluginManager().callEvent(new InitiateMatchStartCountdown(match, remaining));
```
KingOfSquares, line 55: You can just store the `Match` given in the constructor and do `match.callEvent()` here

core/src/main/java/InitiateMatchStartCountdown/InitiateMatchStartCountdown.java
```diff
@@ -0,0 +1,31 @@
+package InitiateMatchStartCountdown;
+
+import java.time.Duration;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.event.MatchEvent;
+
+public class InitiateMatchStartCountdown extends MatchEvent {
```
KingOfSquares, line 8: This classname should end in `event`

core/src/main/java/InitiateMatchStartCountdown/InitiateMatchStartCountdown.java
```diff
@@ -0,0 +1,31 @@
...
+
+  private final Duration duration;
+
+  public InitiateMatchStartCountdown(Match match, Duration duration) {
+    super(match);
+    this.duration = duration;
+  }
+
+  public Duration duration() {
```
KingOfSquares, line 17: This and the similar method in the other event should probably be `Duration getDuration()/getRemaining()`

core/src/main/java/tc/oc/pgm/command/graph/MatchPlayerProvider.java
```diff
@@ -11,7 +11,7 @@
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.util.text.TextException;
 
-final class MatchPlayerProvider implements BukkitProvider<MatchPlayer> {
+public final class MatchPlayerProvider implements BukkitProvider<MatchPlayer> {
```
KingOfSquares, line 14: Unsure if this is supposed to be a part of the API in PGM? What case do you have for needing this?

KingOfSquares (commented):
Both events should also probably be located at `core/src/main/java/tc/oc/pgm/countdowns/events` or some other common folder

core/src/main/java/tc/oc/pgm/command/graph/MatchPlayerProvider.java
dentmaged: For Ingame, we're using `MatchPlayerProvider` and `MatchProvider` in the command framework. These classes need to be public because our `CommandModule` is not in the `tc.oc.pgm.command.graph` package.

core/src/main/java/InitiateMatchStartCountdown/InitiateMatchStartCountdown.java
```diff
@@ -0,0 +1,31 @@
+package InitiateMatchStartCountdown;
```
EricZeiberg, line 1: Per convention, no camel case in package names. Also, not too crazy about this name, too long imo

core/src/main/java/tc/oc/pgm/command/graph/MatchPlayerProvider.java
KingOfSquares: Yeah, I still think that either these providers should all(in contrast to only some few) be made public as a part of the API, or you could make your own providers through `PGM.get()`. You can use bukkit `World`s and/or `Entity`s to get the correct `Match` and `MatchPlayer`s.

Anyone else have any opinions on this?

core/src/main/java/InitiateMatchStartCountdown/InitiateMatchStartCountdown.java
dentmaged: I had to rebase the changes and while reimplenting them, Eclipse for some reason named the package that. I’ll make changes later today.

PR #558, opened 2020-07-10
Force users to always use JDK 8

AustinLMayes:
Maven source/target compilation configuration options only make a best effort attempt to compile for the specified version and won't always compile working code if libraries from different java versions are used. This forces users to only compile against JDK 8.


Electroid (approved):
Sure

PR #561, opened 2020-07-11
Licenses for maps!

### LINKED ISSUE #496: Add license tag to XML
Allow map creators to specify a `<license>` tag for their map. Should be parsed and limited to https://creativecommons.org/choose/ for now.

KingOfSquares:
R.I.P Sherman, fixes #496 

Made under new: "OCC plays IntelliJ" show

Any other licenses to add?

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapCommand.java
@@ -209,10 +209,10 @@ public final class MapCommand {
     }
 
     audience.sendMessage(
-            TextComponent.builder()
-                    .append(mapInfoLabel("map.info.license"))
-                    .append(TextComponent.of(map.getLicense().getFullName(), TextColor.GOLD))
-                    .build());
+        TextComponent.builder()
+            .append(mapInfoLabel("map.info.license"))
+            .append(TextComponent.of(map.getLicense().getFullName(), TextColor.GOLD))
+            .build());
 
     audience.sendMessage(createTagsComponent(map.getTags()));
 
--- a/core/src/main/java/tc/oc/pgm/command/MapCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapCommand.java
@@ -208,6 +208,12 @@ public final class MapCommand {
               .build());
     }
 
+    audience.sendMessage(
+            TextComponent.builder()
+                    .append(mapInfoLabel("map.info.license"))
+                    .append(TextComponent.of(map.getLicense().getFullName(), TextColor.GOLD))
+                    .build());
+
     audience.sendMessage(createTagsComponent(map.getTags()));
 
     if (PGM.get().getMapOrder() instanceof MapPoolManager) {
--- a/core/src/main/java/tc/oc/pgm/command/MapCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapCommand.java
@@ -212,6 +212,8 @@ public final class MapCommand {
         TextComponent.builder()
             .append(mapInfoLabel("map.info.license"))
             .append(TextComponent.of(map.getLicense().getFullName(), TextColor.GOLD))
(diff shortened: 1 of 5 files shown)


KingOfSquares:
Closes #496 btw

core/src/main/java/tc/oc/pgm/map/MapLicense.java
```diff
@@ -0,0 +1,29 @@
+package tc.oc.pgm.map;
+
+public enum MapLicense {
+  CC_BY("Attribution 4.0 International"),
+  CC_BY_ND("Attribution-NoDerivatives 4.0 International"),
+  CC_BY_SA("Attribution-ShareAlike 4.0 International"),
+  CC_BY_NC("Attribution-NonCommercial 4.0 International"),
+  CC_BY_NC_ND("Attribution-NonCommercial-NoDerivatives 4.0 International"),
+  CC_BY_NC_SA("Attribution-NonCommercial-ShareAlike 4.0 International"),
+  CUSTOM(""),
```
Electroid: While I see the desire for a custom license, I'm not convinced we should allow this. This problem is similar to custom map rules: they cause confusion. I don't want map makers defining their own licenses, that's murky water. Creative Commons is used by basically everyone, and it has good legal standing.

core/src/main/java/tc/oc/pgm/map/MapLicense.java
```diff
@@ -0,0 +1,29 @@
+package tc.oc.pgm.map;
+
+public enum MapLicense {
```
Electroid, line 3: Isn't there a public domain license from Creative Commons?

Electroid: Is there anyway we can use their translations? Example: https://github.com/creativecommons/cc.i18n/blob/master/cc/i18n/po/fr_CA/cc_org.po#L639

core/src/main/java/tc/oc/pgm/command/MapCommand.java
```diff
@@ -208,6 +208,12 @@ public void map(Audience audience, CommandSender sender, @Text MapInfo map) {
               .build());
     }
 
+    audience.sendMessage(
+        TextComponent.builder()
+            .append(mapInfoLabel("map.info.license"))
+            .append(TextComponent.of(map.getLicense().getFullName(), TextColor.GOLD))
```
Electroid, line 214: Can we get a hover click to view the license? Also can we make the license more human-readable / understandable? Tags like "Commercial" or something?

Assume the user doesn't know much about CC.

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/map/MapLicense.java
KingOfSquares: Yes! Was wondering about this as well. I am going to remove it, as the implementation is kinda weird as well.

BGMP:
@Shrman007658 would be proud

core/src/main/java/tc/oc/pgm/command/MapCommand.java
KingOfSquares: Yes, great ideas!

core/src/main/java/tc/oc/pgm/map/MapLicense.java
KingOfSquares: So a way to fetch those at runtime(assuming we don't want to bundle these translations with the rest)?

core/src/main/java/tc/oc/pgm/command/MapCommand.java
KingOfSquares: I tried changing them a little, let me know if this is what you were thinking. I think keeping the words as they are in the license is logical because if you google "Attribute license" the CC license will appear.

KingOfSquares:
I can't see myself doing this In while. Will re-open if anything happenes with this project

PR #565, opened 2020-07-12
Add option to show kill count next to names on death messages

mrcookieunderscore13:
This is a feature to show a person's kills next to his/her name on death messages. It can be turned off in settings, but defaults to true.
Here is an example:

Notes: 
If the person has `/toggle stats` set to false, then when that person is involved in a death message, his/her kill count is not included in the message for everybody. 
Also, if a person dies to the void/lava/etc. it works correctly. 
Signed-off-by: mrcookie

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
+++ b/core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
@@ -18,6 +18,7 @@ public enum SettingKey {
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
   DEATH(
       Arrays.asList("death", "dms"), DEATH_ALL, DEATH_OWN), // Changes which death messages are seen
+  KILLS("kills", KILL_COUNT_ON, KILL_COUNT_OFF),
   PICKER("picker", PICKER_AUTO, PICKER_ON, PICKER_OFF), // Changes when the picker is displayed
   JOIN(Arrays.asList("join", "jms"), JOIN_ON, JOIN_OFF), // Changes if join messages are seen
   MESSAGE(
--- a/core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
+++ b/core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
@@ -21,6 +21,9 @@ public enum SettingValue {
   DEATH_OWN("death", "own"), // Only send death messages involving self
   DEATH_ALL("death", "all"), // Send all death messages, highlight your own
 
+  KILL_COUNT_ON("kills", "on"), // Show the amount of kills a player has in death messages
+  KILL_COUNT_OFF("kills", "off"), // Don't show the amount of kills a player has in death messages
+
   PICKER_AUTO("picker", "auto"), // Display after cycle, or with permissions.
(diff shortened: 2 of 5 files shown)


applenick:
Hey @mrcookieunderscore13 

Personally I’m not a fan of merging stats with death messages. It introduces an element that might not immediately be understood by new users. I understand how you allowed for people who have stats to toggle this off, but that will create inconsistencies with the formatting.  It should be either all or none, allowing choice for something that is broadcasted is really less than ideal. 

Perhaps you can look into an alternative method for players being able to determine the active kill count, that is in a less global way.

Nice work though, always great to see people adding new features 😄

qlados:
I think it would be better if you either see it for everybody or nobody (which is then toggleable).

Indicardo:
I think this would work better if the user could toggle between *seeing* all kills (doesn't matter if other people don't have the option toggled). That would make it more consistent, and instead of displaying the announcement to everybody, you could make it so that only the people that are interested in it see it.

Pablete1234:
When you toggle stats off, it means that you do not want your stats to be tracked. This is taken into account by pgm, and that means others shouldn't be able to know your stats either, they're not tracked

mrcookieunderscore13:
To address @applenick
The feature can default to false if it might be confusing to new/existing pgm players. However, this is not a new concept--Badlion had this exact same feature on their pgm-like gamemodes. And I do not see an issue adding a choice for something that is broadcasted globally, because right now there is a feature to flat out turn off death messages. This feature was really intended for stratus, because it is a lot more applicable to competitive matches. I just added it here because there seems to be a push for upstream prs (which I totally agree with, its just how should features more suited for stratus look in main PGM? My go-to is to make it toggleable--but maybe it should be left to a server config instead of letting the users decide, you tell me).

Too address @qlados and @Indicardo 
I think I was not clear enough when explaining how the system works. If someone turns `/toggle kills` to false, that does not impact anybody other than themselves--other people who have the setting to true will still see the kill count. The only time someone can impact other people's death messages is if they have `toggle stats` set to false. When that setting is on false, PGM does not track stats for that person, so it is impossible to show their kill count. (Unless we can remove that setting, I personally do not see a reason to allow players to turn their stats off).

Brottweiler:
> Badlion had this exact same feature on their pgm-like gamemodes.

Didn't they just show MMR? I think the `[1234] Username` is common in the competitive scene (in all games) to show the players MMR.

mrcookieunderscore13:
> > Badlion had this exact same feature on their pgm-like gamemodes.
> 
> Didn't they just show MMR? I think the `[1234] Username` is common in the competitive scene (in all games) to show the players MMR.

Here is an example of what I was talking about:
https://www.youtube.com/watch?v=ylsZofGUxYY

applenick:
Thanks for the reply @mrcookieunderscore13,
I understand how this would be useful for competitive servers and saw the example you provided from Badlion. 

Imo, the chat in that video looks extremely messy though and would take away from the elegant formatting PGM already has.

For the sake of keeping death messages simple, I strongly recommend an alternative to including the count in death messages be found.

 Perhaps you could modify the tab list or see how an API option could be introduced so that third party plugins can modify death messages themselves, to include additional data such as the kill count in this case.

Electroid:
Thanks so much for submitting this change! I think the issue is death messages are already verbose, as in, even as a setting it's just still hard to justify. So I'm going to close this up.

PR #576, opened 2020-07-30
Use correct materials when filtering fluids

Pablete1234:
Currently, if you wanted to prevent a cobblestone  generator, you would need to filter preventing certain blockstates of lava being created, this is annoying and buggy since those states may be used in certain flows.
This makes it so you're able to deny stone/cobble being placed and it prevents "cobblestone generator" behaviour


Electroid (approved):
Interesting

PR #577, opened 2020-07-30
Update running instructions

Brottweiler:
The current running instructions gives you very old SportPaper and PGM versions, and the `sportpaper.yml` 404's.

This PR "fixes" that but it's not optimal since it's "hardcoding" v0.8, and would become outdated when a new version is released. But it's better than how it is currently.


Electroid:
Now fixed.

PR #580, opened 2020-07-31
Allow customization of tablist footer

### LINKED ISSUE #573: Add IP and server name to tab footer
OCN showed the IP and server name in the tab list footer.

image

With a few servers branching into more of a network than a single PGM instance this could be useful.

These vales should be config options and not [hardcoded](https://github.com/StratusNetwork/projectares/pull/26#issuecomment-319222932).

applenick:
# Custom Tablist Footer

Resolves #573 

New config section:
```yaml
tablist:
  left: ""
  right: ""

```
## Screenshots
Example of footer shown with additional stats
footer-match

Example of footer shown as an observer
footer-hello-world

### DIFF
--- a/core/src/main/java/tc/oc/pgm/tablist/MatchFooterTabEntry.java
+++ b/core/src/main/java/tc/oc/pgm/tablist/MatchFooterTabEntry.java
@@ -3,10 +3,12 @@ package tc.oc.pgm.tablist;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
+import net.kyori.text.Component;
 import net.kyori.text.TextComponent;
 import net.kyori.text.TranslatableComponent;
 import net.kyori.text.format.TextColor;
 import net.md_5.bungee.api.chat.BaseComponent;
+import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchScope;
 import tc.oc.pgm.api.player.MatchPlayer;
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/tablist/MatchFooterTabEntry.java
```diff
@@ -60,13 +62,24 @@ public BaseComponent getContent(TabView view) {
...
+
     content
         .append(TranslatableComponent.of("match.info.time", TextColor.GRAY))
         .append(": ", TextColor.GRAY)
         .append(
             TimeUtils.formatDuration(match.getDuration()),
             this.match.isRunning() ? TextColor.GREEN : TextColor.GOLD);
 
+    if (rightContent != null) {
+      content.append(" - ", TextColor.DARK_GRAY).append(rightContent);
```
Pablete1234:
```suggestion
      content.append(" - ", TextColor.DARK_GRAY).append(rightContent.colorIfAbsent(TextColor.WHITE));
```

applenick: Oh wow whoops 😅

Electroid:
While customizing the footer is nice, I'm not sure having the left/right split makes much sense (other than for legacy reasons). It's kinda odd, but I'm not sure of a different solution.

Pablete1234:
I think just having right/left is the easiest way of just letting any generic text in there.
The alternative would be starting to get into what shows and what doesn't show (as in, allowing to remove things like the time) but i think that's a completely different topic, and this one is simple and useful enough

applenick:
I’d also like to point out we do allow similar customization of the sidebar 🤷‍♂️ 

I decided to go with left/right, as @Pablete1234 mentioned, it’s the easiest way of implementing this feature without getting into more complex integrations. 

It’s likely in our favor to offer some some form of this feature and perhaps additional tablist design choices going forward. I imagine that will be an appreciated feature once PGM is in the hands of more general server owners who want to brand their servers further.

Let me know if you have any suggestions on the config or alternatives. I’d be happy to adjust this however you see fit 👍

Electroid (approved):

PR #581, opened 2020-07-31
Add leak completion to cores

### LINKED ISSUE #196: Core leak distance info
I think it would be really helpful, if, instead of just "You damaged Blue's Core." if it could say "You damaged Blue's Core.  You need to leak the lava four blocks down."  Or just _something_ like that, because it's very very unclear to public players as to how leaking cores exactly works, and since it varies per map it's annoying.

Pablete1234:
The description says it all, now cores will no longer be just a touch scoring as 50% of an objective, instead, they will have a % of completion based on how far down the lava flows.
The amount required to flow is from the lowest lava block in the core, to the y level of the leak, wich is the bottom of the core region - leak distance, and the one extra block for the actual leak

### DIFF
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -8,34 +8,42 @@ import javax.annotation.Nullable;
 import net.kyori.text.Component;
 import net.kyori.text.TextComponent;
 import net.kyori.text.TranslatableComponent;
+import org.bukkit.ChatColor;
 import org.bukkit.Location;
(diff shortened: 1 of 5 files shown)


Electroid (approved):
Sweet!

Pugzy:
@Pablete1234 Do these changes need mentioning in the docs? I can see a possible new `show-progress` attribute.

Pablete1234:
@Pugzy `show-progress` is the only change doc-wise, everything else is automatic. Works the same way as destroyables, if you show progress it will display as a % instead of just touch

PR #585, opened 2020-08-02
Add option to show ping in tablist

Pablete1234:
Adds a config option to enable displaying real players' ping on the tablist, will not work on the 1.7 legacy tablist and there are no plans to ever port it there

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -79,6 +79,7 @@ public final class PGMConfig implements Config {
   // ui.*
   private final boolean showSideBar;
   private final boolean showTabList;
+  private final boolean showTabListPing;
   private final boolean showProximity;
   private final boolean showFireworks;
   private final boolean participantsSeeObservers;
@@ -163,6 +164,7 @@ public final class PGMConfig implements Config {
     this.showProximity = parseBoolean(config.getString("ui.proximity", "false"));
     this.showSideBar = parseBoolean(config.getString("ui.sidebar", "true"));
     this.showTabList = parseBoolean(config.getString("ui.tablist", "true"));
+    this.showTabListPing = parseBoolean(config.getString("ui.ping", "true"));
     this.participantsSeeObservers =
         parseBoolean(config.getString("ui.participants-see-observers", "true"));
     this.showFireworks = parseBoolean(config.getString("ui.fireworks", "true"));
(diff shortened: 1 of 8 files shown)


Electroid:
I guess.. why not always show it? If we're caching it, I don't see a reason to disable it. Just seems like a silly config option.

Pablete1234:
You may want players not to be annoying about what other players' ping are, this has happened before and i think it may be in some public servers' interest not to show it

Pugzy:
My previous comments on this feature over at https://github.com/PGMDev/PGM/pull/492

Pablete1234:
The ping is periodically updated and kept on sync, so that one isn't an issue. That is how vanilla minecraft does it too, just send a packet to update ping every so often.

Regarding consistency and visual clutter, some players actually seem to care about knowing this value, if you're lagging and you see yourself on lower than 5 bars you know ur not doing good.

Precisely because there can be some controversy arround this, i've made it a server setting that, in fact, is disabled by default. But i've herd a few memberos of the community ask for this feature and if they actually want it, and many use BLC or other mods that show actual ping values, they'd rather see this.

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -163,6 +164,7 @@
     this.showProximity = parseBoolean(config.getString("ui.proximity", "false"));
     this.showSideBar = parseBoolean(config.getString("ui.sidebar", "true"));
     this.showTabList = parseBoolean(config.getString("ui.tablist", "true"));
+    this.showTabListPing = parseBoolean(config.getString("ui.tablist-ping", "false"));
```
Electroid: I think the default should be `true`.

core/src/main/resources/config.yml
```diff
@@ -67,10 +67,11 @@ gameplay:
 
 # Toggles various user interfaces.
 ui:
-  sidebar: true    # Enable the side bar?
-  tablist: true    # Enable the tab list?
-  proximity: false # Should the proximity of objectives be visible?
-  fireworks: true  # Spawn fireworks after objectives are completed?
+  sidebar: true       # Enable the side bar?
+  tablist: true       # Enable the tab list?
+  tablist-ping: false # Should tab list show real ping?
```
Electroid: Just call this `ping`.

Electroid (changes requested):

Electroid (approved):

PR #587, opened 2020-08-02
Start reworking bossbars

Pablete1234:
Rework bossbars to make use of viaversion's API for 1.9+ clients.

In a future (separate) pr support for PGM bossbars defining colors should be added, currently they're all assuming blue and that's how it'll stay for the scope of this PR

Renamed `BossBar` to `BossBarSource` not to collide with Viaversion's `BossBar`, 1.12 PGM had that class renamed too


Electroid:
FYI, we're going to use https://github.com/KyoriPowered/adventure-platform for this (boss bars, books, etc.)

It will be ready in about a week, so I'd hold off on refactoring boss bars to use Via. Since it will support Via, Bukkit's boss bar, and even fake entity hacks.

Pablete1234:
@Electroid However, this is already done, up & running.
Until someone starts integrating adventure this is a relatively easy change of how the view is rendered, that lets us start working on having color support and multi-bossbar, wich is what the BossBarStack doesn't currently support.

Better to have this now, and later move to adventure, than to keep on waiting in the first place

Electroid (approved):
Sure, looking though the code seems fine. Should be ok to add adventure later.

PR #594, opened 2020-08-05
Update legacy flag beams with proposed changes from #550

jasoryeh:
Changes from discussion on #550 
- Config option for 1.8+ clients to see wool beams (from @Pablete1234 )
- Increases wool beam height to approximately match in-game (effective) particle height (@jasoryeh)
- Respects Flag Definition's beam setting (other)


core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
```diff
@@ -50,6 +53,13 @@ public LegacyFlagBeamMatchModule(Match match) {
     this.beams = new HashMap<>();
   }
 
+  private boolean shouldShowBeams(Flag flag) {
```
Pablete1234: Is this unused? did you mean to call this from trackFlag?

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
```diff
@@ -188,7 +204,8 @@ public void run() {
       this.flag = flag;
       this.bukkit = bukkit;
       this.segments =
-          range(0, 32)
+          range(0, 100) // ~100 blocks is the height which the particles appear to be reasonably
```
Pablete1234: Having 100 zombie entities may be a bit too much on the client, and i doubt in any reasonable scenario this will actually be a factor that limits/helps one player or another, maybe 64 because 32 does look a bit on the lower side?

core/src/main/resources/config.yml
```diff
@@ -63,7 +63,8 @@ join:
 
 # Changes various gameplay mechanics.
 gameplay:
-  refill-wool: true # Should wool in wool rooms be automatically refilled?
+  refill-wool: true              # Should wool in wool rooms be automatically refilled?
+  global-legacy-flag-beams: true # Should everyone see legacy flag beams?
```
Pablete1234: Looks a bit on the long side, maybe just `flag-beams`?
Should probably move to the `ui` section instead of `gameplay` ?
I think false as a default makes more sense, as this is supposed to be for legacy (1.7) clients

```suggestion
  flag-beams: false # Should everyone see floating wool flag beams?
```

Pablete1234 (commented):
Thanks for submitting the update

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
jasoryeh: Ah this is my bad, forgot to remove this and didn't end up using it.

jasoryeh: Resolved in new commit.

core/src/main/resources/config.yml
jasoryeh: I think flag-beams gives off more of a should I have any flag beams at all, thought I agree it is kind of long, but just `flag-beams` is not descriptive enough in my opinion.

The `ui` section seems reasonable.

I'm open to changes if anyone has a shorter key.

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
jasoryeh: I get what you mean, however at times a beam of 32 zombies can sometimes be a bit too short for maps which have underground play areas, and these beams at times stuck out not as high as the particle beams.

Pablete1234: Yeah, that's why i suggested maybe make it 64, i'm hoping that should be more than enough, 32 *can be* short at times

jasoryeh: 64 is okay, but I believe the official flag beams(particle beams( are supposed to run 48 blocks tall if I'm reading https://github.com/PGMDev/PGM/blob/7bd6dd2dcb628f7fdce00c8bd378faf6a9c5d61d/core/src/main/java/tc/oc/pgm/flag/state/Spawned.java#L62-L72 correctly, but in-game these particles exist no matter how high you go, they just decrease in amount the higher you go.

Pablete1234: Some particles aren't an exact science, this ones spawn 56 blocks above the flag, and go "roughly" 24 blocks up/down, this all has randomness to it and the particles themselves will spawn randomly and fall too.
Technically speaking, particles should spawn from 32 blocks above the flag to 80 blocks above the flag, but they fall and cover a larger area

core/src/main/java/tc/oc/pgm/api/Config.java
```diff
@@ -202,6 +202,14 @@
    */
   boolean shouldRefillWool();
 
+  /**
+   * Whether the wool flag beams created for older versions (pre-1.8) should be shown to all
+   * players.
+   *
+   * @return If the wool flag beams should be shown to players >=1.8
+   */
+  boolean shouldShowLegacyFlagBeamsGlobally();
```
Electroid: `useLegacyFlagBeams`

core/src/main/resources/config.yml
Electroid: `ui.flag-beams` is reasonable

Electroid (changes requested):

Electroid:
Can you also rebase this PR to the `dev` branch instead of `0.9`. You'll also need to include the previous changes

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
jasoryeh: Yea, the particles are weird since they spawn and fall like that. Even though they are supposed to spawn anywhere from 32 to 80 blocks above, theres a weird thing where a few particles still spawn about 130 blocks above the flag (in my testing flag spawns at y=20, particles appear to stop completely above y=150)

core/src/main/resources/config.yml
jasoryeh: Decided to just go to `ui.flag-beams`

core/src/main/java/tc/oc/pgm/flag/Flag.java
```diff
@@ -40,11 +41,7 @@
 import tc.oc.pgm.api.region.Region;
 import tc.oc.pgm.flag.event.FlagCaptureEvent;
 import tc.oc.pgm.flag.event.FlagStateChangeEvent;
-import tc.oc.pgm.flag.state.BaseState;
-import tc.oc.pgm.flag.state.Captured;
-import tc.oc.pgm.flag.state.Completed;
-import tc.oc.pgm.flag.state.Returned;
-import tc.oc.pgm.flag.state.State;
+import tc.oc.pgm.flag.state.*;
```
Electroid: Can we not juse `*`?

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -67,6 +67,7 @@
 
   // gameplay.*
   private final boolean woolRefill;
+  private final boolean globalLegacyFlagBeams;
```
Electroid: `flagBeams`

Electroid (changes requested):
2 small things, then ready to go.

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
```diff
@@ -0,0 +1,263 @@
...
+import tc.oc.pgm.flag.state.Spawned;
+import tc.oc.pgm.util.bukkit.ViaUtils;
+import tc.oc.pgm.util.inventory.ItemBuilder;
+import tc.oc.pgm.util.nms.NMSHacks;
+
+@ListenerScope(MatchScope.LOADED)
+public class LegacyFlagBeamMatchModule implements MatchModule, Listener {
+
+  private static int UPDATE_DELAY = 0;
+  private static int UPDATE_FREQUENCY = 50;
```
KingOfSquares: can both be `final` right?

core/src/main/java/tc/oc/pgm/api/Modules.java
```diff
@@ -159,6 +160,9 @@ static void registerAll() {
     register(StatsMatchModule.class, StatsMatchModule::new);
     register(MapmakerMatchModule.class, MapmakerMatchModule::new);
 
+    // Modules that depend help older player versions
```
KingOfSquares: I think there is some typo here? 🤔

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
```diff
@@ -0,0 +1,263 @@
...
+    ItemStack wool() {
+      return new ItemBuilder()
+          .material(Material.WOOL)
+          .enchant(Enchantment.DURABILITY, 1)
+          .color(flag.getDyeColor())
+          .build();
+    }
+
+    void show() {
+      if (carrier().map(carrier -> carrier.equals(bukkit)).orElse(false)) return;
```
KingOfSquares, line 233: If I am understanding this correctly the carrying player can not see their own beam?

core/src/main/resources/config.yml
```diff
@@ -63,7 +63,7 @@ join:
 
 # Changes various gameplay mechanics.
 gameplay:
-  refill-wool: true # Should wool in wool rooms be automatically refilled?
+  refill-wool: true              # Should wool in wool rooms be automatically refilled?
```
KingOfSquares: Very spacious

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
jasoryeh: Was going to play around with this a bit before, but I guess it can be.

core/src/main/java/tc/oc/pgm/api/Modules.java
jasoryeh: 🤔

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
jasoryeh: I believe if the flag can be seen by the carrier then there might be a zombie in them which can block the vision.

core/src/main/resources/config.yml
jasoryeh: Don't remember doing that but oops

core/src/main/java/tc/oc/pgm/flag/Flag.java
jasoryeh: As you wish.

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
jasoryeh: Changed to 64.

Electroid (approved):
Solid work, thanks for porting this patch and fixing it up. Really appreciate it, and I know the community does too <3

PR #595, opened 2020-08-05
Legacy fly after respawn fix

Pablete1234:
Prevents a bug where 1.7 clients can fly after respawning, if they spam the spacebar. The reason as to why this happens is still unknown, the player abilities packet to set the clients not to fly is sent, but seems to be ignored.
Never sending them player abilities as if they can fly in the first place fixes this issue.

It has the side effect of 1.7 players not being able to fly while dead on maps where while respawning you're allowed to spectate, this is however, a very small subset of maps where players on just one version will lose some functionality


Electroid:
This is awesome. Hopefully servers can now more confidently support 1.7 with Via. Thanks for looking into this <3

Electroid (approved):

Issue #250, opened 2020-01-21
Mercy rules for score module


Xerocoles:
This would basically be a thing in xml that ends games when a certain team is up by enough points.  If it's a CTF with timelimit but no flag max, then you could set it to be like `mercy rule="5"` within the flags module, and if a team is leading 5-0, or 6-1, etc, then the match ends.  Each map would pick its own number, due to some being more high-scoring than others.  This was either already implemented in OA or about to be, but it would be very useful in competitive blowouts where nobody even tries at a certain point.

RuedigerLP:
it hink it would be an attribute for the score module

Pablete1234:
Yes, this, if implemented, i think should definetly be part of the score module.

May need to consider if percentage-based makes sense, would you want it to just end the match if one team is having, for example, 3 times as many points as the other? 
This obviously wouldn't make sense for the match start cause 0 to 3 could be recoverable but 10 to 30 is alot harder, if percentage-based was added, it probably would need to be in combination with some minimum, but it may not be worth the trouble.

Xerocoles:
> Yes, this, if implemented, i think should definetly be part of the score module.
> 
> May need to consider if percentage-based makes sense, would you want it to just end the match if one team is having, for example, 3 times as many points as the other?
> This obviously wouldn't make sense for the match start cause 0 to 3 could be recoverable but 10 to 30 is alot harder, if percentage-based was added, it probably would need to be in combination with some minimum, but it may not be worth the trouble.

yeah sorry score module

No i dont think we want percentages, I fail to see why that is better in any way, considering each map can customize the number.  Also, when playing a match, it should be easy to tell how many to need to be up by, rather than doing some math like wait ok we need 2 more, unless they get one, and then we need 4, and - ya know, just, over-complicated for the sake of what?  I get that it could help sort out if a certain match is playing really high or low scoring, but most competitive matches per map won't vary too much, so i think it's too confusing in-game to be worth it.  Does that make sense?

Xerocoles:
@Pugzy sup mate wanna do this? 💞

Pugzy:
Hacked this together on https://github.com/Pugzy/PGM/tree/mercy-score-rules which seems to work. 

Went for the option to set a `mercy` value in the XML which also optionally respects a predefined `limit`. If a team leads by the `mercy` amount they will win, otherwise its first to the score `limit`. As an added bonus this figure is set within the score limit so it means the scoreboard is updated too!

[...]

PR #601, opened 2020-08-08
Add mercy option to score module

Pugzy:
Allow a `mercy` option within the score module. This can be used alongside a `limit` or standalone. As referenced in #250. 

### Example

This XML example will result in a match that progresses like the scoreboard screenshots. 
The winner is the first team to get a mercy lead of 5 or reach the limit of 10.

```xml
<score>
    <mercy>5</mercy>
    <limit>10</limit>
</score>
```

image

*Blue team take a storming lead with just one flag away from the win, but wait, red team pulls it back and forces the score limit. 
What an epic gamer moment, so glad I got to watch this historic match on Nartica ™️ .*

### Notes

If a `limit` is not set then a team will need to lead by the set mercy amount or fallback on another condition such as time.

The example above uses CTF but as it is the `ScoreModule` it would work with TDMs and Hills.

This just hooks into the scoring module so no changes to the scoreboard etc. I'd imagine this feature only being used in competitive environments where teams are aware of the score limit and mercy rules (from scrimming), otherwise, the scoreboard limit changing "randomly" may be a bit confusing. Maybe some UI work could be done to make this clearer.

### Code

I have toyed around with the idea of including the feature directly in the score module but decided to separate it out into its own `MercyRule` class so I can have a few helper methods to make it a bit more readable. The topic of micro optimisation was mentioned in the linked issue, the current implementation only re-sorts all team scores if the lead or trail have their score decreased.

Wanted to get the PR made for this for feedback rather than having a useless finished branch that I had not touched in 3 months. 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/score/MercyRule.java
+++ b/core/src/main/java/tc/oc/pgm/score/MercyRule.java
@@ -1,0 +1,103 @@
+package tc.oc.pgm.score;
+
+import java.util.AbstractMap;
+import java.util.Map;
+import tc.oc.pgm.api.party.Competitor;
+import tc.oc.pgm.api.party.event.CompetitorScoreChangeEvent;
+
+public class MercyRule {
+
+  private final ScoreMatchModule scoreMatchModule;
+  private final int scoreLimit;
+  private final int mercyLimit;
+
+  private Map.Entry<Competitor, Double> leader;
+  private Map.Entry<Competitor, Double> trailer;
+
+  public MercyRule(ScoreMatchModule scoreMatchModule, int scoreLimit, int mercyLimit) {
+    this.scoreMatchModule = scoreMatchModule;
+    this.scoreLimit = scoreLimit;
+    this.mercyLimit = mercyLimit;
+
+    calculateLeaders();
+  }
+
+  private double getLeaderScore() {
+    return leader.getValue();
+  }
+
+  private double getTrailerScore() {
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/score/MercyRule.java
```diff
@@ -0,0 +1,100 @@
...
+import tc.oc.pgm.api.party.event.CompetitorScoreChangeEvent;
+
+public class MercyRule {
+
+  private final ScoreMatchModule scoreMatchModule;
+  private final int scoreLimit;
+  private final int mercyLimit;
+
+  private AbstractMap.SimpleEntry<Competitor, Double> leader;
+  private AbstractMap.SimpleEntry<Competitor, Double> trailer;
```
Pablete1234:
```suggestion
  private Map.Entry<Competitor, Double> leader;
  private Map.Entry<Competitor, Double> trailer;
```
Just define the generric types here

Pablete1234 (commented):
One small change, but lgtm

Pugzy:
- Updated to allow for negative scoring which will also affect the required score.
- Removed reliance on `CompetitorScoreChangeEvent` and call mercy rules directly from the `incrementScore` method.
- Populate the score module with a default score of 0 for all teams (previously didn't exist in map until points scored).

Pablete1234 (approved):

Electroid (approved):
👍

PR #605, opened 2020-08-10
Make minimal changes needed for Ingame to work in tournaments

dentmaged:
Adds three new events (to support /ready and /unready);
- `CountdownStartEvent`
- `CountdownEndEvent`
- `CountdownCancelEvent`

Changes to `CountdownRunner` are not needed for tournaments.


Pugzy:
Could the countdown not be passed to the event rather than the remaining duration?

This way in-game can get the remaining time from it and also makes it more generic for other plugins to use differently.

Not taken a deep dive but could the events also not be generic for all countdowns and not just start ones?

Pablete1234 (approved):

Electroid (approved):
Looking forward to TM integrations!

PR #606, opened 2020-08-11
Prevent team griefing on scored matches

Pablete1234:
A new config option for under how many (negative) points, a player should be kicked out of the match and not allowed to rejoin. This prevents team griefing in TDM dying to the enemy team doesn't substract score so that won't affect this, only purposely suiciding will end up triggering it

image

### DIFF
--- a/util/src/main/i18n/templates/join.properties
+++ b/util/src/main/i18n/templates/join.properties
@@ -35,6 +35,8 @@ join.err.noPermission = You are not allowed to join this match
 
 join.err.alreadyJoined = You have already joined the match
 
+join.err.teamGrief = You made a negative impact for your team and can no longer play this match
+
 # {0} = team name (e.g. "Observers")
 join.err.alreadyJoined.team = You have already joined {0}
 
--- a/util/src/main/i18n/templates/join.properties
+++ b/util/src/main/i18n/templates/join.properties
@@ -35,7 +35,7 @@ join.err.noPermission = You are not allowed to join this match
 
 join.err.alreadyJoined = You have already joined the match
 
-join.err.teamGrief = You made a negative impact for your team and can no longer play this match
+join.err.teamGrief = You caused too much negative impact to play this match
 
 # {0} = team name (e.g. "Observers")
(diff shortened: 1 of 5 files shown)


Indicardo:
I think the default -10 "grief score" is too much, as for a group of 4 that would result in -40 points

Pablete1234:
> I think the default -10 "grief score" is too much, as for a group of 4 that would result in -40 points

While i agree it may be a bit too much, i'd rather keep the default on a number that is pretty much guaranteed not to be an issue that will affect anyone, can be tweaked/lowered for different scenarios, like public matches

KingOfSquares:
This feels like a moderation thing and not a PGM feature. Maybe put it in community?

Pablete1234:
> This feels like a moderation thing and not a PGM feature. Maybe put it in community?

I think this is as much moderation as preventing breaking your own core is, or tnt defusing, it doesn't issue any punishment it just prevents you from keeping on doing it

KingOfSquares:
Then maybe a rephrasing, `You made a negative impact` sound like the plugin judging you for being bad :joy:

Maybe something more `[...]because of intentional misfit[...]` or something giving the vibe of intentional bad actions

Pablete1234:
> Then maybe a rephrasing, `You made a negative impact` sound like the plugin judging you for being bad 
> 
> Maybe something more `[...]because of intentional misfit[...]` or something giving the vibe of intentional bad actions

Well the point is the plugin isn't 100% sure, so it's just a bit more of an open "you weren't helping your team with how you played", unsure if intentional or not

util/src/main/i18n/templates/join.properties
```diff
@@ -35,6 +35,8 @@ join.err.noPermission = You are not allowed to join this match
 
 join.err.alreadyJoined = You have already joined the match
 
+join.err.teamGrief = You made a negative impact for your team and can no longer play this match
```
Electroid: You caused too much negative impact to play this match

Electroid (changes requested):

PR #608, opened 2020-08-13
Add group hover info to chat

### LINKED ISSUE #603: Allow player prefix and suffix to be given hover text info
### Issue

Servers using custom flairs for ranks and roles make it confusing for new players to understand what each of them are.

Once you know an orange `❖` is the admin staff rank things become easy enough to understand, but providing the players with this info is difficult. Currently, OCC has `/ranks` which helps with this but currently needs to be maintained alongside ranks so they match and players need to know of or be instructed to use the command.

Some servers opt for using the ranks full name as the prefix `[Admin]` but this adds text chat clutter which flairs are designed to remove (especially with longer ranks e.g. `[EventCoordinator]`) and again once you know you know so no need for the constant reminder messing up chat.

### Solution

It would be cool if you could add hover over text to the prefix/suffix to see what they mean.

Perhaps a name, description, and optional link (openable by clicking on the prefix).

image

These already exist on servers like Hypixel where you can hover over a players `[MVP] Pugzy` to show network level, achievement points, and guild so this is nothing new for players just a "hover to get more info" and exists in PGM on player names for teleporting.

### LINKED ISSUE #615: Incompatibility of using Bold Prefixes (Probably suffixes too)
If you use "&l" in a prefix, the prefix will work in the chat however it will not work and format the prefix white (ignoring any other chat formatting) in the tablist, administrative commands & end of match stats.

Looking at it, the problem seems to be with the PlayerComponent Class.

Image below shows what works and what doesn't.
image

applenick:
# Add group hover info to chat

This PR resolves #603 and resolves #615 

A brief overview, now when a user hovers over a flair (if defined) information explaining what the rank is will be shown.

To add this feature I did a few things. First created a new `Flair` class in the config to hold info related to prefix/suffix including the attributes for the hover message. Then the main change was made in `ChatDispatcher`, where now instead of sending a formatted string, a component message is created and sent to players. Allowing for the hover/click events to work as intended. 

## Config
New configurations values added:
```yaml
groups:
  # A special group for server operators.
  op:
    prefix: "&6❖"
    suffix: ""
    # Display name of the group - Displayed when hovering over prefix/suffix              
    display-name: "&6Server Operator"
    # Description of group - Displayed when hovering over prefix/suffix
    description: "&7Administrator of the server"
    # URL shown below description when hovering over prefix/suffix
    click-link: "&bhttps://pgm.dev/"

    # If you want to further customize prefix/suffix flairs, provide components here to override format
    # Only use if you know what you're doing
    prefix-component: ""
    suffix-component: “"
```

## Screenshots
Example of defining just a `display-name`
hello-1
Example of defining a `display-name` and `description`
hello-2
Example using all hover attributes including `click-link`
hello-3

## Suggestions
I applied the feedback given in #603 regarding the config layout, if there are any other suggestions just let me know here and I’ll be glad to adjust this. 

As always this PR was tested and all features work as intended 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationProvider.java
+++ b/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationProvider.java
@@ -2,6 +2,7 @@ package tc.oc.pgm.namedecorations;
 
 import java.util.UUID;
 import net.kyori.text.Component;
+import net.kyori.text.TextComponent;
 
 /** This interface is intended to return what prefix and suffix a player should have */
 public interface NameDecorationProvider {
@@ -10,7 +11,11 @@ public interface NameDecorationProvider {
 
   String getSuffix(UUID uuid);
 
-  Component getPrefixComponent(UUID uuid);
+  default Component getPrefixComponent(UUID uuid) {
+    return TextComponent.of(getPrefix(uuid));
+  }
 
-  Component getSuffixComponent(UUID uuid);
+  default Component getSuffixComponent(UUID uuid) {
+    return TextComponent.of(getSuffix(uuid));
+  }
 }
--- a/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationProvider.java
+++ b/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationProvider.java
@@ -1,6 +1,7 @@
 package tc.oc.pgm.namedecorations;
 
 import java.util.UUID;
+import net.kyori.text.Component;
 
 /** This interface is intended to return what prefix and suffix a player should have */
 public interface NameDecorationProvider {
@@ -8,4 +9,8 @@ public interface NameDecorationProvider {
   String getPrefix(UUID uuid);
 
   String getSuffix(UUID uuid);
+
+  Component getPrefixComponent(UUID uuid);
+
+  Component getSuffixComponent(UUID uuid);
 }
--- a/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistry.java
+++ b/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistry.java
@@ -1,6 +1,7 @@
 package tc.oc.pgm.namedecorations;
 
 import net.kyori.text.Component;
+import net.kyori.text.TextComponent;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Listener;
 import tc.oc.pgm.api.party.Party;
@@ -25,7 +26,9 @@ public interface NameDecorationRegistry extends Listener {
    * @param party The party this player is currently in
    * @return The name, decorated, in component form
    */
-  Component getDecoratedNameComponent(Player player, Party party);
+  default Component getDecoratedNameComponent(Player player, Party party) {
+    return TextComponent.of(getDecoratedName(player, party));
+  }
 
   /**
    * Set what name decoration provider this registry should use
--- a/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistry.java
+++ b/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistry.java
@@ -1,5 +1,6 @@
 package tc.oc.pgm.namedecorations;
 
+import net.kyori.text.Component;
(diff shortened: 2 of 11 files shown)


core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationProvider.java
```diff
@@ -1,11 +1,16 @@
...
 /** This interface is intended to return what prefix and suffix a player should have */
 public interface NameDecorationProvider {
 
   String getPrefix(UUID uuid);
 
   String getSuffix(UUID uuid);
+
+  Component getPrefixComponent(UUID uuid);
+
+  Component getSuffixComponent(UUID uuid);
```
Pablete1234:
```suggestion
  default Component getPrefixComponent(UUID uuid) {
      return TextComponent.of(getPrefix(uuid));
  }

  default Component getSuffixComponent(UUID uuid) {
      return TextComponent.of(getSuffix(uuid));
  }
```
Keeping this interface simple to implement for 3rd parties is relevant IMO, and it's an easy enough default. (Maybe do a component parsing from legacy text instead of just a plain text with the text)

core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistry.java
```diff
@@ -15,6 +16,17 @@
...
+  /**
+   * Get the fully decorated name as a Component
+   *
+   * <p>Note: Allows for prefix/suffix hover events
+   *
+   * @param player The player to decorate
+   * @param party The party this player is currently in
+   * @return The name, decorated, in component form
+   */
+  Component getDecoratedNameComponent(Player player, Party party);
```
Pablete1234: I'd do the same as above, provide a fallback default, even if the registry isn't really something that should be reimplemented by 3rd parties often

Electroid:
I'd rather not invent a new format (for chat messaging). Can we use `GsonComponentSerializer` and allow users to specify any `Component` for the flair? That way they can literally customize it in infinite ways?

`flair: {text: "...", hover: { ... }}` You get the idea.

applenick:
So just had a few mins to test some things out.

I’m not sure if (strictly from a user-friendly point of view) we should switch this to using raw components in the config. 

Here’s the output I got from using `GsonComponentSerializer#serialize` on the sample format I included in my screenshots...
```json
{"text":"","extra":[{"text":"§6❖"}],"clickEvent":{"action":"open_url","value":"https://oc.tc/staff"},"hoverEvent":{"action":"show_text","value":{"text":"","extra":[{"text":"§6§lAdmin"},{"text":"\n"},{"text":"This is a cool hover"},{"text":"\n"},{"translate":"chat.clickLink","with":[{"text":"https://oc.tc/staff","color":"aqua","underlined":true}],"color":"dark_aqua"}]}}}
```
I can already foresee us receiving errors from users as to why their flair is not rendering properly due to them not understanding the formatting intuitively.

The three fields I outlined (display-name, description, link), IMO, should be enough customization for a feature as simple as this. However if you do wish I continue looking into using Components instead, do you have any recommendations on how the format could be simplified so the general user can easily define the different parts of a flair?

Pablete1234:
I belive we should aim for the user-friendly option, and maybe add an option for power-users who wish to define their very specific format, but that should never be the way we expect regular end users to configure this.

Keep the 3 fields, make those generate the component, and, alternatively, allow configuring a prefix-component and a suffix-component wich is this tellraw format

applenick:
Just an update, I’ve made the changes @Pablete1234 suggested. 
To reinforce I really think we should stick with a user-friendly format for this feature, but the idea to allow additional customization via an override was a great suggestion. I’ve added that and should work as intended 👍 

If there are any other recommendations just let me know and I’d be glad to adjust it.

applenick:
Another update:
This PR now resolves #615 

Added some detection for text decorations and handled them. There’s probably a cleaner way to handle this, but I’m stumped. Any recommendations on how to improve the handling of ChatColor -> TextColor/Decorations, would be appreciated. 

Updated Screenshot to showcase multiple formatting code being used
fixed-decorations
Admin broadcast message & tablist matched the format shown here too

Electroid (approved):

PR #611, opened 2020-08-14
Implement overtime

### LINKED ISSUE #529: /timelimit doesn't automatically set a victory condition
I remember this being a thing on OCN, but basically when you set a timelimit it should take into account the existing win condition (e.g. players remaining on a blitz map)

### LINKED ISSUE #547: Overtime
Introduce a new (optional) child element of the score module called overtime:
```
<score>
    <overtime/>
</score>
```
If the time runs out while the score is tied, overtime activates. It is announced in the chat and also displayed in the bossbar. Overtime ends whenever the tie is broken (one team gets a higher score than the other team), ending the match regularly with the winning condition specified in the time module.

Pablete1234:
Addresses https://github.com/PGMDev/PGM/issues/473#issuecomment-626084693 , and fixes #547 & fixes #529 

Maps can define a overtime as well as a max-overtime on their timelimit, it can also be defined via the `/tl` command, once the time limit runs out, if no team is winning (there's a tie) overtime starts.

As soon as one team breaks the tie time starts ticking down, if a tie or a comeback (other team starts winning) the time refreshes back to start ticking down from full overtime.

Once max overtime has passed, time will no longer keep refreshing, this avoids infinite matches.

Has been tested and works correctly, there is *one* known bug, if you're approaching the max overtime limit, time may not refresh if it had started ticking down already (even if it should refresh, but to less than the overtime value).
Eg: max-overtime 15m, overtime 1m, at 13:45 overtime starts ticking down, at 14:15, overtime is at 30 seconds left, and a tie occurs again or one team makes a comeback, overtime will not refresh, but it should have refreshed to 0:45.

Some refactoring of countdowns was required or otherwise this "continuously-refreshing-countdown" would cause the bossbar/countdown (including wither packets) to keep on being re-created

### DIFF
--- a/util/src/main/i18n/templates/match.properties
+++ b/util/src/main/i18n/templates/match.properties
@@ -111,8 +111,8 @@ broadcast.matchStart = The match has started!
 
 broadcast.startCancel = Match start cancelled
 
-broadcast.overtime = The match was tied, overtime has started, once the tie is broken time will start ticking
-broadcast.overtime.limit = If the match is still tied after {0} it will result in a tie.
+broadcast.overtime = Overtime! The timer will start after the tie is broken
+broadcast.overtime.limit = If there is no winner after {0}, the match will be tied.
 
 broadcast.gameOver = Game over!
 
--- a/util/src/main/i18n/templates/match.properties
+++ b/util/src/main/i18n/templates/match.properties
@@ -111,6 +111,9 @@ broadcast.matchStart = The match has started!
 
 broadcast.startCancel = Match start cancelled
 
+broadcast.overtime = The match was tied, overtime has started, once the tie is broken time will start ticking
+broadcast.overtime.limit = If the match is still tied after {0} it will result in a tie.
+
 broadcast.gameOver = Game over!
 
 broadcast.gameOver.teamWon = Your team won!
--- a/util/src/main/i18n/templates/misc.properties
+++ b/util/src/main/i18n/templates/misc.properties
@@ -54,6 +54,8 @@ misc.timeAgo = {0} ago
 # {0} = duration (e.g. "1:00" for 1 minute left, "15:25" for 15 minutes and 25 seconds left)
 misc.timeRemaining = Time Remaining: {0}
 
+misc.overtime = Overtime {0}
+
 # {0} = player name
 misc.createdBy = Created by {0}
 
--- a/core/src/main/java/tc/oc/pgm/api/match/Match.java
+++ b/core/src/main/java/tc/oc/pgm/api/match/Match.java
@@ -329,6 +329,13 @@ public interface Match extends MatchPlayerResolver, MultiAudience, ModuleContext
   Collection<Competitor> getCompetitors();
 
   /**
+   * Get all the currently winning {@link Competitor}s in the {@link Match}.
+   *
+   * @return All the winning {@link Competitor}s.
+   */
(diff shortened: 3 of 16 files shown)


util/src/main/i18n/templates/misc.properties
```diff
@@ -54,6 +54,8 @@ misc.timeAgo = {0} ago
 # {0} = duration (e.g. "1:00" for 1 minute left, "15:25" for 15 minutes and 25 seconds left)
 misc.timeRemaining = Time Remaining: {0}
 
+misc.overtime = Overtime {0}
```
Electroid, line 57: `Overtime: {0}`

util/src/main/i18n/templates/match.properties
```diff
@@ -111,6 +111,9 @@ broadcast.matchStart = The match has started!
 
 broadcast.startCancel = Match start cancelled
 
+broadcast.overtime = The match was tied, overtime has started, once the tie is broken time will start ticking
```
Electroid: Overtime! The timer will start after the tie is broken

util/src/main/i18n/templates/match.properties
```diff
@@ -111,6 +111,9 @@ broadcast.matchStart = The match has started!
...
+broadcast.overtime.limit = If the match is still tied after {0} it will result in a tie.
```
Electroid: If there is no winner after {0}, the match will be tied.

Electroid (changes requested):

util/src/main/i18n/templates/misc.properties
Pablete1234: We tested this format and it did not look the best, due to overtime being in bold, and that causing the `:` to also be bolded

image
image

Electroid (approved):
Overtime!

harvanchik:
Add an option for sudden death overtime (aka next to score wins).  This is specifically ideal for TDM or FAA games.
Sudden death is nice so that the game is not extended further than it needs to go.

I also noticed that the overtime attributes were not added to the pgm.dev xml docs.

Pablete1234:
@harvanchik Set overtime to 1 second, that's sudden death for you

harvanchik:
@Pablete1234 So I tried setting overtime to 1s; however, it didn't work as I imagined.  I have a FFA map.  The first player to score 4 points wins.  It's a five minute game, but I want to make it so that if the score is tied at three or below (between any amount of players), it becomes sudden death.  Here's my time element.
`<time result="score" overtime="1s" max-overtime="5m">5m</time>`
The issue is that once overtime begins, if a player dies, the game just ends, even though the result is set to score.  Dying in the match doesn't add or remove any points.
Is this a bug or am I misusing the attributes?

Pablete1234:
"if the score is tied at three or below (between any amount of players)"
Overtime only takes into account the leading competitor(s). If the one in #1 position and the one in #2 position are tied, overtime kicks in. The moment the ranked set returns a single winner, it ticks down

harvanchik:
@Pablete1234 That doesn't exactly answer my question.  I am saying that in overtime, the game is ending if a leading competitor dies.  This should not happen, because the winner is based on the score, and deaths do not contribute to the score.

Pablete1234:
@harvanchik Consider it a bug, open an issue with it and whatever data you can provide

PR #616, opened 2020-08-17
Fix mercy rules when leader scores 0 points

Pablete1234:
Fixes bugs with mercy rules, one that causes mercy rules to overlap leader and trailer if the leader got 0 points, completely breaking mercy rules. 

The other is on cycle sometimes negative infinity was displayed as score limit

### DIFF
--- a/core/src/main/java/tc/oc/pgm/score/MercyRule.java
+++ b/core/src/main/java/tc/oc/pgm/score/MercyRule.java
@@ -68,7 +68,9 @@ public class MercyRule {
 
     // Score larger than trailing score
     if (event.getNewScore() > getTrailerScore()) {
-      setTrailer(event.getCompetitor(), event.getNewScore());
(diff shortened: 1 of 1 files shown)


Pugzy:
Should the score event even be called when the change value is literally 0? It gets called for deaths even when deaths do not modify scores (no value set in XML).

Pablete1234:
> Should the score event even be called when the change value is literally 0? It gets called for deaths even when deaths do not modify scores (no value set in XML).

Fixing mercy rurles by just assuming no score change will ever be 0 seems like a naive approach. Mercy rules independently from the rest of the code, should be able to work no matter what event you feed it it.

If no events with 0 score change should be handled would be scope for a differernt issue and up for debate, if you enter a scorebox with 0 points maybe that should still update something? unsure. I'd need to look into it further to make sure nothing breaks from that

PR #621, opened 2020-08-19
Add names next to scores in FFA in /match and sort by score.

### LINKED ISSUE #619: Re-format FFA scores in /match
Currently FFA scores under `/match` are formatted in a not so useful list:
scores

Would be nice to update the format to include a mix of usernames and scores. Ideally all participants should be displayed similar to `/list`, while sorting from highest to lowest score. 

An alternative simple fix would be to simply add a hover message with the player name over the scores. Though may not be that intuitive/useful for those who want a quick glance.

mrcookieunderscore13:
fixes https://github.com/PGMDev/PGM/issues/619

Typing `/match` on FFA now looks like this:
Screen Shot 2020-08-19 at 3 05 41 PM

Signed-off-by: mrcookie

### DIFF
--- a/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
@@ -122,15 +122,19 @@ public class ScoreMatchModule implements MatchModule, Listener {
       }
     }
     TextComponent returnMessage =
-        TextComponent.of("Score: ", TextColor.DARK_AQUA)
-            .append(TextFormatter.list(scoreMessages, TextColor.GRAY));
+        TextComponent.builder()
+            .append(TranslatableComponent.of("match.info.score").color(TextColor.DARK_AQUA))
+            .append(TextComponent.of(": ", TextColor.DARK_AQUA))
+            .append(TextFormatter.list(scoreMessages, TextColor.GRAY))
+            .build();
     if (matchPlayer != null && ffamm != null) {
       returnMessage =
           returnMessage.append(
               TextComponent.builder()
                   .color(TextColor.GRAY)
                   .append(" | ")
-                  .append("You: ")
+                  .append(TranslatableComponent.of("match.info.you"))
+                  .append(": ")
                   .color(TextColor.DARK_AQUA)
                   .append(
                       TextComponent.of(
--- a/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
@@ -132,10 +132,10 @@ public class ScoreMatchModule implements MatchModule, Listener {
           returnMessage.append(
               TextComponent.builder()
                   .color(TextColor.GRAY)
-                  .append(" | ")
+                  .append(" | ", TextColor.GRAY)
                   .append(TranslatableComponent.of("match.info.you"))
                   .append(": ")
-                  .color(TextColor.DARK_AQUA)
+                  .color(TextFormatter.convert(matchPlayer.getCompetitor().getColor()))
                   .append(
                       TextComponent.of(
                           (int) scores.get(matchPlayer).doubleValue(), TextColor.WHITE))
--- a/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
@@ -2,19 +2,20 @@ package tc.oc.pgm.score;
 
 import static com.google.common.base.Preconditions.checkState;
 
-import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
 import java.time.Instant;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
+import net.kyori.text.Component;
 import net.kyori.text.TextComponent;
 import net.kyori.text.TranslatableComponent;
 import net.kyori.text.format.TextColor;
-import org.bukkit.ChatColor;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
@@ -34,10 +35,12 @@ import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.api.player.event.MatchPlayerDeathEvent;
 import tc.oc.pgm.events.ListenerScope;
 import tc.oc.pgm.events.PlayerParticipationStartEvent;
+import tc.oc.pgm.ffa.FreeForAllMatchModule;
 import tc.oc.pgm.util.chat.Sound;
 import tc.oc.pgm.util.collection.DefaultMapAdapter;
 import tc.oc.pgm.util.material.matcher.SingleMaterialMatcher;
 import tc.oc.pgm.util.named.NameStyle;
+import tc.oc.pgm.util.text.TextFormatter;
 
 @ListenerScope(MatchScope.RUNNING)
 public class ScoreMatchModule implements MatchModule, Listener {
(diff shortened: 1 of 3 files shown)


Pablete1234 (changes requested):
If you're using text components, don't concatenate TextColors/ChatColors as strings, there are a couple places where you're doing stuff like `ChatColor.DARK_AQUA + "Score: "`, that should go away.

Also make use of a `TextComponent.builder`, it will simplify the code alot

applenick:
Nice job!  I agree with what @Pablete1234 suggested

Also regarding the formatting, would be better to perhaps differentiate the player from the score by color. Perhaps making the score white. Also making the `:` gray would improve the look too.

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +86,47 @@ public double getScore(Competitor competitor) {
...
-    return ChatColor.DARK_AQUA + "Score: " + Joiner.on(" ").join(scores);
+    return TextComponent.of(ChatColor.DARK_AQUA + "Score: ")
+        .append(TextFormatter.list(scoreMessages, TextColor.GRAY));
   }
 
   /** Gets the status message for the match. */
-  public String getStatusMessage() {
-    StringBuilder message = new StringBuilder(this.getScoreMessage());
+  public TextComponent getStatusMessage() {
+    TextComponent message = this.getScoreMessage();
```
KingOfSquares:
```suggestion
  public Component getStatusMessage() {
    Component message = this.getScoreMessage();
```

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +86,47 @@ public double getScore(Competitor competitor) {
...
+              x ->
+                  scoreMessages.add(
+                      (TextComponent)
+                          x.getKey()
+                              .getName(NameStyle.VERBOSE)
+                              .append(TextComponent.of(x.getKey().getColor().toString() + ": "))
+                              .append(
+                                  TextComponent.of(
+                                      x.getKey().getColor().toString()
+                                          + ((int) (double) x.getValue())))));
```
KingOfSquares: `(int) x.getValue().doubleValue()`

same for the `scorePair.getValue` call

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +86,47 @@ public double getScore(Competitor competitor) {
   }
 
   /** Gets the score message for the match. */
-  public String getScoreMessage() {
-    List<String> scores = Lists.newArrayList();
-    for (Entry<Competitor, Double> scorePair : this.scores.entrySet()) {
-      scores.add(scorePair.getKey().getColor().toString() + ((int) (double) scorePair.getValue()));
+  public TextComponent getScoreMessage() {
+    List<TextComponent> scoreMessages = Lists.newArrayList();
+    FreeForAllMatchModule ffamm = match.getModule(FreeForAllMatchModule.class);
```
KingOfSquares:
```suggestion
  public Component getScoreMessage() {
    final List<Component> scoreMessages = Lists.newArrayList();
    final FreeForAllMatchModule ffamm = match.getModule(FreeForAllMatchModule.class);
```

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -2,14 +2,10 @@
 
 import static com.google.common.base.Preconditions.checkState;
 
-import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
 import java.time.Instant;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
KingOfSquares: Remove this wildcard import

Electroid:
Maybe we should only show the top 10 scores? For a 50 player FFA, `/match` will become unreadable

Pablete1234:
> Maybe we should only show the top 10 scores? For a 50 player FFA, `/match` will become unreadable

Possibly show top 9 + yours, so you can always check your own scores

mrcookieunderscore13:
I limited the list to 10 entires, fixed the concatenation issues, and changed the colors around as to AppleNick's comment. For pablete's suggestion to add your own score, I think the best way is to add another line to the `/match` command that shows your score--rather than sticking it to the end of the list. Maybe `You: <score>`. Thoughts?

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +90,50 @@ public double getScore(Competitor competitor) {
...
+                      TextComponent.builder()
+                          .append(x.getKey().getName(NameStyle.VERBOSE))
+                          .append(
+                              TextComponent.builder().append(": ").color(TextColor.GRAY).build())
+                          .append(TextComponent.of((int) x.getValue().doubleValue()))
+                          .color(TextColor.WHITE)
+                          .build()));
+      while (scoreMessages.size() > 10) {
+        scoreMessages.remove(scoreMessages.size() - 1);
+      }
```
Pablete1234: use a `.limit(10)` on the stream instead

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +90,50 @@ public double getScore(Competitor competitor) {
...
+          .forEachOrdered(
+              x ->
+                  scoreMessages.add(
+                      TextComponent.builder()
+                          .append(x.getKey().getName(NameStyle.VERBOSE))
+                          .append(
+                              TextComponent.builder().append(": ").color(TextColor.GRAY).build())
+                          .append(TextComponent.of((int) x.getValue().doubleValue()))
+                          .color(TextColor.WHITE)
+                          .build()));
```
Pablete1234: Use a proper reducer instead of forEachOrdered

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +90,50 @@ public double getScore(Competitor competitor) {
...
+    Component message = this.getScoreMessage();
     if (this.config.scoreLimit > 0) {
       message
-          .append("  ")
-          .append(ChatColor.GRAY)
-          .append("[")
-          .append(this.config.scoreLimit)
-          .append("]");
+          .append(TextComponent.of("  "))
+          .append(TextComponent.of(TextColor.GRAY + "["))
```
Pablete1234: Don't use chat colors as strings, use them on the components instead

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
KingOfSquares: When working with `TextComponent.of` you can use the color in the primary constructor. 
What I mean is instead of 
`TextComponent.of("Text").color(TextColor.WHITE);`
you do 
`TextComponent.of("Text", TextColor.WHITE);`

Also, on line 105, why are you using a builder instead of just a simple `TextComponent.of()`?

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +90,50 @@ public double getScore(Competitor competitor) {
...
+                          .build()));
+      while (scoreMessages.size() > 10) {
+        scoreMessages.remove(scoreMessages.size() - 1);
+      }
+    } else {
+
+      for (Entry<Competitor, Double> scorePair : this.scores.entrySet()) {
+        scoreMessages.add(
+            TextComponent.of(
+                scorePair.getKey().getColor().toString()
```
KingOfSquares: Use `TextFormatter#convert` on the color to use it as a TextColor instead.

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -88,26 +90,50 @@ public double getScore(Competitor competitor) {
...
       message
-          .append("  ")
-          .append(ChatColor.GRAY)
-          .append("[")
-          .append(this.config.scoreLimit)
-          .append("]");
+          .append(TextComponent.of("  "))
+          .append(TextComponent.of(TextColor.GRAY + "["))
+          .append(TextComponent.of(this.config.scoreLimit))
+          .append(TextComponent.of("]"));
```
KingOfSquares: Any reason why you are splitting up these symbols into different `append`s? Seems like they could be a single one(since you are not doing any newlines)

mrcookieunderscore13:
This is what the message looks like now:
Screen Shot 2020-08-23 at 2 03 54 PM

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -92,26 +95,60 @@ public double getScore(Competitor competitor) {
...
+              .collect(Collectors.toList());
+    } else {
+
+      for (Entry<Competitor, Double> scorePair : this.scores.entrySet()) {
+        scoreMessages.add(
+            TextComponent.of(
+                ((int) scorePair.getValue().doubleValue()),
+                TextFormatter.convert(scorePair.getKey().getColor())));
+      }
     }
```
Pablete1234, line 123: You can simplify this using a reducer instead of a collector:
```java
    TextComponent result =
        this.scores.entrySet().stream()
            .sorted(Map.Entry.comparingByValue(Comparator.reverseOrder()))
            .limit(10)
            .reduce(
                TextComponent.builder(),
                (builder, entry) ->
                    builder
                        .append(entry.getKey().getName(NameStyle.VERBOSE))
                        .append(": ", TextColor.GRAY)
                        .append("" + entry.getValue().intValue(), TextColor.WHITE),
                ComponentBuilder::append)
            .build();
```

Pablete1234 (changes requested):

applenick:
@mrcookieunderscore13 
Related to the new screenshot, format looks great! 
Perhaps you could make `You:` the same color as the player’s name (as in if I was participating and my name was dark_red, the `You` would match)? Also consider making `Score` and `You` translatable.

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
mrcookieunderscore13: Well, I need the output to be a list, so I can use `TextFormatter.list(...)` to list the names in a nice format: `Player1, Player2, ... , and PlayerN`. Using a collector seems like the proper way right?

Pablete1234: A joining collector may be a better alternative then, look at how the `Collectors.joining` collector works for strings and do a similar one for this

mrcookieunderscore13: Using Collectors.joining, there is no easy way to have the last element separated like `<element n-1>, and <element n>`, rather than just a comma. I don't understand what is wrong with the method I am currently using.

KingOfSquares: @mrcookieunderscore13 Maybe you are looking for something like what is in ```TextFormatter#list``` ?

mrcookieunderscore13: @KingOfSquares I do use that method. I use the stream to make an ordered list of the players, and then a few lines down use `TextFormatter#list` to format it properly.

mrcookieunderscore13:
Applenick I made your changes, but the computer I dev on died, so thats why I haven't commited the change yet. I should be good to go in about a week.

applenick:
@mrcookieunderscore13 Ah thanks for the update!

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
KingOfSquares: Right, then I think I misunderstood the argument between you and Pablo

Pablete1234 (approved):

PR #623, opened 2020-08-21
Fix vanish chat formatting

applenick:
# Fix vanish chat formatting
Another formatting fix related to the recent chat changes. While vanished, sending a message to global or team chat should re-format it as an admin chat message. While currently the message is still sent to the proper recipients, the formatting is not applying the admin chat prefix. This PR resolves that issue.


Pablete1234 (commented):
Looking thru this, it looks like being vanished enforcing admin chat is simply an illusion, wich is an issue. None of the logic (like permissions) is being applied.

/g and /t should call the command /a method if vanished, instead of changing the prefix on one place and who can see it on another

PR #626, opened 2020-08-25
Fix bukkit command permissions not working

Brottweiler:
The `bukkit.*` commands are not actually a thing and instead `minecraft.` should be used.


Electroid (approved):
Thanks for catching this

PR #627, opened 2020-08-28
Improvements to custom voting options

applenick:
# Improvements to custom voting options

Recently some staff on OCC have reported some issues with maps added via the `/vote` command not showing up in the current poll. The main source of this issue stems from the custom map vote selection data being stored in the individual VotingPool instances rather than the MapPoolManager. As when the server switches pools to accommodate player sizes, the custom vote will not carry over. Instead, the selected maps will not show up until the server switches back to the pool in which the maps were selected.

To resolve this I refactored the custom vote code into a `VotingPoolOptions` class, and an instance of that is created and stored in `MapPoolManager`

Also included are several small text fixes that improve some error commands that staff were confused by. An example being attempting to use any `/vote` command while a vote was in progress would display a disabled message, instead I’ve adjusted that to `Unable to modify vote at this time`

This should resolve all issues that have been reported regarding the `/vote` command, however if there are any other suggestions just let me know 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rotation/VotingPoolOptions.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/VotingPoolOptions.java
@@ -1,0 +1,64 @@
+package tc.oc.pgm.rotation;
+
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import java.util.Map;
+import java.util.Set;
+import tc.oc.pgm.api.map.MapInfo;
+
+public class VotingPoolOptions {
+
+  // Set of maps to be used in custom vote selection
+  private Set<MapInfo> customVoteMaps;
+
+  // Whether custom map selection should replace existing entries
+  private boolean replace;
+
+  public VotingPoolOptions() {
+    this.customVoteMaps = Sets.newHashSet();
+    this.replace = true;
+  }
+
+  public boolean shouldOverride() {
+    return customVoteMaps.size() >= VotingPool.MIN_CUSTOM_VOTE_OPTIONS && !replace;
+  }
+
+  public boolean isReplace() {
+    return replace;
+  }
+
+  public boolean toggleMode() {
+    this.replace = !replace;
+    return replace;
+  }
+
+  public boolean addVote(MapInfo map) {
+    if (customVoteMaps.size() < VotingPool.MAX_VOTE_OPTIONS) {
+      this.customVoteMaps.add(map);
+      return true;
+    }
+    return false;
+  }
+
+  public boolean removeMap(MapInfo map) {
+    return this.customVoteMaps.remove(map);
+  }
+
+  public Set<MapInfo> getCustomVoteMaps() {
+    return customVoteMaps;
+  }
+
+  public boolean isAdded(MapInfo info) {
+    return customVoteMaps.stream().anyMatch(s -> s.getName().equalsIgnoreCase(info.getName()));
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/rotation/VotingPoolOptions.java
```diff
@@ -0,0 +1,64 @@
...
+    return replace;
+  }
+
+  public boolean toggleMode() {
+    this.replace = !replace;
+    return replace;
+  }
+
+  public boolean addVote(MapInfo map) {
+    if (customVoteMaps.size() < VotingPool.MAX_VOTE_OPTIONS) {
```
KingOfSquares:
```suggestion
    if (customVoteMaps.size() <= VotingPool.MAX_VOTE_OPTIONS) {
```
Or else max would be 4

core/src/main/java/tc/oc/pgm/rotation/VotingPoolOptions.java
```diff
@@ -0,0 +1,64 @@
+package tc.oc.pgm.rotation;
+
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+import java.util.Map;
+import java.util.Set;
+import tc.oc.pgm.api.map.MapInfo;
+
+public class VotingPoolOptions {
```
KingOfSquares: `CustomVotingPoolOptions`?

I think it should reflect that it's only used for the custom overrides, since `MapPoll` holds the default poll

core/src/main/java/tc/oc/pgm/rotation/VotingPoolOptions.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public void clear() {
+    customVoteMaps.clear();
+  }
+
+  public Map<MapInfo, Double> getCustomVoteMapWeighted() {
+    Map<MapInfo, Double> maps = Maps.newHashMap();
+    customVoteMaps.forEach(map -> maps.put(map, VotingPool.DEFAULT_WEIGHT));
+    return maps;
+  }
```
KingOfSquares:
```suggestion
  public Map<MapInfo, Double> getCustomVoteMapWeighted() {
    return customVoteMaps.stream().collect(Collectors.toMap(map -> map, VotingPool.DEFAULT_WEIGHT));
  }
```

I like the one liner at least :)

util/src/main/i18n/templates/map.properties
```diff
@@ -58,6 +58,8 @@ map.setNext.revert = {0} has removed {1} as the next map
 
 map.noNextMap = No next map
 
+map.noMapsFound = No maps found
```
KingOfSquares: If I understand this correctly, this is displayed if you do `/vote clear` and there is no maps to clear.
I think this is too vague if I am correct. Maybe something like `There are no custom vote options to clear` | `There are no maps to clear from the vote` | `No vote options to clear` | `No maps to clear`

core/src/main/java/tc/oc/pgm/rotation/VotingPoolOptions.java
applenick: Exactly! 0-4 is 5 😄

applenick:
Thanks for the suggestions @KingOfSquares!
Should be good to go now 👍

Electroid (approved):

PR #628, opened 2020-08-28
Color the team picker

applenick:
# Color the team picker 🎨 

Just a minor visual update I thought would look neat and be helpful for users. Now the team picker item will be dyed the same color as the team you have chosen. 

## Screenshot
colors


Electroid (approved):
Very cool!

PR #632, opened 2020-08-29
Implement rendering arbitrary team amounts on tablist

Pablete1234:
Allow rendering arbitrary amount of teams on the tablist, instead of showing only the first 4 teams


Electroid (approved):
🎉

PR #636, opened 2020-09-01
Remove legacy tablist and use modern tablist across versions

Pablete1234:
Removes all the legacy tablist that had a custom display, and instead uses the same code that 1.8+ uses, to render the tablist for legacy players, with 4 columns, and using the first 2 rows for header (only on legacy clients).

Some screenshots:
image
image
image


Electroid (approved):
Awesome!

PR #643, opened 2020-09-11
Disable arbitrary sorting of sidebar competitors

alexsosnovsky:
This change enables teams to be sorted by objective completion on the scoreboard for everyone, not just observers. Of course, players on a team will still see their own team at the top.

This solves the issue on many-team maps where sometimes the winning team is not seen on the scoreboard by any other teams, and is therefore not seen as a threat.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
@@ -2,7 +2,6 @@ package tc.oc.pgm.scoreboard;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Ordering;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


Pugzy:
This ~~makes it harder to know what team you're on and~~ removes any possible suspense during tournament matches.

Pablete1234:
> This ~makes it harder to know what team you're on and~ removes any possible suspense during tournament matches.

It does not. Your team will always be the highest, it's other teams that appear lower, sorted. In the case of 2-team maps, this has no effect whatsoever.

alexsosnovsky:
As stated earlier, this would not affect normal 2 team tournament matches. If it's really that important to have arbitrary sorting for tournament matches with more than 2 teams, maybe it could be enabled if scoreboard wool proximity is enabled? As that is a feature typically enabled on tournament servers.

Pablete1234 (approved):

Electroid (approved):

PR #646, opened 2020-09-13
Support modifybowprojectile in 1.4 protocol

Pablete1234:
Modifybowprojectile was deprecated in 1.4 proto for no apparent good reason, there's no valid replacement for it and it just means if anyone wants to use it they have to downgrade their map xml to 1.3.x, for no real reason.

Until there's a better alternative, maps should be allowed to use it


Electroid (approved):
I'm good with this.

PR #647, opened 2020-09-13
Fix rounding errors when formatting durations

### LINKED ISSUE #464: Rounding errors when formatting durations
Screen Shot 2020-05-07 at 3 01 48 PM

See `TimeUtils`

lPoloBlash:
Fixes #464


KingOfSquares:
Why not lower the `minQuantity` to `1`? This fix makes the use of the changed method _always_ return the duration in seconds?

Or you could literally do something like 
```java
if(duration.toMillis == 1000) return formatPeriod(ChronoUnit.SECONDS, 1)
```

Pablete1234:
> Why not lower the `minQuantity` to `1`? This fix makes the use of the changed method _always_ return the duration in seconds?

It makes it _never_ return in millis, that doesn't mean it will be _always_ in seconds, it just means seconds will be the lowest unit used

Electroid (approved):

PR #658, opened 2020-09-23
Added option to click a link to download SportsPaper.

benrobson:
I've added an option for users to just click a link to download SportsPaper, in the case that they don't know how to use bash (for the novice user they wouldn't know).

Signed off by Ben Robson


KingOfSquares:
Good idea! To get DCO to pass you have to sign off your commits :)

benrobson:
So I edited my original commit message and added a sign-off, should update eventually?

benrobson:
Well getting DCO to sign off the commit took a while but I think I got it.

Electroid:
I updated the README during last release.

PR #663, opened 2020-09-30
Prevent locked slots from clearing after player reset

### LINKED ISSUE #662: Locked armor slot items lose locked attribute after death
An item that is set to the armor slot with the locked attribute will lose this attribute after the player respawns.  This happens when the item in the slot is in the keepitem module of course.
`<helmet material="stone button" locked="true">`
`<itemkeep>
    <item>stone button</item>
</itemkeep>`

chatasma:
Fixes #662 

Signed-off-by: delyte <nodeadbatteries@gmail.com>


harvanchik:
Thank you!

Electroid:
@Pablete1234 See any potential bugs with this change?

Pablete1234:
I'm pretty sure this will break if you're first given a kit with non-removable armor, and after death you're given another kit with removable armor.

The issue feels more related to itemkeep not re-applying locking

Electroid:
For the reasons above, going to close this. Although, thanks so much for submitting a PR!

PR #665, opened 2020-10-01
Reintroduce projectile trails

applenick:
# Reintroduce Projectile trails

This PR reintroduces a fun feature from the old PGM. Arrow trails spawn a small particle trail matching your teams color. I adapted the code that made this possible from the old PGM as well as added the ability to toggle the arrow trails on and off via a new setting`/toggle effects` Note: I also made the setting “effects” that way in the future if we choose to allow more customized effects they can fall under this setting category. 

## Screenshot
arrowsyay

## Feedback
This is a feature I personally enjoy and adds a bit of uniqueness to PGM. However if desired I can add an option to the config to completely disable this feature. However I felt that having simply the new setting toggle would be enough. 

Any feedback is appreciated :+1:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
@@ -45,34 +45,55 @@ public class ProjectileTrailMatchModule implements MatchModule, Listener {
               } else {
                 final Color color = (Color) projectile.getMetadata(TRAIL_META, PGM.get()).value();
 
-                match.getPlayers().stream()
-                    .filter(
-                        pl ->
-                            pl.getSettings().getValue(SettingKey.EFFECTS)
-                                == SettingValue.EFFECTS_ON)
-                    .forEach(
-                        player -> {
-                          player
-                              .getBukkit()
-                              .spigot()
-                              .playEffect(
-                                  projectile.getLocation(),
-                                  Effect.COLOURED_DUST,
-                                  0,
-                                  0,
-                                  rgbToParticle(color.getRed()),
-                                  rgbToParticle(color.getGreen()),
-                                  rgbToParticle(color.getBlue()),
-                                  1,
-                                  0,
-                                  50);
-                        });
+                for (MatchPlayer player : match.getPlayers()) {
+                  boolean colors =
+                      player
+                          .getSettings()
+                          .getValue(SettingKey.EFFECTS)
+                          .equals(SettingValue.EFFECTS_ON);
+                  if (colors) {
+                    player
+                        .getBukkit()
+                        .spigot()
+                        .playEffect(
+                            projectile.getLocation(),
+                            Effect.COLOURED_DUST,
+                            0,
+                            0,
+                            rgbToParticle(color.getRed()),
+                            rgbToParticle(color.getGreen()),
+                            rgbToParticle(color.getBlue()),
+                            1,
+                            0,
+                            50);
+                  } else {
+                    // Play the critical effect to those who have effects off, to replicate original
+                    // arrow behavior
+                    if (isCriticalArrow(projectile)) {
+                      player
+                          .getBukkit()
+                          .spigot()
+                          .playEffect(
+                              projectile.getLocation(), Effect.CRIT, 0, 0, 0, 0, 0, 1, 0, 50);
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
```diff
@@ -0,0 +1,110 @@
...
+    if (player != null) {
+      final Projectile projectile = event.getEntity();
+      projectile.setMetadata(
+          TRAIL_META, new FixedMetadataValue(PGM.get(), player.getParty().getFullColor()));
+      // Set critical metadata to false in order to remove default particle trail.
+      // The metadata will be restored just before the arrow hits something.
+      if (projectile instanceof Arrow) {
+        final Arrow arrow = (Arrow) projectile;
+        arrow.setMetadata(CRITICAL_META, new FixedMetadataValue(PGM.get(), arrow.isCritical()));
+        arrow.setCritical(false);
```
KingOfSquares, line 111: This would also remove critical effects for people with /toggle effects off? Unsure how to solve, maybe add white trails manually?

applenick: Yeah I'm not sure, maybe instead of making it a toggle per player, perhaps it could be a server-wide config value. I'm not sure if a white trail would be the right call, since the point of the toggle is to disable the effects. Uh any suggestions you have would be welcome.

KingOfSquares: I was trying to say you could "simulate" a critical effect? `Effect.CRIT`

KingOfSquares: I don't think the crits can have colors, so you could just replace `Effect.COLORED_DUST` with the crit one for players with the setting toggled off?

Pugzy: How does it look with both the crit and color? 
Maybe simulate the critical effect as king says but to a less extent than vanilla does.

applenick: Oh that’s actually a really cool idea. I’ll experiment with that later and let you guys know the results.

applenick: Thanks again for the suggestion @KingOfSquares & @Pugzy. 
I adjusted it and works flawlessly 👍 Much better alternative to having them not show up at all.

Electroid:
I remember this being a controversial change, should it be brought back?

applenick:
Personally I like it, feels it gives adds a fun feature to PGM. Plus with being able to toggle it off and retain the critical effect, I don't see the harm in it. (Whereas the original implementation did not offer a setting value)

Though the question I guess is should it be enabled by default or not.

Pablete1234:
> I remember this being a controversial change, should it be brought back?

It was controversial because there was no way to toggle it on pgm, i added it to cardinal with a toggle and no one complained, and the same exists here

KingOfSquares:
Just remembered, i've experienced lag with setting related features(most notably death messages) already so i can only imagine that this will make it trigger more often. (Really many requests) i dont know if playersettings are cached right now, maybe OCC has a bad database, anyway this will probably trigger alit

applenick:
@KingOfSquares 
https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/db/CacheDatastore.java
They are cached 👍

KingOfSquares:
> @KingOfSquares
> https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/db/CacheDatastore.java
> They are cached 👍

I wonder what is making it so slow then

applenick:
@KingOfSquares 
I’ve seen what you’ve mentioned happen on OCC before during times when a lot of players are online. My only guess is network issues/or trying to keep up with all the online players at once. Since OCC does use a remote database to store everything. On 95% of PGM servers, I don’t believe it will be an issue. However if fetching items from the database becomes more of an issue someone will have to investigate it more.

Electroid (approved):

PR #666, opened 2020-10-01
Add /votebook command

applenick:
# Add a vote book command

This PR introduces a new command, `/votebook`, which allows for players who have lost the current voting book or who have joined after a vote has begun to receive the item again.

## Screenshot
votebook

## Feedback
While there is already the `/votenext -o` flag, I felt that method of forcing open the vote book to not be as intuitive. And this command solves that by simply giving another book. 

If you have any further suggestions just let me know :+1:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
@@ -284,21 +284,36 @@ public final class MapPoolCommand {
       @Switch('o') boolean forceOpen,
       @Text MapInfo map)
       throws CommandException {
+    MapPoll poll = getVotingPool(player, sender, mapOrder);
+    if (poll != null) {
+      boolean voteResult = poll.toggleVote(map, ((Player) sender).getUniqueId());
+      Component voteAction =
+          TranslatableComponent.of(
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
```diff
@@ -284,21 +284,36 @@ public static void voteNext(
...
+      player.sendMessage(voteAction);
+      poll.sendBook(player, forceOpen);
+    }
+  }
+
+  @Command(aliases = "votebook", desc = "Spawn a vote book")
+  public void voteBook(MatchPlayer player, CommandSender sender, MapOrder mapOrder)
+      throws CommandException {
+    MapPoll poll = getVotingPool(player, sender, mapOrder);
+    if (poll != null) {
```
KingOfSquares, line 304: Maybe a message if you try to do `/votebook` during a match? We should have a generic `You can not use this command while the match is running` anyway IMO.

applenick: The `sendBook` method already does that, it says something like `There is nothing to vote for at this time`, so no need to add an additional message to the command 😄

Electroid (approved):

PR #667, opened 2020-10-01
Reintroduce /velocity command

applenick:
# Reintroduce /velocity command

In an effort to make PGM more useful for map developers, I've reintroduced the `/velocity` which can be used to test features such as jump-pads and allow for the mapmaking process to progress more efficiently. 

I hope to bring a number of other map dev specific commands into PGM, but decided to start with this one. 

## Screenshot
velocity

## Feedback/Suggestions
Once again always open to feedback and look forward to any ideas people may have for other mapmaking commands/features that would prove useful :+1:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Permissions.java
+++ b/core/src/main/java/tc/oc/pgm/api/Permissions.java
@@ -93,6 +93,7 @@ public interface Permissions {
               .put(GAMEPLAY, true)
               .put(DEBUG, true)
               .put(RELOAD, true)
+              .put(MAPDEV, true)
               .build());
 
   Permission ALL =
--- a/core/src/main/java/tc/oc/pgm/api/Permissions.java
+++ b/core/src/main/java/tc/oc/pgm/api/Permissions.java
@@ -37,6 +37,7 @@ public interface Permissions {
   String BAN = ROOT + ".ban"; // Access to the /ban command
   String FREEZE = ROOT + ".freeze"; // Access to the /freeze command
   String VANISH = ROOT + ".vanish"; // Access to /vanish command
+  String MAPDEV = ROOT + ".mapdev"; // Access to mapdev related commands
 
   String MAPMAKER = GROUP + ".mapmaker"; // Permission group for mapmakers, defined in config.yml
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/api/Permissions.java
```diff
@@ -37,6 +37,7 @@
   String BAN = ROOT + ".ban"; // Access to the /ban command
   String FREEZE = ROOT + ".freeze"; // Access to the /freeze command
   String VANISH = ROOT + ".vanish"; // Access to /vanish command
+  String MAPDEV = ROOT + ".mapdev"; // Access to mapdev related commands
```
KingOfSquares, line 40: You should probably add this permission to `Permission DEVELOPER`. Also to `registerAll()`

util/src/main/i18n/templates/command.properties
```diff
@@ -81,3 +81,5 @@ command.message.blocked = {0} is not accepting messages at this time.
 
 # {0} = map name
 command.maps.hover = Click to view {0}
+
+command.mapdev.velocity = Applying Velocity
```
KingOfSquares, line 85: A reason you did not do the text like this; `Applying Velocity ({0}, {1}, {2})` ?

core/src/main/java/tc/oc/pgm/api/Permissions.java
applenick: Good catch, will do 👍

util/src/main/i18n/templates/command.properties
applenick: I thought it was best to leave some formatting out of the translations, but if you think it would be better I'll switch it to that.

KingOfSquares: I was just wondering. I think both works, but that with the formatting the string makes more sense? It's the norm with the other strings and you would save the string building in the command method. I think both works, just wanted to check you didn't forget the option.

applenick: I was going to switch it around, but then I took a look and actually I wanted to hide the player name if it’s executed on yourself, and show if the target is another player. Seems easier this way 😄

Electroid:
I thought we weren't adding fluff to PGM? :( I know this isn't much of a change, but still

applenick:
@Electroid You’re right, this would be best in a MapDev specific plugin. I’m gonna close it.

PR #671, opened 2020-10-04
Add offset filter

### LINKED ISSUE #664: Add <standing-on> element to the Filters module
### Feature Explanation
Sometimes, you want to know if a player is standing on a certain block, but you don't want to make 10, 20 or even more regions for all the places where that block/material is.  Adding a `<standing-on>` element to the Filters module would make this so much easier.

### Use Case & Implementation
A possible use case using this feature would be to kill a player when standing on a specific block.  Instead of creating regions all around the map for every single occurance of the "kill block," the `<standing-on>` in the filter would do all the work for you.
```xml
<filters>
    <standing-on id="on-stone">
        <material>stone</material>
    </standing-on>
</filters>
```
Then apply a kit everywhere (or wherever this should take affect) to set their health to zero.

Whether or not it is implemented as an element like `<standing-on>` or another way, it would be nice to have a way of detecting this.  In addition to `<standing-on>` it would make sense to add a `<standing-beneath>` (or under) -- and possibly a way of detecting blocks to the right, left, in front, and behind (or synthesize that into a `<touching>` element.

I'm not sure that adding them as elements is the best way of doing it, but I'll leave that up to the more experienced to decide.

### Streamlining
I also wonder if it would be possible to streamline it so that you could also check for things such as if a player is placing a block that is above or below another block/material.

KingOfSquares:
Closes #664

This PR adds the generic `QueryModifier` and a `LocationQueryModifier`. A `QueryModifier` takes a type of query, and modifies it before passing it to its child filter. The `LocationQueryModifier` takes any query with a `Location` and transforms it into a `BlockQuery` with a custom `Location`. 

```xml
<filters>
  <offset id="hardFilter" location="10.5, 5, 15.5">
        <material>bedrock</material>
    </offset>

  <offset id="blockUnderFeetFilter" location="~0, ~-1, ~0">
        <material>cobblestone</material>
    </offset>

  <offset id="inFrontFilter" location="^0,^0,^1">
        <material>water</material>
    </offset>

</filters>
```

Normal numbers(`hardFilter`) and numbers with `~`(`blockUnderFeetFilter`) are world coordinates.
Numbers with `^`(`inFrontFilter`) are local coordinates. (`^<left/right>, ^<up/down>, ^<forwards/backwards`)

World coordinates can be mixed with each other, but you can not mix world and local coordinates.

`hardFilter` looks for a bedrock block on the specific coordinates given

`blockUnderFeetFilter` looks for a cobblestone block under the players feet, `^0,^-1,^0` would also yield the same result

`inFrontFilter` looks for a water block in front of the player, switching out `1` for `-1` would look a block behind instead

It is worth noting since dynamic filters is currently very limited (see `FilterMatchModule`) this will only be checked once the check is triggered, and will not trigger anything by itself. (e.g if this is a kit filter where the kit is applied to a region, the filter will only be checked once the player enters the region, and not continuously as the player is in the region). This puts the current usability of this filter low, but the issue is not something covered by this PR.


Pugzy:
Do regular regions allow for spaces in their coordinates? is `10.5, 5, 15.5` valid?

Pablete1234:
This is currently implemented as a filter, however, it should be a query modifier. What that means is the children shouldn't be a list of materials, the child should be a filter, that you pass the modified query to

harvanchik:
This is great, but like you said, it's still limited since it's not dynamic. I wonder if doing what @Pablete1234 said fixes this issue. 

I do like that you added the ability to mix the three types of coordinates, something Mojang is incapable of.

Pablete1234:
> I do like that you added the ability to mix the three types of coordinates, something Mojang is incapable of.

The difference is mojang implemented them properly.
Absolute and relative coordinates can be mixed, and that's fine.
View-relative coordinates can't be mixed at the same time, but can be if you use subsequent `execute at` commands, but that's because they literally can't be, you can't move "one block towards the direction the player is moving, and one block sideways in the X direction". The 3 `^` coordinates mean something completely different to x,y,z, and can not be mixed.

Further reading about caret notation: <https://minecraft.gamepedia.com/Commands#Local_coordinates:_Caret_notation>

KingOfSquares:
> > I do like that you added the ability to mix the three types of coordinates, something Mojang is incapable of.
> 
> The difference is mojang implemented them properly.
> Absolute and relative coordinates can be mixed, and that's fine.
> View-relative coordinates can't be mixed at the same time, but can be if you use subsequent `execute at` commands, but that's because they literally can't be, you can't move "one block towards the direction the player is moving, and one block sideways in the X direction". The 3 `^` coordinates mean something completely different to x,y,z, and can not be mixed.
> 
> Further reading about caret notation: https://minecraft.gamepedia.com/Commands#Local_coordinates:_Caret_notation

I have seem to misunderstood, I thought `^` was related to the specific block the player is looking at, and not the direction they are looking, will fix...

I will also make an abstract class `QueryModifierFilter` which takes a child filter, and make `OffsetFilter` extend this class

harvanchik:
> I will also make an abstract class `QueryModifierFilter `which takes a child filter, and make `OffsetFilter `extend this class

@KingOfSquares Does doing this make the filter dynamic?  For example, can this be achieved then? -- applying the kit when `cobblestone` is one block below the player and removing the kit when not.

KingOfSquares:
> > I will also make an abstract class `QueryModifierFilter `which takes a child filter, and make `OffsetFilter `extend this class
> 
> @KingOfSquares Does doing this make the filter dynamic? For example, can this be achieved then? -- applying the kit when `cobblestone` is one block below the player and removing the kit when not.

No, that would have to be implemented by its own (dynamic filters is not yet in this version of PGM)

KingOfSquares:
I fixed the implementation of `^`locations. The function I found to calculate the correct position relative to the player seems to have 0-2° error on pitch and yaw, so for this to work more flawlessly it should probably be used as a dynamic filter.

core/src/main/java/tc/oc/pgm/filters/LocationQueryModifier.java
```diff
@@ -0,0 +1,96 @@
...
+    } else {
+      newVector =
+          new Vector(
+              parseRelativeLocation(coords[0], location.getX()),
+              parseRelativeLocation(coords[1], location.getY()),
+              parseRelativeLocation(coords[2], location.getZ()));
+    }
+
+    return new BlockQueryCustomLocation(
+        query.getEvent(), newVector.toLocation(query.getMatch().getWorld()));
```
Pablete1234: You're losing pitch & yaw information when doing this.

Use `location.set(x,y,z)` to keep yaw/pitch, but watch out of unintended side-effects, i doubt it will be a thing since player#getLocation always returns a new location, but if other places create this location it could be keeping it modified even after leaving the child filter.

Alternatively just copy pitch & yaw alongisde world

KingOfSquares: this is solved with a `clone()`, good catch!

core/src/main/java/tc/oc/pgm/filters/LocationQueryModifier.java
```diff
@@ -0,0 +1,97 @@
...
+import org.bukkit.Location;
+import org.bukkit.event.Event;
+import org.bukkit.util.Vector;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.filter.query.LocationQuery;
+import tc.oc.pgm.filters.query.BlockQuery;
+
+public class LocationQueryModifier extends QueryModifier<LocationQuery> {
+
+  private final String[] coords;
```
Electroid: Break this up into multiple classes, right now you are parsing this field during every query. Same with the relative vs. absolute coordinates. Do parsing at XML time, but it creates different classes with different functionality.

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -443,4 +446,46 @@ public StructuralLoadFilter parseStructuralLoad(Element el) throws InvalidXMLExc
...
     return new TimeFilter(XMLUtils.parseDuration(el, null));
   }
+
+  // Methods for parsing QueryModifiers
+
+  @MethodParser("offset")
+  public LocationQueryModifier parseOffsetFilter(Element el) throws InvalidXMLException {
+    String value = el.getAttributeValue("location");
+    // Check vector format
+    XMLUtils.parseVector(new Node(el), value.replaceAll("[\\^~]", ""));
```
Pablete1234: I belive you could just pass this vector arround instead of a double[] with 3 values.
In case of the WorldLocationQueryModifier, pass both the vector and if x, y or z are relative or absolute

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -453,4 +457,46 @@ public TimeFilter parseTimeFilter(Element el) throws InvalidXMLException {
...
+        throw new InvalidXMLException("Cannot mix world & local coordinates", el);
+    }
+
+    if (local == null) throw new InvalidXMLException("No coordinates provided", el);
+
+    if (local) {
+      double x = Double.parseDouble(coords[0].substring(1));
+      double y = Double.parseDouble(coords[1].substring(1));
+      double z = Double.parseDouble(coords[2].substring(1));
+      return new LocalLocationQueryModifier(parseChild(el), new Vector(x, y, z));
```
Pablete1234: You're already running an XMLUtils.parseVector all the way at the top, no need to parse individual coords in substrings

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -453,4 +457,46 @@ public TimeFilter parseTimeFilter(Element el) throws InvalidXMLException {
...
+      double[] coordinates = new double[3];
+      for (int i = 0; i < coords.length; i++) {
+        String coord = coords[i];
+        boolean isRelative = coord.startsWith("~");
+        relative[i] = isRelative;
+        coordinates[i] = Double.parseDouble(isRelative ? coord.substring(1) : coord);
+      }
+
+      return new WorldLocationQueryModifier(
+          parseChild(el), new Vector(coordinates[0], coordinates[1], coordinates[2]), relative);
```
Pablete1234: kinda same thing here, no need for coordinates, just use the vector and just check for ~

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
KingOfSquares: Right, forgot about that

Pablete1234 (approved):

Electroid (approved):

PR #678, opened 2020-10-13
Fix match announcer subtitle check using team type

### LINKED ISSUE #677: End of match crashes players
Due to PR #674 which recently changed this [line](https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java#L65) of code, an issue now occurs whenever a match ends. As passing a null subtitle value to `viewer.showTitle(title, subtitle, 0, 40, 40);` is giving issues (Caused most online players to crash/be kicked from the server). 

@Pugzy since you were the last to modify it, I’ll leave it for you to fix. Otherwise I’ll take care of it tomorrow. Thanks

Pugzy:
Fixes https://github.com/PGMDev/PGM/issues/677.

Move the subtitle message check to be based on team Observers rather than initialising as null.


Electroid (approved):
Thanks for fixing this!

PR #680, opened 2020-10-17
Prevent invalid tablist on vanish with forced teams

Pablete1234:
Currently vanish/unvanish assumes you'll be doing so from observers, since it forces you to observers. The issue comes when the event is prevented by a separate plugin that enforces teams (like the events plugin), causing the tablist to be in an invalid state where the player appears to be in obs after unvanishing, but is actually in the team.

This PR does proper updating of the lists where the player could be to ensure consistency when vanishing/unvanishing


Electroid (dismissed):

Electroid:
@Pablete1234 You'll need to format, this and the other PR

Pablete1234:
> @Pablete1234 You'll need to format, this and the other PR

Yeah, i didn't notice yesterday, pushed the fix now :+1:

Electroid (approved):

PR #689, opened 2020-11-01
Add a match state filter

KingOfSquares:
~~Sorry I didn't test I wanted to get this in time for Hacktoberfest, will of course test before its merged~~
It already ended 🤦 


Pugzy:
Proto 1.4.2 added filters that allow you to check the match state.

These can be found under the [Matcher Elements](http://docs.oc.tc/modules/filters) section and are `<match-started/>`, `<match-running/>` and `<match-finished/>`.

Not sure why one implementation would be favoured over another (others may able to comment) but you should probs add the ability to define them in that way to bring us closer to that sweet sweet 1.4.2 support.

Pablete1234:
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/parser/FilterDefinitionParser.java#L463
I belive we should try to re-implement what's missing in a way that is compatible with what existed before

core/src/main/java/tc/oc/pgm/filters/MatchStateFilter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.function.Predicate;
+import tc.oc.pgm.api.filter.query.MatchQuery;
+import tc.oc.pgm.api.match.Match;
+
+public class MatchStateFilter extends TypedFilter<MatchQuery> {
+
+  private final Predicate<Match> matchPredicate;
+
+  public MatchStateFilter(Predicate<Match> matchPredicate) {
```
Pablete1234: I don't get the obsession for a predicate here. Just use a `Set<MatchPhase>`

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -443,4 +446,52 @@ public StructuralLoadFilter parseStructuralLoad(Element el) throws InvalidXMLExc
...
+        break;
+      case "before":
+        matchPredicate = m -> m.getPhase() == MatchPhase.IDLE;
+        break;
+      case "started":
+        matchPredicate =
+            m -> m.getPhase() == MatchPhase.RUNNING || m.getPhase() == MatchPhase.FINISHED;
+    }
+    if (matchPredicate == null)
+      throw new InvalidXMLException("Invalid or no match state found", el);
```
Electroid: Seems like this should be a private function `parseMatchPhase(String) -> MatchPhase`

core/src/main/java/tc/oc/pgm/filters/MatchStateFilter.java
Electroid: See my other comment, this constructor can then be `MatchPhaseFilter(MatchPhase)`

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/filters/MatchStateFilter.java
KingOfSquares: Yes! Predicate was an old thing before i actually used all MatchPhases (was using method refrences instead like Match::isRunning etc.)

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -443,4 +442,52 @@ public StructuralLoadFilter parseStructuralLoad(Element el) throws InvalidXMLExc
...
+  public MatchPhaseFilter parseMatchRunning(Element el) throws InvalidXMLException {
+    return parseMatchPhaseFilter("running", el);
+  }
+
+  @MethodParser("match-finished")
+  public MatchPhaseFilter parseMatchFinished(Element el) throws InvalidXMLException {
+    return parseMatchPhaseFilter("finished", el);
+  }
+
+  private MatchPhaseFilter parseMatchPhaseFilter(String matchState, Element el)
```
Electroid: I have an idea to make this easier:

* Change the signature of `parseMatchPhaseFilter` return a generic `Filter`
* Change `MatchPhaseFilter` to only accept 1 `MatchPhase` and not a `Set`
* For the "started" case, return an `OrFilter(MatchPhaseFilter(RUNNING), MatchPhaseFilter(FINISHED))`

See what I'm getting at?

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
KingOfSquares: Yes! Great idea 👍

Electroid (dismissed):

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -453,4 +454,50 @@ public TimeFilter parseTimeFilter(Element el) throws InvalidXMLException {
...
+        break;
+      case "starting":
+        matchPhase = MatchPhase.STARTING;
+        break;
+      case "idle":
+        matchPhase = MatchPhase.IDLE;
+        break;
+      case "started":
+        return AnyFilter.of(
+            new MatchPhaseFilter(MatchPhase.RUNNING), new MatchPhaseFilter(MatchPhase.FINISHED));
```
Pablete1234:
```suggestion
      case "running": return MatchPhaseFilter.RUNNING;
      case "finished": return MatchPhaseFilter.FINISHED;
    // ...
```
As suggested in discord you should make static final fields in MatchPhaseFilter and reuse the same instances

Pablete1234 (changes requested):

Pablete1234 (approved):

Pablete1234:
@Electroid could we get this merged?

Electroid (approved):

PR #690, opened 2020-11-03
Refactor: Remove old sendMessage/showTitle

KingOfSquares:
- Moves away from old methods, properly references the console, fixes a couple of sound names

Next commit: Replacing some `StringBuilder`s with Components


core/src/main/java/tc/oc/pgm/command/ClassCommand.java
```diff
@@ -65,10 +65,12 @@ public void classList(Match match, MatchPlayer player) {
     final PlayerClass currentClass = classes.getSelectedClass(player.getId());
 
     player.sendMessage(
-        LegacyFormatUtils.dashedChatMessage(
-            ChatColor.GOLD + TextTranslations.translate("match.class.title", player.getBukkit()),
-            "-",
-            ChatColor.RED.toString()));
+        text(
+            LegacyFormatUtils.dashedChatMessage(
```
applenick, line 69: Switch this to use [TextFormatter#horizontalLineHeading](https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/text/TextFormatter.java#L110) 😄

KingOfSquares: Yeah I know, I just thought it would make sense to do such a big refactor step by step.

I was thinking TextFormatter should be refactored to use `MatchPlayer`s instead anyways

applenick: Ah just be careful about that, since if an outside plugin wants to use the utils you don’t want them to be forced to use PGM just to get the MatchPlayer.

KingOfSquares: Right, then the bukkit `Player` would also work, yeah? `player.getBukkit()`

applenick: It depends on the method. For example I use the horizontalLineHeading methods in my requests plugin [here](https://github.com/applenick/Requests/blob/master/src/main/java/tc/oc/occ/Requests/RequestCommand.java#L245). If it were switched to player, then would no longer be able to run the command from the console.

KingOfSquares: Then what about a `Locale` ? This seems to be the only need for the sender

applenick: I don’t have time atm to look into what the sender is needed for, but awesome yeah if it’s something more general like `Locale` that would be the best path to take 👍

core/src/main/java/tc/oc/pgm/PGMAudiences.java
```diff
@@ -0,0 +1,30 @@
...
+
+import net.kyori.adventure.audience.Audience;
+import net.kyori.adventure.key.Key;
+import net.kyori.adventure.platform.bukkit.BukkitAudiences;
+import net.kyori.adventure.sound.Sound;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.bukkit.plugin.Plugin;
+
+public class PGMAudiences {
```
Electroid: Instead of having something like `PGMAudiences`, I would like this to live in `util`. 

`tc.oc.util.text.Audience` would be both an interface, with things like `sendWarning` wrapper, and 1-stop-shop to create audiences from players or groups of players.

KingOfSquares: I see, any reason you decided to deprecate the provider that was in there?

The way im understanding this is that the provider also should stay in this Audience class

KingOfSquares:
Thanks for taking my contribution into review Electroid. I hope this is closer to what you imagined

Electroid:
Great progress, willing to merge this in.

PR #691, opened 2020-11-05
Refactor: more components and new book

KingOfSquares:
- Adding static imports whenever i remember
- use new openBook (remove old NMS hack)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rotation/VotingBookListener.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/VotingBookListener.java
@@ -1,0 +1,37 @@
+package tc.oc.pgm.rotation;
+
+import org.bukkit.Material;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.block.Action;
+import org.bukkit.event.player.PlayerInteractEvent;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.ListenerScope;
+
+@ListenerScope(MatchScope.LOADED)
+final class VotingBookListener implements Listener {
+
+  private final VotingPool votingPool;
+  private final Match match;
+
+  public VotingBookListener(VotingPool votingPool, Match match) {
+    this.votingPool = votingPool;
+    this.match = match;
+  }
+
+  @EventHandler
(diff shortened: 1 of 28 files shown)


core/src/main/java/tc/oc/pgm/rotation/VotingBookListener.java
```diff
@@ -0,0 +1,37 @@
...
+  public VotingBookListener(VotingPool votingPool, Match match) {
+    this.votingPool = votingPool;
+    this.match = match;
+  }
+
+  @EventHandler
+  public void openVote(PlayerInteractEvent event) {
+    MatchPlayer player = match.getPlayer(event.getPlayer());
+    if (isRightClick(event.getAction())
+        && event.getMaterial() == Material.ENCHANTED_BOOK
```
Electroid, line 31: Use metadata to mark the book as a voting book. Or else *any* enchanted book becomes a vote book.

KingOfSquares:
I tried to implement your feed back Electroid, but it posed some challenges,

`ItemStack`s can not have metadata like entities(they don't implement the `Metadatable`(or whatever))

I tried adding a custom enchantment as a metadata-ish validator, that didn't persist in some weird cases so I threw out that idea. (The enchantment would disappear if a player fetched an item from the creative menu and put it into the hotbar??)

I ended up using a localized info string as a validator in a similar way to how it's done with metadata, it's a little hacky but imo the best solution that does not look weird for the player.

Also added a little tip to the book lore for all player with the `EXTRA_VOTE` permission.
(all strings in the picture is ofc translatable)
Skjermbilde 2020-11-06 kl  23 24 21

Electroid:
@KingOfSquares This code allows you tag an `ItemStack`: https://github.com/PGMDev/PGM/blob/1a6867f7f87c121a1f238fd32f50f5987c38c1ea/util/src/main/java/tc/oc/pgm/util/inventory/tag/ItemTag.java#L60

KingOfSquares:
Seems to work fine, it only looks a little weird (An empty box)
Skjermbilde 2020-11-07 kl  10 59 45

I have to omit the extra voting feature thing If I am going to use this, I think it would look to weird with the space.
What do you think?

Electroid:
Then I guess for now, not using the metadata is fine.

Electroid (approved):
Looks great!

Electroid:
I'll merge this into the feature branch, feel free to make another PR. Maybe one that replaces all the usages with static imports? Maybe you can do that with IntelliJ find & replace?

PR #694, opened 2020-11-11
Fix transitive dependency of bungee chat

AustinLMayes:
Via requires bungee chat upstream, which breaks builds if you don't have it already resolved locally.


Electroid (approved):
Thanks, get your credentials verified though!

PR #699, opened 2020-11-14
Re-introduce player count filter

AustinLMayes:
Re-adds the `<players>` filter with some more configurability than the legacy syntax. 

The filter now supports a child filter to determine which players should be counter. Legacy options get converted to the new filter.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
+++ b/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
@@ -1,5 +1,6 @@
 package tc.oc.pgm.filters;
 
+import com.google.common.collect.BoundType;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Range;
 import java.lang.reflect.Method;
@@ -8,6 +9,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
+import org.bukkit.Bukkit;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.event.entity.CreatureSpawnEvent.SpawnReason;
@@ -514,6 +516,18 @@ public abstract class FilterParser {
       countFilter = ParticipatingFilter.OBSERVING;
     }
 
-    return new PlayerCountFilter(XMLUtils.parseNumericRange(el, Integer.class), countFilter);
+    Range<Integer> range = XMLUtils.parseNumericRange(el, Integer.class);
+
+    if (range.hasLowerBound() && range.lowerEndpoint() <= 0) throw new InvalidXMLException("Lower limit must be at least 0", el);
+    if (range.hasUpperBound() && range.upperEndpoint() < 0) throw new InvalidXMLException("Upper limit must be greater than 0", el);
+
+    // Shrink the range to be at least 0 - the maximum possible number of players that can be online
+    // For example (assuming the max player count is 20):
+    // - If the supplied range is -inf...50, the result is 0...20
+    // - If the supplied range is 5...10, the result is 5...10
+    // - If the supplied range is -inf...inf, the result is 0...20
+    range = Range.open(0, Bukkit.getMaxPlayers()).intersection(range);
+
+    return new PlayerCountFilter(range, countFilter);
   }
 }
--- a/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
+++ b/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
@@ -14,6 +14,7 @@ import org.bukkit.util.Vector;
 import org.jdom2.Attribute;
 import org.jdom2.Element;
 import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.filter.query.PlayerQuery;
 import tc.oc.pgm.api.map.factory.MapFactory;
 import tc.oc.pgm.api.player.PlayerRelation;
 import tc.oc.pgm.classes.ClassModule;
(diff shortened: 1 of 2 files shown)


CoWinkKeyDinkInc:
>The filter now supports a child filter to determine which players should be counter.

Does this mean that this filter would match for a certain number of players within a region? This would come really helpful to a map that I am trying to make

AustinLMayes:
> > The filter now supports a child filter to determine which players should be counter.
> 
> Does this mean that this filter would match for a certain number of players within a region? This would come really helpful to a map that I am trying to make

It accepts location filters, yes.

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -443,4 +444,30 @@ public StructuralLoadFilter parseStructuralLoad(Element el) throws InvalidXMLExc
...
   }
+
+  @MethodParser("players")
+  public PlayerCountFilter parsePlayerCountFilter(Element el) throws InvalidXMLException {
+    Filter countFilter = parseFilterProperty(el, "count-filter", StaticFilter.ALLOW);
+
+    // TODO: Legacy syntax, maybe bump proto to deprecate
+    boolean participants = XMLUtils.parseBoolean(el.getAttribute("participants"), true);
+    boolean observers = XMLUtils.parseBoolean(el.getAttribute("observers"), false);
+    countFilter =
```
Pablete1234: use `ParticipatingFilter.PARTICIPATING` or `ParticipatingFilter.OBSERVING` instead of this annonymous filter

core/src/main/java/tc/oc/pgm/filters/PlayerCountFilter.java
```diff
@@ -0,0 +1,31 @@
...
+    return MatchQuery.class;
+  }
+
+  @Override
+  public QueryResponse queryTyped(MatchQuery query) {
+    int count = 0;
+    for (MatchPlayer player : query.getMatch().getPlayers()) {
+      if (this.countFilter.query(player.getQuery()) == QueryResponse.ALLOW) count++;
+    }
+    return QueryResponse.fromBoolean(range.contains(count));
```
Pablete1234: I'd consider some cases where you can short-circuit.

With both the lower bound, and the upper bound, you can break after `n` players allow, or if `remaining + count < n` (it either won't reach minimum, or it can't overflow the maximum).

Some examples:
 - range min=1, count==1 -> can allow
 - range min=10, 5 players remaining to check, count == 4 -> can disallow
 - range max=10, count==11 -> can disallow
 - range max=10, 5 players remaining to check, count == 4 -> can allow

(keep in mind that you need both min and max range to match of the break conditions, or be unbounded)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/filters/PlayerCountFilter.java
```diff
@@ -0,0 +1,49 @@
...
+  @Override
+  public QueryResponse queryTyped(MatchQuery query) {
+    int count = 0;
+    Collection<MatchPlayer> allPlayers = query.getMatch().getPlayers();
+    int total = allPlayers.size();
+    boolean hasUpperBound = range.hasUpperBound();
+    boolean hasLowerBound = range.hasLowerBound();
+
+    // Not even enough players to consider checking
+    if (hasLowerBound && range.lowerEndpoint() > total) return QueryResponse.DENY;
```
Pablete1234, line 33: This can happen for both lower and upper bound.

if you have a range from -inf to 5, and only 3 players online, you know you'll always match

core/src/main/java/tc/oc/pgm/filters/PlayerCountFilter.java
```diff
@@ -0,0 +1,49 @@
...
+
+    for (MatchPlayer player : allPlayers) {
+      total--;
+      if (this.countFilter.query(player.getQuery()) == QueryResponse.ALLOW) {
```
Pablete1234, line 37: Even if failing, you can run the short-circuiting code too, right now you're only short circuiting after an allow

core/src/main/java/tc/oc/pgm/filters/PlayerCountFilter.java
```diff
@@ -0,0 +1,49 @@
...
+        count++;
+        // Too high, give up
+        if (hasUpperBound && range.upperEndpoint() < count) return QueryResponse.DENY;
+
+        // In the range - even if every other player passes the filter we won't go out of bounds
+        if (hasUpperBound && range.contains(count) && count + total <= range.upperEndpoint())
+          return QueryResponse.ALLOW;
+      }
```
Pablete1234: What if there's no upper bound? you can short-circuit that too

Pablete1234 (changes requested):
I feel like the short-circuiting logic is a bit iffy, needs to be a bit more solid.

To short-circuit, one of 3 conditions needs to happen, for BOTH min AND max:
 - Already reached (eg, min/max=3 && count=4)
 - Will never reach (eg, min/max=10, count=2 & remaining=7)
 - It's unbound

If both lower & upper allow short-circuit (they meet one of the 3 above criteria) then you can short-circuit. 
Once short-circuited, you simply return what the current count says (allow or deny), you effectively have assumed that increasing count will not move it in or out of the range

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -448,4 +450,38 @@ public TimeFilter parseTimeFilter(Element el) throws InvalidXMLException {
...
+      throw new InvalidXMLException("Lower limit must be at least 0", el);
+    if (range.hasUpperBound() && range.upperEndpoint() < 0)
+      throw new InvalidXMLException("Upper limit must be greater than 0", el);
+
+    // Shrink the range to be at least 0 - the maximum possible number of players that can be online
+    // For example (assuming the max player count is 20):
+    // - If the supplied range is -inf...50, the result is 0...20
+    // - If the supplied range is 5...10, the result is 5...10
+    // - If the supplied range is -inf...inf, the result is 0...20
+    range = Range.open(0, Bukkit.getMaxPlayers()).intersection(range);
```
Pablete1234, line 528: What's the point of doing this?

PR #702, opened 2020-11-14
End single-player blitz matches on start

AustinLMayes:
There's no way for an external non-staff player to end the match if the only playing player is AFK.


Pablete1234:
Wouldn't it be smarter to not start blitz matches with a single player, unless forced?

Early ending the match means it's impossible to test a blitz map without 2 players

Electroid:
Other points are good ones, going to pass on this one.

PR #705, opened 2020-11-14
Sanity check time limits

AustinLMayes:
Prevents negative time limits and time limits that could overflow some different features.

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
+++ b/util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
@@ -499,7 +499,7 @@ public final class XMLUtils {
     try {
       return TextParser.parseDuration(node.getValueNormalize(), validRange);
     } catch (TextException e) {
-      throw new InvalidXMLException("invalid time format: " + e.getLocalizedMessage(), node);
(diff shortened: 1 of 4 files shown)


util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
```diff
@@ -482,17 +482,22 @@ public static Boolean parseBoolean(@Nullable Attribute attr, Boolean def)
...
+      throws InvalidXMLException {
     if (node == null) {
       return def;
     }
     try {
-      return TextParser.parseDuration(node.getValueNormalize());
+      return TextParser.parseDuration(node.getValueNormalize(), validRange);
     } catch (TextException e) {
-      throw new InvalidXMLException("invalid time format", node, e);
+      throw new InvalidXMLException("invalid time format: " + e.getLocalizedMessage(), node);
```
Electroid: `getLocalizedMessage` will return the `en_US` translation. If you pass the exception through, I believe that formater will properly translate it.

Electroid (changes requested):

Pablete1234 (approved):

PR #706, opened 2020-11-15
Ensure blitz life count is at least 1

AustinLMayes:
Just in case 😉


core/src/main/java/tc/oc/pgm/blitz/BlitzModule.java
```diff
@@ -47,6 +47,9 @@ public BlitzModule parse(MapFactory factory, Logger logger, Document doc)
       for (Element blitzEl : blitzElements) {
         boolean broadcastLives = XMLUtils.parseBoolean(blitzEl.getChild("broadcastLives"), true);
         int lives = XMLUtils.parseNumber(Node.fromChildOrAttr(blitzEl, "lives"), Integer.class, 1);
```
Electroid: Instead of adding explicit checks, you can change the signature of this instance of `parseNumber` and provide a `Range`. See `XMLUtils#parseNumber`

Electroid (changes requested):

Electroid (approved):

PR #710, opened 2020-11-15
Fixup Party min/max

AustinLMayes:
Ensure party max is always at least or greater than the minimum needed players.


core/src/main/java/tc/oc/pgm/teams/TeamModule.java
```diff
@@ -135,6 +135,10 @@ private static TeamFactory parseTeamDefinition(Element el, MapFactory factory)
       throw new InvalidXMLException("Max overfill can not be less then max players.", el);
     }
 
+    if (maxPlayers <= minPlayers) {
+      throw new InvalidXMLException("Min players can not be less (or equal) to max players.", el);
```
Pablete1234, line 139: Min can definetly be equal to max.
If you want a map to play 8v8, with no more and no less players, you set both min and max to 8 in each team

core/src/main/java/tc/oc/pgm/ffa/FreeForAllModule.java
```diff
@@ -86,6 +86,15 @@ public FreeForAllModule parse(MapFactory factory, Logger logger, Document doc)
           colors = XMLUtils.parseBoolean(Node.fromAttr(elPlayers, "colors"), colors);
         }
 
+        if (minPlayers > maxPlayers) {
+          if (elPlayers != null)
+            throw new InvalidXMLException("\"max\" must be greater than \"min\"", elPlayers);
```
Pablete1234, line 91: max can also be equal to min, not necessarily greater

core/src/main/java/tc/oc/pgm/ffa/FreeForAllModule.java
```diff
@@ -86,6 +86,15 @@ public FreeForAllModule parse(MapFactory factory, Logger logger, Document doc)
...
+          else {
+            if (maxPlayers > 1) minPlayers = maxPlayers / 2;
```
Pablete1234, line 93: I belive it makes more sense to force minPlayers to maxPlayers (bukkit max) if it goes over the limit. That's the biggest minimum that will be able to be met

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/ffa/FreeForAllModule.java
Electroid: Same, PGM should not care about `Bukkit#getMaxPlayers` anywhere.

Electroid:
Closing, this should be done with XML parsing ranges.

PR #711, opened 2020-11-15
Fix some ItemModify strangeness

AustinLMayes:
Applies item mods to picked up arrows and fixes unmodified items being marked as modified.


core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
```diff
@@ -31,9 +31,9 @@ public boolean applyRules(ItemStack stack) {
     if (stack == null || stack.getType() == Material.AIR || APPLIED.has(stack)) {
       return false;
     } else {
-      APPLIED.set(stack, true);
       for (ItemRule rule : rules) {
         if (rule.matches(stack)) {
+          APPLIED.set(stack, true);
```
Pablete1234, line 36: This doesn't seem like a good change.
Applied should be set regardless of if any rule applies to the item, so that it is not checked again. Also, if multiple rules match you're applying setting the applied nbt tag multiple times for no good reason

PR #712, opened 2020-11-15
Spawn block drops in the center instead of the corner

AustinLMayes:
Matches the behavior of internal block drop mechanics.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
@@ -85,7 +85,7 @@ public class BlockDropsMatchModule implements MatchModule, Listener {
     Random random = match.getRandom();
     for (Map.Entry<ItemStack, Double> entry : drops.items.entrySet()) {
       if (random.nextFloat() < yield * entry.getValue()) {
-        location.getWorld().dropItemNaturally(location.toCenterLocation(), entry.getKey());
+        location.getWorld().dropItemNaturally(location, entry.getKey());
       }
     }
   }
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
```diff
@@ -85,15 +85,18 @@ private void dropItems(BlockDrops drops, Location location, double yield) {
     Random random = match.getRandom();
     for (Map.Entry<ItemStack, Double> entry : drops.items.entrySet()) {
       if (random.nextFloat() < yield * entry.getValue()) {
-        location.getWorld().dropItemNaturally(location, entry.getKey());
+        location.getWorld().dropItemNaturally(location.toCenterLocation(), entry.getKey());
```
Electroid: I feel like the `toCenterLocation` should be done before calling `dropItems`, same with `dropExperience`

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
```diff
@@ -93,7 +93,10 @@ private void dropItems(BlockDrops drops, Location location, double yield) {
   private void dropExperience(BlockDrops drops, Location location) {
     if (drops.experience != 0) {
       ExperienceOrb expOrb =
-          (ExperienceOrb) location.getWorld().spawnEntity(location, EntityType.EXPERIENCE_ORB);
+          (ExperienceOrb)
+              location
+                  .getWorld()
+                  .spawnEntity(location.toCenterLocation(), EntityType.EXPERIENCE_ORB);
```
Electroid: This should not call `toCenterLocation`

PR #713, opened 2020-11-15
Re-introduce filtered respawn options

AustinLMayes:
Might eventually bump proto to get rid of the legacy stuff.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
@@ -9,11 +9,14 @@ import java.util.logging.Logger;
 import net.kyori.text.Component;
 import org.jdom2.Document;
 import org.jdom2.Element;
+import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.map.MapModule;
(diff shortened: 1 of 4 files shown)


Electroid (approved):

core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
```diff
@@ -68,29 +74,46 @@ public SpawnModule parse(MapFactory factory, Logger logger, Document doc)
...
+      for (Element elRespawn :
+          XMLUtils.flattenElements(doc.getRootElement(), "respawns", "respawn")) {
+        respawnOptions.add(parseOptions(elRespawn, true, false));
+      }
+
+      respawnOptions.add(
+          parseOptions(
+              doc.getRootElement().getChild("respawn"),
+              false,
+              doc.getRootElement().getChild("autorespawn") == null));
```
Pablete1234, line 88: This is incorrect, there is not a single respawn child, there's N several respawn children that all generate a default respawn option. 

Consider this code the correct one:

https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/spawns/SpawnModule.java#L106

PR #714, opened 2020-11-15
Allow owned goals to operate outside of team-based environments

AustinLMayes:
Re-adds support for flags in FFA environments

Null checks were added just in case implementations instantiate features programmatically.


Electroid (approved):
This is hacky, not sure I'm fully supportive of this approach (I'd rather just switch the `Goal` system to only use `Competitor`s).

That said, FFA CTF is cool and want to enable that.

PR #715, opened 2020-11-15
Add some potion creation XML helpers

AustinLMayes:
Adds some helpers to make creation potions easier
`splash` - If the potion should be a splash potion
`display-type` - Type of potion the stack should be seen as, regardless of effects

Creates a splash potion of poison that actually gives invisibility. 
```xml
<item slot="1" name="name" potions="invisibility:1000:1:false" material="potion" splash="true" display-type="poison">
    <effect duration="50s" amplifier="2">invisibility</effect>
</item>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
+++ b/core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
@@ -4,7 +4,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
-import org.bukkit.Bukkit;
 import org.bukkit.Material;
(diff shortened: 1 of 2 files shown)


CoWinkKeyDinkInc:
I've always been confused by the `potions` attribute in items. I think all potion related configuration should be moved to the `<effect>` tag. Other than that I think this is a great fix. I already written the docs page to accommodate for the `damage` value, but when this is merged I can update this.

AustinLMayes:
> I've always been confused by the `potions` attribute in items. I think all potion related configuration should be moved to the `<effect>` tag. Other than that I think this is a great fix. I already written the docs page to accommodate for the `damage` value, but when this is merged I can update this.

Basically `potions` is the same as using sub-tags. It's disgusting yea but a lot of old maps use it so a proto update would be needed to get rid of it.

PR #720, opened 2020-11-21
Fix previews for non-player inventory holders

BGMP:
When observers right click the players of a match, they are shown a preview of the their current inventory, and this preview is kept updated so that if the inventory changes, the preview will also reflect these changes. This behaviour was not being applied, however, to other inventory holders such as chests, hoppers, or anything that was not a player, and observers would have to re-right click the container to update the contents of their preview.

This commit makes non-player inventory holders's previews update lively, just like when previewing a player's inventory.


applenick (approved):
Tested this with BGM on my dev server 👍

Electroid (approved):

PR #722, opened 2020-11-21
Support for Community

applenick:
# Community

## Update: July 10, 2022
A complete redesign of Community is in the works! Goal will be to clean up the codebase and reduce the complexity that has arisen from countless changes being applied overtime. I foresee it being done by the end of 2022 📆 Perhaps then we'll be able to get this merged!

## Update: December 4th, 2021
Wow! This branch has been in-dev for over a year 📆. Unfortunately, due to the dire need of the contained features on [Overcast Community](https://oc.tc) , this PR has become rather messy over time. Massive refactoring will have to take place, likely introducing little features here and there until complete. That being said, this branch of PGM is the primary one in use on [Overcast Community](https://oc.tc) and is generally stable. If you wish to run this version of PGM follow these steps:

1.  Clone the [Community](https://github.com/applenick/PGM.git)  branch of PGM
2. Build the community branch with `mvn clean install`

If you have any questions please feel free to ask in the [PGMDev](https://discord.gg/F4WPvmgqvS) discord. 

⭐ Learn more about [Community](https://github.com/PGMDev/Community) ⭐ 
 
---
**The following information may be outdated, but is preserved from the original PR in November 2020**

## Overview
The long awaited Community plugin is almost ready! In order for it to function properly, I've made a number of PGM changes required to support it. I believe most of these changes should be straightforward and make sense in the long term.

Due to this PR containing a large number of changes, I'll highlight the two most notable ones below:

## Removal of Moderation Commands & Reports :warning:
Preparing for the release of Community, we are removing most moderation commands from PGM. The exception to this will be freeze & vanish, which have been moved to new homes within PGM. Vanish to `tc.oc.pgm.vanish` and has been given a configuration option, while `FreezeMatchModule` is now located with other match modules. The `PlayerVanishEvent` has been moved as well (in case any third party plugins were listening to this event). Some commands such as `/staff` which was located inside of the moderation commands class have been saved, as they are still relevant and fit nicely with PGM.

Reports have also been removed from PGM and will now be located in the Community plugin.

When moderation commands were originally added to PGM earlier this year, it was always meant as a temporary solution. Now with Community, we have a dedicated plugin that will work together with PGM and allow us to simplify things and properly handle scope. This is why moderation commands & reports have been completely removed instead of working to keep a default implementation.  

## Friend Support :heart:
With this PR comes the return of friends to PGM. The main features related to friends include
- friend only death messages
- friend only join messages
- friend only visible observers
- italicized names in the tablist

No friend commands or persistence of friends will be located in PGM. External plugins (such as Community) will make use of the new `FriendProvider` api in order to let PGM know who your friends are. I believe this keeps friends within scope of PGM and will be an appreciated feature to all users. 

## Feedback
As always, I welcome feedback of any kind and will work to accommodate suggestions when possible. Within the coming days these changes will be coming to OCC to act as a showcase of what the Community plugin can do.

## Repo
Visit the [Community](https://github.com/PGMDev/Community) repo to learn more about the project


core/src/main/java/tc/oc/pgm/command/FreezeCommand.java
```diff
@@ -0,0 +1,79 @@
...
+    FreezeMatchModule fmm = match.getModule(FreezeMatchModule.class);
+    MatchPlayer player = match.getPlayer(target);
+    if (player != null) {
+      fmm.setFrozen(sender, player, !fmm.isFrozen(player), silent);
+    }
+  }
+
+  private Component formatFrozenList(String key, int count, Component names) {
+    return TranslatableComponent.of(
+        key, TextColor.GRAY, TextComponent.of(Integer.toString(count), TextColor.AQUA), names);
```
KingOfSquares: Dont need the `Integer.toString`

core/src/main/java/tc/oc/pgm/api/setting/SettingKey.java
```diff
@@ -17,14 +17,21 @@
 public enum SettingKey {
   CHAT("chat", CHAT_TEAM, CHAT_GLOBAL, CHAT_ADMIN), // Changes the default chat channel
   DEATH(
-      Arrays.asList("death", "dms"), DEATH_ALL, DEATH_OWN), // Changes which death messages are seen
+      Arrays.asList("death", "dms"),
+      DEATH_ALL,
+      DEATH_OWN,
+      DEATH_FRIENDS), // Changes which death messages are seen
```
KingOfSquares, line 33: Following the logic from the other settings this should probably be ALL, FRIENDS, OWN

util/src/main/java/tc/oc/pgm/util/friends/FriendStatusChangeEvent.java
```diff
@@ -0,0 +1,30 @@
+package tc.oc.pgm.util.friends;
+
+import java.util.UUID;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+/** Called when friendship status is updated for a player * */
+public class FriendStatusChangeEvent extends Event {
+
```
KingOfSquares: This seems a little confusing, does a friendship not always include 2 parties? Why not include both parties here instead of throwing two events.

core/src/main/java/tc/oc/pgm/friends/FriendRegistry.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.bukkit.entity.Player;
+import org.bukkit.event.Listener;
+import tc.oc.pgm.util.friends.FriendProvider;
+
+/**
+ * The FriendRegistry is responsible for managing a {@link FriendProvider} to consistently supply
+ * whether two players are friends
+ */
+public interface FriendRegistry extends Listener, FriendProvider {
```
KingOfSquares: Seems weird to have these friend things in PGM while not being able to interact with it without having a separate
plugin. All things useable with the friend feature seems to care if to parties is related in a certain way. Why not have the settings have an option to add custom relation checks in a way. This way other plugins could use their own groups as well (based on ranks, teams, staff...)

KingOfSquares: I thought a little about this, and I think the best way would be to have a way to register some `PartyRelation`.

This way whenever the settings that could care about parties is queried they can look for any custom `PartyRelation`s in addition to the hardcoded options.

I believe this solves the support for friends, while not locking the possibility to only friends.

applenick: @KingOfSquares 
I like your thinking on this topic. I was initially unsure when implementing this how specific to get. 

I think the settings area of PGM is in dire need of a comprehensive update, cleanup to the commands and perhaps less reliance on hard coded values. That was not something I was planning to include in this PR, but could very well be a future project.

I’d like some other feedback on whether switching everything to `PartyRelation` would be the right way to go. If you had any ideas in regards to the specifics though, let me know! 

Atm, I’m not sure what other sort of groups would use the same features as friends though? It may be confusing if more than one plugin utilized the same formatting of italicizing friend names, I feel making it specific reserves that formatting consistency. But I understand the desire to make it more generic. 

Thanks again for the feedback! 😄

core/src/main/java/tc/oc/pgm/command/FreezeCommand.java
```diff
@@ -0,0 +1,79 @@
...
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.modules.FreezeMatchModule;
+import tc.oc.pgm.util.chat.Audience;
+import tc.oc.pgm.util.named.NameStyle;
+
+public class FreezeCommand {
```
Electroid: I feel like freezing is a community thing?

core/src/main/java/tc/oc/pgm/command/ListCommand.java
```diff
@@ -59,6 +59,37 @@ public void list(CommandSender sender, Audience viewer, Match match) {
                 Integer.toString(getSize(match.getPlayers(), false)), TextColor.GREEN)));
   }
 
+  @Command(
```
Electroid: This command is most definitely for community.

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/command/FreezeCommand.java
applenick: I have to check when I have a chance and see if it can be safely moved over, last I thought there was one PGM specific requirement needed. But I’ll get back to you on this 👍

core/src/main/java/tc/oc/pgm/command/ListCommand.java
applenick: Will do 👍

core/src/main/java/tc/oc/pgm/command/FreezeCommand.java
applenick: Was able to completely remove freeze from PGM, has been implemented in Community 😎

Electroid:
Going to hold this until we get the `refactor/adventure` stuff in, but excited that this is almost ready to go. Maybe you can rebase to that branch?

applenick:
@Electroid Excellent, yeah I’d rather wait for that to merge first too 👍 
Sure thing will start rebase to that branch. Last major change I’d like to add to this branch before getting it merged would be nick-name support. Will let you know once more progress has been made on that 😃If it’s too much can always split up into a different PR, but will see how many changes are required.

Electroid:
Let's rebase to `dev`

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -19,19 +19,22 @@
   CHAT_ADMIN("chat", "admin"), // Send to all server operators
 
   DEATH_OWN("death", "own"), // Only send death messages involving self
+  DEATH_FRIENDS("death", "friends"), // Only send death messages involving yourself or friends
```
Electroid: You need to add all of these new settings at the end to prevent messing up existing settings.

core/src/main/java/tc/oc/pgm/friends/FriendRegistry.java
```diff
@@ -0,0 +1,20 @@
...
+
+  default boolean areFriends(Player viewer, Player other) {
```
Electroid: Let's use `UUID`s here.

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
applenick: Oh shoot that's odd, thought I fixed all the settings. Will do 👍

core/src/main/java/tc/oc/pgm/friends/FriendRegistry.java
applenick: Sure thing!

applenick:
I've been continually rebasing this, perhaps it's time to get this merged 👍 
@Electroid if you have any suggestions let me know and I'll fix it asap, otherwise should be fine. We've been running this branch on OCC for over a month and no issues.

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -90,6 +92,8 @@
   private ScheduledExecutorService executorService;
   private ScheduledExecutorService asyncExecutorService;
   private VanishManager vanishManager;
+  private FriendRegistry friendRegistry;
+  private NickRegistry nickRegistry;
```
Electroid: I've been thinking about the integration point between PGM and other plugins (like Community, et al.) 

Here are some of the things plugins may want to customize, that PGM needs to know the answer to:
- Is this player vanished?
- Is this player nicked? If so, what is their nickname?
- Are these players friends?
- What is this player's prefix/suffixes?
- What is this player's priority on the tab list?
- Does this player have X setting? If so, what is the value?

It want to keep this integration very simple, so I'm thinking of a simple `Integration` class and using Bukkit metadata under-the-hood to implement it.

```java
import tc.oc.pgm.api;

// This class is a simple helper class for other plugins to integrate with PGM.
// They can either import the PGM API, or simply copy-paste this file.
// Which means this class should not reference any PGM classes, only Bukkit.
class PGMIntegration {
   public static final String VANISH_KEY = "pgm-vanish";
   public static final String NICKNAME_KEY = "pgm-nick";

   PGMIntegration(Plugin owner) {
     this.owner = owner;
   }

   // Invoked by 3rd party integration
   public void setFriends(Player a, Player b) {
       a.setMetadata(VANISH_KEY + "-" + b.getUniqueId(), new FixedMetadataValue(owner, true));
   }

   public void setNickname(Player a, @Nullable String nickname) {
       a.setMetadata(NICKNAME_KEY, new FixedMetadataValue(owner, nickname));
   }

   // Checked by PGM
   public static boolean isFriends(UUID a, UUID b) {
       // ...
   }
}
```

I gave a few examples, but I think you can see where I'm going with this. I'd like to keep the integration code in 1 area, both so it's easy to show to other plugin devs how to integrate, but also for maintainability. 

(/cc @Pablete1234, @Indicardo, @applenick)

core/src/main/java/tc/oc/pgm/api/player/VanishManager.java
```diff
@@ -33,4 +34,9 @@
 
   /** Called when the vanish manager is disabled */
   void disable();
+
+  @Nonnull
+  VanishManager getManager();
+
+  void setManager(VanishManager manager);
```
Electroid: See my `Integration` proposal. It doesn't make to set the manager, from the manager.

core/src/main/resources/config.yml
```diff
@@ -96,6 +96,10 @@ tablist:
   left: ""
   right: ""
 
+# Enables vanish features such as /vanish and auto vanish joining with 'vanish.<server.ip>'
+vanish:
```
Electroid, line 104: I think this should be Community, see integration proposal.

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/PGMPlugin.java
applenick: @Electroid 
`VANISH_KEY` for `setFriends` is a just a typo right? 

The idea you'd like to see would look something along the lines of this if I'm not mistaken :
```java
public static final String FRIEND_KEY = "pgm-friend";

public void setFriends(Player a, Player b) {
  a.setMetadata(FRIEND_KEY + "-" + b.getUniqueId(), new FixedMetadataValue(owner, true));
}
```

Then `isFriends(...)` will basically check if the users have the matching meta `FRIEND_KEY-UUID`

Electroid: Yes, I made some typos, it's pseudo-code. But that's the idea.

applenick: Just making sure 😄 
I think it'll work well, will get started on adjusting things soon and let you know!

integration/src/main/java/tc/oc/pgm/integration/FriendIntegration.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public static final String FRIEND_KEY = "pgm-friend";
+
+  default void setFriend(Player player, Set<UUID> playerIds) {
+    player.setMetadata(FRIEND_KEY, new FixedMetadataValue(getOwner(), playerIds));
+  }
+
+  static boolean isFriend(Player a, Player b) {
+    if (a.hasMetadata(FRIEND_KEY)) {
+      Set<UUID> playerIds = (Set<UUID>) a.getMetadata(FRIEND_KEY).get(0).value();
```
Electroid: Let's cast as arbitrary as possible, `Collection<UUID>`

core/src/main/java/tc/oc/pgm/listeners/VanishListener.java
```diff
@@ -0,0 +1,53 @@
...
+import tc.oc.pgm.listeners.JoinLeaveAnnouncer.JoinVisibility;
+
+public class VanishListener implements Listener {
+
+  @EventHandler(priority = EventPriority.NORMAL)
+  public void setVanishState(PlayerVanishEvent event) {
+    MatchPlayer player = event.getPlayer();
+
+    // Ensure player is an observer
+    event.getMatch().setParty(player, event.getMatch().getDefaultParty());
```
Pablete1234, line 20: This is a tricky one, but you may not be allowed to join obs (eg: death tracker says you're in danger) which means you'll vanish while on blue/red team.

Could make it so if you do not change party it doesn't set vanished, but the message that you left has already been sent.

Making this switch before sending the message is also not possible because then you'd always appear as in obs in the leave message when vanishing.

It's a tricky one

applenick: Had not thought of this! Good observation. 

Indeed, very tricky.

Could we perhaps modify `CombatLogTracker#getImminentDeath` so we can call that from `VanishCommand` and just do a quick check before vanishing? 

Or an alternative, maybe split up vanishing into multiple events. `PrePlayerVanishEvent` Allow for us to cancel the vanish before the leave messages are shown.

core/pom.xml
```diff
@@ -67,10 +67,6 @@
                             <pattern>app.ashcon.intake</pattern>
                             <shadedPattern>tc.oc.pgm.lib.app.ashcon.intake</shadedPattern>
                         </relocation>
-                        <relocation>
-                            <pattern>net.kyori</pattern>
-                            <shadedPattern>tc.oc.pgm.lib.net.kyori</shadedPattern>
-                        </relocation>
```
Electroid: Why did this need to be changed?

core/src/main/java/tc/oc/pgm/api/Permissions.java
```diff
@@ -31,12 +31,10 @@
...
   String RELOAD = ROOT + ".reload"; // Reload the PGM configuration
-  String KICK = ROOT + ".kick"; // Access to the /kick command
-  String WARN = ROOT + ".warn"; // Access to the /warn command
-  String MUTE = ROOT + ".mute"; // Access to the /mute command
-  String BAN = ROOT + ".ban"; // Access to the /ban command
-  String FREEZE = ROOT + ".freeze"; // Access to the /freeze command
   String VANISH = ROOT + ".vanish"; // Access to /vanish command
+  String SKIN = ROOT + ".skin"; // Access to /skin command
+  String ADMIN = ROOT + ".admin"; // Access to admin features
+  String VOTE_MULTIPLIER = JOIN + ".votecount"; // Extra map voting multiplier
```
Electroid, line 37: I don't think this should be attached to join.

core/src/main/java/tc/oc/pgm/api/event/FriendStatusChangeEvent.java
```diff
@@ -0,0 +1,30 @@
+package tc.oc.pgm.api.event;
+
+import java.util.UUID;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+/** Called when friendship status is updated for a player * */
+public class FriendStatusChangeEvent extends Event {
```
Electroid, line 8: Ideally, this class is defined by `Community`. Why does PGM need this?

core/src/main/java/tc/oc/pgm/api/event/PlayerVoteEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.UUID;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.map.MapInfo;
+
+public class PlayerVoteEvent extends Event {
+
+  private UUID playerId;
+  private MapInfo map;
+  private boolean add;
```
Electroid, line 12: I would rename this to `choice`, and `getChoice`.

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
```diff
@@ -0,0 +1,105 @@
...
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.integration.FriendIntegrationImpl;
+import tc.oc.pgm.integration.NickIntegrationImpl;
+import tc.oc.pgm.integration.PunishmentIntegrationImpl;
+import tc.oc.pgm.integration.RequestIntegrationImpl;
+import tc.oc.pgm.integration.TranslationIntegrationImpl;
+import tc.oc.pgm.integration.VanishIntegrationImpl;
+import tc.oc.pgm.util.translation.Translation;
+
+public interface Integration {
```
Electroid, line 19: Let's change this to a `final class` with a private, empty constructor.

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
```diff
@@ -0,0 +1,105 @@
...
+  static final AtomicReference<VanishIntegration> VANISH =
+      new AtomicReference<VanishIntegration>(new VanishIntegrationImpl());
+  static final AtomicReference<PunishmentIntegration> PUNISHMENTS =
+      new AtomicReference<PunishmentIntegration>(new PunishmentIntegrationImpl());
+  static final AtomicReference<RequestIntegration> REQUESTS =
+      new AtomicReference<RequestIntegration>(new RequestIntegrationImpl());
+  static final AtomicReference<TranslationIntegration> TRANSLATIONS =
+      new AtomicReference<TranslationIntegration>(new TranslationIntegrationImpl());
+
+  public static void setFriendIntegration(FriendIntegration integration) {
```
Electroid, line 34: These methods should to a `requireNonNull` check.

core/src/main/java/tc/oc/pgm/api/integration/NickIntegration.java
```diff
@@ -0,0 +1,10 @@
+package tc.oc.pgm.api.integration;
+
+import javax.annotation.Nullable;
+import org.bukkit.entity.Player;
+
+public interface NickIntegration {
+
+  @Nullable
+  String getNick(Player player);
```
Electroid, line 9: `UUID` instead of `Player`

core/src/main/java/tc/oc/pgm/api/integration/RequestIntegration.java
```diff
@@ -0,0 +1,12 @@
+package tc.oc.pgm.api.integration;
+
+import java.util.List;
+import net.kyori.adventure.text.Component;
+import tc.oc.pgm.api.map.MapInfo;
+
+public interface RequestIntegration {
+
+  public List<Component> getExtraMatchInfoLines(MapInfo map);
```
Electroid, line 9: What is this used for?

core/src/main/java/tc/oc/pgm/api/integration/NickIntegration.java
Electroid: This is because we should probably pre-fetch the nick on async login, when `Player` does not exist yet.

core/src/main/java/tc/oc/pgm/api/player/event/MatchPlayerFullJoinEvent.java
```diff
@@ -0,0 +1,23 @@
+package tc.oc.pgm.api.player.event;
+
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+/** MatchPlayerFullJoinEvent - Called when a {@link MatchPlayer} attempts to join a full match * */
+public class MatchPlayerFullJoinEvent extends MatchPlayerEvent {
```
Electroid, line 7: Why is this event needed?

core/src/main/java/tc/oc/pgm/command/SkinCommand.java
```diff
@@ -0,0 +1,79 @@
...
+import app.ashcon.intake.parametric.annotation.Switch;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.bukkit.Skin;
+import org.bukkit.entity.Player;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.Audience;
+
+public final class SkinCommand {
```
Electroid: I don't think we should have these commands in PGM.

core/src/main/java/tc/oc/pgm/events/MapVoteWinnerEvent.java
```diff
@@ -0,0 +1,29 @@
+package tc.oc.pgm.events;
+
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.map.MapInfo;
+
+public class MapVoteWinnerEvent extends Event {
```
Electroid, line 7: Why is this event needed? Also, could this be a `MatchEvent`?

core/src/main/java/tc/oc/pgm/integration/PunishmentIntegrationImpl.java
```diff
@@ -0,0 +1,23 @@
...
+public class PunishmentIntegrationImpl implements PunishmentIntegration {
+
+  @Override
+  public boolean isMuted(Player player) {
+    return false;
+  }
+
+  @Override
+  public @Nullable String getMuteReason(Player player) {
+    return "";
```
Electroid, line 16: Why not return `null` here?

core/src/main/java/tc/oc/pgm/integration/RequestIntegrationImpl.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.List;
+import net.kyori.adventure.text.Component;
+import tc.oc.pgm.api.integration.RequestIntegration;
+import tc.oc.pgm.api.map.MapInfo;
+
+public class RequestIntegrationImpl implements RequestIntegration {
+
+  @Override
+  public List<Component> getExtraMatchInfoLines(MapInfo map) {
+    return Lists.newArrayList();
```
Electroid, line 13: Use `Collections.emptyList`

core/src/main/java/tc/oc/pgm/integration/TranslationIntegrationImpl.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import org.bukkit.entity.Player;
+import tc.oc.pgm.api.integration.TranslationIntegration;
+import tc.oc.pgm.util.translation.Translation;
+
+public class TranslationIntegrationImpl implements TranslationIntegration {
+
+  @Override
+  public CompletableFuture<Translation> translate(Player sender, String message) {
```
Electroid: Is there a reason we need to use `CompletableFuture`s here?

core/src/main/java/tc/oc/pgm/listeners/SkinCache.java
```diff
@@ -0,0 +1,146 @@
...
+  public void onPlayerJoin(PlayerJoinEvent event) {
+    offlineSkins.invalidate(event.getPlayer().getUniqueId());
+  }
+
+  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+  public void refreshNamesOnLogin(PlayerJoinEvent event) {
+    refreshPlayer(event.getPlayer());
+  }
+
+  // SPORTPAPER STUFF - TODO: Add alternative method and check if server is running SportPaper to
```
Electroid, line 73: What happens on Spigot? Let's make sure it will safely fail.

Electroid (commented):
Left some comments, didn't go through it all.

core/src/main/java/tc/oc/pgm/api/event/FriendStatusChangeEvent.java
Pablete1234: PGM needs the event to be able to react to state changes (eg: your tablist view re-rendering after you add someone as friend), the event is only launched externally tho. It's a similar mechanism to player prefixes/suffixes, if you have a change in state you should launch the event and pgm will re-render those names in tablist of make any updates that are needed to reflect the new prefix/suffix

core/src/main/java/tc/oc/pgm/api/integration/RequestIntegration.java
Pablete1234: My guess is sponsors "Sponsored by XXX" added at the end of the match info, shown when you log-in or when you get teleported into the map, the typical message with:
```
----- Map Name -----
Here goes the map objective in
light blue color

Extra lines, Sponsored by Player
---------------------
```

core/src/main/java/tc/oc/pgm/events/MapVoteWinnerEvent.java
Pablete1234: Probably used to return the sponsor token if your sponsored map wins the vote

core/src/main/java/tc/oc/pgm/integration/TranslationIntegrationImpl.java
Pablete1234: if Translations are going to be resolved in an external service they should be handled in a nonblocking way, you can't block the whole server while you wait for an external api to translate a message

core/src/main/java/tc/oc/pgm/listeners/SkinCache.java
Pablete1234: on spigot this functionality will simply not be supported as spigot doesn't have per-viewer fake display names and skins

If the methods don't exist this should be all just skipped

applenick:
## Update: A major refactor is now in progress 🚧 

Will be reviewing all previous changes and cleaning up to finally get this PR merged. Started by squashing all previous commits into a single one, so new changes will be easier to see.

Spoke with Electroid and we're going to try and remove many of the unnecessary changes that have accumulated over time. Call from community will no longer go through the `Integration` interface, instead opting to call methods directly. 

Expect to see some progress being made over the next few weeks! Setting a soft due-date of November 1st, hopefully we can get this done before the 3 year anniversary of this PR 🤣

core/src/main/java/tc/oc/pgm/api/match/event/MatchStatsEvent.java
```diff
@@ -1,19 +1,26 @@
...
+import tc.oc.pgm.stats.PlayerStats;
 
 public class MatchStatsEvent extends MatchEvent implements Cancellable {
 
   private boolean cancelled = false;
   private boolean showBest;
   private boolean showOwn;
 
-  public MatchStatsEvent(Match match, boolean showBest, boolean showOwn) {
+  private Map<UUID, PlayerStats> stats;
```
Pablete1234, line 16: This should not be in this event at all, stats can be obtained via getter on the stats module of the match. This event is expected to be thrown by 3rd parties to signal pgm to show the match-end stats in chat.

core/src/main/java/tc/oc/pgm/api/match/Match.java
```diff
@@ -407,6 +407,10 @@ default boolean finish() {
    */
   boolean calculateVictory();
 
+  boolean isFriendlyFire();
+
+  void setFriendlyFire(boolean enabled);
+
```
Pablete1234, line 413: imo this does not belong in match. MapInfo has a flag for it, if you want to modify it at runtime it should probably be part of whatever module handles it, not match itself. (maybe either scoreboard module since it handles teams, or some other damage-related module)

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -228,6 +223,9 @@ public void onEnable() {
       asyncExecutorService.scheduleAtFixedRate(new ShouldRestartTask(), 0, 1, TimeUnit.MINUTES);
     }
 
+    // Set name provider
+    PlayerComponentProvider.PROVIDER.set(new NameComponentProviderImpl());
```
Pablete1234, line 225: The whole of name rendering is absurdly messy in community branch, that needs to be sorted out. Needs proper `PlayerComponent`s which render to per-player view, but not via a text-replace

core/src/main/java/tc/oc/pgm/scoreboard/ScoreboardMatchModule.java
```diff
@@ -250,6 +250,14 @@ public void onPlayerChangeParty(PlayerPartyChangeEvent event) {
     changePlayerScoreboard(event.getPlayer(), event.getOldParty(), event.getNewParty());
   }
 
+  public void updateAll() {
```
Pablete1234, line 253: this screams performance issue, why is it needed?

core/src/main/java/tc/oc/pgm/integration/VanishIntegrationImpl.java
```diff
@@ -0,0 +1,18 @@
+package tc.oc.pgm.integration;
+
+import org.bukkit.entity.Player;
+import tc.oc.pgm.api.integration.VanishIntegration;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+public class VanishIntegrationImpl implements VanishIntegration {
```
Pablete1234, line 7: this needs a better than no-op default, even if community plugin isn't used vanish shouldn't just *not work*

applenick:
Closed in favor of #1111 🤞

PR #724, opened 2020-11-27
Add bonus-points module

CoWinkKeyDinkInc:
This adds a `bonus-points` module that can let control points give a set number of points when captured.
```xml
   <!-- Gives exactly 20 points when captured -->
    <control-point name="The Hill" id="pi-cp" capture-time="1s" bonus-points="20" points="0">
        <capture><region id="hill"/></capture>
        <progress><region id="hill"/></progress>
        <captured><region id="hill"/></captured>
    </control-point>

   <!-- Gives exactly 10 points when captured, then awards 1 point per second -->
    <control-point name="The Hill" id="pi-cp" capture-time="1s" bonus-points="10">
        <capture><region id="hill"/></capture>
        <progress><region id="hill"/></progress>
        <captured><region id="hill"/></captured>
    </control-point>
```
Still a work in progress because it gives the team the points the second they start capturing the hill, rather than when capturing but otherwise it still works.


Pablete1234:
Read https://github.com/PGMDev/PGM/issues/716#issuecomment-733680488 if you're implementing this feature

PR #725, opened 2020-11-27
Add owner-points module

### LINKED ISSUE #716: Reimplement "owner-points" for control points
My idea for "flat-points" is for a control point to give out a set number of points when captured, instead of a linear amount of points like the default 1 point per second.
Control points can use a permanent module making it unable to be captured by another team once the control point is captured, and flat points might help make permanent control points more useful, though flat-points could also be used in hills that can be always captured.

I was thinking something like this for the XML. I'm going to take a try and this one.

```xml
<control-points capture-rule="lead" incremental="true" show-progress="true">
    <control-point name="Center" capture-time="20s" permanent="true" flat-points="20">
        <capture><cylinder base="0.5,23,7.5" radius="7" height="5"/></capture>
        <progress><cylinder base="0.5,22,7.5" radius="7" height="4"/></progress>
        <captured><cylinder base="0.5,64,7.5" radius="7" height="5"/></captured>
    </control-point>
</control-points>
```

CoWinkKeyDinkInc:
This adds a `owner-points` module that can let control points give a set number of points when captured.
```xml
    <!-- Gives exactly 20 points when captured, removes 20 points when control point lost -->
    <control-point name="The Hill" id="hill" capture-time="1s" owner-points="20" points="0">
        <capture><region id="hill"/></capture>
        <progress><region id="hill"/></progress>
        <captured><region id="hill"/></captured>
    </control-point>

    <!-- Holds 10 points when captured, then awards 1 point per second -->
    <control-point name="The Hill" id="second-hill" capture-time="1s" owner-points="10">
        <capture><region id="hill-2"/></capture>
        <progress><region id="hill-2"/></progress>
        <captured><region id="hill-2"/></captured>
    </control-point>

    <!-- Holds 40 points when captured, then awards 3 points per second -->
    <control-point name="The Hill" id="third-hill" capture-time="1s" owner-points="40" points="3">
        <capture><region id="hill-3"/></capture>
        <progress><region id="hill-3"/></progress>
        <captured><region id="hill-3"/></captured>
    </control-point>
```
Still a work in progress because it gives the team the points the moment they start capturing the hill, rather than when capturing but otherwise it still works, just needs another case added into it but what I've tried so far didn't work. I'll check the code @Pablete1234 was talking about later. Fixes #716

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
@@ -355,8 +355,8 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
           new ControllerChangeEvent(this.match, this, oldControllingTeam, this.controllingTeam));
 
       ScoreMatchModule scoreMatchModule = this.getMatch().getModule(ScoreMatchModule.class);
-      // Gives a set number of owner points to a team when captured.
-      if (this.controllingTeam != null && scoreMatchModule != null) {
+      // Gives a set number of owner points to a team when captured, lost when an enemy captures
+      if (scoreMatchModule != null) {
         if (oldControllingTeam != null) {
           scoreMatchModule.incrementScore(
               oldControllingTeam, getDefinition().getPointsOwner() * -1);
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
@@ -348,17 +348,23 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
     if (oldCapturingTeam != this.capturingTeam) {
       this.match.callEvent(
           new CapturingTeamChangeEvent(this.match, this, oldCapturingTeam, this.capturingTeam));
-      ScoreMatchModule scoreMatchModule = this.getMatch().getModule(ScoreMatchModule.class);
-      // Gives a set number of Owner points to a team when captured.
-      if (this.capturingTeam != null && scoreMatchModule != null) {
-        scoreMatchModule.incrementScore(this.capturingTeam, this.getDefinition().getPointsOwner());
-      }
     }
 
     if (oldControllingTeam != this.controllingTeam) {
       this.match.callEvent(
           new ControllerChangeEvent(this.match, this, oldControllingTeam, this.controllingTeam));
 
+      ScoreMatchModule scoreMatchModule = this.getMatch().getModule(ScoreMatchModule.class);
+      // Gives a set number of owner points to a team when captured.
+      if (this.controllingTeam != null && scoreMatchModule != null) {
+        if (oldControllingTeam != null) {
+          scoreMatchModule.incrementScore(
+              oldControllingTeam, getDefinition().getPointsOwner() * -1);
+        }
+        if (this.controllingTeam != null) {
+          scoreMatchModule.incrementScore(this.controllingTeam, getDefinition().getPointsOwner());
+        }
+      }
       if (this.controllingTeam == null) {
         this.match.callEvent(new GoalCompleteEvent(this.match, this, oldControllingTeam, false));
       } else {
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
```diff
@@ -87,6 +87,8 @@ public static ControlPointDefinition parseControlPoint(
     boolean permanent = XMLUtils.parseBoolean(elControlPoint.getAttribute("permanent"), false);
     float pointsPerSecond =
         XMLUtils.parseNumber(elControlPoint.getAttribute("points"), Float.class, 1f);
+    float pointsBonus =
+        XMLUtils.parseNumber(elControlPoint.getAttribute("bonus-points"), Float.class, 1f);
```
KingOfSquares: The default should probably be 0

CoWinkKeyDinkInc: Thanks for catching that, at the time I did this a few weeks ago I didn't realize this sets the default value.

Pablete1234 (changes requested):
This property needs to be called owner points for compatibility with 1.9+ PGM

CoWinkKeyDinkInc:
> This property needs to be called owner points for compatibility with 1.9+ PGM

Would it be appropriate to have this as an alias? I think `bonus-points` is more clear for newer mapmakers.

Pablete1234:
I belive `bonus-points` is not only ambiguous as to what it's behaviour may be but also potentially confusing if you do not know at what point this "bonus" is applied.

Additionally, having multiple names for things is generally bad. It makes it so you have 2 XMLs which are functionally identical yet are different, this does happen across pgm but it is usually result of backwards compatibility (an oddly-named attribute is renamed, and the old is kept as an alias) however this isn't such a case since it's a new feature.

`owner-points` is pretty unambiguous as in, it gives points to the (current) owner of the point, and it's pretty hard to misinterpret that

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -348,6 +348,11 @@ private void dominateAndFireEvents(@Nullable Competitor dominantTeam, Duration d
     if (oldCapturingTeam != this.capturingTeam) {
       this.match.callEvent(
           new CapturingTeamChangeEvent(this.match, this, oldCapturingTeam, this.capturingTeam));
+      ScoreMatchModule scoreMatchModule = this.getMatch().getModule(ScoreMatchModule.class);
+      // Gives a set number of Owner points to a team when captured.
+      if (this.capturingTeam != null && scoreMatchModule != null) {
+        scoreMatchModule.incrementScore(this.capturingTeam, this.getDefinition().getPointsOwner());
+      }
```
Pablete1234: The points are added on capture but never removed when uncaptured, see:

https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java#L349-L353

Also i belive this code should be a bit lower, when the controlling team changes, not when the capturing team does

Pablete1234 (changes requested):

CoWinkKeyDinkInc:
The scoring system now works properly after moving the mechanic to the `if (oldControllingTeam != this.controllingTeam)` from the `if (oldCapturingTeam != this.capturingTeam)`. I am thinking of adding another XML option that will let the points stack, so you can choose between removing the enemy points on capture or not.

Pablete1234:
The original issue #716 mentioned itself that this was intended for permanent points that you won't uncap.

I belive the option of gaining points per each point you capture is a bit stupid gameplay-wise, it incentivizes LOSING your own point just so that you can earn more points?

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -354,6 +354,17 @@ private void dominateAndFireEvents(@Nullable Competitor dominantTeam, Duration d
       this.match.callEvent(
           new ControllerChangeEvent(this.match, this, oldControllingTeam, this.controllingTeam));
 
+      ScoreMatchModule scoreMatchModule = this.getMatch().getModule(ScoreMatchModule.class);
+      // Gives a set number of owner points to a team when captured.
+      if (this.controllingTeam != null && scoreMatchModule != null) {
```
```suggestion
      if (scoreMatchModule != null) {
```
If you do not apply this logic when the controlling team is null, capturing won't give points since the current controlling team is null. In general points with a netrual state won't have this feature working as intended

Pablete1234 (changes requested):

CoWinkKeyDinkInc:
Fair enough, I was thinking the same thing later on, I thought maybe there was some way that it could be used reasonably. I won't add it in here, consider this PR for the most part complete! :tada: Feels great to make a real pull request for this project.

CoWinkKeyDinkInc:
This should also have the reintroduce label added to it btw

Pablete1234 (approved):

Electroid (approved):
Awesome, excited to see this back!

PR #729, opened 2020-12-03
Menu fixes and refactors

KingOfSquares:
See commit messages for changelog


util/src/main/java/tc/oc/pgm/util/menu/pattern/Dynamic5RowMenuArranger.java
```diff
@@ -0,0 +1,233 @@
...
+
+public class Dynamic5RowMenuArranger extends MenuArranger {
+
+  @Override
+  public List<InventoryMenuItem> arrangeItems(List<InventoryMenuItem> itemsWithoutSpaces) {
+
+    // Figure out which slots to use
+    int[] slots = new int[Math.min(itemsWithoutSpaces.size(), 18)];
+
+    switch (slots.length) {
```
Electroid: Is there a better way to do this? This it totally nebulous.

KingOfSquares: It felt the most intuitive way to have a different display for each amount. I could always reduce the "customization" for less code?

Pablete1234: You can create a similar algorithm without hand-picked slots

Pugzy: This could just be my opinion but the checkerboard pattern for player heads just seems strange. I don't see any reason for it as it just increases the number of pages required to display a full team due to the excess padding.

image

I think the teams should use some logic to spread them evenly but like you say I'm sure some math can be done for this.

image

Electroid:
I'd prefer, as well, if you rebase this to `refactor/adventure`. Since it uses lots of `Component`

core/src/main/java/tc/oc/pgm/command/StatsCommand.java
```diff
@@ -21,21 +21,23 @@
       aliases = {"stats"},
       desc = "Show your stats for the match")
   public void stats(Audience audience, CommandSender sender, MatchPlayer player, Match match) {
+    if (player.getSettings().getValue(SettingKey.STATS).equals(SettingValue.STATS_ON)) {
+      throw TextException.of("match.stats.disabled");
```
Electroid: Why not just let them see their stats? My thought was if they disable stats it won't automatically display at the end of the match, but if they really want to see it, by all means let them.

util/src/main/java/tc/oc/pgm/util/menu/pattern/Dynamic5RowMenuArranger.java
Electroid: Seems like the algorithm would just start in the upper left hand corner and skip a slot?

core/src/main/java/tc/oc/pgm/command/StatsCommand.java
KingOfSquares: That makes sense, ill remove this code from the command

Electroid:
Rebase this to `dev`

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenu.java
```diff
@@ -0,0 +1,315 @@
...
+import com.google.common.collect.Lists;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+import net.kyori.adventure.text.Component;
+import net.md_5.bungee.api.ChatColor;
+import org.bukkit.Bukkit;
+import org.bukkit.Material;
+import org.bukkit.World;
+import org.bukkit.craftbukkit.v1_8_R3.util.WeakCollection;
```
Electroid: We can't depend on this because it's `nms`. Use a `WeakHashMap` with `Boolean` values.

core/src/main/java/tc/oc/pgm/observers/ObserverToolsMatchModule.java
```diff
@@ -80,14 +83,14 @@ public void onToolClick(PlayerInteractEvent event) {
       MatchPlayer player = match.getPlayer(event.getPlayer());
 
       if (item.getType().equals(TOOL_MATERIAL) && player != null && canUse(player)) {
-        this.toolItem.onInventoryClick(null, player, ClickType.RIGHT);
+        this.toolItem.onInventoryClick(null, player.getBukkit(), ClickType.RIGHT);
```
Pablete1234: Why not just `event.getPlayer()` ?

core/src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
```diff
@@ -32,37 +33,37 @@ public ChatColor getColor() {
...
     toggleObserverGameMode(player);
     menu.refreshWindow(player);
   }
 
-  public void toggleObserverGameMode(MatchPlayer player) {
+  public void toggleObserverGameMode(Player player) {
     player.setGameMode(getOppositeMode(player.getGameMode()));
     if (player.getGameMode() == GameMode.SPECTATOR) {
-      player.sendWarning(getToggleMessage());
+      PGM.get().getMatchManager().getPlayer(player).sendWarning(getToggleMessage());
```
Pablete1234: Is there no simpler way to get an audience so you can call send warning? Kinda liked how most of the menu stuff no longer relies on MatchPlayer at all

core/src/main/java/tc/oc/pgm/observers/tools/GamemodeTool.java
```diff
@@ -32,37 +33,37 @@ public ChatColor getColor() {
...
     } else if (isCreative(player)) {
       // Note: When WorldEdit is present, this executes a command to ensure the player is not stuck
       if (Bukkit.getPluginManager().isPluginEnabled("WorldEdit")) {
```
Pablete1234: Should probably check for permission too here

util/src/main/java/tc/oc/pgm/util/menu/InventoryMenu.java
```diff
@@ -0,0 +1,369 @@
...
+      currentPage.add(null);
+      currentPage.add(null);
+      currentPage.add(null);
+      currentPage.add(
+          // Is there a next page?
+          pages.size() - 1 == i
+              ? null
+              : new PageInventoryMenuItem(pages.get(i + 1), menuArranger, true));
+      currentPage.add(null);
+      currentPage.add(null);
```
Pablete1234:
```suggestion
      List<InventoryMenuItem> currentPage = pages.get(i);
      for (int item = 0; item < ROW_WIDTH; item++) currentPage.add(null);
      if (i > 0) currentPage.set(2, new PageInventoryMenuItem(pages.get(i - 1), menuArranger, false));
      if (i < pages.size() - 1) currentPage.set(6, new PageInventoryMenuItem(pages.get(i - 1), menuArranger, false));
```

You can simply add 9 items on a loop and then set the 2 overrides for previous/next pages, adding so many currentPage.add(null) is a bit dirty

Pablete1234 (changes requested):

Pablete1234 (approved):

Pablete1234:
@Electroid Could you give this a review as it looks ready to merge?

Electroid (approved):

PR #731, opened 2020-12-06
Move to adventure BossBars

KingOfSquares:
Moves to the new bossBar system..

There is one bug that I'm clueless about, when you cycle to a new map the unready bossBar(pictured below) seems to disappear after 2-3 seconds. I can not find any place any showing/hiding of bossBars happening at the same time that it goes away. This only happens on the initial showing it seems.

(Waiting for 1 player to join) 

I've also noticed whenever I hide a boss bar it seems to take a few seconds before it actually goes away on my screen, sometimes it even shows "Wither" a little before disappearing. (disregard the weird monument names, I'm going to research this bug by itself) 

https://streamable.com/gm2c2g
https://streamable.com/azu3xu

Should I also remove the UnreadyTimeout stuff? It seems that right now its hardcoded as disabled, but still has some code left.


Electroid (approved):
Good to go, will need another PR to fix the sidebar.

PR #733, opened 2020-12-07
Fix control point visual-material attribute name closes #723

### LINKED ISSUE #723: Capture point visual-materials attribute doesn't function
It allowed you to filter which blocks would change when a hill was captured. Either the way it's applied no longer works, or support was removed. The filter is ignored, and all default blocks (wool, carpet, clay, glass, etc.) change even when otherwise specified 

Refer to https://github.com/CrazyisCreeps/maps.stratus.network/blob/master/other/tournament/conquest/cp/factori/map.xml#L72 for an example of the usage

Pugzy:
As discussed in #723 this attribute is references as `visual-material` in all documentation `docs.oc.tc` and `pgm.dev`.

I have checked the entire OCC and OCN maps repository (plus others) and found only one instance of `visual-world` used by Bolt who I assume source dived to debug an issue where the original `visual-material` did not exist finding this alternate name.

For the above reasons and the low usage of the feature (3 instances the whole of OCN), I believe it is safe to drop support for the assumed typo error made during the open-source process. 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
@@ -50,7 +50,7 @@ public abstract class ControlPointParser {
     Filter playerFilter = filterParser.parseFilterProperty(elControlPoint, "player-filter");
 
     Filter visualMaterials;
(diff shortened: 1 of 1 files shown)


Pablete1234 (approved):

Electroid:
This was an IntelliJ refactor gone terribly wrong! Thanks

Electroid (approved):

PR #735, opened 2020-12-11
Make length of death animation configurable

mrcookieunderscore13:
This PR makes a config option for how long the death animation should be shown before the player is vanished. The default value in PGM is 15 ticks. Here is a video showing 7 ticks and 3 ticks respectively: [https://youtu.be/0-oW-7wEpQA](url)

For context, this helps an issue where a dead player's animation blocks sword hits.
Signed-off-by: mrcookie

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -166,7 +166,7 @@ public final class PGMConfig implements Config {
     this.griefScore =
         parseInteger(config.getString("gameplay.grief-score", "-10"), Range.atMost(0));
     this.deathTicks =
-        parseInteger(config.getString("gameplay.death-tick", "15"), Range.atLeast(0));
+        parseInteger(config.getString("gameplay.death-ticks", "15"), Range.closed(0, 20));
 
     this.minPlayers = parseInteger(config.getString("join.min-players", "1"));
     this.limitJoin = parseBoolean(config.getString("join.limit", "true"));
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -69,6 +69,7 @@ public final class PGMConfig implements Config {
   // gameplay.*
   private final boolean woolRefill;
   private final int griefScore;
+  private final int deathAnimationLength;
 
   // join.*
   private final long minPlayers;
@@ -164,6 +165,8 @@ public final class PGMConfig implements Config {
     this.verboseStats = parseBoolean(config.getString("ui.verbose-stats", "false"));
     this.griefScore =
         parseInteger(config.getString("gameplay.grief-score", "-10"), Range.atMost(0));
+    this.deathAnimationLength =
+        parseInteger(config.getString("gameplay.death-animation-length", "15"));
 
     this.minPlayers = parseInteger(config.getString("join.min-players", "1"));
     this.limitJoin = parseBoolean(config.getString("join.limit", "true"));
@@ -516,6 +519,11 @@ public final class PGMConfig implements Config {
   @Override
   public int getGriefScore() {
     return griefScore;
+  }
+
+  @Override
+  public int getDeathAnimationLength() {
+    return deathAnimationLength;
   }
 
   @Override
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -69,7 +69,7 @@ public final class PGMConfig implements Config {
   // gameplay.*
   private final boolean woolRefill;
   private final int griefScore;
-  private final int deathAnimationLength;
+  private final int deathTicks;
 
   // join.*
   private final long minPlayers;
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -164,6 +165,8 @@
     this.verboseStats = parseBoolean(config.getString("ui.verbose-stats", "false"));
     this.griefScore =
         parseInteger(config.getString("gameplay.grief-score", "-10"), Range.atMost(0));
+    this.deathAnimationLength =
+        parseInteger(config.getString("gameplay.death-animation-length", "15"));
```
Electroid: `gameplay.death-ticks`

core/src/main/java/tc/oc/pgm/api/Config.java
```diff
@@ -253,6 +253,13 @@
    */
   int getGriefScore();
 
+  /**
+   * Gets the length in ticks for how long the death animation is shown
+   *
+   * @return length of death animation
+   */
+  int getDeathAnimationLength();
```
Electroid: `getDeathTicks`

Electroid (changes requested):
Also add `death-ticks: 15` to the `main/resources/config.yml`, with a comment explanation of what it does.

core/src/main/java/tc/oc/pgm/PGMConfig.java
Electroid: Oh, also make sure it is `Range.atLeast(0)`. See the `griefScore` above for an example.

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
```diff
@@ -32,6 +33,7 @@ public Dead(SpawnMatchModule smm, MatchPlayer player) {
 
   public Dead(SpawnMatchModule smm, MatchPlayer player, long deathTick) {
     super(smm, player);
+    CORPSE_ROT_TICKS = PGM.get().getConfiguration().getDeathAnimationLength();
```
Pablete1234: see comment below

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -164,6 +165,8 @@
     this.verboseStats = parseBoolean(config.getString("ui.verbose-stats", "false"));
     this.griefScore =
         parseInteger(config.getString("gameplay.grief-score", "-10"), Range.atMost(0));
+    this.deathTicks =
+        parseInteger(config.getString("gameplay.death-tick", "15"), Range.atLeast(0));
```
Electroid: `death-ticks`, missing an `s`

core/src/main/resources/config.yml
```diff
@@ -65,6 +65,7 @@ join:
 gameplay:
   refill-wool: true # Should wool in wool rooms be automatically refilled?
   grief-score: -10 # Score under which players should be kept out of the match
+  death-tick: 15 # how long should the death animation be shown for?
```
Electroid: `death-ticks: 15 # Number of ticks the death animation should last`

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/PGMConfig.java
Pablete1234: This must be between 0 and under the minimum respawn time which currently in pgm is 20 ticks. If you set it to a higher number and the map has a low respawn time you'd start seeing some bad behaviour where a player may respawn before their death animation has ended.

Pablete1234 (changes requested):
See review below.

core/src/main/java/tc/oc/pgm/PGMConfig.java
Pablete1234: This must be between 0 and under the minimum respawn time which currently in pgm is 20 ticks. If you set it to a higher number and the map has a low respawn time you'd start seeing some bad behaviour where a player may respawn before their death animation has ended.

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
Pablete1234: This is not a per-player configuration, you can have this as a static variable without having every Dead state object need to grab it

Pablete1234 (changes requested):
In general, i'm not even sure that this really has to be a setting since it's so specific on how it's relied upon by other parts of the plugin even if there's no direct connection (eg: respawn times)

Also i belive having each server use a different number will generate inconsistencies and when a proper replacement animation is pushed thru this will be troublesome. If a particle-based animation is to take place as the replacement, we will not want to have to deal with differing animation lengths when handling that, which is why i see this setting as something that will inevitably be short-lived.

Electroid (approved):

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
```diff
@@ -33,7 +33,7 @@ public Dead(SpawnMatchModule smm, MatchPlayer player) {
 
   public Dead(SpawnMatchModule smm, MatchPlayer player, long deathTick) {
     super(smm, player);
-    CORPSE_ROT_TICKS = PGM.get().getConfiguration().getDeathAnimationLength();
+    CORPSE_ROT_TICKS = PGM.get().getConfiguration().getDeathTicks();
```
Pablete1234, line 36: This is still copying the value every time a player dies, which is unneeded

Issue #488, opened 2020-05-14
Improve 1.7 support


Palmidence:
So the repo doesn't get clogged with issues and details that pertain to an unsupported version and instead focuses on the current capabilities of PGM, I think that this would be the best compromise when reporting any issues for legacy versions. 

1.7.x
--
- [ ] Player helmets disappear upon respawn if flag is captured (see #483)
- [x] Particle beams used for marking flags do not display when held (see #485 , #594 )
- [x] Players who respawn and spam space in 1.7 can end up keeping the ability to fly
- [x] On death players will appear to glitch thru the world (see #736)
- [x] Missing "Click to respawn" message upon death - send as chat message
- [x] `/maps` command is unformatted (all maps/authors/etc.) on same line
- [X] 1.7 tablist support
- [x] Send votes to click via chat instead of book for 1.7 players

sharkskill:
The helmet bug described in #483 isn’t 1.7 specific, it happens in all versions.

rafibaum:
If 1.7 is unsupported, do we add code which fixes legacy issues related to it? Imo 1.7 should be an unofficial target. We should add code to fix legacy issue but offer no guarantees for stability.

Pablete1234:
IMO 1.7 support isn't a priority, and some things will just never work on it (eg: voting book) because it simply isn't possible, but pgm should still aim to work in 1.7 where reasonably doable

Electroid:
I would like to maintain basic 1.7 support, but 1.8 is prioritized right now.

Pablete1234:
Going to close this as everything that was 1.7 specific has been fixed, i'll reopen #483 to keep track of the helmet bug that was said to affect all versions

PR #736, opened 2020-12-12
Fix several issues on player states

Pablete1234:
Fixed a few issues relating to player states arround the playerbase, notably:
 - Fixed players switching gamemode multiple times on the same tick
 - Fixed bugged players on death animation being left on blitz
 - Fixed legacy (1.7) players glitching thru the world or falling when frozen, see #488 
 - Reverted mitigation of legacy players using adventure mode to prevent flying bug, the root issue is [fixed upstream](https://github.com/ViaVersion/ViaRewind/pull/231)


Electroid (approved):
🚀

PR #737, opened 2020-12-12
Dynamic tablist update delay

Pablete1234:
Modifies tablist defered update times to be dynamic, when few updates are occurring (eg: just you joining a team) the update will be quicker (100ms now, 250ms before) however, if the updates are being slow, tablist won't update more frequently than spending 1/20th of the time on tablist rendering.

If the tablist takes 20ms to re-layout, re-render etc, no update will be scheduled for sooner than 20 * 20 = 400ms since the update ended, however if it takes 5ms to re-layout, 5*20 = 100, and in 2 ticks it would be allowed to re-render.


Electroid (approved):
Smart, excited to see this in action.

PR #738, opened 2020-12-13
Refactor:  fixing bugs

KingOfSquares:
Removed a duplicate class
Fixed scoreboard bug with numbers

currently working on bossbar bugs


Electroid (approved):
Nice catch

PR #739, opened 2020-12-14
Make player death animation hitbox tiny

Pablete1234:
A few changes to death animations: 
 - Increased death ticks from 15 to 20, matching vanilla and getting the particle at the end of the animation
 - Removed option to configure length of death animation
 - Minimum respawn time in xml is now 1,5s (down from 2s), can't make it one second or animation wouldn't have time to finish
 - Fake the corpse entering a bed in order to make clients see them as a tiny hitbox


Electroid (dismissed):
Who knew the solution was to put everyone to sleep

PR #744, opened 2020-12-18
Implement filters in blitz

### LINKED ISSUE #365: Filters and show attribute for Blitz module
show="false" would just hide the blitz configuration from the scoreboard,
and filters would allow games like bedwars or more project ares oriented variants with the bedwars machanics to work with less hacky stuff

Pablete1234:
Fixes #365 

Adds a filter attribute or child to blitz, if the filter denies, your life counter does not go down.

Example, you will only lose a life when dying after 30s from match start:
```
<blitz>
  <lives>5</lives>
  <filter>
    <time>30s</time>
  </filter>
</blitz>
```

Adding it as an attribute is also valid (note, the filter will have to be defined elsewhere):
```
<blitz filter="time-30-sec">
  <lives>5</lives>
</blitz>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blitz/BlitzConfig.java
+++ b/core/src/main/java/tc/oc/pgm/blitz/BlitzConfig.java
@@ -2,13 +2,21 @@ package tc.oc.pgm.blitz;
 
 import static com.google.common.base.Preconditions.checkArgument;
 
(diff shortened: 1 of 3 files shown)


Electroid (approved):
Sweet!

Pablete1234:
This was a draft because i have done 0 testing on this feature so far

PR #745, opened 2020-12-18
Assign team columns proportional to team size in tab

Pablete1234:
Currently the amount of columns to assign per team is static to all teams, with this PR each team will get their own column size depending on the max players the team can accomodate

Example of a 10v50 match, blue takes one column, red takes 3:
image


Electroid:
Is it proportional to the max size or current size?

Pablete1234:
It's proportional to the max size of the teams, how many players have actually joined isn't relevant because we don't want people joining/leaving causing column sizes to shift, and because proportions will roughly match the max either way

Electroid (approved):

PR #749, opened 2020-12-20
Allow dead players to use picker helmet

Pugzy:
Currently, you can not open the team picker helmet you are given when you die as the interact event is cancelled.
This picker lets you change teams, join observers and change classes (when classes exist).

image

Looking into this it seems like OCN PGM had a slightly modified Observer event system which resolves this.
Rejigging the event priorities and listening for the ObserverInteractEvent on dead players allows the event to be caught and used.

This also fixes an issue in the `EventFilterMatchModule` where the check for cancelling an event was incorrect.

There is a strange when the `picker` setting is set to `off` as the picker will not popup which means the death helmet will not work but I think the current picker settings need a review before a solution is made.


Pablete1234 (dismissed):

Electroid (dismissed):

Electroid:
Looks good, can you resolve the conflicts?

Electroid (approved):

PR #751, opened 2020-12-22
Add query modifier for checking acting player info

Pugzy:
Let's say you wanted to prevent a player from opening chests unless they're in a specific region (the wool room maybe).

The current way of filtering location on a `block-break` or `use` event target the location of the physical block (which is always in the wool room). In this instance, I want to query the location of the actor (player) rather than the block in question.

For this, I have created a `PlayerBlockQueryModifier` which sets focus on the player in the event vs the block. This also slightly modifies the newly created base `QueryModifier` class allowing for unmodified queries to abstain by returning null. In this case, the player within the `PlayerQuery` can be null (when offline) so a fallback was required.

See the example below where the player can only use items within the yellow wool room if they too are in that region.

```xml
<!--- filter with region -->
<player id="player-in-yellow-wool">
    <rectangle id="yellow-wool-room" min="534.5,-354" max="544,-339"/>
</player>

<!--- application -->
<apply use="player-in-yellow-wool" region="yellow-wool-room" message="You can not interact with this from here."/>
```

Open to suggestions if this can be completed an alternative way (existing XML or improved implementation). 
Discussed with Pablo and Simon in Discord so ty for help.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/modifier/PlayerBlockQueryModifier.java
+++ b/core/src/main/java/tc/oc/pgm/filters/modifier/PlayerBlockQueryModifier.java
@@ -1,0 +1,27 @@
+package tc.oc.pgm.filters.modifier;
+
+import javax.annotation.Nullable;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.filter.query.PlayerQuery;
+import tc.oc.pgm.api.filter.query.Query;
+
+public class PlayerBlockQueryModifier extends QueryModifier<PlayerQuery> {
+
+  public PlayerBlockQueryModifier(Filter filter) {
+    super(filter);
+  }
+
+  @Nullable
+  @Override
+  protected Query modifyQuery(PlayerQuery query) {
+    // Abstain when no player can be found, eg: they disconnected
+    if (query.getPlayer() == null) return null;
+
+    return query.getPlayer().getQuery();
+  }
+
+  @Override
+  public Class<? extends PlayerQuery> getQueryType() {
+    return PlayerQuery.class;
+  }
+}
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/filters/modifier/PlayerBlockQueryModifier.java
```diff
@@ -0,0 +1,26 @@
...
+public class PlayerBlockQueryModifier extends QueryModifier<PlayerQuery> {
+
+  public PlayerBlockQueryModifier(Filter filter) {
+    super(filter);
+  }
+
+  @Nullable
+  @Override
+  protected Query modifyQuery(PlayerQuery query) {
+    if (query.getPlayer() == null) return null;
```
Pablete1234, line 18:
```suggestion
    // Abstain when no player can be found, eg: they disconnected
    if (query.getPlayer() == null) return null;
```
I'd add a clarifying comment for future mantainers

Pablete1234 (dismissed):

core/src/main/java/tc/oc/pgm/filters/modifier/QueryModifier.java
```diff
@@ -10,6 +11,8 @@
  * Takes a query and modifies it before passing it onto its child filter. Might transform query type
  * so {@link #getQueryType()} on this modifier can be different from its child filter
  *
+ * <p>If the modified query returns null the query will abstain.
```
KingOfSquares:
```suggestion
 * <p>If the modified query returns {@code null} this modifier will abstain without querying the child filter.
```

A query can't abstain, unsure about the phrasing.

Pablete1234 (approved):

Electroid (approved):
Awesome! Excited to see the use-cases

PR #752, opened 2020-12-22
Add repeat-posts tag to flags

mrcookieunderscore13:
This option would allow map makers to prevent a flag from respawning to the same post two times in a row. 
For example, with `repeat-posts="false"`, a flag will not respawn to the same post twice in a row. To preserve previous behavior, this option defaults to true.

This will help reduce the RNG impact on kotf maps. 

Signed-off-by: mrcookie

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/Flag.java
+++ b/core/src/main/java/tc/oc/pgm/flag/Flag.java
@@ -6,9 +6,11 @@ import static net.kyori.adventure.text.Component.translatable;
 
 import com.google.common.collect.ImmutableSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Optional;
 import java.util.Random;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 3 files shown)


Pablete1234:
sequential="true" ?

mrcookieunderscore13:
> sequential="true" ?

Sorry, I'm not sure what you mean by this

Pablete1234:
If the previous flag post doesn't repeat, isn't it just the same as sequential posts?

mrcookieunderscore13:
> If the previous flag post doesn't repeat, isn't it just the same as sequential posts?

Oh, no it is different. With sequential, it will follow a set order like ABCABCABC.
With `repeat-posts="false"`, after it spawns on post A, it will randomly choose between B and C. So, a hypothetical order could look like this: ACABCBA. It is essentially the random option, but it prevents the same post from being chosen twice in a row.

Pablete1234:
I think this may be a bit too much of specific attributes to very specific things. It would probably make more sense to add a filter to posts that define if the flag can respawn there, and adding a filter that lets you query for the last post a flag was in

mrcookieunderscore13:
> I think this may be a bit too much of specific attributes to very specific things. It would probably make more sense to add a filter to posts that define if the flag can respawn there, and adding a filter that lets you query for the last post a flag was in

Yeah that method makes more sense--I'll implement it that way.
Closing the PR.

PR #758, opened 2020-12-25
Add Selectable Monument Modes

CoWinkKeyDinkInc:
This pull request adds a `mode` attribute to `<core>` and `<destroyable>`, and an `id` attribute to `<mode>` so that monument modes can now be selectable instead of defaulting to all cores and monuments. To maintain compatability, only modes with no `id` can affect cores and destroyables with no `mode`. Modes have a one to many relationship with the objectives. This means that monuments made of more complex materials can now be changed to different materials using multiple `<destroyable>` modules. Maps that use them as "stages" or extra platforms like Miner Sixty Niner, CannonCube, CannonDuel, and Royal Garden CTW will also greatly benefit from this addition, as more can be added, or the platforms can take on more complex patterns.

```xml
<cores material="obsidian" leak="2">
    <core team="red-team" region="left-core" mode-changes="true" mode="glass-mode"/>
    <core team="blue-team" region="right-core" mode-changes="true" mode="gold-mode"/>
</cores>
<destroyables>
    <destroyable id="red-mon" name="red one" owner="red-team" mode-changes="true" mode="glass-mode" region="mon-red" materials="obsidian"/>
    <destroyable id="blue-mon" name="blue one" owner="blue-team" mode-changes="true" mode="gold-mode" region="mon-blu" materials="obsidian"/>
</destroyables>
<modes>
    <mode id="glass-mode" after="10s" material="glass" name="`eGlass Mode" show-before="5s"/>
    <mode id="gold-mode" after="20s" material="gold block" name="`eGold Mode" show-before="5s"/>
</modes>
```

In this example, the Red Core and Red Monument will turn to glass from obsidian in 10 seconds, Blue Core and Blue Monument will turn to gold blocks from obsidian in 20 seconds. You can download this test map as an example to check it out these changes.
[testworld.zip](https://github.com/PGMDev/PGM/files/5741182/testworld.zip)
 
**EDIT:** This deprecates the `mode-changes` attribute as something that defines the Core and Destroyable instance, it remains in the XML parser for compatibility reasons.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -245,7 +245,6 @@ public class Core extends TouchableGoal<CoreFactory>
         && block.getData() == this.material.getData();
   }
 
-  @Override
   public String getModeChangeMessage(Material material) {
     return ModeUtils.formatMaterial(material) + " CORE MODE";
   }
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -6,6 +6,7 @@ import static net.kyori.adventure.text.Component.translatable;
 
 import com.google.common.collect.ImmutableList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -160,8 +161,8 @@ public class Core extends TouchableGoal<CoreFactory>
     this.leaked = true;
   }
 
-  public String getMode() {
-    return this.definition.getMode();
+  public List<String> getModeList() {
+    return this.definition.getModeList();
   }
 
   public boolean hasLeaked() {
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -28,6 +28,7 @@ import tc.oc.pgm.goals.Contribution;
 import tc.oc.pgm.goals.IncrementalGoal;
 import tc.oc.pgm.goals.ModeChangeGoal;
 import tc.oc.pgm.goals.TouchableGoal;
+import tc.oc.pgm.modes.Mode;
 import tc.oc.pgm.modes.ModeUtils;
 import tc.oc.pgm.regions.CuboidRegion;
 import tc.oc.pgm.regions.FiniteBlockRegion;
@@ -161,7 +162,7 @@ public class Core extends TouchableGoal<CoreFactory>
     this.leaked = true;
   }
 
-  public List<String> getModeList() {
+  public List<Mode> getModeList() {
     return this.definition.getModeList();
   }
 
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -5,12 +5,11 @@ import static net.kyori.adventure.text.Component.text;
 import static net.kyori.adventure.text.Component.translatable;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import java.util.Collections;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-
-import com.google.common.collect.ImmutableSet;
 import net.kyori.adventure.text.Component;
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
@@ -134,6 +133,10 @@ public class Core extends TouchableGoal<CoreFactory>
     return proximityLocations;
   }
 
+  public ImmutableSet<Mode> getModes() {
+    return this.definition.getModes();
+  }
+
   public MaterialData getMaterial() {
     return this.material;
   }
@@ -163,10 +166,6 @@ public class Core extends TouchableGoal<CoreFactory>
     this.leaked = true;
   }
 
-  public ImmutableSet<Mode> getModes() {
-    return this.definition.getModes();
-  }
-
   public boolean hasLeaked() {
     return this.leaked;
   }
@@ -189,11 +188,6 @@ public class Core extends TouchableGoal<CoreFactory>
   @Override
   public boolean isCompleted(Competitor team) {
     return this.leaked && this.canComplete(team);
-  }
-
-  @Override
-  public boolean isAffectedByModeChanges() {
-    return this.definition.hasModeChanges();
   }
 
   @Override
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -6,7 +6,6 @@ import static net.kyori.adventure.text.Component.translatable;
 
 import com.google.common.collect.ImmutableList;
 import java.util.Collections;
-import java.util.List;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -24,11 +23,11 @@ import tc.oc.pgm.api.party.Party;
 import tc.oc.pgm.api.player.MatchPlayerState;
 import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.api.region.Region;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 import tc.oc.pgm.goals.Contribution;
 import tc.oc.pgm.goals.IncrementalGoal;
 import tc.oc.pgm.goals.ModeChangeGoal;
 import tc.oc.pgm.goals.TouchableGoal;
-import tc.oc.pgm.modes.Mode;
 import tc.oc.pgm.modes.ModeUtils;
 import tc.oc.pgm.regions.CuboidRegion;
 import tc.oc.pgm.regions.FiniteBlockRegion;
@@ -162,7 +161,7 @@ public class Core extends TouchableGoal<CoreFactory>
     this.leaked = true;
   }
 
-  public List<Mode> getModeList() {
+  public Set<SelfIdentifyingFeatureDefinition> getModeList() {
     return this.definition.getModeList();
   }
 
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -9,6 +9,8 @@ import java.util.Collections;
 import java.util.Set;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+
+import com.google.common.collect.ImmutableSet;
 import net.kyori.adventure.text.Component;
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
@@ -23,11 +25,11 @@ import tc.oc.pgm.api.party.Party;
 import tc.oc.pgm.api.player.MatchPlayerState;
 import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.api.region.Region;
-import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 import tc.oc.pgm.goals.Contribution;
 import tc.oc.pgm.goals.IncrementalGoal;
 import tc.oc.pgm.goals.ModeChangeGoal;
 import tc.oc.pgm.goals.TouchableGoal;
+import tc.oc.pgm.modes.Mode;
 import tc.oc.pgm.modes.ModeUtils;
 import tc.oc.pgm.regions.CuboidRegion;
 import tc.oc.pgm.regions.FiniteBlockRegion;
@@ -161,8 +163,8 @@ public class Core extends TouchableGoal<CoreFactory>
     this.leaked = true;
   }
 
-  public Set<SelfIdentifyingFeatureDefinition> getModeList() {
-    return this.definition.getModeList();
+  public ImmutableSet<Mode> getModes() {
+    return this.definition.getModes();
   }
 
   public boolean hasLeaked() {
--- a/core/src/main/java/tc/oc/pgm/core/Core.java
+++ b/core/src/main/java/tc/oc/pgm/core/Core.java
@@ -160,6 +160,10 @@ public class Core extends TouchableGoal<CoreFactory>
     this.leaked = true;
   }
 
+  public String getMode() {
+    return this.definition.getMode();
+  }
+
   public boolean hasLeaked() {
     return this.leaked;
   }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -157,7 +157,7 @@ public class CoreMatchModule implements MatchModule, Listener {
       if (core.isAffectedByModeChanges()) {
         if (core.getMode() == null) {
           core.replaceBlocks(event.getMode().getMaterialData());
-        } else if (core.getMode().equals(event.getMode().getStage())) {
+        } else if (core.getMode().equals(event.getMode().getId())) {
           core.replaceBlocks(event.getMode().getMaterialData());
         }
       }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -154,10 +154,8 @@ public class CoreMatchModule implements MatchModule, Listener {
   @EventHandler(priority = EventPriority.MONITOR)
   public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Core core : this.cores) {
-      if (core.isAffectedByModeChanges()) {
-        if (core.getModeList() == null || core.getModeList().contains(event.getMode())) {
-          core.replaceBlocks(event.getMode().getMaterialData());
-        }
+      if (core.getModeList() == null || core.getModeList().contains(event.getMode())) {
+        core.replaceBlocks(event.getMode().getMaterialData());
       }
     }
   }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -157,7 +157,7 @@ public class CoreMatchModule implements MatchModule, Listener {
       if (core.isAffectedByModeChanges()) {
         if (core.getMode() == null) {
           core.replaceBlocks(event.getMode().getMaterialData());
-        } else if (core.getMode() != null && core.getMode().equals(event.getMode().getId())) {
+        } else if (core.getMode() != null && core.getMode().equals(event.getMode().getStage())) {
           core.replaceBlocks(event.getMode().getMaterialData());
         }
       }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -155,9 +155,10 @@ public class CoreMatchModule implements MatchModule, Listener {
   public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Core core : this.cores) {
       if (core.isAffectedByModeChanges()) {
-        if (core.getMode() == null) {
+        if (core.getModeList() == null || core.getModeList().isEmpty()) {
           core.replaceBlocks(event.getMode().getMaterialData());
-        } else if (core.getMode().equals(event.getMode().getId())) {
+          // if a mode on modeList has one that matches get id
+        } else if (core.getModeList().contains(event.getMode().getId())) {
           core.replaceBlocks(event.getMode().getMaterialData());
         }
       }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -155,10 +155,7 @@ public class CoreMatchModule implements MatchModule, Listener {
   public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Core core : this.cores) {
       if (core.isAffectedByModeChanges()) {
-        if (core.getModeList() == null || core.getModeList().isEmpty()) {
-          core.replaceBlocks(event.getMode().getMaterialData());
-          // if a mode on modeList has one that matches get id
-        } else if (core.getModeList().contains(event.getMode().getId())) {
+        if (core.getModeList() == null || core.getModeList().contains(event.getMode())) {
           core.replaceBlocks(event.getMode().getMaterialData());
         }
       }
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -157,7 +157,7 @@ public class CoreMatchModule implements MatchModule, Listener {
       if (core.isAffectedByModeChanges()) {
         if (core.getMode() == null) {
           core.replaceBlocks(event.getMode().getMaterialData());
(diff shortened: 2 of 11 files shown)


CoWinkKeyDinkInc:
I have tried to add an XML checker that would pull up an error if a `mode` was set without `mode-changes` being enabled. But for some reason IntelliJ seems to think that this would always set to false, and did something with the compiler so it hard set it to false. Is there a way to make it ignore this?

Pablete1234:
As this is right now, it is impossible to define a core that goes gold at 15m, glass at 30m, and a monument that turns into endstone at 20m.

You need cores and monuments to be able to have a set of modes they're affected by, that list is either all modes or none of the modes if mode-changes= true/false, or, alternatively, a list of mode ids if you use modes="mode-1 mode-2 mode-3", that way you also avoid the "you require one for the other", instead they're mutually exclusive, you can not have both mode-changes and modes attributes

CoWinkKeyDinkInc:
Couldn't you just have another `<mode>` with the same `id` and that would accomplish the same thing? Modesetting just has to be at a different time and not conflict with another mode.

Pablete1234:
> Couldn't you just have another `<mode>` with the same `id` and that would accomplish the same thing? Modesetting just has to be at a different time and not conflict with another mode.

IDs are, by definition, **unique**. No 2 modes should be using the same id, and in general, no 2 features in a map (this is, filters, regions, cores, flags, nets, etc etc) should have the same id

CoWinkKeyDinkInc:
That's fair, maybe `id` can be replaced with something like `stage` or `set` so it is more clear that it can be non unique.

CoWinkKeyDinkInc:
The `id` attribute on `<mode>` has been replaced with `stage` to emphasize that the modes can be grouped as a set, or a "stage", as the selectable monuments can have multiple modes. `mode` in the objectives has been renamed to `modes` to show map makers that they can follow a set of modes that share a common `stage`.

Pablete1234:
This still has an unintended side effect, say we want:
Core: obsidian -> gold (15m) -> glass (30m)
Monument endstone -> glass (30m)

You'll have to define 3 modes, 2 of them with stage "core-modes"  and 1 with "monument-modes", there are 3 modes in total, you will get 3 countdowns (bossbars) in total, when in reality you should only get 2, one for "Gold core mode" and one for "Glass mode" (for everything)

CoWinkKeyDinkInc:
Bossbar countdowns can be disabled, you can have one mode that has a boss bar enabled with a name that mentions both of them. When I was testing it it seems to flick back and forth but yeah it does have buggy behaviour with two boss bars running at the same time.

Also two modes still can't happen at the same time, maybe this is not a needed restriction anymore.

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -126,7 +126,12 @@ public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Destroyable destroyable : this.destroyables) {
       if (destroyable.isAffectedByModeChanges()) {
         double oldCompletion = destroyable.getCompletion();
-        destroyable.replaceBlocks(event.getMode().getMaterialData());
+        if (destroyable.getMode() == null) {
+          destroyable.replaceBlocks(event.getMode().getMaterialData());
+        } else if (destroyable.getMode() != null
+            && destroyable.getMode().equals(event.getMode().getStage())) {
```
applenick:
```suggestion
        } else if (destroyable.getMode().equals(event.getMode().getStage())) {
```
Since the previous if statement is already checking if the mode is null, no need to check again if it's not null 😄

core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
```diff
@@ -155,7 +155,11 @@ public void lavaProtection(final BlockTransformEvent event) {
   public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Core core : this.cores) {
       if (core.isAffectedByModeChanges()) {
-        core.replaceBlocks(event.getMode().getMaterialData());
+        if (core.getMode() == null) {
+          core.replaceBlocks(event.getMode().getMaterialData());
+        } else if (core.getMode() != null && core.getMode().equals(event.getMode().getStage())) {
+          core.replaceBlocks(event.getMode().getMaterialData());
```
applenick:
```suggestion
        } else if (core.getMode().equals(event.getMode().getStage())) {
          core.replaceBlocks(event.getMode().getMaterialData());
```
See other comment

Pablete1234:
> Also two modes still can't happen at the same time, maybe this is not a needed restriction anymore.

why is that?

Pugzy:
> The `id` attribute on `<mode>` has been replaced with `stage` to emphasize that the modes can be grouped as a set, or a "stage", as the selectable monuments can have multiple modes. `mode` in the objectives has been renamed to `modes` to show map makers that they can follow a set of modes that share a common `stage`.

I wouldn't make this change.

I think nesting `<modes>` would be better for this purpose (if possible) like so.
```xml
<modes>
    <modes id="core-modes">
        <mode after="10s" material="gold block"/>
        <mode after="20s" material="glass"/>
    </modes>
    <mode id="monument-mode" after="10s" material="endstone"/>
</modes>
```

Pablete1234:
> > The `id` attribute on `<mode>` has been replaced with `stage` to emphasize that the modes can be grouped as a set, or a "stage", as the selectable monuments can have multiple modes. `mode` in the objectives has been renamed to `modes` to show map makers that they can follow a set of modes that share a common `stage`.
> 
> I wouldn't make this change.
> 
> I think nesting `<modes>` would be better for this purpose (if possible) like so.
> 
> ```
> <modes>
>     <modes id="core-modes">
>         <mode after="10s" material="gold block"/>
>         <mode after="20s" material="glass"/>
>     </modes>
>     <mode id="monument-mode" after="10s" material="endstone"/>
> </modes>
> ```

While i agree that stage is bad overall, if you set an id on a parent `modes`, all pgm does is transfer that id down to the individual modes, so your issue is the same.

I still belive the best solution is to have a list of mode ids for what modes you want your core/monument to be affected by

CoWinkKeyDinkInc:
`modes` can now accept multiple Mode IDs using a space as a separator. `id` is returned to `<mode>`. Right now the block application does not work, the `gold-mode` still applies to the two destroyables. It's probably the `List<String>` not being created correctly or something but I wasted too much time on this today so yeah.

```xml
<cores material="obsidian" leak="2">
    <core team="red-team" region="left-core" mode-changes="true" modes="glass-mode gold-mode"/>
    <core team="blue-team" region="right-core" mode-changes="true" modes="glass-mode gold-mode"/>
</cores>
<destroyables>
    <destroyable id="red-mon" name="red one" owner="red-team" mode-changes="true" modes="glass-mode" region="mon-red" materials="obsidian"/>
    <destroyable id="blue-mon" name="blue one" owner="blue-team" mode-changes="true" modes="glass-mode" region="mon-blu" materials="obsidian"/>
</destroyables>
<modes>
    <mode id="glass-mode" after="10s" material="glass" name="`eGlass Mode" show-before="5s"/>
    <mode id="gold-mode" after="20s" material="gold block" name="`eGold Mode" show-before="5s"/>
</modes>
```

mitchts:
This is coming from someone that has never touched PGM code, but I especially liked how Avicus did destroyable modes which addressed different destroyable objectives having varying modes but with the added flexibility of filters which could allow for some creative match dynamics that are simply not possible right now.

See [here](https://xml.avicus.net/modules/destroyable-phases) for reference (or the [Walrus version](https://docs.walrus.gg/facets/destroy#mode-changes) which uses PGM terms). Of course, such an implementation is quite a bit more than what is being worked on above and would require a proto update, but I believe it provides much more flexibility and avenues for map concepts.

A simple example would be something like:

```xml
<monuments materials="obsidian" first-mode="beacon-mode">
    <monument owner="blue" name="Monument A" region="blue-monument-a"/>
    <monument owner="red" name="Monument A" region="red-monument-a"/>
</monuments>
<modes>
    <mode id="beacon-mode" pass-mode="coal-mode" delay="15m">
        <materials>
            <material find="obsidian" replace="beacon"/>
        </materials>
    </mode>
    <mode id="coal-mode" delay="5m">
        <materials>
            <material find="beacon" replace="coal block"/>
        </materials>
    </mode>
</modes>
```

Something to consider.

Pablete1234:
> This is coming from someone that has never touched PGM code, but I especially liked how Avicus did destroyable modes which addressed different destroyable objectives having varying modes but with the added flexibility of filters which could allow for some creative match dynamics that are simply not possible right now.
> 
> See [here](https://xml.avicus.net/modules/destroyable-phases) for reference (or the [Walrus version](https://docs.walrus.gg/facets/destroy#mode-changes) which uses PGM terms). Of course, such an implementation is quite a bit more than what is being worked on above and would require a proto update, but I believe it provides much more flexibility and avenues for map concepts.
> 
> A simple example would be something like:
> 
> ```xml
> <monuments materials="obsidian" first-mode="beacon-mode">
>     <monument owner="blue" name="Monument A" region="blue-monument-a"/>
>     <monument owner="red" name="Monument A" region="red-monument-a"/>
> </monuments>
> <modes>
>     <mode id="beacon-mode" pass-mode="coal-mode" delay="15m">
>         <materials>
>             <material find="obsidian" replace="beacon"/>
>         </materials>
>     </mode>
>     <mode id="coal-mode" delay="5m">
>         <materials>
>             <material find="beacon" replace="coal block"/>
>         </materials>
>     </mode>
> </modes>
> ```
> 
> Something to consider.

In this specific example, i do not see why this would be any different than what is currently proposed, except for it being more verbose, and, if i may say, unintuitive. There's no need for XML to define a weirdly made chain of modes that may be harder to follow.

Is it any different to having this?
 ```xml
<destroyables material="obsidian" modes="beacon-mode coal-mode">
    <destroyable owner="blue" name="Monument A" region="blue-monument-a"/>
    <destroyable owner="red" name="Monument A" region="red-monument-a"/>
</destroyables>
<modes>
    <mode id="beacon-mode" after="15m" material="beacon"/>
    <mode id="coal-mode" after="20m" material="coal block"/>
</modes>
```

**Review by Pablete1234 — CHANGES_REQUESTED** [review]
You're using strings to handle ids and references, however, you should be using definitions for those.

As a pretty similar example, Nets in CTF have a list of flags that can be captured in them:
https://pgm.dev/docs/modules/objectives/ctf/#nets (look at `flags`)

You can see in the code that the `flags` node (the attribute in xml) is handled as returning a set of flag definitions:
https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/flag/FlagParser.java#L105

Essentially, you need to make `Mode` extend `SelfIdentifyingFeatureDefinition`, and register it as a feature on creation via `factory.getFeatures().addFeature(el, mode);`, then when creating cores or monuments, you need to get the feature, which will make it so pgm is correcting for ids not defined or duplicate ids on modes or other features

File: core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java [hunk 3]
```diff
@@ -155,10 +155,7 @@ public void lavaProtection(final BlockTransformEvent event) {
   public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Core core : this.cores) {
       if (core.isAffectedByModeChanges()) {
```
Pablete1234: You no longer need isAffectedByModeChanges.

If the list of modes is an empty collection, that's your validation that this is in fact, not affected by modes.

On the parsing side, make sure to make it so that if `mode-changes` is mutually exclusive with `modes`, and if `modes` isn't included you fall-back to either an empty set of modes (not affected by modes) or a null one or (possibly one that includes all modes) if it is mode-change affected

core/src/main/java/tc/oc/pgm/core/Core.java
```diff
@@ -161,7 +162,7 @@ public void markLeaked() {
     this.leaked = true;
   }
 
-  public List<String> getModeList() {
+  public List<Mode> getModeList() {
```
Pablete1234:
```suggestion
  public Set<Mode> getModeList() {
```
Across the code, replace List with Set, you're primarily going to be doing `contains` on this collection, sets are better for that

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -118,13 +119,14 @@ public CoreModule parse(MapFactory context, Logger logger, Document doc)
...
         String modes = coreEl.getAttributeValue("modes");
-        List<String> modeList = new ArrayList<>();
+        List<Mode> modeList = new ArrayList<>();
         if (modes != null) {
-          Node node = Node.fromAttr(coreEl, modes);
-          if (node != null) {
-            for (String mode : Splitter.on(" ").split(node.getValue())) {
-              modeList.add(mode);
-            }
+          for (String mode : Splitter.on(" ").split(modes)) {
```
Pablete1234:
```suggestion
          for (String mode : Splitter.on("\\s").split(modes)) {
```
use whitespace as a separator, similar to the example in https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/flag/FlagParser.java#L105

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -118,13 +119,14 @@ public CoreModule parse(MapFactory context, Logger logger, Document doc)
         }
 
         String modes = coreEl.getAttributeValue("modes");
-        List<String> modeList = new ArrayList<>();
+        List<Mode> modeList = new ArrayList<>();
```
Pablete1234: Make this an ImmutableSet<Mode>, like the example in https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/flag/FlagParser.java#L105

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -118,13 +119,14 @@ public CoreModule parse(MapFactory context, Logger logger, Document doc)
...
-          if (node != null) {
-            for (String mode : Splitter.on(" ").split(node.getValue())) {
-              modeList.add(mode);
-            }
+          for (String mode : Splitter.on(" ").split(modes)) {
+            Iterable bigcontext = context.getFeatures().getAll(Mode.class);
+            System.out.println(bigcontext);
+            Mode mode1 = context.getFeatures().get(mode, Mode.class);
+            modeList.add(mode1);
+            System.out.println(mode1);
```
Pablete1234: You left out some debugs here, also, you need to check that the mode returned isn't null or else it means that mode hasn't been found, again, you have a perfect 1 to 1 example in https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/flag/FlagParser.java#L105

Pablete1234 (changes requested):

Pugzy:
Crazy idea but.. (mentioned this in Discord) what about combining the `mode-changes` and `modes` into one attribute.
So `mode-changes="true"` (applying all modes) would work and so would `mode-changes="endstone-mode gold-mode glass-mode"`.

The only caveat would be you couldn't have a mode named `true` which isn't that big of an issue. Thoughts?

Pablete1234:
> Crazy idea but.. (mentioned this in Discord) what about combining the `mode-changes` and `modes` into one attribute.
> So `mode-changes="true"` (applying all modes) would work and so would `mode-changes="endstone-mode gold-mode glass-mode"`.
> 
> The only caveat would be you couldn't have a mode named `true` which isn't that big of an issue. Thoughts?

I don't think it's a good idea, boolean parsing should be boolean parsing, and mode ids should be mode ids. Ambiguity is bad, i belive booleans could accept more than "true" and "false", imagine "on" or "off", it just creates issues for us when we do not need them. The mode-changes is the "old" true/false setting, and "modes" is the alternative more modern one, where you say exactly what modes

core/src/main/java/tc/oc/pgm/core/Core.java
CoWinkKeyDinkInc: Is there a reason why these shouldn't be ImmutableSets? These are after the sets are created and passed to the core object.

CoWinkKeyDinkInc:
This commit is a fully working commit, for both cores and monuments. Cores and monuments can have multiple modes selected in the `modes` attribute, separated by a space. Hopefully updating the branch goes well.

```xml
<cores material="obsidian" leak="2">
    <core team="red-team" region="left-core" mode-changes="true" modes="glass-mode gold-mode" name="red core"/>
    <core team="blue-team" region="right-core" mode-changes="true" modes="glass-mode gold-mode" name="blue core"/>
</cores>
<destroyables>
    <destroyable id="red-mon" name="red one" owner="red-team" mode-changes="true" modes="glass-mode" region="mon-red" materials="obsidian"/>
    <destroyable id="blue-mon" name="blue one" owner="blue-team" mode-changes="true" modes="glass-mode" region="mon-blu" materials="obsidian"/>
</destroyables>
<modes>
    <mode id="glass-mode" after="10s" material="glass" name="`eGlass Mode" show-before="5s"/>
    <mode id="gold-mode" after="20s" material="gold block" name="`eGold Mode" show-before="5s"/>
</modes>
```

core/src/main/java/tc/oc/pgm/core/Core.java
```diff
@@ -160,6 +161,10 @@ public void markLeaked() {
     this.leaked = true;
   }
 
+  public Set<SelfIdentifyingFeatureDefinition> getModeList() {
```
Pablete1234:
```suggestion
  public ImmutableSet<Mode> getModes() {
```

And make this change across the places with this set, it's a set of Modes, not a set of any type of feature definition, also just call it `getModes`, and the variable in the definition `modes`,  it's not a list to start with (it's a set) and that's also the naming used in the xml, so easiest to keep it consistent.

You also need to move this up the hierarchy in interfaces to `ModeChangeGoal`, replacing `isAffectedByModeChanges`

Pugzy: You can replace instances of `SelfIdentifyingFeatureDefinition` `Mode` in this and a few other classes when used in return types and lists as it extends it.

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -71,6 +71,20 @@ public MatchModule createMatchModule(Match match) {
...
+      ImmutableSet.Builder<SelfIdentifyingFeatureDefinition> modes = ImmutableSet.builder();
+      for (String modeId : node.getValue().split("\\s")) {
+        SelfIdentifyingFeatureDefinition mode =
+            context.getFeatures().get(modeId, SelfIdentifyingFeatureDefinition.class);
+        if (mode == null) {
+          throw new InvalidXMLException("No mode with ID '" + modeId + "'", node);
+        }
+        modes.add(mode);
+      }
+      return modes.build();
```
Pablete1234:
```suggestion
      ImmutableSet.Builder<Mode> modes = ImmutableSet.builder();
      for (String modeId : node.getValue().split("\\s")) {
        Mode mode = context.getFeatures().get(modeId, Mode.class);
        if (mode == null) {
          throw new InvalidXMLException("No mode with ID '" + modeId + "'", node);
        }
        modes.add(mode);
      }
      return modes.build();
```
Get modes, not `SelfIdentifyingFeatureDefinition`s

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -1,10 +1,9 @@
 package tc.oc.pgm.core;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.List;
+import java.util.*;
```
Pugzy: Avoid wildcard imports (can be turned off in Intellij).

core/src/main/java/tc/oc/pgm/destroyable/DestroyableModule.java
```diff
@@ -111,6 +124,13 @@ public DestroyableModule parse(MapFactory context, Logger logger, Document doc)
         Set<SingleMaterialMatcher> materials =
             XMLUtils.parseMaterialPatternSet(
                 Node.fromRequiredAttr(destroyableEl, "materials", "material"));
+
+        ImmutableSet<SelfIdentifyingFeatureDefinition> modeList = null;
+        if (!destroyableEl.getAttributeValue("modes").isEmpty()) {
```
Pugzy: Should use `if (destroyableEl.getAttributeValue("modes") != null) {` rather than `isEmpty` and `!` not.

`getAttributeValue` returns the `Attribute` or `null` so calling `isEmpty` produces an NPE on maps without modes defined.

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -71,6 +71,20 @@ public MatchModule createMatchModule(Match match) {
       return ImmutableList.of(RegionModule.class, TeamModule.class);
     }
 
+    public ImmutableSet<SelfIdentifyingFeatureDefinition> parseModeSet(
+        Node node, MapFactory context) throws InvalidXMLException {
```
Pablete1234:
```suggestion
    public ImmutableSet<Mode> parseModeSet(Node node) throws InvalidXMLException {
```
Make the factory a field in the class, no need to pass it as parameter

core/src/main/java/tc/oc/pgm/core/CoreModule.java
```diff
@@ -115,6 +129,11 @@ public CoreModule parse(MapFactory context, Logger logger, Document doc)
           serialNumbers.put(owner, serial + 1);
         }
 
+        ImmutableSet<SelfIdentifyingFeatureDefinition> modeList = null;
+        if (!coreEl.getAttributeValue("modes").isEmpty()) {
+          modeList = parseModeSet(Objects.requireNonNull(Node.fromAttr(coreEl, "modes")), context);
+        }
+
         boolean modeChanges = XMLUtils.parseBoolean(coreEl.getAttribute("mode-changes"), false);
```
Pablete1234:
```suggestion
        ImmutableSet<Mode> modes;
        Node modes = Node.fromAttr(coreEl, "modes");
        if (modes != null) {
            if (coreEl.getAttribute("mode-changes") != null)
                throw new InvalidXMLException("Cannot combine modes and mode-changes");
                
            modes = parseModeSet(modes, context); // Specific set of modes
        } else if (XMLUtils.parseBoolean(coreEl.getAttribute("mode-changes"), false)) {
            modes = null; // All modes
        } else {
            modes = ImmutableSet.of(); // No modes
        }
```
Make parsing for `modes` properly support the legacy mode-changes, also make sure to throw an exception if they're combined.

You also need to remove all traces of `modeChanges` in the rest of the codebase, it is no longer a thing since we have a specific set now

Pablete1234 (changes requested):
The same comments i've done on core need to be set on Destroyable as well.

Keep in mind that `isAffectedByModeChanges` needs to completely go away from the codebase

core/src/main/java/tc/oc/pgm/goals/ModeChangeGoal.java
```diff
@@ -11,6 +11,4 @@
   boolean isObjectiveMaterial(Block block);
 
   String getModeChangeMessage(Material material);
-
-  boolean isAffectedByModeChanges();
```
Pablete1234, line 15: Consider adding here `ImmutableSet<Mode> getModes` to replace the previous method

Pablete1234 (commented):
Except for this last suggestion, lgtm

Pablete1234 (dismissed):

CoWinkKeyDinkInc:
@Pablete1234 I forgot to reformat code, please reapprove this PR.

Pablete1234 (approved):

Pablete1234:
@Electroid Could you give this a review as it looks ready to merge?

Electroid (approved):

PR #759, opened 2020-12-26
New post filters

### LINKED ISSUE #543: Allow flag posts to be filtered
Currently a flag can have multiple posts, but you cant disable a certain by filtering it.

Code example:
```
<flags>
    <flag id="flag" name="Flag" post="post" points-rate="1" shared="true" pickup-kit="flag-kit" drop-kit="reset-health" sequential="true">
        <post id="mid" name="Mid" return-time="0s" respawn-time="20s" filter="only-first-spawn">
            <block>-351.5,12,510.5</block>
        </post>
        <post id="post-a" name="Post A" return-time="0s" respawn-time="20s">
            <block>-375.5,12,510.5</block>
        </post>
        <post id="post-b" name="Post B" return-time="0s" respawn-time="20s">
            <block>-328.5,12,510.5</block>
        </post>		
    </flag>
</flags>
<filters>
	<not id="only-first-spawn">
		<time>1s</time>
	</not>
</filters>
```

In this example the flag would be in mid at first, but never respawn back there because of the time filter. Instead it would just alternate between Post A and Post B.

mrcookieunderscore13:
This adds the `respawn-filter` attribute to posts, which determine if a flag can respawn to a given post or not. It also adds the filter `previous-post` ex `<previous-post>post_id</previous-post>` which returns true if the post_id matches the post_id of the given flag's previous respawn post. 

I also added a `FlagQuery` class. 

Example:
In KOTF, to prevent flags from respawning to the same post twice in a row, an XML could look like this:

filter

post

Each post would need its own filter. 

If there are no valid posts for a flag to respawn to, it will respawn to its default post. 

Closes https://github.com/PGMDev/PGM/issues/543

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/PreviousPostFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/PreviousPostFilter.java
@@ -1,0 +1,27 @@
+package tc.oc.pgm.filters;
+
+import tc.oc.pgm.api.feature.FeatureReference;
+import tc.oc.pgm.filters.query.FlagQuery;
+import tc.oc.pgm.flag.Flag;
+import tc.oc.pgm.flag.Post;
+
+public class PreviousPostFilter extends TypedFilter<FlagQuery> {
+  private final FeatureReference<? extends Post> postReference;
+
+  public PreviousPostFilter(FeatureReference<? extends Post> post) {
+    this.postReference = post;
+  }
+
+  @Override
+  public Class<? extends FlagQuery> getQueryType() {
+    return FlagQuery.class;
+  }
+
+  @Override
+  protected QueryResponse queryTyped(FlagQuery query) {
+    final Post post = this.postReference.get();
+    final Flag flag = query.getFlag();
+    if (post == null) return QueryResponse.ABSTAIN;
+    return QueryResponse.fromBoolean(flag.getPreviousPost().equals(post));
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/filters/query/FlagQuery.java
+++ b/core/src/main/java/tc/oc/pgm/filters/query/FlagQuery.java
@@ -1,0 +1,26 @@
+package tc.oc.pgm.filters.query;
+
+import tc.oc.pgm.flag.Flag;
+
+public class FlagQuery extends GoalQuery {
+  private final Flag flag;
+
+  public FlagQuery(Flag flag) {
+    super(flag);
+    this.flag = flag;
+  }
+
+  public Flag getFlag() {
+    return flag;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof FlagQuery)) return false;
+    if (!super.equals(o)) return false;
+    FlagQuery query = (FlagQuery) o;
+    if (!flag.equals(query.flag)) return false;
+    return true;
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/flag/Flag.java
+++ b/core/src/main/java/tc/oc/pgm/flag/Flag.java
@@ -290,7 +290,6 @@ public class Flag extends TouchableGoal<FlagDefinition> implements Listener {
       if (definition.isSequential()) {
         // Start posts collection at sequential count
         Collections.rotate(possiblePosts, -sequentialPostCounter);
-        sequentialPostCounter++;
       } else {
         Collections.shuffle(possiblePosts);
       }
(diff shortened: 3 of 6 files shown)


core/src/main/java/tc/oc/pgm/flag/Flag.java
```diff
@@ -264,27 +277,61 @@ public boolean canDropAt(Location location) {
...
+              .query(query)
+              .isDenied()
+          && i < posts.size()) {
+        i++;
+      }
+      if (i >= posts.size()) {
+        returnPost = definition.getDefaultPost();
+      } else {
+        returnPost = posts.get((i + sequentialPostCounter++) % posts.size());
+        sequentialPostCounter = sequentialPostCounter + i;
```
KingOfSquares:
```suggestion
        sequentialPostCounter += i;
```

Pugzy: lots of duplicated code here too, it seems like the only difference is the collection gets randomised, maybe extra to another method passing the posts (randomised or not) as an argument.

core/src/main/java/tc/oc/pgm/flag/Flag.java
```diff
@@ -264,27 +278,48 @@ public boolean canDropAt(Location location) {
...
+
+    previousPost = returnPost;
+    return returnPost;
+  }
+
+  private Post getNextPost(Collection<Post> posts) {
+    return posts.stream()
+        .filter(
+            post -> {
+              sequentialPostCounter++;
```
Pugzy, line 307: This `sequentialPostCounter++;` is in a different location to the PR I made on your repo.

I'm not sure if this still does what you were intending. 
What is the expected functionality of the sequential post counter when posts fail the filter?

mrcookieunderscore13: I moved it here so that the counter skips over posts whose filter returns false. For example, without the counter there, for the posts A, B, and C, if A's respawn filter is always false, the spawning sequence will be BCBBCBBCBB etc. because the post counter doesn't increment extra times when it iterates over a filter returning false.

core/src/main/java/tc/oc/pgm/filters/PreviousPostFilter.java
```diff
@@ -0,0 +1,27 @@
...
+  @Override
+  public Class<? extends FlagQuery> getQueryType() {
+    return FlagQuery.class;
+  }
+
+  @Override
+  protected QueryResponse queryTyped(FlagQuery query) {
+    final Post post = this.postReference.get();
+    final Flag flag = query.getFlag();
+    if (post == null) return QueryResponse.ABSTAIN;
```
Pablete1234, line 24: This shouldn't be a possibility at all, it means a filter of an invalid post was created, some sort of exception should occur, not just an abstain

core/src/main/java/tc/oc/pgm/filters/query/FlagQuery.java
```diff
@@ -0,0 +1,26 @@
+package tc.oc.pgm.filters.query;
+
+import tc.oc.pgm.flag.Flag;
+
+public class FlagQuery extends GoalQuery {
```
Pablete1234, line 5: I don't think there's a good reason to have a flag query, it's just extending goal query and adding little to it.

Captured#trySpawn already uses a `GoalQuery` with the flag as the goal being filtered, and i'd stick with it.

On the PreviousPostFilter just make your type a GoalQuery and if the goal isn't instace of a flag simply abstain, this will make it so the nets respawn filter is allowed to filter by previous post too.

core/src/main/java/tc/oc/pgm/flag/Flag.java
Pablete1234: Modifying an external variable inside a filter which is not guaranteed to execute any specific amount of times or in any specific order is a bad idea.
Streams do not guarantee this filter code will only be executed until you find the first valid answer, it should be stateless and indepotent

core/src/main/java/tc/oc/pgm/flag/Flag.java
```diff
@@ -264,27 +278,48 @@ public boolean canDropAt(Location location) {
   private int sequentialPostCounter = 1;
 
   public Post getReturnPost(Post post) {
```
Pablete1234, line 280: I want to point out how overly complex this function has become, it used to be simply:
```
  public Location getReturnPoint(Post post) {
    return post.getReturnPoint(this, this.bannerYawProvider).clone();
  }
```

Now there's `predeterminePost`, `getReturnPoint`, `getReturnPost`, `getBannerYawProvider` to accomplish the same goal, and with this PR it only gets even increasingly more complicated.

This is screaming for a need of simplification.

Showing the next post when the flag is picked up on the carried state is the only thing that needs `predeterminePost` which may be completely incompatible with filters, you can't know what the filter output may say when it's evaluated later, and the filter should only be evaluated when flag needs to decide the next post.

Pablete1234 (changes requested):

Pablete1234:
Most of the new possibilities that this enabled are now enabled by #984, and this has been abandoned for long, so i will be closing it.

PR #760, opened 2020-12-27
Fix scoreboards breaking on long lines

Pablete1234:
When scoreboard entries with very long lines are created the scoreboard stops updating. This is visible in FFA maps since flairs make names go over 30 characters easily, eg: `10/10 ***Name` with colors will easily get to the char limit.

Old PGM sidebar rendering would simply cut them, i'm applying the same


Electroid (approved):

Pugzy:
Closes https://github.com/PGMDev/PGM/issues/756

PR #766, opened 2020-12-31
Fix several bossbar issues

Pablete1234:
Fixes: 
 - On server start, joining a match sends you an unready bar, but it is not updated and is left as just a blank bar
 - If you cycle, then join, it would leave behind the match start bossbar
 - Overtime bossbar is now yellow, and other bossbars now use old PGM bossbar colors


Electroid (approved):
Thanks!

PR #767, opened 2020-12-31
Fix observers not being able to preview villager trades

### LINKED ISSUE #571: Observers can't view villager trades
Clicking on a villager while in observers opens an empty chest GUI rather than the trading options.

BGMP:
Resolves #571.

It is to note that this fix has been tested on the latest revision of [SportPaper](https://github.com/Electroid/SportPaper/commits/master), and that the version publicly available at pgm.dev's [downloads section](https://pgm.dev/downloads) makes right clicking villagers break.

Happy new year :tada:


Electroid (approved):
Indeed, happy New Years!

Issue #54, opened 2019-11-05
Make PGM independent from SportPaper


Electroid:
Most of the custom `SportPaper` patches are either not necessary, or are also supported by `Spigot`. Ideally, `PGM` should be able to run on any, reasonable, version of `Bukkit`. Some reflection will be required, especially for `NMSHacks`, but I think that's okay.

If we can, it would also be nice if PGM could run on later versions of `Bukkit` as well, but we'll have to see if that's feasible.

AustinLMayes:
This won't work without a lot of boilerplate since SP adds a lot of events that PGM listens to

PR #778, opened 2021-01-13
Remove lang3 usage

KingOfSquares:
Necessary for #54

### DIFF
--- a/core/src/main/java/tc/oc/pgm/killreward/KillRewardMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/killreward/KillRewardMatchModule.java
@@ -11,7 +11,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import javax.annotation.Nullable;
-import org.apache.commons.lang3.mutable.MutableInt;
 import org.bukkit.event.Event;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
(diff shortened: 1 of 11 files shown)


core/src/main/java/tc/oc/pgm/killreward/KillRewardMatchModule.java
```diff
@@ -34,19 +33,19 @@
 @ListenerScope(MatchScope.RUNNING)
 public class KillRewardMatchModule implements MatchModule, Listener {
   private final Match match;
-  private final Map<UUID, MutableInt> killStreaks;
+  private final Map<UUID, Integer> killStreaks; // TODO: switch to fastutil Object2IntMap??
```
Pablete1234: Normal map is fine

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -201,7 +200,7 @@ public boolean equals(Object obj) {
 
   @Override
   public String toString() {
-    return new ToStringBuilder(this).append("id", getId()).build();
+    return "MapInfoImpl{" + "id='" + id + '\'' + ", version=" + version + '}';
```
Pablete1234: name or something?

core/src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -54,7 +51,7 @@
 import tc.oc.pgm.util.named.NameStyle;
 import tc.oc.pgm.util.nms.NMSHacks;
 
-public class MatchPlayerImpl implements MatchPlayer, Comparable<MatchPlayer> {
+public class MatchPlayerImpl implements MatchPlayer {
```
Pablete1234, line 54: This looks like a dangerous change

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
KingOfSquares: The method was not used anywhere I could find, it was especially weird that it was scoped to `MatchPlayerImpl` and not `MatchPlayer`

Pablete1234: That's all fine until there's somewhere in the code that uses a TreeSet<MatchPlayer> or similar and you're doomed

KingOfSquares: `Match` does not implement `Comparable`, do you think that is something to include back as well?

Pablete1234: It probably should for ease of use, but the example of something that could break doesn't necessarily need it to work

Electroid (approved):

Electroid:
Can you re-open this?

PR #779, opened 2021-01-13
Remove usage of Physical and EntityAction

KingOfSquares:
Necessary for #54

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
@@ -12,7 +12,6 @@ import org.bukkit.event.block.BlockDamageEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.EntityExplodeEvent;
 import org.bukkit.event.player.PlayerBucketFillEvent;
-import org.checkerframework.checker.nullness.qual.Nullable;
 import tc.oc.pgm.api.filter.query.MatchQuery;
 import tc.oc.pgm.api.tracker.info.*;
 import tc.oc.pgm.api.tracker.info.PotionInfo;
@@ -22,6 +21,8 @@ import tc.oc.pgm.tracker.info.ProjectileInfo;
 import tc.oc.pgm.util.event.GeneralizedEvent;
 import tc.oc.pgm.util.event.PlayerPunchBlockEvent;
 import tc.oc.pgm.util.event.PlayerTrampleBlockEvent;
+
+import javax.annotation.Nullable;
 
 public class CauseFilter extends TypedFilter<MatchQuery> {
 
--- a/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
@@ -1,6 +1,7 @@
 package tc.oc.pgm.filters;
 
 import com.google.common.base.Preconditions;
+import javax.annotation.Nullable;
 import org.bukkit.Material;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.LivingEntity;
@@ -21,8 +22,6 @@ import tc.oc.pgm.tracker.info.ProjectileInfo;
 import tc.oc.pgm.util.event.GeneralizedEvent;
 import tc.oc.pgm.util.event.PlayerPunchBlockEvent;
 import tc.oc.pgm.util.event.PlayerTrampleBlockEvent;
-
-import javax.annotation.Nullable;
 
 public class CauseFilter extends TypedFilter<MatchQuery> {
 
--- a/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
@@ -5,7 +5,6 @@ import org.bukkit.Material;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
-import org.bukkit.event.EntityAction;
 import org.bukkit.event.Event;
 import org.bukkit.event.block.BlockBreakEvent;
 import org.bukkit.event.block.BlockBurnEvent;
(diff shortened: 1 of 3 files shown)


util/src/main/java/tc/oc/pgm/util/event/GeneralizedEvent.java
```diff
@@ -48,14 +50,21 @@ public void setCancelled(final boolean cancel) {
     }
   }
 
-  // TODO: Un-implement Physical and EntityAction, since they are SportPaper only
-  @Override
-  public World getWorld() {
-    return getCause() instanceof Physical ? ((Physical) getCause()).getWorld() : null;
+  public @Nullable World getWorld() throws EventException {
+    if (getCause() == null) return null;
+    if (!getCause().getClass().isAssignableFrom(WorldEvent.class))
```
Pablete1234: Can't you simply instanceOf here?

KingOfSquares: Yeah, you are right

core/src/main/java/tc/oc/pgm/filters/CauseFilter.java
```diff
@@ -5,14 +5,14 @@
...
 import org.bukkit.entity.Player;
-import org.bukkit.event.EntityAction;
 import org.bukkit.event.Event;
 import org.bukkit.event.block.BlockBreakEvent;
 import org.bukkit.event.block.BlockBurnEvent;
 import org.bukkit.event.block.BlockDamageEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.EntityExplodeEvent;
 import org.bukkit.event.player.PlayerBucketFillEvent;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
Electroid: `javax`

util/src/main/java/tc/oc/pgm/util/event/GeneralizedEvent.java
```diff
@@ -48,14 +50,21 @@ public void setCancelled(final boolean cancel) {
...
   }
 
-  // TODO: Un-implement Physical and EntityAction, since they are SportPaper only
-  @Override
-  public World getWorld() {
-    return getCause() instanceof Physical ? ((Physical) getCause()).getWorld() : null;
+  public @Nullable World getWorld() throws EventException {
+    if (getCause() == null) return null;
+    if (!(getCause() instanceof WorldEvent))
+      throw new EventException("Event has no associated world");
```
Electroid: Add the event class name to the message

util/src/main/java/tc/oc/pgm/util/event/GeneralizedEvent.java
```diff
@@ -48,14 +50,75 @@ public void setCancelled(final boolean cancel) {
...
   }
 
-  @Override
-  public Entity getActor() {
-    return getCause() instanceof EntityAction ? ((EntityAction) getCause()).getActor() : null;
+  public @Nullable Entity getActor() {
+    return getActorIfPresent(getCause());
+  }
+
+  public static @Nullable Entity getActorIfPresent(Event event) {
```
Electroid, line 78: These reflection methods are going to be too expensive to run on every event. We're going to need to remember/cache the `MethodHandle` for each event class. We'll definitely need unit tests for this as well.

Electroid: Actually, instead of reflection mess, why don't we just have `instanceof` checks for the specific event sub-classes? Like `PlayerEvent`, `BlockEvent`, etc..

KingOfSquares: My idea was that reflection would cover for non-bukkit events as well.. maybe not worth the job?

KingOfSquares: As in events that just extended `Event` but still had a `World`/`Entity`

Electroid: Probably not worth it, just cover the Bukkit basics.

KingOfSquares: Ok pushed changes, issue with this solution Is that it does __not__ get the `Entity` from a `MatchPlayerEvent`, since it is non-bukkit.

util/src/main/java/tc/oc/pgm/util/event/GeneralizedEvent.java
```diff
@@ -48,14 +62,74 @@ public void setCancelled(final boolean cancel) {
...
```
Electroid, line 78: Remove the extra spaces here.

Electroid: Isn't `getActor` is a SportPaper method, right? You would want to use other getters.

util/src/main/java/tc/oc/pgm/util/event/GeneralizedEvent.java
```diff
@@ -48,14 +62,74 @@ public void setCancelled(final boolean cancel) {
...
+    return null;
+  }
+
+  /**
+   * Tries to extract a {@link World} from a event.
+   *
+   * @param event The event to look for a {@link World} in
+   */
+  public static @Nullable World getWorldIfPresent(Event event) {
+
```
Electroid: Same, collapse spacing

Electroid (changes requested):

util/src/main/java/tc/oc/pgm/util/event/GeneralizedEvent.java
KingOfSquares: > Isn't `getActor` is a SportPaper method, right? You would want to use other getters.

Ah, I did not know that!

Pablete1234 (commented):
Egreed with electroid, too much empty spacing that does little to help

Electroid (approved):

PR #780, opened 2021-01-13
Switch javax nullness with checkerframework

KingOfSquares:
Necessary for #54


Pablete1234 (commented):
Why not just use javax annotations? 
https://mvnrepository.com/artifact/javax.annotation/javax.annotation-api

Modifying every import to switch to checkerframework isn't needed

KingOfSquares:
> Why not just use javax annotations?
> https://mvnrepository.com/artifact/javax.annotation/javax.annotation-api
> 
> Modifying every import to switch to checkerframework isn't needed

My idea was that if we used checkerframework we didn't need to bring in another dependency, since adventure already ships checkerframework

Pablete1234:
> > Why not just use javax annotations?
> > https://mvnrepository.com/artifact/javax.annotation/javax.annotation-api
> > Modifying every import to switch to checkerframework isn't needed
> 
> My idea was that if we used checkerframework we didn't need to bring in another dependency, since adventure already ships checkerframework

Java annotations like notnull are (usually) not kept at runtime so it won't really save anything.
The javax annoations are pretty standard and most things are compatible with it

Electroid:
I'd rather stick with `javax`

PR #791, opened 2021-02-02
./mode is shorthand for ./mode list, likewise ./vote is shorthand for ./vote list

dentmaged:
image

`/mode` and `/modes` without any arguments will run `/mode list`. Running a child command that doesn't exist will send the parent command help (as seen with `/mode a`).

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/ModeCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/ModeCommand.java
@@ -14,6 +14,7 @@ import net.kyori.adventure.text.format.NamedTextColor;
 import org.apache.commons.lang.WordUtils;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.command.graph.DefaultCommand;
 import tc.oc.pgm.countdowns.CountdownContext;
 import tc.oc.pgm.modes.ModeChangeCountdown;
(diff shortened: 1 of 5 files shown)


Pablete1234:
Make it so `mode` and `modes` are just aliases for the same thing, and the default if no subcommand is passed is showing the list

core/src/main/java/tc/oc/pgm/command/ModesCommand.java
```diff
@@ -0,0 +1,18 @@
...
+import app.ashcon.intake.CommandException;
+import app.ashcon.intake.parametric.annotation.Default;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.util.Audience;
+
+public final class ModesCommand {
+
+  @Command(
+      aliases = {"modes"},
+      desc = "List all objectivs modes",
```
CoWinkKeyDinkInc: typo

CoWinkKeyDinkInc:
This was something that has been bothering me, I like this.

Pablete1234 (changes requested):
The implementation of all the command stuff does not belong in PGM, it belongs in intake

AustinLMayes:
`@DefaultCommand` is generic enough to exist in the base intake project itself.

Electroid:
> `@DefaultCommand` is generic enough to exist in the base intake project itself.

Agreed, let's add this to `intake`.

PR #792, opened 2021-02-04
Add team checks when vanishing players

Pugzy:
Vanish logic currently assumes all players are in observers and even attempts to force this inside the `setVanished` method.

Plugins such as `Events` can allow for a situation where a player is joining the match assigned to a team, it does this by modifying the initial team of the `MatchPlayerAddEvent` to the team they are assigned to. 

This can also cause a situation where a vanish player is forced on to a team on match cycle or on registering a tournament team.

This maintains the logic that already exists, however, stops processing the `PlayerJoinEvent` logic if the player is on a team.
And unvanishes the player with a join message if the player is set to a team from a vanished state in the `PlayerJoinEvent`.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
+++ b/core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
@@ -31,6 +31,7 @@ import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.Permissions;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchManager;
+import tc.oc.pgm.api.party.Competitor;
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/community/features/VanishManagerImpl.java
```diff
@@ -198,7 +205,13 @@ public void onUnvanish(PlayerVanishEvent event) {
 
   @EventHandler
```
Pablete1234: This event could be handled before the initial party is set to be a competitor, missing this logic. Make this happen on monitor

Pablete1234 (changes requested):

Pablete1234 (approved):

Electroid (approved):

PR #799, opened 2021-02-11
Reintroduce team switch kit

### LINKED ISSUE #732: /team force cannot bypass blitz
Previously, PGM would allow `/team force` to work during a Blitz match. However, it no longer workers.

applenick:
# Reintroduce Team Switch Kit

Quick PR to reimplement the team switch kit back into PGM. Also funny enough this fixes #732 

Thanks to @CrazyisCreeps for helping me test 😄 

A little background, this kit is primarily used for the Infection gamemode. But also could have some nifty applications when applied to new maps 😁  

## Example XML:
```xml
 <kit id="switch-to-apple-team" force="true">
     <team-switch team="apples" title="true"/>
 </kit>
```

## Video
Updated

https://user-images.githubusercontent.com/3377659/107834175-5ff79280-6d4a-11eb-8a8a-489bf0d86a22.mp4

Original Video

https://user-images.githubusercontent.com/3377659/107683199-25143280-6c56-11eb-8d2d-bd1c71e4fe2f.mp4

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/KitParser.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitParser.java
@@ -678,11 +678,12 @@ public abstract class KitParser {
     Element el = XMLUtils.getUniqueChild(parent, "team-switch");
     if (el == null) return null;
 
+    boolean showTitle = XMLUtils.parseBoolean(el.getAttribute("title"), true);
     TeamFactory team = Teams.getTeam(el.getAttributeValue("team"), factory);
     if (team == null) {
       throw new InvalidXMLException(
           el.getAttributeValue("team") + " is not a valid team name!", el);
     }
-    return new TeamSwitchKit(team);
+    return new TeamSwitchKit(team, showTitle);
   }
 }
--- a/core/src/main/java/tc/oc/pgm/kits/KitParser.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitParser.java
@@ -46,6 +46,8 @@ import tc.oc.pgm.kits.tag.ItemTags;
 import tc.oc.pgm.projectile.ProjectileDefinition;
 import tc.oc.pgm.shield.ShieldKit;
 import tc.oc.pgm.shield.ShieldParameters;
+import tc.oc.pgm.teams.TeamFactory;
+import tc.oc.pgm.teams.Teams;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.material.Materials;
 import tc.oc.pgm.util.xml.InvalidXMLException;
@@ -137,6 +139,7 @@ public abstract class KitParser {
     kits.add(this.parseFlyKit(el));
     kits.add(this.parseGameModeKit(el));
     kits.add(this.parseShieldKit(el));
+    kits.add(this.parseTeamSwitchKit(el));
     kits.addAll(this.parseRemoveKits(el));
 
     kits.removeAll(Collections.singleton((Kit) null)); // Remove any nulls returned above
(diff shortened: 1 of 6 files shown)


CoWinkKeyDinkInc:
That's pretty neat, what sort of things could/has this been used for? :D

RuedigerLP:
> That's pretty neat, what sort of things could/has this been used for? :D

Infection gamemode

core/src/main/java/tc/oc/pgm/kits/TeamSwitchKit.java
```diff
@@ -0,0 +1,23 @@
...
+
+  private final TeamFactory team;
+
+  public TeamSwitchKit(TeamFactory team) {
+    this.team = team;
+  }
+
+  @Override
+  public void applyDelayed(MatchPlayer player, boolean force) {
+    TeamMatchModule tmm = player.getMatch().getModule(TeamMatchModule.class);
```
Pablete1234: i'd just require team match module, there's no point in silently ignoring when someone tries to use this in FFA

CrazyisCreeps:
> That's pretty neat, what sort of things could/has this been used for? :D

I made a freeze tag styled map with this module back on Stratus where when tagged you were swapped to a "Frozen Team" that basically prevented you from moving. I think it used a combination of attacker and victim kits as well to really make it work.

applenick:
> > That's pretty neat, what sort of things could/has this been used for? :D
> 
> I made a freeze tag styled map with this module back on Stratus where when tagged you were swapped to a "Frozen Team" that basically prevented you from moving. I think it used a combination of attacker and victim kits as well to really make it work.

Woah that sounds cool! Would be interested to see what other cool map concepts we can implement with this too 😄

Electroid:
Could we show the `You joined X` message as a sub-title? We want to make it obvious that they have switched teams

Pablete1234:
> Could we show the `You joined X` message as a sub-title? We want to make it obvious that they have switched teams

Would be nice, however, i'd keep an XML tag to be able to turn it off

core/src/main/java/tc/oc/pgm/kits/TeamSwitchKit.java
applenick: Great idea! Will adjust 👍

applenick:
@Electroid sure thing! That's an excellent idea, may have to change how the join messages are sent but should be manageable, I'll work on this in a little bit.

@Pablete1234 also will include a tag to choose

applenick:
@Electroid & @Pablete1234 

Example of two different regions applying the team switch kit
One with `title="true"` the other with `title="false"`

https://user-images.githubusercontent.com/3377659/107834175-5ff79280-6d4a-11eb-8a8a-489bf0d86a22.mp4

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -670,4 +673,17 @@ public ShieldKit parseShieldKit(Element parent) throws InvalidXMLException {
         XMLUtils.parseDuration(el.getAttribute("delay"), ShieldParameters.DEFAULT_DELAY);
     return new ShieldKit(new ShieldParameters(health, rechargeDelay));
   }
+
+  public TeamSwitchKit parseTeamSwitchKit(Element parent) throws InvalidXMLException {
+    Element el = XMLUtils.getUniqueChild(parent, "team-switch");
+    if (el == null) return null;
+
+    boolean showTitle = XMLUtils.parseBoolean(el.getAttribute("title"), true);
```
Electroid: `show-title`

Electroid (dismissed):
Just 1 change, otherwise looks good.

applenick:
@Electroid all good now 👍

Pablete1234 (dismissed):

core/src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
```diff
@@ -573,9 +612,32 @@ public boolean kickPlayerOffTeam(Team kickFrom, boolean forBalance) {
...
+      Component joinMsg = translatable("join.ok.team", event.getNewParty().getName());
+      boolean title = false;
+
+      // Players with team switch kit, check for title value and remove from map
+      if (hasTeamSwitchKit(player)) {
+        title = showTeamSwitchTitle(player);
+        setTeamSwitchKit(player, false, false);
+      }
+
+      if (title) {
```
Pablete1234: 1.7 players can't see titles, they need to keep the chat messages

applenick: Oh good catch, I'll add in a legacy check

Pablete1234 (approved):

Electroid (approved):

PR #804, opened 2021-02-13
Spawner Module: Rework max entity management

alexsosnovsky:
The `max-entities` attribute of the spawner module was implemented in such a way that it only worked in a few ideal scenarios. It used an `isTracked()` method to determine whether or not the amount of spawned entities should be incremented or decremented, and this didn't work properly as the internal value for the amount of spawned entities easily went far down into the negatives. Additionally, there was no way to differentiate items spawned by separate spawners, so picking up one spawner's items affected other spawners.

I have implemented a `numericID` value for each spawner, which is assigned and incremented when the SpawnerDefinitions are being generated. This ID is not specifiable in XML, as it has no value to mapmakers and is for internal usage only. Each item spawned by a PGM spawner is tagged with this ID, so that items can later be traced back to their source spawners. I modified the spawner's event handlers to only change its amount of spawned entities if the item in the event is spawned by it.

There was another issue where throwing an item stackable with the spawned item could mess up the spawner's internal values, if the item stacks were to merge. To alleviate this, I added a listener for the `ItemMergeEvent` which prevents item stacks from merging if they are from different PGM spawners, or if only one of the stacks was generated by a PGM spawner. As a side note, all the items' spawner-identifying information is stripped when they are first picked up, so any spawned item a user has in their inventory is the same as if it never came from the spawner in the first place.


Electroid (approved):
Thanks!

PR #805, opened 2021-02-13
Rework stats and allow showing own stats

Pablete1234:
Moves the stats displaying to an event, which can be cancelled or thrown by 3rd party plugins to decide when stats show, additionally, PGM has a config for when it automatically should be shown, and the event itself has settings for what to show.

image
(the pic is in spanish due to local server but it shouldn't be a big issue)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -87,6 +87,9 @@ public final class PGMConfig implements Config {
   private final boolean showFireworks;
   private final boolean participantsSeeObservers;
   private final boolean verboseStats;
+  private final Integer statsShowAfter;
+  private final boolean statsShowHigh;
+  private final boolean statsShowOwn;
 
   // sidebar.*
   private final Component header;
@@ -161,7 +164,6 @@ public final class PGMConfig implements Config {
     this.matchLimit = parseInteger(config.getString("restart.match-limit", "30"));
 
     this.woolRefill = parseBoolean(config.getString("gameplay.refill-wool", "true"));
-    this.verboseStats = parseBoolean(config.getString("ui.verbose-stats", "false"));
     this.griefScore =
         parseInteger(config.getString("gameplay.grief-score", "-10"), Range.atMost(0));
 
@@ -180,6 +182,10 @@ public final class PGMConfig implements Config {
         parseBoolean(config.getString("ui.participants-see-observers", "true"));
     this.showFireworks = parseBoolean(config.getString("ui.fireworks", "true"));
     this.flagBeams = parseBoolean(config.getString("ui.flag-beams", "false"));
+    this.verboseStats = parseBoolean(config.getString("ui.verbose-stats", "false"));
+    this.statsShowAfter = parseInteger(config.getString("ui.stats.show-after", "120"));
+    this.statsShowHigh = parseBoolean(config.getString("ui.stats.show-high", "true"));
+    this.statsShowOwn = parseBoolean(config.getString("ui.stats.show-own", "true"));
 
     final String header = config.getString("sidebar.header");
     this.header = header == null || header.isEmpty() ? null : parseComponent(header);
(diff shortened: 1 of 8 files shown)


core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -180,6 +182,10 @@
         parseBoolean(config.getString("ui.participants-see-observers", "true"));
     this.showFireworks = parseBoolean(config.getString("ui.fireworks", "true"));
     this.flagBeams = parseBoolean(config.getString("ui.flag-beams", "false"));
+    this.verboseStats = parseBoolean(config.getString("ui.verbose-stats", "false"));
+    this.statsShowAfter = parseInteger(config.getString("ui.stats.show-after", "120"));
+    this.statsShowHigh = parseBoolean(config.getString("ui.stats.show-high", "true"));
```
Pugzy: Using `best` over high may be more explanatory i.e `ui.stats.show-best`

Pugzy (commented):
Using the term `best` over `high` may be more explanatory i.e ui.stats.show-best

core/src/main/java/tc/oc/pgm/api/Config.java
```diff
@@ -221,6 +221,22 @@
...
 
+  /**
+   * Gets whether to show a more verbose representation of the match stats at the end of each match
+   *
+   * @return If verbose stats at the end of the match is enabled
+   */
+  boolean showVerboseStats();
+
+  /** @return How many ticks should wait until showing stats */
+  int showStatsAfter();
```
Electroid: Let's do `Duration` here.

core/src/main/resources/config.yml
```diff
@@ -76,6 +76,10 @@ ui:
   participants-see-observers: true # Can participants see observers in the tab list?
   flag-beams: false # Should everyone see floating wool flag beams?
   verbose-stats: false # Enable more thorough stats at the end of each match
+  stats:
```
Electroid: Let's move this to its own section, so not `ui.stats`, just `stats`.

Electroid:
```yml
stats:
  verbose: true
  show-after: 3s
  show-best: true
  show-own: true
```

core/src/main/java/tc/oc/pgm/stats/StatsDisplayEvent.java
```diff
@@ -0,0 +1,56 @@
+package tc.oc.pgm.stats;
```
Electroid: This needs to be in `tc.oc.pgm.api.match.event`

core/src/main/java/tc/oc/pgm/stats/StatsDisplayEvent.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.event.MatchEvent;
+
+public class StatsDisplayEvent extends MatchEvent implements Cancellable {
```
Electroid: `MatchStatsEvent`

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -211,54 +228,60 @@ public void onMatchEnd(MatchFinishEvent event) {
...
-            },
-            5 + 1, // NOTE: This is 1 second after the votebook appears
-            TimeUnit.SECONDS);
+    List<Component> best = new ArrayList<>();
+    if (event.isShowBest()) {
+      best.add(getMessage("match.stats.kills", sortStats(allKills), NamedTextColor.GREEN));
+      best.add(getMessage("match.stats.killstreak", sortStats(allStreaks), NamedTextColor.GREEN));
+      best.add(getMessage("match.stats.deaths", sortStats(allDeaths), NamedTextColor.RED));
+
+      // bow
```
Electroid: These comments can be removed, they don't add much

Electroid (changes requested):
Also remove the `server/sportpaper.jar`, maybe add it to `.gitignore`

Electroid (approved):

PR #808, opened 2021-02-15
Add queued restart countdown time config option

Pugzy:
This allows you to configure the amount of time a restart countdown will take.
Within ranked we wish to restart the servers quicker than the typical 30 seconds but this option is not currently configurable.

This config option already exists for start, cycle, and huddle countdowns. This extends it to restart countdowns too.

```
countdown:
  start: "30s"    # After a match cycles or /start
  cycle: "30s"    # After a match ends or /cycle
  huddle: -1      # Before a match starts (only recommended for "ranked" matches)
  restart: "30s"  # After a restart countdown is queued or /qr
```

This value will be used as the default for the `queuerestart` command (unless a time is specified) and for automated restarts.

Happy for suggestions on the config hint and `Config` interface docblock to avoid confusion with the `getMatchLimit` value.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Config.java
+++ b/core/src/main/java/tc/oc/pgm/api/Config.java
@@ -82,6 +82,13 @@ public interface Config {
   Duration getCycleTime();
 
   /**
+   * Gets a duration to wait before restarting the server.
+   *
+   * @return A duration.
+   */
+  Duration getRestartTime();
+
+  /**
    * Gets a duration to wait before the server should restart.
    *
    * @return A duration, disabled if non-positive.
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -61,6 +61,7 @@ public final class PGMConfig implements Config {
   private final Duration startTime;
   private final Duration huddleTime;
   private final Duration cycleTime;
+  private final Duration restartTime;
 
   // restart.*
   private final Duration uptimeLimit;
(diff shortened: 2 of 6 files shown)


core/src/main/java/tc/oc/pgm/api/Config.java
```diff
@@ -81,6 +81,13 @@
    */
   Duration getCycleTime();
 
+  /**
+   * Gets a duration to countdown before restarting the server.
```
applenick:
```suggestion
   * Gets a duration to wait before restarting the server.
```
I think this would align with the other duration descriptions better.

Pugzy: Had it like this originally but it could be muddled with the idle time regardless will switch it back.

applenick (approved):

alexsosnovsky:
Would be nice if we could set separate cycle times for regular cycles and voted cycles, since regular cycles typically don't need as much time as voted cycles. To not break people's configs, if no voted cycle time is specified, the regular time could be used for all cycles.

Something like:
`cycle: "10s"`
`voted-cycle: "30s"`

Indicardo:
> Would be nice if we could set separate cycle times for regular cycles and voted cycles, since regular cycles typically don't need as much time as voted cycles. To not break people's configs, if no voted cycle time is specified, the regular time could be used for all cycles.
> 
> Something like:
> `cycle: "10s"`
> `voted-cycle: "30s"`

That's something probably out of this PR's reach. I would create an issue for this if you're interested

Electroid (approved):

PR #809, opened 2021-02-15
Add Lives Filter

CoWinkKeyDinkInc:
This adds a basic filter that matches players with a certain amount of lives. This is used the same way as the `<kill-streak>` filter, though there is no `repeat` attribute.

```xml
<regions>
    <apply enter="live-block" region="lives-thingy" message="You need two lives to enter here."/>
</regions>
<filters>
    <all id="live-block">
        <lives count="2"/>
    </all>
</filters>
```


Electroid (approved):
Looks good!

PR #810, opened 2021-02-16
Add Effect Filter

alexsosnovsky:
The new `effect` filter returns `ALLOW` if the player has an effect of the same type as the filter. If a mapmaker wants to be more selective, there are three optional attributes: `min-duration`, `max-duration`, and `amplifier`.

Syntax example:
```xml
<filters>
    <effect id="filter1" min-duration="10s" amplifier="1">saturation</effect>
    <effect id="filter2">saturation</effect>
    <effect id="filter3" max-duration="30s" amplifier="2">night vision</effect>
    <effect id="filter4" amplifier="1">water breathing</effect>
    <effect id="filter5" min-duration="10s" max-duration="20s" amplifier="2">night vision</effect>
</filters>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/EffectFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/EffectFilter.java
@@ -1,22 +1,21 @@
 package tc.oc.pgm.filters;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Range;
 import java.util.Collection;
 import org.bukkit.potion.PotionEffect;
 import tc.oc.pgm.api.filter.query.PlayerQuery;
 import tc.oc.pgm.api.player.MatchPlayer;
 
-public class HasEffectFilter extends ParticipantFilter {
+public class EffectFilter extends ParticipantFilter {
   protected final PotionEffect base;
-  // min/max duration are stored in ticks
-  protected final long minDuration;
-  protected final long maxDuration;
+  // duration is stored in ticks
+  protected final Range<Long> duration;
   protected final boolean amplifier;
 
-  public HasEffectFilter(PotionEffect base, long minDuration, long maxDuration, boolean amplifier) {
+  public EffectFilter(PotionEffect base, Range<Long> duration, boolean amplifier) {
     this.base = Preconditions.checkNotNull(base);
-    this.minDuration = minDuration;
-    this.maxDuration = maxDuration;
+    this.duration = duration;
     this.amplifier = amplifier;
   }
 
@@ -31,8 +30,7 @@ public class HasEffectFilter extends ParticipantFilter {
 
       if (effect.getType().equals(base.getType())
           && (!amplifier || (effect.getAmplifier() == base.getAmplifier()))
-          && ((minDuration == -1) || (effect.getDuration() >= minDuration))
-          && ((maxDuration == -1) || (effect.getDuration() <= maxDuration))) {
+          && duration.contains((long) effect.getDuration())) {
         return QueryResponse.ALLOW;
       }
     }
--- a/core/src/main/java/tc/oc/pgm/filters/EffectFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/EffectFilter.java
@@ -10,10 +10,10 @@ import tc.oc.pgm.api.player.MatchPlayer;
 public class EffectFilter extends ParticipantFilter {
   protected final PotionEffect base;
   // duration is stored in ticks
-  protected final Range<Long> duration;
+  protected final Range<Integer> duration;
   protected final boolean amplifier;
 
-  public EffectFilter(PotionEffect base, Range<Long> duration, boolean amplifier) {
+  public EffectFilter(PotionEffect base, Range<Integer> duration, boolean amplifier) {
     this.base = Preconditions.checkNotNull(base);
     this.duration = duration;
     this.amplifier = amplifier;
@@ -30,7 +30,7 @@ public class EffectFilter extends ParticipantFilter {
 
       if (effect.getType().equals(base.getType())
           && (!amplifier || (effect.getAmplifier() == base.getAmplifier()))
-          && duration.contains((long) effect.getDuration())) {
+          && duration.contains(effect.getDuration())) {
         return QueryResponse.ALLOW;
       }
     }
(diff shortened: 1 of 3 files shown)


Pablete1234:
I do not see the point in having a filter with effects as children, i think no other filter does this, instead, it's the filter itself that has the properties. 
Also, there's no use for has-*, for example team filter is simply team, even if the name itself is reused elsewhere
```xml
<filters>
    <effect id="filter1" duration="0s" amplifier="4">saturation</effect>
    <effect id="filter2" duration="30s" amplifier="1">water breathing</effect>
    <effect id="filter3" duration="0s" amplifier="2">night vision</effect>
</filters>
```

Regarding duration, a filter with a *specific* duration is of little use, it should probably be a range: `duration="2s..15s"` or `duration="2s.."`

Pugzy:
Not sure if these exists in this version of PGM but it looks like ranges and comparisons were implemented like so https://pgm.dev/docs/modules/mechanics/filters/#random-filter which may be worth replicating as the syntax used in the comment above.

alexsosnovsky:
> I do not see the point in having a filter with effects as children, i think no other filter does this, instead, it's the filter itself that has the properties.
> Also, there's no use for has-*, for example team filter is simply team, even if the name itself is reused elsewhere
> Regarding duration, a filter with a _specific_ duration is of little use, it should probably be a range: `duration="2s..15s"` or `duration="2s.."`

I've updated it to simply use the `<effect/>` tag. I also added optional `min-duration` and `max-duration` attributes so that a range can be created easily. The sample XML in the original post has been updated.


> Not sure if these exists in this version of PGM but it looks like ranges and comparisons were implemented like so https://pgm.dev/docs/modules/mechanics/filters/#random-filter which may be worth replicating as the syntax used in the comment above.

This looks interesting but I felt it was simpler and more intuitive to just have `min-duration` and `max-duration` attributes which take normal duration strings as input.

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -470,6 +470,19 @@ public WearingItemFilter parseWearingItem(Element el) throws InvalidXMLException
     return new WearingItemFilter(factory.getKits().parseRequiredItem(el));
   }
 
+  @MethodParser("effect")
+  public HasEffectFilter parseHasEffect(Element el) throws InvalidXMLException {
+    Attribute minAttribute = el.getAttribute("min-duration");
```
Pablete1234: Make use of PGM nodes  instead of XML attributes, eg:: `XMLUtils.parseDuration(Node.fromAttr(el, "min-duration"))`

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/filters/HasEffectFilter.java
```diff
@@ -0,0 +1,42 @@
+package tc.oc.pgm.filters;
+
+import com.google.common.base.Preconditions;
+import java.util.Collection;
+import org.bukkit.potion.PotionEffect;
+import tc.oc.pgm.api.filter.query.PlayerQuery;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+public class HasEffectFilter extends ParticipantFilter {
```
Pablete1234:
```suggestion
public class EffectFilter extends ParticipantFilter {
```

core/src/main/java/tc/oc/pgm/filters/HasEffectFilter.java
```diff
@@ -0,0 +1,42 @@
...
+  protected final PotionEffect base;
+  // min/max duration are stored in ticks
+  protected final long minDuration;
+  protected final long maxDuration;
```
Pablete1234: Use an actual `Range<Long>`, pass it as part of the constructor

core/src/main/java/tc/oc/pgm/filters/HasEffectFilter.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  protected QueryResponse queryPlayer(PlayerQuery query, MatchPlayer player) {
+    for (PotionEffect effect : getEffects(player)) {
+      if (effect == null) continue;
+
+      if (effect.getType().equals(base.getType())
+          && (!amplifier || (effect.getAmplifier() == base.getAmplifier()))
+          && ((minDuration == -1) || (effect.getDuration() >= minDuration))
+          && ((maxDuration == -1) || (effect.getDuration() <= maxDuration))) {
```
Pablete1234: This logic will be greatly simplified with the range, `range.contains(effect.getDuration())`

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -470,6 +471,18 @@ public WearingItemFilter parseWearingItem(Element el) throws InvalidXMLException
     return new WearingItemFilter(factory.getKits().parseRequiredItem(el));
   }
 
+  @MethodParser("effect")
+  public HasEffectFilter parseHasEffect(Element el) throws InvalidXMLException {
+    Duration minDuration = XMLUtils.parseDuration(Node.fromAttr(el, "min-duration"));
+    Duration maxDuration = XMLUtils.parseDuration(Node.fromAttr(el, "max-duration"));
```
Pablete1234, line 477: Create a Range.atMost, Range.atLeast, or Range.closed based on if the durations are present or not, and use that in the filter

core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -470,6 +471,18 @@ public WearingItemFilter parseWearingItem(Element el) throws InvalidXMLException
...
+    boolean amplifier = Node.fromAttr(el, "amplifier") != null ? true : false;
```
Pablete1234:
```suggestion
    boolean amplifier = Node.fromAttr(el, "amplifier") != null;
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/filters/EffectFilter.java
```diff
@@ -31,8 +30,7 @@ protected QueryResponse queryPlayer(PlayerQuery query, MatchPlayer player) {
 
       if (effect.getType().equals(base.getType())
           && (!amplifier || (effect.getAmplifier() == base.getAmplifier()))
-          && ((minDuration == -1) || (effect.getDuration() >= minDuration))
-          && ((maxDuration == -1) || (effect.getDuration() <= maxDuration))) {
+          && duration.contains((long) effect.getDuration())) {
```
Pablete1234: I was under the impression that getDuration was a long, seems like it's an int, so the Range should also be a `Range<Integer>`, and you should do the cast in the parser instead, but it's a somewhat minor issue

Pablete1234 (approved):

Electroid (approved):

PR #820, opened 2021-02-23
Add WorldTimeModule

CoWinkKeyDinkInc:
Finally! A way to set the time using the map XML!

This adds two tags to PGM, one that sets the map to a certain time, and another that lets the map pick from one of twelve preset times at match load.

```xml
<timeset>18000</timeset>
<randomtime/>
```

If map makers want the day time to cycle when using these elements, the map **must** have timelock disabled, regardless if DoDaylightCycle is enabled in the Minecraft map.
```xml
<randomtime/>
<timelock>off</timelock>
```
The preset random times are listed below. There is 7 bright times and 5 night times, so I think this is fairly balanced. If anyone knows of better times for shaders I'd like to hear them.
```
0
1000  /time set day
6000  noon, sun is at its peak
12000 /time set sunset
12610 dusk
13000 /time set night
13188 first tick when monsters spawn
14000
18000 midnight, moon is at its peak
21000
23000 /time set sunrise
23460 undead mobs start to burn
```


Pugzy:
I don't think these need to be two separate new modules. We could have a single `world` parent element like we have for `terrain`.
https://pgm.dev/docs/modules/environment/world/#terrain

We could then extract any remaining `level.dat` world based features that we may want to configure.

```xml
<world>
    <time random="true" /> <!--- or -->
    <time>noon</time> <!--- or -->
    <time>16000</time> <!--- or -->
    <weather rain="true" thunder="true" />
</world>
```

Or remove the need for both `randomtime` and `timeset` with something like the above (using the same module with both options configured via attributes) but name it `world-time` / `day-time` rather than `time` (to avoid match time conflicts).

Maybe one day in a protocol not too far away we can drop the `level.dat` altogether and replace with defaults and modules.

I'm not sure if there are any other nbt properties that we would need: terrain, game rules, difficulty and borders already exist. 
https://minecraft.gamepedia.com/Java_Edition_level_format#level.dat_format

Electroid:
> I don't think these need to be two separate new modules. We could have a single `world` parent element like we have for `terrain`.
> https://pgm.dev/docs/modules/environment/world/#terrain
> 
> We could then extract any remaining `level.dat` world based features that we may want to configure.
> 
> ```
> <world>
>     <time random="true" /> 
>     <time>noon</time> 
>     <time>16000</time> 
>     <weather rain="true" thunder="true" />
> </world>
> ```
> 
> Or remove the need for both `randomtime` and `timeset` with something like the above (using the same module with both options configured via attributes) but name it `world-time` / `day-time` rather than `time` (to avoid match time conflicts).
> 
> Maybe one day in a protocol not too far away we can drop the `level.dat` altogether and replace with defaults and modules.
> 
> I'm not sure if there are any other nbt properties that we would need: terrain, game rules, difficulty and borders already exist.
> https://minecraft.gamepedia.com/Java_Edition_level_format#level.dat_format

Agree, I like this format, but I'd actually call it `<properties>` and `PropertyModule`.

Pablete1234:
i think `<properties>` is too generic to have any meaning, it feels like it's just random properties you're setting while this is the worlds' level.dat options that you're setting

CoWinkKeyDinkInc:
A lot of things should be moved in this `<world>` tag like `<terrain>`, `<timelock>`, `<dimension>`, `<world-borders>`. Since `<timelock>` was its own isolated element I made these separate to keep it consistent. I think this should be done in the next map proto.

Pablete1234:
> A lot of things should be moved in this `<world>` tag like `<terrain>`, `<timelock>`, `<dimension>`, `<world-borders>`. Since `<timelock>` was its own isolated element I made these separate to keep it consistent. I think this should be done in the next map proto.

Protos are not the tool to advance syntax, they're only the way to deprecate older ones. PGM should support the best syntax no matter what proto the map is in, only breaking changes need a protocol change. In this case, the `<timelock>on</timelock>` attribute would remain as a legacy fallback.

This is similar to how `<autorespawn/>` was deprecated in favor of `<respawn auto="true"/>`, but pgm simply supports both, where the new respawn module will search for the legacy autorespawn tag  to set the default value of "auto"

CoWinkKeyDinkInc:
The timesetting modules are now a subset of a `<world>` tag. Legacy check is added for `<timelock>`

```xml
<world>
    <randomtime/>
    <timelock>off</timelock>
    <timeset>18000</timeset>
</world>
```

Pablete1234:
You've merged all the syntax xml-wise but code-wise you're still using 3 modules for this same feature, i'd say rename `TimeLockModule` to `WorldTimeModule` and  implement all 3 in there, time locking, setting, and setting to random.

Also may consider to allow for a list of `<timeset>` elements and it just picks one randomly from the list

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -278,6 +281,38 @@ public void lockTime(final MatchFinishEvent event) {
...
+      6000  noon, sun is at its peak
+      12000 /time set sunset
+      12610 dusk
+      13000 /time set night
+      13188 first tick when monsters spawn
+      14000
+      18000 midnight, moon is at its peak
+      21000
+      23000 /time set sunrise
+      23460 undead mobs start to burn
```
Pablete1234: Too arbitrary, many of these times are at night, there's 3 times where it's bright and 9 where it's dark, this would arbitrarily make most maps using random times happen during night time

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -278,6 +279,32 @@ public void lockTime(final MatchFinishEvent event) {
...
+      /*
+      1000  /time set day
+      6000  noon, sun is at its peak
+      12610 dusk
+      13000 /time set night
+      14000
+      18000 midnight, moon is at its peak
+      */
+      long[] times = {1000, 6000, 12610, 13000, 14000, 18000};
+      Random rand = new Random();
```
Electroid: Use `Match#getRandom`

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -278,6 +279,32 @@ public void lockTime(final MatchFinishEvent event) {
...
+    if (event.getMatch().getModule(WorldTimeModule.class).isTimeRandom()) {
+      /*
+      1000  /time set day
+      6000  noon, sun is at its peak
+      12610 dusk
+      13000 /time set night
+      14000
+      18000 midnight, moon is at its peak
+      */
+      long[] times = {1000, 6000, 12610, 13000, 14000, 18000};
```
Electroid: Define this in a `private static final` variable

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -50,14 +51,15 @@
...
-import tc.oc.pgm.modules.TimeLockModule;
+import tc.oc.pgm.modules.WorldTimeModule;
 import tc.oc.pgm.util.UsernameFormatUtils;
 import tc.oc.pgm.util.named.NameStyle;
 import tc.oc.pgm.util.text.TemporalComponent;
 import tc.oc.pgm.util.text.TextTranslations;
 
 public class PGMListener implements Listener {
   private static final String DO_DAYLIGHT_CYCLE = "doDaylightCycle";
+  private static final long[] times = {1000, 6000, 12610, 13000, 14000, 18000};
```
Electroid:
```suggestion
  // Move the /* */ description here
  private static final long[] WORLD_TIMES = {1000, 6000, 12610, 13000, 14000, 18000};
```

Electroid (changes requested):

Electroid (approved):

PR #825, opened 2021-03-06
Make completion sounds for control points half as loud

Pablete1234:
Usually in Koth maps the sounds are extremely spammy and very lound, it's deafening. This makes all those sounds half as loud as well as opening the door to other objectives having their own sounds


Electroid:
@Pablete1234 Code is failing to format, otherwise I can merge

Pablete1234:
> @Pablete1234 Code is failing to format, otherwise I can merge

Fixed it, had apparently added one accidental newline after formatting and never checked back

Electroid (approved):

PR #833, opened 2021-03-12
Fix Timelock Evaluation

CoWinkKeyDinkInc:
This fixes maps that intentionally use the DoDayLightCycle gamerule to stop time and do not use `<timelock>` like ~BlockBlock,~ ~BlocksCTW~, BoomBox, etc, and maps that intentionally have timelock turned off but DoDayLightCycle set to false, such as Sunrise Over Paradise and ~Moonlight Summit, both of~ which are to have sunrise and sunset. 

WorldTimeModule will pass a false if it does not find a `<timelock>` module. Bit confusing dealing with two booleans that deal with the same thing innit, so now there's a third boolean `isLockAbsent` to record if `<timelock>` even exists in the XML.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
@@ -271,7 +271,13 @@ public class PGMListener implements Listener {
   @EventHandler
   public void unlockTime(final MatchStartEvent event) {
     boolean unlockTime = false;
-    if (!event.getMatch().getModule(WorldTimeModule.class).isTimeLocked()) {
+    boolean isTimeLocked = event.getMatch().getModule(WorldTimeModule.class).isTimeLocked();
+    boolean isDaylightCycle =
+        Boolean.parseBoolean(event.getMatch().getWorld().getGameRuleValue(DO_DAYLIGHT_CYCLE));
+    // isTimeLocked() will be false if no timelock module is found
+    // if timelock is false and daylightcycle is true, set unlockTime to true
+    // if timelock is false and daylightcycle is false, set to true
+    if ((!isTimeLocked && isDaylightCycle) || !isTimeLocked && !isDaylightCycle) {
       unlockTime = true;
     }
 
--- a/core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
@@ -76,8 +76,6 @@ public class PGMListener implements Listener {
   // Single-write, multi-read lock used to create the first match
   private final ReentrantReadWriteLock lock;
 
-  private boolean isDaylightCycle = false;
-
   public PGMListener(Plugin parent, MatchManager mm, VanishManager vm) {
     this.parent = parent;
     this.mm = mm;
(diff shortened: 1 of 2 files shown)


Pablete1234:
Simply make timeLock a `Boolean` with true, false, or null, depending on if it's on, off, or absent

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -271,7 +271,13 @@ public void lockTime(final MatchLoadEvent event) {
...
-    if (!event.getMatch().getModule(WorldTimeModule.class).isTimeLocked()) {
+    Boolean isTimeLocked = event.getMatch().getModule(WorldTimeModule.class).isTimeLocked();
+    boolean isDaylightCycle =
+        Boolean.parseBoolean(event.getMatch().getWorld().getGameRuleValue(DO_DAYLIGHT_CYCLE));
+    // isTimeLocked() will be null if no timelock module is found
+    // if there is no timelock module and daylight cycle is enabled, unlock time
+    // if there is timelock module and its off, unlock time
+    if (((isTimeLocked == null) && isDaylightCycle) || (isTimeLocked != null && !isTimeLocked)) {
       unlockTime = true;
     }
```
Pablete1234:
```java
Boolean isTimeLocked = event.getMatch().getModule(WorldTimeModule.class).isTimeLocked();
boolean isDaylightCycle =
    Boolean.parseBoolean(event.getMatch().getWorld().getGameRuleValue(DO_DAYLIGHT_CYCLE));
boolean unlockTime = isTimeLocked == null ? isDaylightCycle : !isTimeLocked;
```

This can be heavily simplified

However, if you look at the event on line 268, the gamerule is forced to false to prevent the match traversing while the map is loaded, i think queriying isDaylightCycle will always return false.

Check out what the behavior was prior to all of the time/world changes and keep it that way

CoWinkKeyDinkInc: I have moved the isDayLightCycle boolean so it can be accessible to the first lockTime function, and record the gamerule before it locks the time. For some reason when I check the debugging it always seems to return true, even when I tested maps that have the gamerule to false like Moonlight Summit and Sunrise. This does not make a difference on what maps should have timelock though.

CoWinkKeyDinkInc:
@Pablete1234 The original TimeLockModule [defaulted to on](https://github.com/PGMDev/PGM/blob/c25a41bbf10c2e652beb4d829bc47b5cd6b16451/core/src/main/java/tc/oc/pgm/modules/TimeLockModule.java) despite the documentation saying that it defaults to off. This will involve having to change a lot of maps that do not have a timelock like BlockBlock, to intentionally have one. Very quick map changes, but my older commits unintentionally used this logic. Should we try to preserve compatibility here?

We could just check if the timelock exists, and if it is off and leave it at that.

Pablete1234:
> @Pablete1234 The original TimeLockModule [defaulted to on](https://github.com/PGMDev/PGM/blob/c25a41bbf10c2e652beb4d829bc47b5cd6b16451/core/src/main/java/tc/oc/pgm/modules/TimeLockModule.java) despite the documentation saying that it defaults to off. This will involve having to change a lot of maps that do not have a timelock like BlockBlock, to intentionally have one. Very quick map changes, but my older commits unintentionally used this logic. Should we try to preserve compatibility here?
> 
> We could just check if the timelock exists, and if it is off and leave it at that.

Keep the old behavior, always.
If it defaulted to off, do the same. Don't keep the gamerule value if it didn't before, it will just break maps

Pablete1234 (approved):

Electroid (approved):

PR #834, opened 2021-03-18
Support Spigot 1.8

### LINKED ISSUE #54: Make PGM independent from SportPaper

cswhite2000:
This still compiles against SportPaper, but I've changed enough of the references to SportPaper specific methods for it to be able to run the maps in https://github.com/PGMDev/Maps without issue. I haven't as of yet tested this on maps not in that repository.

I wasn't really sure how best to handle EventHandlers for events not in Spigot, so I've commented those out for now.
I'm also not sure about how best to handle attributes in spigot 1.8, since those weren't added to spigot until 1.9, so those are also commented out for now.


Electroid:
Wow, thanks @cswhite2000, also hello again!

I had a conversation in Discord about this a while ago, but (if it helps) we can target Paper 1.8. I won't have time to review this until the weekend, but will tag @Pablete1234 to take a look as I think he may have some opinions/ideas.

Electroid:
Also, since you've dived into this: how feasible is it to no longer compile against SportPaper? What would you see as the hurdles?

applenick:
@Electroid 
Just wanna inject, for nicknames to be supported in Community & PGM, the FakeName & Skin API in SportPaper is of interest 👀 

Also excellent job @cswhite2000 🎉

cswhite2000:
No longer compiling against SportPaper is feasible. You would lose a few features that require sportpaper, and that can't be replicated with reflection.

There's also probably a number of places that use SportPaper APIs that I've missed, as in things not used by the default maps, and those would have to handled too.

And then it would be fairly easy to go from checking to see if SportPaper is running before running SportPaper specific methods to just removing those method calls.

I think ideally, pgm should compile against Paper, but check before using paper specific methods, so that it still runs on spigot.

Electroid:
So, basically we need a way to listen to events that may or may not exist in the Bukkit runtime.

What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.

Thoughts?

cswhite2000:
That sounds like it would work.

util/src/main/java/tc/oc/pgm/util/collection/ImmutableMaterialSet.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.common.collect.ForwardingSet;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Set;
+import org.bukkit.Material;
+
+/**
+ * Highly efficient immutable container for {@link Material}s. Useful for sharing large material
```
Electroid: Kinda laughed at "highly efficient", given it's just a wrapper ;)

cswhite2000:
> What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.

This doesn't seem to work.

Electroid:
> > What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.
> 
> This doesn't seem to work.

What doesn't work about it? Is the package the same?

Pablete1234:
You could make a separate PGM module that is pgm-sportpaper, that compiles against sportpaper and adds listeners to sportpaper events and launches PGM events.
You can compile pgm against spigot, and that other module compiles against sportpaper, and you are able to compile pgm without that module too if you wish to run it on just spigot

cswhite2000:
> > > What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.
> > 
> > 
> > This doesn't seem to work.
> 
> What doesn't work about it? Is the package the same?

Yeah the package is the same. I copied BlockFallEvent into util but I still got 
```java.lang.NoClassDefFoundError: org/bukkit/event/block/BlockFallEvent``` 
when trying to run it

File: util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java [hunk 2]
```diff
@@ -75,14 +83,45 @@ static void sendPacketToViewers(Entity entity, Object packet) {
...
+
+        Constructor<PacketPlayOutPlayerInfo.PlayerInfoData> constructor =
+            PacketPlayOutPlayerInfo.PlayerInfoData.class.getConstructor(
+                PacketPlayOutPlayerInfo.class,
+                GameProfile.class,
+                int.class,
+                WorldSettings.EnumGamemode.class,
+                IChatBaseComponent.class);
+
+        constructor.setAccessible(true);
```
Pablete1234: Cache the constructor and save it

util/src/main/java/tc/oc/pgm/util/reflect/ReflectionUtils.java
```diff
@@ -25,4 +25,21 @@ public static Object readField(@Nullable Object obj, Field field) {
       field.setAccessible(wasAccessible);
     }
   }
+
+  public static void setField(Class<?> parent, Object base, Object value, String fieldName) {
+    try {
+      setField(base, value, parent.getDeclaredField(fieldName));
```
Pablete1234, line 83: Generally try to avoid repeated calls to get fields, from past experience java isn't very fast when creating `Field`s and it creates a new one each time you call. Prefer calling once to get the `Field` and reusing it, especially for things like the tablist where it already is a decent performance bottleneck from the amount of calls it can produce with high playercounts

Pablete1234 (commented):
I feel like every place where `BukkitUtils.isSportPaper()` is used should contain a comment about what happens about it.

Some features look straight-out disabled, like attribute kits or canplace/candestroy. If these do not work without sportpaper we either need alternatives for other platforms or we need to make it very clear in the documentation that these features will simply not work outside sportpaper.
It's essentially turning them into second-class features that maps could avoid because they don't work, meaning maps may try to avoid using them for being "non-compliant" with the "standard set of features that works everywhere". Should definetly try to look into alternatives.

One decent way to do this would be to move some of these into NMSHacks, and have differing implementations for NMSHacks, one that can work for spigot via NMS or reflection if needed, and another that can make those methods use directy sportpaper apis.

Pablete1234:
> > > > What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.
> > > 
> > > 
> > > This doesn't seem to work.
> > 
> > 
> > What doesn't work about it? Is the package the same?
> 
> Yeah the package is the same. I copied BlockFallEvent into util but I still got
> `java.lang.NoClassDefFoundError: org/bukkit/event/block/BlockFallEvent`
> when trying to run it

Are you adding the event in that package inside PGM?

cswhite2000:
> > > > > What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.
> > > > 
> > > > 
> > > > This doesn't seem to work.
> > > 
> > > 
> > > What doesn't work about it? Is the package the same?
> > 
> > 
> > Yeah the package is the same. I copied BlockFallEvent into util but I still got
> > `java.lang.NoClassDefFoundError: org/bukkit/event/block/BlockFallEvent`
> > when trying to run it
> 
> Are you adding the event in that package inside PGM?

Yes

util/src/main/java/tc/oc/pgm/util/reflect/ReflectionUtils.java
cswhite2000: So would it be best to cache all of the fields used in the packet creation methods in NMSHacks?

Pablete1234: For all the fields that are called often, yeah

cswhite2000:
> Some features look straight-out disabled, like attribute kits or canplace/candestroy. If these do not work without sportpaper we either need alternatives for other platforms or we need to make it very clear in the documentation that these features will simply not work outside sportpaper.
> It's essentially turning them into second-class features that maps could avoid because they don't work, meaning maps may try to avoid using them for being "non-compliant" with the "standard set of features that works everywhere". Should definetly try to look into alternatives.

I've added more comments explaining these. Attributes and canplace/candestroy should be doable with reflection, however they would take a significant amount of code / time, so I should think they should be done in their own pull request.

> One decent way to do this would be to move some of these into NMSHacks, and have differing implementations for NMSHacks, one that can work for spigot via NMS or reflection if needed, and another that can make those methods use directy sportpaper apis.

That does sound like a good way to do it

core/src/main/java/tc/oc/pgm/listeners/AntiGriefListener.java
```diff
@@ -60,10 +59,11 @@ public void rightClickDefuse(final PlayerInteractEntityEvent event) {
     this.participantDefuse(event.getPlayer(), event.getRightClicked());
   }
 
-  @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
-  public void leftClickDefuse(final PlayerAttackEntityEvent event) {
-    this.participantDefuse(event.getPlayer(), event.getLeftClicked());
-  }
+  // TODO: This is only in SportPaper
+  //  @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+  //  public void leftClickDefuse(final PlayerAttackEntityEvent event) {
```
KingOfSquares: Cant this be switched with a `PlayerInteractEvent`?

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -311,13 +312,13 @@ public void freezeWorld(final BlockTransformEvent event) {
...
 
-  @EventHandler
-  public void freezeVehicle(final VehicleUpdateEvent event) {
-    Match match = this.mm.getMatch(event.getWorld());
-    if (match == null || match.isFinished()) {
-      event.getVehicle().setVelocity(new Vector());
-    }
-  }
+  //  @EventHandler
+  //  public void freezeVehicle(final VehicleUpdateEvent event) {
```
KingOfSquares: Can you just capture `VehicleEvent` here?

core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
```diff
@@ -43,7 +44,12 @@ public void checkMatchProjectiles() {
               if (projectile.isDead() || projectile.isOnGround()) {
                 projectile.removeMetadata(TRAIL_META, PGM.get());
               } else {
-                final Color color = (Color) projectile.getMetadata(TRAIL_META, PGM.get()).value();
+                final Color color =
+                    (Color)
+                        (BukkitUtils
+                                .isSportPaper() // SportPaper allows for getting metadata by plugin
+                            ? projectile.getMetadata(TRAIL_META, PGM.get()).value()
```
KingOfSquares: Could not this be abstracted out into something like

`getMetadata(Entity entity, String meta)`

And then in that method you do the either A or B method, to avoid boilerplate.

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
```diff
@@ -226,14 +225,14 @@ public void onObserverInteract(final ObserverInteractEvent event) {
...
-  @EventHandler
-  public void onAttackEntity(final PlayerAttackEntityEvent event) {
-    MatchPlayer player = match.getPlayer(event.getPlayer());
-    if (player != null) {
-      State state = states.get(player);
-      if (state != null) state.onEvent(event);
-    }
-  }
+  //  @EventHandler
+  //  public void onAttackEntity(final PlayerAttackEntityEvent event) {
```
KingOfSquares: On second thought, maybe a `ParticipantInteractEvent` or similar should be added to match `ObserverInteractEvent`, instead of using `PlayerInteractEvent` for these.

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
```diff
@@ -269,8 +268,15 @@ public void matchEnd(final MatchFinishEvent event) {
     }
   }
 
+  //  // TODO: this should only be for initial spawn
```
KingOfSquares: Can you use `PlayerLoginEvent`?

util/src/main/java/tc/oc/pgm/util/tablist/TabManager.java
```diff
@@ -141,11 +140,12 @@ public void onWorldChange(PlayerChangedWorldEvent event) {
...
-    }
-  }
+  // TODO: This is not in spigot
+  //  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+  //  public void onSkinPartsChange(PlayerSkinPartsChangeEvent event) {
+  //    TabEntry entry = this.getPlayerEntryOrNull(event.getPlayer());
+  //    if (entry instanceof PlayerTabEntry) {
+  //      ((PlayerTabEntry) entry).onSkinPartsChange(event);
+  //    }
+  //  }
```
KingOfSquares: This does not seem to be necessary to support, logging out/in is reasonable to expect for a refresh

util/src/main/java/tc/oc/pgm/util/tablist/TabRender.java
```diff
@@ -146,18 +161,37 @@ public void refreshEntry(TabEntry entry, int index) {
...
+              net.md_5.bungee.chat.ComponentSerializer.toString(header)),
+          "a");
+      ReflectionUtils.setField(
+          packet.getClass(),
+          packet,
+          IChatBaseComponent.ChatSerializer.a(
+              net.md_5.bungee.chat.ComponentSerializer.toString(footer)),
+          "b");
+      send(packet);
+    }
```
KingOfSquares: Tablist header and footer is supported by adventure

core/src/main/java/tc/oc/pgm/listeners/AntiGriefListener.java
cswhite2000: It can be but we'd have to draw a ray from the player for every single PlayerInteractEvent, which I don't think would be worth it given how there is still the right click defuse.

core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
cswhite2000: Yeah, that would be a better strategy.

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
cswhite2000: Maybe

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
cswhite2000: Actually this can be safely removed, since the PlayerSpawnLocationEvent does the exact same thing and is called in NMS right after the SportPaper PlayerInitialSpawnEvent.

cswhite2000:
> > What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.
> 
> This doesn't seem to work.

I believe this is why it didn't work
https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/4c766c92ac4d10daeedee256e096106570d0e9b9

Electroid:
> > > What if in `util`, we copy-paste the events we need from `SportPaper` with the same package path, same signature. We can import those events in our code. If the runtime already defines the event class, it does not get overriden. If the runtime doesn't have the class, also no problem, the event just never gets thrown.
> > 
> > 
> > This doesn't seem to work.
> 
> I believe this is why it didn't work
> https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/4c766c92ac4d10daeedee256e096106570d0e9b9

Yep, that’ll be the problem. Can we load the class outside of PluginClassLoader?

util/src/main/java/tc/oc/pgm/util/tablist/TabManager.java
Pablete1234: A re-login will not necessarily cause an update. Skin flags may be sent after login with a settings packet, that would likely be ignored. The result from not having this is no overlay skin layer ever showing in tab

Pablete1234:
> > I believe this is why it didn't work
> > https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/commits/4c766c92ac4d10daeedee256e096106570d0e9b9
> 
> Yep, that’ll be the problem. Can we load the class outside of PluginClassLoader?

I don't think you'll ever be able to load those classes. Bukkit/spigot load plugins with its own class loader, the PluginClassLoader. Due to class loader hierarchy if you try to create your own class loader in PGM to boot those classes yourself, the parent class loader will still prevent them from going thru.

cswhite2000:
So at this point pgm should run on sportpaper with all of the features it had before.

With spigot I haven't implemented the following:

- [x] - Attributes
- [x] - canPlaceOn / canDestroy
- [ ] - None of the events in tc.oc.pgm.util.event.sport are called unless the server is running sportpaper

cswhite2000:
I can't think of any good ways to call any of the events in tc.oc.pgm.util.event.sport when the server isn't running SportPaper.
Aside from that, this should be done.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
```diff
@@ -112,31 +111,19 @@ static void sendPacketToViewers(Entity entity, Object packet) {
         profile.getProperties().putAll(entry.getKey(), entry.getValue());
       }
     }
-
-    if (BukkitUtils.isSportPaper()) {
```
Pablete1234: Don't try to cheap out and use reflection for everything.

Direct access is **significantly** faster, and for tablist that's a must

cswhite2000: That's fair

core/pom.xml
```diff
@@ -45,6 +45,7 @@
                     <artifactSet>
                         <includes>
                             <include>tc.oc.pgm:*:*</include>
+                            <include>it.unimi.dsi:fastutil:*</include>
```
Electroid: Where do we use this? This is a big library, so I'm wondering if there's a way we can do this without it?

Electroid: Maybe we add a utility in `CollectionUtils` (I think that exists?) to create a collection with this, otherwise it fallback to the standard library. Newer server runtimes ship with this and I'd rather not bundle it.

core/src/main/java/tc/oc/pgm/crafting/CraftingMatchModule.java
```diff
@@ -27,8 +29,14 @@ public CraftingMatchModule(
...
-    for (Recipe recipe : customRecipes) {
-      match.getWorld().addRecipe(recipe);
+    if (BukkitUtils.isSportPaper()) {
+      for (Recipe recipe : customRecipes) {
+        match.getWorld().addRecipe(recipe);
+      }
+    } else {
+      for (Recipe recipe : customRecipes) {
+        Bukkit.addRecipe(recipe);
+      }
```
Electroid: Let's move this into a `WorldUtils` or `BukkitUtils` as `addRecipe(World, Recipe)`

core/src/main/java/tc/oc/pgm/crafting/CraftingMatchModule.java
```diff
@@ -37,7 +45,11 @@ public void disable() {
     // Recipe changes affect all worlds on the server, so we make changes at match start/end
     // to avoid interfering with adjacent matches. If we wait until unload() to reset them,
     // the next match would already be loaded.
-    match.getWorld().resetRecipes();
+    if (BukkitUtils.isSportPaper()) {
```
Electroid: Same here.

core/src/main/java/tc/oc/pgm/itemmeta/ItemRule.java
```diff
@@ -42,24 +46,34 @@ public void apply(ItemStack stack) {
 
       InventoryUtils.addEnchantments(meta, this.meta.getEnchants());
 
-      for (String attribute : this.meta.getModifiedAttributes()) {
-        for (AttributeModifier modifier : this.meta.getAttributeModifiers(attribute)) {
-          meta.addAttributeModifier(attribute, modifier);
+      // Since SportPaper handles attributes they don't show up in unhandledTags
```
Electroid: Let's also put this into a utility.

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -126,13 +128,16 @@ protected KitDefinition parseDefinition(Element el) throws InvalidXMLException {
       kits.add(this.parse(child));
     }
 
+    if (BukkitUtils.isSportPaper()) {
+      // This is only possible in SportPaper
+      kits.add(this.parseKnockbackReductionKit(el));
+    }
```
Electroid: Instead of guarding from the call site, it's better to guard within the method. (This applies to other things as well)

Reason being: if the method is called from other places, it makes it harder to misuse.

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -443,8 +449,21 @@ public void parseItemMeta(Element el, ItemMeta meta) throws InvalidXMLException
...
+            entry.getKey(),
+            new org.bukkit.attribute.AttributeModifier(
+                attributeModifier.getUniqueId(),
+                attributeModifier.getName(),
+                attributeModifier.getAmount(),
+                org.bukkit.attribute.AttributeModifier.Operation.fromOpcode(
+                    attributeModifier.getOperation().ordinal())));
+      }
+    } else {
+      NMSHacks.applyAttributeModifiers(parseAttributeModifiers(el), meta);
```
Electroid: Utility method needed.

core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
```diff
@@ -43,7 +46,12 @@ public void checkMatchProjectiles() {
               if (projectile.isDead() || projectile.isOnGround()) {
                 projectile.removeMetadata(TRAIL_META, PGM.get());
               } else {
-                final Color color = (Color) projectile.getMetadata(TRAIL_META, PGM.get()).value();
+                Color color = null;
+                for (MetadataValue mv : projectile.getMetadata(TRAIL_META)) {
+                  if (mv.getOwningPlugin().equals(PGM.get())) {
+                    color = (Color) mv.value();
```
Electroid: Add a `break`

core/src/main/java/tc/oc/pgm/modules/ProjectileTrailMatchModule.java
```diff
@@ -115,15 +132,19 @@ public void onProjectileLaunch(ProjectileLaunchEvent event) {
...
       final Projectile projectile = event.getEntity();
       projectile.removeMetadata(TRAIL_META, PGM.get());
       // Restore critical metadata to arrows if applicable
       if (projectile instanceof Arrow) {
         final Arrow arrow = (Arrow) projectile;
         if (arrow.hasMetadata(CRITICAL_META)) {
-          arrow.setCritical(arrow.getMetadata(CRITICAL_META, PGM.get()).asBoolean());
+          for (MetadataValue mv : arrow.getMetadata(CRITICAL_META)) {
+            if (mv.getOwningPlugin().equals(PGM.get())) {
+              arrow.setCritical(mv.asBoolean());
```
Electroid: Add a `break`

core/src/main/java/tc/oc/pgm/tablist/PlayerOrder.java
```diff
@@ -61,6 +61,6 @@ public int compare(MatchPlayer ma, MatchPlayer mb) {
     }
 
     // All else equal, order the players alphabetically
-    return a.getName(viewer).compareToIgnoreCase(b.getName(viewer));
+    return a.getName().compareToIgnoreCase(b.getName());
```
Electroid, line 64: Is this going to cause problems?

util/src/main/java/tc/oc/pgm/util/attribute/Attribute.java
```diff
@@ -0,0 +1,45 @@
+package tc.oc.pgm.util.attribute;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public enum Attribute {
```
Electroid, line 6: Let's not do enum. Either an interface or class should be fine here.

util/src/main/java/tc/oc/pgm/util/attribute/AttributeInstance.java
```diff
@@ -0,0 +1,70 @@
+package tc.oc.pgm.util.attribute;
+
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class AttributeInstance {
+
+  private final net.minecraft.server.v1_8_R3.AttributeInstance handle;
```
Electroid: We want to move away from importing NMS directly. This looks like a `NMSHacks` worthy thing, using reflection?

util/src/main/java/tc/oc/pgm/util/attribute/AttributeInstance.java
```diff
@@ -0,0 +1,70 @@
+package tc.oc.pgm.util.attribute;
+
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class AttributeInstance {
```
Electroid: Ideally, this is an interface. There is a `SportPaper` implementation, a `Spigot` one, and eventually we would have implementations in newer versions.

util/src/main/java/tc/oc/pgm/util/attribute/AttributeMap.java
```diff
@@ -0,0 +1,41 @@
+package tc.oc.pgm.util.attribute;
+
+import com.google.common.base.Preconditions;
+import net.minecraft.server.v1_8_R3.AttributeMapBase;
+import org.bukkit.craftbukkit.v1_8_R3.entity.CraftPlayer;
+import org.bukkit.entity.Player;
+
+public class AttributeMap {
```
Electroid: Similar abstraction comment with `AttributeInstance`.

util/src/main/java/tc/oc/pgm/util/collection/ImmutableMaterialSet.java
```diff
@@ -0,0 +1,81 @@
...
+ * lists.
+ */
+public class ImmutableMaterialSet extends ForwardingSet<Material> {
```
Electroid: Do we actually need this? Can we just use a `Set<Material>` wherever this is used?

util/src/main/java/tc/oc/pgm/util/compatability/SportPaperListener.java
```diff
@@ -0,0 +1,127 @@
...
+    if (sportEvent instanceof Cancellable) {
+      ((Cancellable) pgmEvent).setCancelled(((Cancellable) sportEvent).isCancelled());
+      Bukkit.getServer().getPluginManager().callEvent(pgmEvent);
+      ((Cancellable) sportEvent).setCancelled(((Cancellable) pgmEvent).isCancelled());
+    } else {
+      Bukkit.getServer().getPluginManager().callEvent(pgmEvent);
+    }
+  }
+
+  @EventHandler(ignoreCancelled = true)
```
Electroid, line 32: Does the event priority matter here?

util/src/main/java/tc/oc/pgm/util/compatability/SportPaperListener.java
```diff
@@ -0,0 +1,127 @@
...
+    handleCall(pgmEvent, sportEvent);
+  }
+
+  @EventHandler(ignoreCancelled = true)
+  public void onSkinPartsChange(org.bukkit.event.player.PlayerSkinPartsChangeEvent sportEvent) {
+    Set<org.bukkit.Skin.Part> sportParts = sportEvent.getParts();
+    Set<Skin.Part> pgmParts = EnumSet.noneOf(Skin.Part.class);
+    for (org.bukkit.Skin.Part part : sportParts) {
+      pgmParts.add(Skin.Part.values()[part.ordinal()]);
+    }
```
Electroid: We never use the skin parts, can we just support the minimum? (Which is just notifying a skin change)

util/src/main/java/tc/oc/pgm/util/event/sport/entity/EntityExtinguishEvent.java
```diff
@@ -0,0 +1,24 @@
+package tc.oc.pgm.util.event.sport.entity;
```
Electroid: We don't need the `.sport.` part. But, maybe we should define a `@SportPaper` annotation?

util/src/main/java/tc/oc/pgm/util/skin/Skin.java
```diff
@@ -0,0 +1,67 @@
+package tc.oc.pgm.util.skin;
+
+/** A self-contained skin */
+public class Skin {
+  // NOTE: ordinals must match bit positions in packet data
+  public static enum Part {
```
Electroid: Stated before, but let's not implement the skin parts, part. Just the data and signature should be ok.

util/src/main/java/tc/oc/pgm/util/tablist/TabRender.java
```diff
@@ -146,18 +160,22 @@ public void refreshEntry(TabEntry entry, int index) {
...
+    getBField(this.updatePingPacket)
+        .add(NMSHacks.playerListPacketData(this.updatePingPacket, entry.getId(), entry.getPing()));
   }
 
   public void setHeaderFooter(BaseComponent[] header, BaseComponent[] footer) {
-    view.getViewer().setPlayerListHeaderFooter(header, footer);
+    BungeeComponentSerializer serializer = BungeeComponentSerializer.get();
+    Audience.get(view.getViewer())
+        .sendPlayerListHeaderAndFooter(
+            serializer.deserialize(header), serializer.deserialize(footer));
```
Electroid: This will be very inefficient. We're going `Component` -> `BaseComponent` -> `Component` -> `BaseComponent` -> `String`. Can we just keep the old way of doing it?

util/src/main/java/tc/oc/pgm/util/text/PlayerComponent.java
```diff
@@ -60,9 +60,11 @@ public static Component player(
 
     NameDecorationProvider provider = NameDecorationProvider.DEFAULT;
     if (player != null) {
-      MetadataValue metadata =
-          player.getMetadata(NameDecorationProvider.METADATA_KEY, BukkitUtils.getPlugin());
-      if (metadata != null) provider = (NameDecorationProvider) metadata.value();
+      for (MetadataValue mv : player.getMetadata(NameDecorationProvider.METADATA_KEY)) {
+        if (mv.getOwningPlugin().equals(BukkitUtils.getPlugin())) {
+          provider = (NameDecorationProvider) mv.value();
```
Electroid: Insert `break`

util/src/main/java/tc/oc/pgm/util/text/PlayerComponent.java
```diff
@@ -60,9 +60,11 @@ public static Component player(
 
     NameDecorationProvider provider = NameDecorationProvider.DEFAULT;
     if (player != null) {
-      MetadataValue metadata =
-          player.getMetadata(NameDecorationProvider.METADATA_KEY, BukkitUtils.getPlugin());
-      if (metadata != null) provider = (NameDecorationProvider) metadata.value();
+      for (MetadataValue mv : player.getMetadata(NameDecorationProvider.METADATA_KEY)) {
```
Electroid: I've seen this pattern a lot, sounds like it needs a utility.

util/src/main/java/tc/oc/pgm/util/text/TextTranslations.java
```diff
@@ -220,6 +221,32 @@ public static long loadKeys(Locale locale) {
...
+    } catch (IllegalArgumentException e) {
+      // ignore
+    }
+
+    // bad locale sent?
+    return java.util.Locale.US;
+  }
+
+  public static Locale getLocale(@Nullable CommandSender sender) {
+    if (sender == null || !(sender instanceof CraftPlayer)) {
```
Electroid: Why `CraftPlayer`? Use `Player` instead?

Electroid (changes requested):
This is really great work, @cswhite2000. You're taking this in the right direction, but I think we need more abstractions (separate interfaces/implementations).

Right now, a lot of the pattern is:
```java
if (BukkitUtils.isSportPaper()) {
 // SportPaper stuff
} else {
 // Fallback, sometimes invoking NMS
}
```

I worry about too much of that in the codebase, because we have two medium-term goals:
1. Don't depend on SportPaper API (reflection only)
2. Potentially support newer server versions

Obviously, we aren't going to get there 100% in this PR, but we need to keep that in mind. 

For most of my comments, I suggest moving the logic into a utility method (either in `BukkitUtils` or `NMSHacks` depending on what it does). But for `Attribute`, I think we should try a interface/implementation approach (you can take some inspiration from `ItemTag`)

It would be great if we no longer import the `SportPaper` APIs, and just use reflection, but if you think that's too much for now, we can punt that to a v2.

Pablete1234:
> It would be great if we no longer import the `SportPaper` APIs, and just use reflection, but if you think that's too much for now, we can punt that to a v2.

Keep in mind that reflection is a way slower alternative to using a direct API. PGM would aim to support both, a package that is specific to sportpaper and implements NMSHacks by calling sportpaper APIs, and a generic NMSHacks that works by using reflection. Also for 1.16 support, there could be a specific implementation that makes use of the apis in spigot 1.16 etc

core/src/main/java/tc/oc/pgm/tablist/PlayerOrder.java
cswhite2000: This will cause some strange ordering if another plugin is using the SportPaper fake name api

util/src/main/java/tc/oc/pgm/util/compatability/SportPaperListener.java
cswhite2000: Not unless another plugin is handling SportPaper specific events

util/src/main/java/tc/oc/pgm/util/tablist/TabRender.java
cswhite2000: Well I think the alternative without doing a fair bit of refactoring is 
```
    if (BukkitUtils.isSportPaper()) {
      view.getViewer().setPlayerListHeaderFooter(header, footer);
    } else {
      PacketPlayOutPlayerListHeaderFooter packet = new PacketPlayOutPlayerListHeaderFooter();
      ReflectionUtils.setField(
          packet.getClass(),
          packet,
          IChatBaseComponent.ChatSerializer.a(
              net.md_5.bungee.chat.ComponentSerializer.toString(header)),
          "a");
      ReflectionUtils.setField(
          packet.getClass(),
          packet,
          IChatBaseComponent.ChatSerializer.a(
              net.md_5.bungee.chat.ComponentSerializer.toString(footer)),
          "b");
      send(packet);
    }
```
Not sure which is better

util/src/main/java/tc/oc/pgm/util/attribute/AttributeInstance.java
cswhite2000: I copied this from the SportPaper CraftAttributeInstance class. This is class is a wrapper for the NMS AttributeInstance class. Not sure how we'd move this to NMSHacks since we do need to store the NMSAttributeInstance.
We could change the handle type to object and move all the calls to NMShack and just cast the handle back to AttributeInstance in NMSHacks

util/src/main/java/tc/oc/pgm/util/attribute/AttributeInstance.java
cswhite2000: We won't need a separate implementation for this for Spigot and SportPaper. I'll make an interface for this though.

util/src/main/java/tc/oc/pgm/util/attribute/Attribute.java
cswhite2000: Are you sure?
The enum methods are used in XMLParsing to make sure its a valid attribute

```    
    tc.oc.pgm.util.attribute.Attribute attribute = tc.oc.pgm.util.attribute.Attribute.byName(text);
    if (attribute != null) return attribute;

    attribute = tc.oc.pgm.util.attribute.Attribute.byName("generic." + text);
    if (attribute != null) return attribute;
```
And for going through and reseting all the attributes on the player in MatchPlayerImpl::reset

util/src/main/java/tc/oc/pgm/util/tablist/TabRender.java
cswhite2000: Ignore above, refactoring was easier than I though.

cswhite2000:
> It would be great if we no longer import the SportPaper APIs, and just use reflection, but if you think that's too much for now, we can punt that to a v2.

Yeah, I think that's too much for now

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -93,11 +91,7 @@ public void onPrePlayerLogin(final AsyncPlayerPreLoginEvent event) {
     // Create the match when the first player joins
     if (lock.writeLock().tryLock()) {
       // If the server is suspended, need to release so match can be created
-      final Server server = parent.getServer();
-      // Non Sport servers cannot be suspended
-      if (BukkitUtils.isSportPaper() && server.isSuspended()) {
-        server.setSuspended(false);
-      }
+      NMSHacks.unSuspendServer();
```
Electroid: `resumeServer`

Electroid (changes requested):

Electroid (dismissed):
Made a minor comment, but this is ready to go. I was testing it locally and works like a charm. If you have any other minor tweaks you can commit them. Otherwise, let me know and this is good to merge! 🚢

Electroid:
Oh, also rebase to `dev`

cswhite2000:
> Made a minor comment, but this is ready to go. I was testing it locally and works like a charm. If you have any other minor tweaks you can commit them. Otherwise, let me know and this is good to merge! 🚢

All good 👍 @Electroid

core/src/main/java/tc/oc/pgm/inventory/ViewInventoryMatchModule.java
```diff
@@ -333,7 +332,7 @@ protected void previewPlayerInventory(Player viewer, PlayerInventory inventory)
     Player holder = (Player) inventory.getHolder();
     // Ensure that the title of the inventory is <= 32 characters long to appease Minecraft's
     // restrictions on inventory titles
-    String title = StringUtils.substring(holder.getDisplayName(viewer), 0, 32);
+    String title = StringUtils.substring(holder.getDisplayName(), 0, 32);
```
Pablete1234: This *will* reveal nicked players. Make use of a PlayerComponent to render the name:

`TextTranslations.translateLegacy(player(holder, NameStyle.CONCISE, viewer), viewer) `
with a static import for PlayerComponent.player

core/src/main/java/tc/oc/pgm/itemmeta/ItemRule.java
```diff
@@ -42,24 +42,22 @@ public void apply(ItemStack stack) {
...
       stack.setItemMeta(meta);
     }
   }
-
-  private static ImmutableMaterialSet unionMaterials(
-      ImmutableMaterialSet a, ImmutableMaterialSet b) {
-    if (a.containsAll(b)) return a;
-    if (b.containsAll(a)) return b;
-    return ImmutableMaterialSet.of(Sets.union(a, b));
-  }
```
Pablete1234: Consider that this was likely a thing for performance reasons. If one of the sets contains the other, making a new one and adding everything will likely be way slower. I'm not sure how often this runs but likely often enough that it's worth keeping this logic for performance

core/src/main/java/tc/oc/pgm/listeners/AntiGriefListener.java
```diff
@@ -97,7 +97,7 @@ private void participantDefuse(Player player, Entity entity) {
                 + TextTranslations.translate(
                     "moderation.defuse.player",
                     clicker.getBukkit(),
-                    owner.getBukkit().getDisplayName(clicker.getBukkit()) + ChatColor.RED));
+                    owner.getBukkit().getDisplayName() + ChatColor.RED));
```
Pablete1234: Same as above here, this should use a player component instead

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -428,17 +428,16 @@ public void send(
...
   }
 
   private MatchPlayer getApproximatePlayer(Match match, String query, CommandSender sender) {
     return StringUtils.bestFuzzyMatch(
         query,
         match.getPlayers().stream()
-            .collect(
-                Collectors.toMap(
-                    player -> player.getBukkit().getName(sender), Function.identity())),
+            .collect(Collectors.toMap(player -> player.getBukkit().getName(), Function.identity())),
```
Pablete1234, line 440: This again will be an annoyance with nicks, but it's not something that would have to be resolved with a player component either way. It will just be something to change for integrations.

Can stay as it is for now.

core/src/main/java/tc/oc/pgm/spawner/Spawner.java
```diff
@@ -122,44 +119,30 @@ public void onItemMerge(ItemMergeEvent event) {
...
-      for (MetadataValue mv : event.getItem().getMetadata(METADATA_KEY)) {
-        if (mv.getOwningPlugin().equals(PGM.get()) && mv.asInt() == definition.numericID) {
-          event.getItem().removeMetadata(METADATA_KEY, PGM.get());
-          spawnedEntities -= event.getItem().getItemStack().getAmount();
-          spawnedEntities = Math.max(0, spawnedEntities);
-          return;
-        }
-      }
-    }
+    if (event.isCancelled()) return;
```
Pablete1234: Why not just `ignoreCancelled = true` in the annotation?

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
```diff
@@ -506,20 +851,343 @@ static double getAbsorption(LivingEntity entity) {
   }
 
   static int getProtocolVersion(Player player) {
-    return ((CraftPlayer) player).getHandle().playerConnection.networkManager.protocolVersion;
+    if (BukkitUtils.isSportPaper()) {
+      return ((CraftPlayer) player).getHandle().playerConnection.networkManager.protocolVersion;
+    } else {
+      return ViaUtils.VERSION_1_8; // Assume 1.8 if not sportpaper.
+    }
```
Pablete1234: Tbh this can just be removed, it never worked to start with.
`ViaUtils` is where protocol versions are actually resolved, the `protocolVersion` in sportpaper is ALWAYS 1.8 from what i've seen if viaversion API fails to be loaded as soft dependency (which happens if you run PGM from the IDE)

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
```diff
@@ -506,20 +851,343 @@ static double getAbsorption(LivingEntity entity) {
...
+
+  // Not relevant if not SportPaper
+  static void resumeServer() {
+    if (BukkitUtils.isSportPaper() && Bukkit.getServer().isSuspended())
+      Bukkit.getServer().setSuspended(false);
+  }
+
+  static void setAffectsSpawning(Player player, boolean affectsSpawning) {
+    if (BukkitUtils.isSportPaper()) {
+      player.spigot().setAffectsSpawning(affectsSpawning);
```
Pablete1234, line 1168: Is this not a spigot method?

util/src/main/java/tc/oc/pgm/util/tablist/TabEntry.java
```diff
@@ -54,6 +56,13 @@
    */
   BaseComponent[] getContent(TabView view);
 
+  /**
+   * Content to show in the entry
+   *
+   * @return
+   */
+  Component getComponent(TabView view);
+
```
Pablete1234: Having two methods to get the component is ambiguous and generally a bad idea. 
It was changed to `BaseComponent[]` for performance reasons when rendering complex components (like playerrnames) to avoid extra layers of nesting on serialization of the component.

If you wish, you can have this as an abstract method in `SimpleTabEntry`, and implement the `BaseComponent[]` version there, that by default will convert the `Component` to `BaseComponent[]` (which is what you're doing either way) but it should not be part of the interface.

With the current changes, any external tab entry implementation that implements `BaseComponent[]` but not the `Component` version will be either different for 1.7 players than 1.8 players, or it will simply break, depending on if it extends pgm's tab entry or not.

util/src/main/java/tc/oc/pgm/util/tablist/TabView.java
```diff
@@ -229,21 +230,14 @@ private void renderLegacy() {
           if (this.slots[i] == this.rendered[i] && !this.slots[i].isDirty(this)) continue;
           this.rendered[i] = this.slots[i];
 
-          this.display.set(x, y, toPlain(this.rendered[i].getContent(this)));
+          this.display.set(
+              x, y, TextTranslations.translateLegacy(this.rendered[i].getComponent(this), viewer));
```
Pablete1234: See the coment above regarding this, it is bound to break from differing implementations.

Pablete1234 (changes requested):

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
cswhite2000: This is in Paper 1.8, but not spigot 1.8

util/src/main/java/tc/oc/pgm/util/tablist/TabEntry.java
cswhite2000: Every single implementation of getContent for TabEntry in this repository makes an adventure component and then calls TextTranslations.toBaseComponentArray to turn it into a BaseComponent. I added the Component version so that we could call the adventure set playerlistheaderfooter method, which actually cuts down on serialization/deserialization for that. Also, at least for the pgm implementations, this won't affect performance for the rest of the tab list.

util/src/main/java/tc/oc/pgm/util/tablist/TabView.java
cswhite2000: This cuts down on serialization and deserialization.

util/src/main/java/tc/oc/pgm/util/tablist/TabEntry.java
cswhite2000: I've removed the original getContent method, which resolves the ambiguity. This will still break any external tab entry that implements the BaseComponent version.

cswhite2000:
This now depends on https://github.com/Electroid/SportPaper/pull/66 
With this, the tab list should be faster, with one step less of serialization/deserialization, and PGM will be using adventure Components for the tab list.

Also, this should make multi version support easier since later versions don't support BaseComponents for the tab list either

util/src/main/java/tc/oc/pgm/util/tablist/TabView.java
Pablete1234: It does, but it's wrong. Editing the result of BaseComponent[] does not reflect here, and it's something used by extending implementations

util/src/main/java/tc/oc/pgm/util/tablist/TabEntry.java
Pablete1234: Take this as an example:
https://github.com/bolt-rip/ingame/blob/master/src/main/java/rip/bolt/ingame/utils/RankedTeamTabEntry.java

It works fast due to being able to edit the component by simply using array manipulation, works fast and works well. This isn't the only place it's used

util/src/main/java/tc/oc/pgm/util/tablist/TabView.java
cswhite2000: I see your point

util/src/main/java/tc/oc/pgm/util/tablist/TabEntry.java
cswhite2000: So then the way I see it is it's a tradeoff, is it worth making a breaking API change for this.

Switching the tab list to using Adventure Components reduces serialization and deserialization of components.

As an aside, the linked RankedTeamTabEntry class would still be faster if it used an Adventure Component; in its current state the TeamTabEntry superclass makes an Adventure Component then converts it to a BaseComponent Array then the RankedTeamTabEntry class modifies that array, and then that is serialized by SportPaper to a gson string. It would be faster to override the Adventure Component version from this pull request which would then be directly converted to a gson string. Although you do know better than me what else this would break.

I think it's worth it to get rid of BaseComponents from the Tab list and increment the PGM version number for a breaking API change.

@Electroid  thoughts?

util/src/main/java/tc/oc/pgm/util/text/TextTranslations.java
```diff
@@ -310,9 +309,8 @@ public static String translate(String key, @Nullable CommandSender sender, Objec
     return LegacyComponentSerializer.legacySection().serialize(translate(text, locale));
   }
 
-  public static BaseComponent[] toBaseComponentArray(
-      Component component, @Nullable CommandSender viewer) {
+  public static String toMinecraftGson(Component component, @Nullable CommandSender viewer) {
     Component translated = translate(component, viewer == null ? SOURCE_LOCALE : getLocale(viewer));
-    return BungeeComponentSerializer.get().serialize(translated);
+    return GsonComponentSerializer.gson().serialize(translated);
```
Electroid: Need to use the color downsampling version of `GsonComponentSerializer`?

cswhite2000: Oh, yeah you're right

Electroid:
Some people will try using an older version of `SportPaper`, does PGM handle the case where that patch isn't introduced?

cswhite2000:
> Some people will try using an older version of `SportPaper`, does PGM handle the case where that patch isn't introduced?

It'll fall back to the spigot compatible version if the field doesn't exist

PR #837, opened 2021-03-23
Switch titles to ticks over milliseconds

Pugzy:
**Change the use of `ofMillis` to `fromTicks` on two instances where this was not updated**

These titles would previously not display. Noticed this earlier during the match start 3.. 2.. 1.. when nothing appeared it felt odd, turns out its been like this for 3 months 🤷 .

**Send both warn message and title to non-legacy players**

Noticed the warn message also used the old miliseconds method and did not show in chat for non-legacy players. A longer warning message or message that is sent at the same as another title would not be readable to the player so having it in chat for them too seems logical.

Another thing I noticed is the "Go!" now displays for everyone even the observers which previously did not happen (as seen in this [tournament match from 4 months ago](https://youtu.be/2E3SXMAiZVM). Open to ideas on what this could be as I liked that as an observer I wasn't told to go.


Electroid (approved):
Nice catch, thanks!

PR #840, opened 2021-03-28
Better spectate

KingOfSquares:
Fixes the bug where players in spectating mode would bug out whenever the player they are spectating died or changed teams.

A couple of QOL changes for the spectator as well:
- Kills/Deaths related to the player you are spectating will show up in bold in chat. (And on death setting OWN, so always)
- When the player you are spectating dies the you will see the same death message as the player you are spectating


core/src/main/java/tc/oc/pgm/spawns/states/Spawning.java
```diff
@@ -100,11 +100,13 @@ protected boolean trySpawn() {
...
   public void updateTitle() {
-    player.showTitle(
+    Title title =
         title(
             getTitle(),
             getSubtitle().color(NamedTextColor.GREEN),
-            Title.Times.of(Duration.ZERO, fromTicks(3), fromTicks(3))));
+            Title.Times.of(Duration.ZERO, fromTicks(3), fromTicks(3)));
+    player.showTitle(title);
+    player.getSpectators().forEach(p -> p.showTitle(title));
```
applenick: This will display the same respawn title correct?
Maybe we could slightly alter the message based on the viewer, so instead of saying 
`Left click to respawn`

Either hide the subtitle for spectators or add a custom one to mention `{0} is respawning`

Although a majority of players will understand it's to mirror the target, someone may accidentally click or get confused 😆 

Just an idea to consider, well done @KingOfSquares 👍

KingOfSquares: I thought about that as well, but could not think of something that would be as concise as the normal death message, which would also include info from the original death message (time until respawn, player can left click to spawn, can't respawn yet etc.)

KingOfSquares: I like the " is respawning" one, will implement

core/src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -433,6 +434,18 @@ public void tick(Match match, Tick tick) {
     return audience;
   }
 
+  @Nullable
+  @Override
+  public MatchPlayer getSpectatorTarget() {
+    if (getBukkit() == null) return null;
+    return match.getPlayer(getBukkit().getSpectatorTarget());
```
TheMolkaPL: Looks like a race condition.

```suggestion
    Player bukkit = getBukkit();
    return bukkit == null ? null : match.getPlayer(bukkit.getSpectatorTarget());
```

core/src/main/java/tc/oc/pgm/modules/SpectateMatchModule.java
```diff
@@ -0,0 +1,103 @@
...
+import tc.oc.pgm.spawns.events.ParticipantPostSpawnEvent;
+import tc.oc.pgm.util.collection.DefaultMapAdapter;
+
+@ListenerScope(MatchScope.LOADED)
+public class SpectateMatchModule implements MatchModule, Listener {
+
+  private final Match match;
+
+  // Stores which players (the list) is spectating a player(the uuid)
+  private final Map<UUID, List<UUID>> spectators = new DefaultMapAdapter<>(new ArrayList<>(), true);
```
TheMolkaPL: Can be replaced with an `ArrayListMultimap`.

core/src/main/java/tc/oc/pgm/modules/SpectateMatchModule.java
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Get the {@link MatchPlayer}s currently spectating the {@link MatchPlayer} that holds the given
+   * {@link UUID}, if any.
+   */
+  public List<MatchPlayer> getSpectating(UUID player) {
+    final List<UUID> list = spectators.get(player);
+    if (list == null) return ImmutableList.of();
+    return Collections.unmodifiableList(
+        list.stream().map(match::getPlayer).filter(Objects::nonNull).collect(Collectors.toList()));
```
TheMolkaPL, line 77:
```suggestion
    return list.stream().map(match::getPlayer).filter(Objects::nonNull).collect(ImmutableList.toImmutableList());
```

TheMolkaPL (changes requested):

core/src/main/java/tc/oc/pgm/modules/SpectateMatchModule.java
```diff
@@ -0,0 +1,103 @@
...
+          // While testing delaying by 1 tick also sometimes bugged out, so the
+          // re-enter to spectate is delayed by 3 ticks for safety.
+          player.getBukkit().setSpectatorTarget(null);
+          player.getBukkit().teleport(event.getPlayer().getBukkit());
+          match
+              .getExecutor(MatchScope.LOADED)
+              .schedule(
+                  () -> player.getBukkit().setSpectatorTarget(event.getPlayer().getBukkit()),
+                  3 * 50,
+                  TimeUnit.MILLISECONDS);
```
Pablete1234: Consider that you could make it so that visibility isn't removed while dead if you're spectating. in other words, editing the `canSee` function so you can continue to see the dead player

KingOfSquares: visibility removed? Not sure I understand what you mean

Pablete1234: Players being able to see eachother is what i mean by "visibility". Normally when a player dies, no one is able to see them, despawn packets for that entity are sent to all viewers. You could change the `canSee` function so that if you're spectating that player it does not despawn for you, so tit doesn't generate that "broken camera" look at all

core/src/main/java/tc/oc/pgm/modules/SpectateMatchModule.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @EventHandler
+  public void onPlayerTeleport(PlayerTeleportEvent event) {
+    if (event.getCause() == PlayerTeleportEvent.TeleportCause.UNKNOWN) {
+      MatchPlayer player = match.getPlayer(event.getPlayer());
+      if (player != null) {
+        MatchPlayer spectating = player.getSpectatorTarget();
+        if (spectating != null) { // Player is going into spectate
+          spectators.get(spectating.getId()).add(player.getId());
```
TheMolkaPL, line 61: I think you should normally use https://guava.dev/releases/23.0/api/docs/com/google/common/collect/Multimap.html#put-K-V-

TheMolkaPL (dismissed):
Looks good.

core/src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -199,6 +200,8 @@ public boolean canInteract() {
 
   @Override
   public boolean canSee(MatchPlayer other) {
+    @Nullable MatchPlayer spectatorTarget = this.getSpectatorTarget();
+    if (spectatorTarget != null && spectatorTarget.getId().equals(other.getId())) return true;
     if (!other.isVisible()) return false;
     if (other.isParticipating()) return true;
     if (other.isVanished() && !getBukkit().hasPermission(Permissions.VANISH)) return false;
```
Pablete1234: If someone you're spectating vanishes you still see them? nope

Don't return true in this case, rather ignore the specifically other not visible clausule, leave everything else as-is.

```suggestion
    @Nullable MatchPlayer spectatorTarget = this.getSpectatorTarget();
    boolean isSpectatorTarget = spectatorTarget != null && spectatorTarget.getId().equals(other.getId());
    if (!other.isVisible() && !isSpectatorTarget) return false;
    if (other.isParticipating()) return true;
    if (other.isVanished() && !getBukkit().hasPermission(Permissions.VANISH)) return false;
```

Pablete1234 (changes requested):

Pablete1234 (dismissed):

Pablete1234 (dismissed):

Electroid (approved):

PR #842, opened 2021-04-01
Fix several issues with flag beams

### LINKED ISSUE #811: Maintain a player's nametag when they pick up a flag
When a player picks up a flag their nametag just disappears. It seems rather odd and you can't seem them through walls like usual  (if you're in a building for example and the beam isn't visible)

Pablete1234:
Prevents legacy flag beams loading if there are no flags, additionally, it includes a number of bugfixes and improvements to flags like showing the head item as wool for legacy players and moving the up a few blocks so the hitboxes don't block hits from above

Closes #811

image


Electroid (approved):
Nice!

PR #844, opened 2021-04-01
Include observers and offline players when showing match stats

KingOfSquares:
Also re-orders and improves some logic

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Datastore.java
+++ b/core/src/main/java/tc/oc/pgm/api/Datastore.java
@@ -1,6 +1,7 @@
 package tc.oc.pgm.api;
 
 import java.util.UUID;
+import org.bukkit.Skin;
 import tc.oc.pgm.api.map.MapActivity;
 import tc.oc.pgm.api.player.Username;
 import tc.oc.pgm.api.setting.Settings;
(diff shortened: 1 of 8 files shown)


core/src/main/java/tc/oc/pgm/db/CacheDatastore.java
```diff
@@ -59,6 +70,16 @@ public Settings getSettings(UUID id) {
...
+  @Override
+  public void setSkin(UUID uuid, Skin skin) {
+    skins.put(uuid, skin);
+  }
+
+  @Override
+  public Skin getSkin(UUID id) {
+    return skins.getUnchecked(id);
+  }
+
```
KingOfSquares, line 82: Should this store a `DatabaseSkin` object to be similar to the other stuff?

Pablete1234 (approved):

core/src/main/java/tc/oc/pgm/api/Datastore.java
```diff
@@ -24,6 +25,22 @@
...
+   */
+  void setSkin(UUID uuid, Skin skin);
+
+  /**
+   * Get the skin for a given player {@link UUID}.
+   *
+   * @param uuid The {@link UUID} of a player
+   * @return A {@link Skin}
+   */
+  Skin getSkin(UUID uuid);
```
Electroid, line 42: Skins will explode the size of the database.. are we sure?

Electroid: Oh, I just realized it's only implemented by cache, ok.

KingOfSquares: Yes, my thought as well :+1:

PR #845, opened 2021-04-02
Fix match start message being shown on finished matches

Pablete1234:
Currently if a match is ended before it starts, the message about it starting is sent regardless

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
@@ -59,6 +59,7 @@ public class MatchAnnouncer implements Listener {
   @EventHandler(priority = EventPriority.MONITOR)
   public void onMatchBegin(final MatchStartEvent event) {
     Match match = event.getMatch();
+    if (match.isFinished()) return;
     match.sendMessage(translatable("broadcast.matchStart", NamedTextColor.GREEN));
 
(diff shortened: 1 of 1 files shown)


Electroid:
I'm confused.. how has the match finished.. in the `MatchStartEvent`? It sounds like we shouldn't be firing the event in the first place?

Pablete1234:
> I'm confused.. how has the match finished.. in the `MatchStartEvent`? It sounds like we shouldn't be firing the event in the first place?

This can happen if a match start happens, and anything during the match start events decides to /end it. If running integrations this is possible

The result ends up being the game over appearing first, then the match start message

Electroid:
I'm just worried this is going to cause all sorts of trouble. There are lots of things that happen at start, if the match is actually considered finished at that point, it's going to cause all sorts of edge-cases: like this one. I'd rather defer match ends to the next tick, and integrations shouldn't start the match if they anticipate on ending it immediately?

Pablete1234:
> I'm just worried this is going to cause all sorts of trouble. There are lots of things that happen at start, if the match is actually considered finished at that point, it's going to cause all sorts of edge-cases: like this one. I'd rather defer match ends to the next tick, and integrations shouldn't start the match if they anticipate on ending it immediately?

It's been done for a long while, it works just fine, the only issue is the message being shown. The match state changes are all atomic and they can happen one inside another, pgm does not bug out from this and handles it well

Electroid (approved):
Alright, good to go then.

PR #847, opened 2021-04-05
Track burning kills from being punched into lava

Pablete1234:
Makes it so if you're punched into lava and end up burning to death, you'll be reported as a kill for who knocked you into lava, awarding kill rewards to them


Electroid (approved):
Nice!!

PR #849, opened 2021-04-11
Fix flag beams dying in the void

Pablete1234:
Due to the recent refactor that made it so flag beams don't re-create each time, it means that if they're lost in the void they will never come back.

This shows them one tick later on cycle to ensure they always show (otherwise they'd sometimes not load at first) and prevents the beam going under Y=-64, where the wither skull for 1.7 players despawns client-side making the beam be lost and never re-created


Electroid (dismissed):

Electroid:
@Pablete1234 Formatting error

PR #853, opened 2021-04-19
Add long render distance tnt

Pablete1234:
Renders TNT entities far away as blocks to the clients. Additionally, it also spawns explosion particles for that case, where vanilla would just not show anything at all

image

### DIFF
--- a/core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
@@ -1,0 +1,102 @@
+package tc.oc.pgm.tntrender;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import org.bukkit.Effect;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.entity.TNTPrimed;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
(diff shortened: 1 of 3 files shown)


CoWinkKeyDinkInc:
I didn't know PGM could fix this, great work!

core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
```diff
@@ -0,0 +1,101 @@
...
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.nms.NMSHacks;
+
+public class TNTRenderMatchModule implements MatchModule, Listener {
+
+  private static final double MAX_DISTANCE = Math.pow(64d, 2);
+
+  private final Match match;
+  private final List<PrimedTnt> entities = new ArrayList<>();
```
Electroid: Move to constructor?

Electroid: Also, can we clear it on `disable`?

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
Pablete1234: I'll move it to constructor, i can not clear it on disable or it'd leave tnt hanging if match ended before it explodes for people far away, which reminds me i need to change the executor from running to LOADED. 
As the TNT finishes blowing up it will cleanup itself.

PR #856, opened 2021-04-23
Add "team-color" attribute to Armor Kits

destial:
When ever I create maps, most of the time I have to create a separate kit for the teams while trying to find the correct color that corresponds to their color.

Like so:
```xml
<helmet color="cd00cd" unbreakable="true" material="leather helmet"/>
```

So, I added an attribute that automatically uses the team's color so that I can just use the default kit and it will automatically use the team's color when applied.

Like so:
```xml
<helmet team-color="true" unbreakable="true" material="leather helmet"/>
```

I made it so "team-color" overrides the "color" attribute since it is applied during the match and not beforehand.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
+++ b/core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
@@ -40,12 +40,13 @@ public class ArmorKit extends AbstractKit {
     for (Map.Entry<ArmorType, ArmorItem> entry : this.armor.entrySet()) {
       int slot = entry.getKey().ordinal();
       if (force || wearing[slot] == null || wearing[slot].getType() == Material.AIR) {
-        if (entry.getValue().teamColor) {
-          LeatherArmorMeta meta = (LeatherArmorMeta) entry.getValue().stack.getItemMeta();
-          meta.setColor(player.getParty().getFullColor());
-          entry.getValue().stack.setItemMeta(meta);
-        }
         wearing[slot] = entry.getValue().stack.clone();
+
+        if (entry.getValue().teamColor) {
+          LeatherArmorMeta meta = (LeatherArmorMeta) wearing[slot].getItemMeta();
+          meta.setColor(player.getParty().getFullColor());
+          wearing[slot].setItemMeta(meta);
+        }
 
         KitMatchModule kitMatchModule = player.getMatch().getModule(KitMatchModule.class);
         if (kitMatchModule != null) {
--- a/core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
+++ b/core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
@@ -4,12 +4,14 @@ import java.util.List;
 import java.util.Map;
 import org.bukkit.Material;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.LeatherArmorMeta;
 import tc.oc.pgm.api.player.MatchPlayer;
 
 public class ArmorKit extends AbstractKit {
   public static class ArmorItem {
     public final ItemStack stack;
     public final boolean locked;
+    public boolean teamColor = false;
 
     public ArmorItem(ItemStack stack, boolean locked) {
       this.stack = stack;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
```diff
@@ -37,6 +39,11 @@ public void applyPostEvent(MatchPlayer player, boolean force, List<ItemStack> di
     for (Map.Entry<ArmorType, ArmorItem> entry : this.armor.entrySet()) {
       int slot = entry.getKey().ordinal();
       if (force || wearing[slot] == null || wearing[slot].getType() == Material.AIR) {
+        if (entry.getValue().teamColor) {
+          LeatherArmorMeta meta = (LeatherArmorMeta) entry.getValue().stack.getItemMeta();
+          meta.setColor(player.getParty().getFullColor());
+          entry.getValue().stack.setItemMeta(meta);
+        }
```
Pablete1234: This is modifying the item that should be immutable, apply this to the cloned stack instead (one line lower).
Also, i don't think you need to `setItemMeta` if you're getting the one and setting color

CoWinkKeyDinkInc:
I don't really see what is the issue here that this resolves, a colour picker is easy to find and search online while referring to the documentation in another tab while writing XML. Since they are hex colours, what would be the default values you would use for the team colours?

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -203,7 +203,11 @@ public FlyKit parseFlyKit(Element el) throws InvalidXMLException {
     }
     ItemStack stack = parseItem(el, true);
     boolean locked = XMLUtils.parseBoolean(el.getAttribute("locked"), false);
-    return new ArmorKit.ArmorItem(stack, locked);
+    ArmorKit.ArmorItem armorItem = new ArmorKit.ArmorItem(stack, locked);
+    if (stack.getItemMeta() instanceof LeatherArmorMeta) {
+      armorItem.teamColor = XMLUtils.parseBoolean(el.getAttribute("team-color"), false);
```
Pablete1234: Pass this as part of the constructor, and keep ArmorItem immutable, making teamColor final.
You can do something along the lines of:
`boolean autoColor = stack.getIteamMeta() instanceof LeatherArmorMeta && XMLUtils.parseBoolean(el.getAttribute("team-color"), false);`

It may be worth considering throwing an XML exception if a map combines team-color attribute with non-leather armor

Pablete1234 (changes requested):

Pablete1234:
> I don't really see what is the issue here that this resolves, a colour picker is easy to find and search online while referring to the documentation in another tab while writing XML. Since they are hex colours, what would be the default values you would use for the team colours?

It's simply convenience, also removes the need of having 3 kits, one main one and one per each team that have the other as parent, it simplifies XML

core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
destial: I've tried it without setItemMeta and it gives me uncolored armor.

core/src/main/java/tc/oc/pgm/kits/KitParser.java
destial: I have added the changes you requested

core/src/main/java/tc/oc/pgm/kits/ArmorKit.java
KingOfSquares: > Also, i don't think you need to `setItemMeta` if you're getting the one and setting color

`getItemMeta` returns a clone and not a reference of the meta

Pugzy:
Did the Stratus (or was it Walrus) implementation work with blocks too (glass, clay, wool etc)?

I recall some funky repair integration too. Anyone got a link to their docs?

Electroid:
Since this is syntax sugar, it'd be nice to make it even simpler: `color=<hex|true|false>`

destial:
color is parsed under `parseItemMeta` and set's the item meta once passed. it doesnt take the armorkit as a parameter, so it would be more complicated.

Electroid (approved):
On second thought, I think the syntax you've proposed is good.

Electroid:
Congrats on your first PGM feature! (and one of the most requested ones by map makers!)

PR #857, opened 2021-04-23
Fix class confirm message

KingOfSquares:
Fixes this bug:

Didn't test but I mean :shrug:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
@@ -693,10 +693,8 @@ public class PickerMatchModule implements MatchModule, Listener {
           if (cmm.getCanChangeClass(player.getId())) {
             cmm.setPlayerClass(player.getId(), cls);
             player.sendMessage(
-                text()
-                    .append(translatable("match.class.ok", NamedTextColor.GOLD))
-                    .append(text(name, NamedTextColor.GREEN))
-                    .build());
+                translatable(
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
```diff
@@ -694,8 +694,9 @@ private void handleInventoryClick(
             cmm.setPlayerClass(player.getId(), cls);
             player.sendMessage(
                 text()
-                    .append(translatable("match.class.ok", NamedTextColor.GOLD))
-                    .append(text(name, NamedTextColor.GREEN))
+                    .append(
+                        translatable("match.class.ok", NamedTextColor.GOLD),
+                        text(name, NamedTextColor.GREEN))
```
Pablete1234:
```suggestion
                        translatable("match.class.ok", NamedTextColor.GOLD,
                        text(name, NamedTextColor.GREEN))
```

Pablete1234 (changes requested):
You should probably test it

core/src/main/java/tc/oc/pgm/picker/PickerMatchModule.java
KingOfSquares: wait... 🤔

KingOfSquares:
Lets pretend like I managed to do this first try....

Pablete1234 (approved):

Electroid (approved):

PR #864, opened 2021-05-08
Add decay-time, recovery-time, and owned-decay-time

mrcookieunderscore13:
Addresses https://github.com/PGMDev/PGM/issues/803

Edit: below is outdated, check comment for up-to-date description
I added decay-time, recovery-time, and owned-decay-time attributes to control points. All of these attributes are in units of time. So, for example, if `recover-time="5s"`, assuming a captured point that is 50% uncaptured, it will take 2.5 seconds to recover to fully captured. 99% uncaptured will take ~5s.

Another thing to note is that if an equal number of defenders and attackers(non-dominating team) are standing on the point, the point will decay/recover/owned-decay. 

Signed-off-by: mrcookie

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
@@ -49,6 +49,11 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
 
   protected final Vector centerPoint;
 
+  // time values to increment by for decay, recovery, and owned-decay
+  protected final float decayIncrement;
+  protected final float recoveryIncrement;
+  protected final float ownedDecayIncrement;
+
   // This is set false after the first state change if definition.permanent == true
   protected boolean capturable = true;
 
@@ -80,6 +85,24 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
     this.playerTracker = new ControlPointPlayerTracker(match, this.getCaptureRegion());
 
     this.blockDisplay = new ControlPointBlockDisplay(match, this);
+
+    this.decayIncrement =
+        definition.getDecayTime().equals(Duration.ZERO)
+            ? 0
+            : (float) definition.getTimeToCapture().toMillis()
+                / (float) definition.getDecayTime().toMillis();
+
+    this.recoveryIncrement =
+        definition.getRecoveryTime().equals(Duration.ZERO)
+            ? 0
+            : (float) definition.getTimeToCapture().toMillis()
+                / (float) definition.getRecoveryTime().toMillis();
+
+    this.ownedDecayIncrement =
+        definition.getOwnedDecayTime().equals(Duration.ZERO)
+            ? 0
+            : (float) definition.getTimeToCapture().toMillis()
+                / (float) definition.getOwnedDecayTime().toMillis();
   }
 
   public void registerEvents() {
@@ -413,7 +436,6 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
     }
 
     ControlPointDefinition definition = this.getDefinition();
-
     if (this.controllingTeam != null && definition.hasNeutralState()) {
       // Point is owned and must go through the neutral state before another team can capture it
       if (dominantTeam == this.controllingTeam) {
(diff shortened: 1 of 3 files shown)


Pablete1234 (changes requested):
When implementing recovery & decay, `incremental` has to be mutually-exclusive with them, and code-wise incremental needs to go away, since it's equivalent to instant decay & recovery.

Visit http://docs.oc.tc/modules/control_points for the 1.9+ PGM documentation about how to properly havee incremental, recovery & decay attributes play together:

There's also the parsing code from 1.9+ PGM regarding this topic: https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java#L80
Also the dominate code for reference: https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java#L383

mrcookieunderscore13:
Everything was changed to rates rather than units of time. Also, the functionality was changed to mimic that of 1.9 PGM. Here is a quick run-down of what each attribute defines:

`recovery-rate`: Point recovers to its current state. Applies when the point owner is dominant or when a non-capturing team is dominant. A team must be dominant for this to take effect.

`decay-rate`: Point "decays" to its current state. Applies when nobody is dominating the point (zero players on point or lead = 0)

`owned-decay-rate`: Point decays from captured to neutral. Applies when nobody is dominating the point, same as decay.


`owned-decay-rate` takes precedent over `decay-rate`. So, if both have defined values and nobody is dominating the point, the point will decay towards neutral rather than "decaying" towards fully captured.

Also, here is an explanation of what rate implies:  if time to capture is 10 seconds, and the owned-decay-rate is 2, then it will decay to neutral in 5 seconds. 

Also worth noting, most of the code here was not my own, I copied a lot of it from the 1.9 PGM codebase with some minor adjustments.

core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
```diff
@@ -77,11 +77,36 @@ public static ControlPointDefinition parseControlPoint(
...
+        throw new InvalidXMLException(
+            "Cannot combine this attribute with incremental", attrRecovery);
+      if (attrOwnedDecay != null)
+        throw new InvalidXMLException(
+            "Cannot combine this attribute with incremental", attrOwnedDecay);
+
+      final boolean incremental = XMLUtils.parseBoolean(attrIncremental, koth);
+      recoveryRate = incremental ? 1.0 : Double.POSITIVE_INFINITY;
+      ownedDecayRate = 0.0;
+      decayRate = incremental ? 0.0 : Double.POSITIVE_INFINITY;
```
Pablete1234:
```suggestion
      recoveryRate = incremental ? 1.0 : Double.POSITIVE_INFINITY;
      decayRate = incremental ? 0.0 : Double.POSITIVE_INFINITY;
      ownedDecayRate = 0.0;
```
For consistency, keep the order of the 3

core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
```diff
@@ -77,11 +77,36 @@ public static ControlPointDefinition parseControlPoint(
...
+      ownedDecayRate = XMLUtils.parseNumber(attrOwnedDecay, Double.class, 0.0);
+    } else {
+      if (attrDecay != null)
+        throw new InvalidXMLException("Cannot combine this attribute with incremental", attrDecay);
+      if (attrRecovery != null)
+        throw new InvalidXMLException(
+            "Cannot combine this attribute with incremental", attrRecovery);
+      if (attrOwnedDecay != null)
+        throw new InvalidXMLException(
+            "Cannot combine this attribute with incremental", attrOwnedDecay);
```
Pablete1234:
```suggestion
      if (attrDecay != null || attrRecovery != null || attrOwnedDecay != null)
        throw new InvalidXMLException("Cannot combine this attribute with incremental", 
            attrDecay != null ? attrDecay : attrRecovery != null ? attrRecovery : attrOwnedDecay);
```
You can just merge it into a single if statement

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -403,37 +403,39 @@ private void dominateAndFireEvents(@Nullable Competitor dominantTeam, Duration d
...
+        // owner is recovering the point
+        recover(dominantTeam, dominantTime);
       } else if (dominantTeam != null) {
-        this.progressUncapture(dominantTeam, dominantTime);
-      } else if (!definition.isIncrementalCapture()) {
-        // No team is dominant and point is not incremental, so reset the time
-        this.capturingTime = Duration.ZERO;
+        // non-owner is uncapturing the point
+        uncapture(dominantTeam, dominantTime);
+      } else if (definition.getOwnedDecayRate() > 0.0) {
```
```suggestion
      } else if (definition.getOwnedDecayRate() > 0) {
```
I'm pretty sure 0.0 is not required here

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -403,37 +403,39 @@ private void dominateAndFireEvents(@Nullable Competitor dominantTeam, Duration d
...
+        // nobody on point so decay to neutral state
+        ownedDecay(dominantTime);
```
Pablete1234: The owned decay rate is only ever used IF there's a neutral state, which means it should probably have an XML error in case you don't have neutral state and you try to use owned decay rate

Pablete1234 (changes requested):
It's looking pretty good, make sure you test this with multiple existing maps as well as modified xmls to test the new features and make sure they all both work as intended, and fit the needs of the original issue

mrcookieunderscore13:
Update: just added Pablo's changes, and tested on lots of cp maps. Existing maps work correctly and the new features work correctly as well 😄.

Pablete1234 (approved):

Electroid (approved):
Thanks for tracking this down and working with Pablo, looks ready to roll! :rocket:

PR #873, opened 2021-05-19
Add <gamemode> module

CoWinkKeyDinkInc:
This re-adds the `<gamemode>` module back to PGM (if it ever was there). It works similar to how it is already described in the [docs](https://pgm.dev/docs/modules/general/main/#gamemode-ids), as I noticed that there is no code in PGM that uses the tags. You can't do that weird adding thing it talks about. This uses the translatable objective names so maps that are based around "Attack & Defend" like BoomBox or a "Scorebox" like Zap and Miner Sixty Niner, or maps with monument mode trickery no longer need to use the untranslatable `<game>` tag. `<game>` was known as `gamemode` in the code so this was changed to `game`. No new translatable components were added.

```xml
<!-- Sets Scorebox as game title -->
<gamemode>scorebox</gamemode>
<!-- Sets Attack/Defend as game title -->
<gamemode>ad</gamemode>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
@@ -1,0 +1,44 @@
+package tc.oc.pgm.api.map;
+
+public enum Gamemode {
+  ATTACK_DEFEND("ad", "Attack/Defend"),
+  ARCADE("arcade", "Arcade"),
+  BLITZ("blitz", "Blitz"),
+  BLITZ_RAGE("br", "Blitz: Rage"),
+  CAPTURE_THE_FLAG("ctf", "Capture the Flag"),
+  CONTROL_THE_POINT("cp", "Control the Point"),
+  CAPTURE_THE_WOOL("ctw", "Capture the Wool"),
+  DESTROY_THE_CORE("dtc", "Destroy the Core"),
+  DESTROY_THE_MONUMENT("dtm", "Destroy the Monument"),
+  FREE_FOR_ALL("ffa", "Free For All"),
+  FLAG_FOOTBALL("ffb", "Flag Football"),
+  KING_OF_THE_HILL("koth", "King of the Hill"),
+  KING_OF_THE_FLAG("kotf", "King of the Flag"),
+  MIXED("mixed", "Mixed"),
+  RAGE("rage", "Rage"),
+  RACE_FOR_WOOL("rfw", "Race for Wool"),
+  SCOREBOX("scorebox", "Scorebox"),
+  DEATHMATCH("tdm", "Deathmatch");
+
+  private final String id;
+  private final String name;
+  private final MapTag mapTag;
+
+  Gamemode(String id, String name) {
+    this.id = id;
+    this.name = name;
+    this.mapTag = new MapTag(id, name, true, false);
+  }
+
+  public String getId() {
+    return id;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public MapTag getMapTag() {
+    return mapTag;
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
@@ -1,44 +1,32 @@
 package tc.oc.pgm.api.map;
 
 public enum Gamemode {
-  ATTACK_DEFEND("ad", "Attack/Defend"),
-  ARCADE("arcade", "Arcade"),
-  BLITZ("blitz", "Blitz"),
-  BLITZ_RAGE("br", "Blitz: Rage"),
-  CAPTURE_THE_FLAG("ctf", "Capture the Flag"),
-  CONTROL_THE_POINT("cp", "Control the Point"),
-  CAPTURE_THE_WOOL("ctw", "Capture the Wool"),
-  DESTROY_THE_CORE("dtc", "Destroy the Core"),
-  DESTROY_THE_MONUMENT("dtm", "Destroy the Monument"),
-  FREE_FOR_ALL("ffa", "Free For All"),
-  FLAG_FOOTBALL("ffb", "Flag Football"),
-  KING_OF_THE_HILL("koth", "King of the Hill"),
-  KING_OF_THE_FLAG("kotf", "King of the Flag"),
-  MIXED("mixed", "Mixed"),
-  RAGE("rage", "Rage"),
-  RACE_FOR_WOOL("rfw", "Race for Wool"),
-  SCOREBOX("scorebox", "Scorebox"),
-  DEATHMATCH("tdm", "Deathmatch");
+  ATTACK_DEFEND("ad"),
+  ARCADE("arcade"),
+  BLITZ("blitz"),
+  BLITZ_RAGE("br"),
+  CAPTURE_THE_FLAG("ctf"),
+  CONTROL_THE_POINT("cp"),
+  CAPTURE_THE_WOOL("ctw"),
+  DESTROY_THE_CORE("dtc"),
+  DESTROY_THE_MONUMENT("dtm"),
+  FREE_FOR_ALL("ffa"),
+  FLAG_FOOTBALL("ffb"),
+  KING_OF_THE_HILL("koth"),
+  KING_OF_THE_FLAG("kotf"),
+  MIXED("mixed"),
+  RAGE("rage"),
+  RACE_FOR_WOOL("rfw"),
+  SCOREBOX("scorebox"),
+  DEATHMATCH("tdm");
 
   private final String id;
-  private final String name;
-  private final MapTag mapTag;
 
-  Gamemode(String id, String name) {
+  Gamemode(String id) {
     this.id = id;
-    this.name = name;
-    this.mapTag = new MapTag(id, name, true, false);
   }
 
   public String getId() {
     return id;
-  }
-
-  public String getName() {
-    return name;
-  }
-
-  public MapTag getMapTag() {
-    return mapTag;
   }
 }
--- a/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
@@ -26,6 +26,16 @@ public enum Gamemode {
     this.id = id;
   }
 
+  public static Gamemode byId(String gamemodeId) {
+    Gamemode gm = null;
+    for (Gamemode gamemode : Gamemode.values()) {
+      if (gamemode.getId().equals(gamemodeId)) {
+        gm = gamemode;
+      }
+    }
+    return gm;
+  }
+
   public String getId() {
     return id;
   }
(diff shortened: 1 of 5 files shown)


CoWinkKeyDinkInc:
We should expand on this and have more translatable gamemode titles, here are some others I have in mind:
```
kotf: King of the Flag
ffb: Flag Football
rfw: Race for Wool
br: Blitz: Rage
cp: Control Points
ctp: Control the Point
ctps: Control the Points
```

Pablete1234:
> ctps: Control the points

Game names are all in singular, you still call it "Capture the wool" when there's 2 wools, `Control the points` makes no sense, it's `Control the point`

CoWinkKeyDinkInc:
New translatable gamemode titles are added (I call dibs on Pirate English and lolcat translations :P)

Electroid:
Though, map tags were designed to replace game mode? Since it makes things more extensible?

CoWinkKeyDinkInc:
@Electroid map tags are automated, so they are automatically generated by what tags a map uses. `<gamemode>` forces a gamemode title since PGM can't or doesn't evaluate certain game types like ~~"Scorebox" or~~ "King of the Flag".

Pugzy:
> @Electroid map tags are automated, so they are automatically generated by what tags a map uses. `<gamemode>` forces a gamemode title since PGM can't or doesn't evaluate certain game types like ~"Scorebox" or~ "King of the Flag".

Can map tags not be defined in XML? This does seem like a duplicate of what map tags are for and support for translatable map tags already exist. Maybe they just need parsing and if exist override the auto-generated ones.

CoWinkKeyDinkInc:
This PR will ~~give them the tags and~~ change the objective name on the scoreboard. It doesn't look like any tag is explicitly set in XML, ~~so this would be the first.~~

CoWinkKeyDinkInc:
The PR doesn't deal with the tags anymore. Thanks to some help with Pablete it's a bit better now. You can now use two gamemode types that will display on the scoreboard title. Now the documentation is just like how it is already described on the docs.
```xml
<!--- DTC & CTW -->
<gamemode>dtc</gamemode>
<gamemode>ctw</gamemode>
```

core/src/main/java/tc/oc/pgm/api/map/MapInfo.java
```diff
@@ -109,12 +109,19 @@ default String getStyledNameLegacy(MapNameStyle style, @Nullable CommandSender s
...
+   * Get a {@link Component} that represents this map's custom game title.
    *
    * @return A component of the gamemode name if defined or null.
    */
   Component getGamemode();
 
+  /**
+   * Get a {@link Collection<Gamemode>} that represents this map's gamemodes.
+   *
+   * @return A component of gamemodes if defined or null.
```
Pablete1234:
```suggestion
   * @return The list of defined gamemodes, empty list if none are defined.
```

Don't lie in the javadocs, makes the APIs hard to use

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -243,6 +254,18 @@ public Component getStyledName(MapNameStyle style) {
     return rules;
   }
 
+  private static List<Gamemode> parseGamemodes(Element root) throws InvalidXMLException {
+    List<Gamemode> gamemodes = new ArrayList<>();
+    for (Element gamemodeEl : root.getChildren("gamemode")) {
+      for (Gamemode gamemode : Gamemode.values()) {
+        if (gamemode.getId().equals(gamemodeEl.getText())) {
```
Pablete1234: Consider moving this to a static method in gamemode, so that you can call it by doing:
```
Gamemode gm = Gamemode.byId(gamemodeEl.getText());
if (gamemode == null) throw new InvalidXMLException("Unknown gamemode", gamemodeEl);
else gamemodes.add(gamemode);
```

Currently if the gamemode doesn't exist you're just ignoring it, it's an error that needs to be reported or else you may set your map to be "wools" (instead of ctw), assume that's valid, never check-back, and pgm will never say a thing about it.

Also test what happens when you just define a `<gamemode/>` tag without any text, does it blow up, or does it simply complain?

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -258,9 +260,29 @@ private Component renderTitle(final Config config, final MapInfo map) {
...
+            Component.translatable("gamemode." + secondgamemodeId + ".acronym");
+        return firstComponent
+            .append(Component.text(" & "))
+            .append(secondComponent)
+            .colorIfAbsent(NamedTextColor.AQUA);
+      } else {
+        return Component.translatable("gamemode." + firstgamemodeId + ".name")
+            .colorIfAbsent(NamedTextColor.AQUA);
+      }
     }
```
Pablete1234: This is a very odd way to do this, you're essentially doing an if-hell of if it's one, or if it's 2 gamemodes, ignoring any other case, additionally, you're using a hard-coded `&`, which has to be localized for different languages.

Consider a more iterative approach like the one with map tags below

core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
```diff
@@ -0,0 +1,44 @@
+package tc.oc.pgm.api.map;
+
+public enum Gamemode {
+  ATTACK_DEFEND("ad", "Attack/Defend"),
```
Pablete1234:
```suggestion
  ATTACK_DEFEND("ad"),
```
Since these are simply translated in the string files, they're not needed in the code, you can remove all references to name in general

core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
```diff
@@ -0,0 +1,44 @@
...
+  DEATHMATCH("tdm", "Deathmatch");
+
+  private final String id;
+  private final String name;
+  private final MapTag mapTag;
+
+  Gamemode(String id, String name) {
+    this.id = id;
+    this.name = name;
+    this.mapTag = new MapTag(id, name, true, false);
```
Pablete1234: You do nothing with these tags, either make it so the map will override all tags to enforce these, or remove them for now. They can always be added later on

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public static Gamemode byId(String gamemodeId) {
+    Gamemode gm = null;
+    for (Gamemode gamemode : Gamemode.values()) {
+      if (gamemode.getId().equals(gamemodeId)) {
+        gm = gamemode;
+      }
+    }
+    return gm;
```
Pablete1234:
```suggestion
    for (Gamemode gamemode : Gamemode.values()) {
      if (gamemode.getId().equals(gamemodeId)) {
        return gamemode;
      }
    }
    return null;
```

core/src/main/java/tc/oc/pgm/api/map/MapInfo.java
```diff
@@ -109,12 +109,19 @@ default String getStyledNameLegacy(MapNameStyle style, @Nullable CommandSender s
   Collection<MapTag> getTags();
 
   /**
-   * Get a {@link Component} that represents this map's gamemode name.
+   * Get a {@link Component} that represents this map's custom game title.
    *
-   * @return A component of the gamemode name if defined or null.
+   * @return The list of defined gamemodes, empty list if none are defined.
```
Pablete1234: This does not look like a list to me

core/src/main/java/tc/oc/pgm/api/map/MapInfo.java
Pablete1234: Doesn't look resolved

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -255,6 +266,19 @@ public Component getStyledName(MapNameStyle style) {
...
 
+  private static List<Gamemode> parseGamemodes(Element root) throws InvalidXMLException {
+    List<Gamemode> gamemodes = new ArrayList<>();
+    for (Element gamemodeEl : root.getChildren("gamemode")) {
+      Gamemode gm = Gamemode.byId(gamemodeEl.getText());
+      if (gm == null) {
+        throw new InvalidXMLException("Unknown gamemode", gamemodeEl);
+      } else {
+        gamemodes.add(gm);
+      }
```
Pablete1234:
```suggestion
      if (gm == null)
        throw new InvalidXMLException("Unknown gamemode", gamemodeEl);
    gamemodes.add(gm);
```
no need for the else, since the exception will already stop further execution

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -258,9 +260,55 @@ private Component renderTitle(final Config config, final MapInfo map) {
...
+      for (Gamemode value : gamemodes) {
+        Component gmComponent;
+        // Make this smarter using MAX_LENGTH
+        if (gamemodes.size() <= 1) {
+          gmComponent = Component.translatable("gamemode." + value.getId() + ".name");
+        } else {
+          gmComponent = Component.translatable("gamemode." + value.getId() + ".acronym");
+        }
+        gamemodeComponents.add(gmComponent);
+      }
```
Pablete1234:
```suggestion
      String suffix = gamemodes.size() <= 1 ? ".name" : ".acronym";
      List<Component> gmComponents = gamemodes.stream()
              .map(gm  -> translatable("gamemode." + gm.getId() + suffix))
              .collect(Collectors.toList());      
      Component gamemodeTitle = null;
```

You can simplify all of that into this
There's no need for an if-check on each iteration to decide what suffix to use (moved it out of the loop), and there's no need to have a list that resizes on each iteration

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -258,9 +260,55 @@ private Component renderTitle(final Config config, final MapInfo map) {
...
+        case 4:
+          gamemodeTitle =
+              Component.translatable(
+                      "misc.list.start", gamemodeComponents.get(0), gamemodeComponents.get(1))
+                  .append(
+                      Component.translatable(
+                          "misc.list.end", gamemodeComponents.get(2), gamemodeComponents.get(3)))
+                  .colorIfAbsent(NamedTextColor.AQUA);
+          break;
+      }
```
Pablete1234: I'm 100% sure this has to be a util somewhere, if it really isn't, make it one.

Also, what about over 4? i know it's not going to be common but you still have to support N items on the list

Pablete1234 (changes requested):

CoWinkKeyDinkInc:
Found what Pablo was talking about, it's TextFormatter.java I'll use this function soon.

Pablete1234 (commented):
The comment on getGamemodes is still wrong. It says it returns a component when it returns a list of gamemodes

Pablete1234 (dismissed):

CoWinkKeyDinkInc:
@Pablete1234 @Electroid Ready to merge!

Electroid (approved):

PR #879, opened 2021-05-27
Check FireTick during match load and start

CoWinkKeyDinkInc:
While I was helping @technodono find out why The Dropper Community map was damaged, I discovered that fire spreads before the match starts. This PR fixes this simple bug to make it consistent with other world events, like PGM locking mob movement and time. DoFireTick will now default to on (as it always has) when the match **starts**.


core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -258,6 +260,26 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
 
+  @EventHandler
+  public void lockFireTick(final MatchLoadEvent event) {
+    event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, Boolean.toString(false));
+  }
+
+  // unlock firespread if not disabled by <gamerule>
+  @EventHandler
+  public void unlockFireTick(final MatchStartEvent event) {
+    ImmutableMap<String, String> lockFireTick =
```
Pablete1234: These are gamerules, NOT lockFireTick, give proper names to variables

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -258,6 +260,26 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
+        event.getMatch().getModule(GameRulesMatchModule.class).getGameRules();
+    if (lockFireTick.isEmpty() || Boolean.parseBoolean(lockFireTick.get(DO_FIRE_TICK))) {
+      event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, Boolean.toString(true));
```
Pablete1234: You're setting it to TRUE if it's defined in `<gamerules>` but what if it's not defined at all? Currently you're breaking any map that defines this in the world data and not in xml (older maps)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
CoWinkKeyDinkInc: I forgot to simplify this.

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -77,6 +79,7 @@
 
   // Single-write, multi-read lock used to create the first match
   private final ReentrantReadWriteLock lock;
+  private boolean oldFireTick = true;
```
Pablete1234: this looks like a pretty odd place for this.

Consider having GameRulesMatchModule have logic for getting all gamerules in the level.dat, then ammending the ones defined in the XML ontop of them, and being able to provide them via getter, that way you can simply use them anywhere without having to save random booleans all across the codebase

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -258,6 +261,30 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
+  public void lockFireTick(final MatchLoadEvent event) {
+    oldFireTick = Boolean.parseBoolean(event.getMatch().getWorld().getGameRuleValue(DO_FIRE_TICK));
+    event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, Boolean.toString(false));
+  }
+
+  // if no doFireTick <gamerule> is found, then use doFireTick setting saved in world
+  @EventHandler
+  public void unlockFireTick(final MatchStartEvent event) {
+    ImmutableMap<String, String> gameRules =
+        event.getMatch().getModule(GameRulesMatchModule.class).getGameRules();
```
Pablete1234: I'm not sure why `getGameRules` returning the whole map exists to start with, consider having a more specific getter that just tells you the value of the requested gamerule directly, will simplify this logic and logic elsewhere

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -258,6 +261,30 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
+    String newFireTick = gameRules.get(DO_FIRE_TICK);
+    if (!gameRules.isEmpty() && newFireTick != null) {
+      event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, newFireTick);
+    } else {
+      event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, Boolean.toString(oldFireTick));
+    }
```
Pablete1234: don't copy-paste code in 2 diff invocations depending on an if for a boolean parameter, make them part of the parameter passed directly.

```suggestion
    event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, 
            !gameRules.isEmpty() && newFireTick != null ? newFireTick : Boolean.toString(oldFireTick));
```
however, this logic should be even simpler with what i proposed above

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -258,6 +261,30 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
+    oldFireTick = Boolean.parseBoolean(event.getMatch().getWorld().getGameRuleValue(DO_FIRE_TICK));
+    event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, Boolean.toString(false));
+  }
+
+  // if no doFireTick <gamerule> is found, then use doFireTick setting saved in world
+  @EventHandler
+  public void unlockFireTick(final MatchStartEvent event) {
+    ImmutableMap<String, String> gameRules =
+        event.getMatch().getModule(GameRulesMatchModule.class).getGameRules();
+    String newFireTick = gameRules.get(DO_FIRE_TICK);
```
Pablete1234: you're doing a `get` here just to later ignore the result if is empty, bit of a weird one, could simply `getOrDefault` to oldFireTick and simplify this further

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
```diff
@@ -26,4 +33,7 @@ public void load() {
   public ImmutableMap<String, String> getGameRules() {
     return ImmutableMap.copyOf(gameRules);
   }
+  public ImmutableMap<String, String> getMapGameRules() {
+    return ImmutableMap.copyOf(mapGameRules);
+  }
```
Pablete1234: Merge this into one call to get a certain gamerule.

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -263,21 +261,15 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
-        event.getMatch().getModule(GameRulesMatchModule.class).getGameRules();
-    String newFireTick = gameRules.get(DO_FIRE_TICK);
-    if (!gameRules.isEmpty() && newFireTick != null) {
-      event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, newFireTick);
-    } else {
-      event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, Boolean.toString(oldFireTick));
-    }
+    String farts = event.getMatch().getModule(GameRulesMatchModule.class).getMapGameRules().get(DO_FIRE_TICK);
+    String newFireTick = event.getMatch().getModule(GameRulesMatchModule.class).getGameRules().get(DO_FIRE_TICK);
+    event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, newFireTick != null ? newFireTick : farts);
```
Pablete1234:
```suggestion
    String fireTick = event.getMatch().needModule(GameRulesMatchModule.class).getGameRule(DO_FIRE_TICK);
    event.getMatch().getWorld().setGameRuleValue(DO_FIRE_TICK, fireTick);
```
Once you create a single method that does the logic, you'll be able to use this. You must change getModule to needModule, because you will otherwise NPE if for some reason it was null

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
```diff
@@ -11,7 +10,7 @@
 
   private final Match match;
   private final Map<String, String> gameRules;
-  private Map<String, String> mapGameRules = new HashMap<>();
+  private boolean doDaylightCycle = false;
```
Pablete1234: Keep this as it was. MapGameRules, do not handle them specifically

Pablete1234 (changes requested):

CoWinkKeyDinkInc:
This PR now adds gamerules set by `level.dat` not mentioned by `<gamemode>` into the GameRulesMatchModule, so now it takes gamemodes intentionally set in `level.dat` to account.

core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
```diff
@@ -21,9 +20,17 @@ public void load() {
     for (Map.Entry<String, String> gameRule : this.gameRules.entrySet()) {
       this.match.getWorld().setGameRuleValue(gameRule.getKey(), gameRule.getValue());
     }
+
+    // gets gamerule values from level.dat after being set
+    // does not save doDaylightCycle to gameRules
```
Pablete1234: why not?
Either explain why (and it needs to make sense) or do save it

If you're avoiding saving it because it may already have been set by PGM before this, consider that this module should probably be saving before changing daylight cycle

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -60,6 +60,7 @@
 import tc.oc.pgm.util.text.TextTranslations;
 
 public class PGMListener implements Listener {
+  private static final String DO_FIRE_TICK = "doFireTick";
   private static final String DO_DAYLIGHT_CYCLE = "doDaylightCycle";
```
Pablete1234: consider moving these constant somewhere else where they can be reused, maybe a util class so `Gamemode.DO_FIRE_TICK` or something along the lines can be used

core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
CoWinkKeyDinkInc: PGM by default locks the the time when the match starts unless it's specifically unlocked with `<timelock>` or `<gamerule>`. So the original state of the DaylightCycle gamerule from level.dat does not matter

Pablete1234: That it does not matter doesn't mean it should be explicitly not saved on the map, just have it there and ignore or overwrite it, but having a comment mentioning what a line of code bellow does without explaining why is a bit redundant

core/src/main/java/tc/oc/pgm/api/map/GameRule.java
```diff
@@ -0,0 +1,40 @@
...
+  showDeathMessages
+   */
+
+  private final String id;
+
+  GameRule(String id) {
+    this.id = id;
+  }
+
+  public static GameRule byId(String gameRuleID) {
```
Pablete1234:
```suggestion
  public static GameRule byId(String gameRuleId) {
```

core/src/main/java/tc/oc/pgm/api/map/GameRule.java
```diff
@@ -0,0 +1,40 @@
...
+    for (GameRule gameRule : GameRule.values()) {
+      if (gameRule.getId().equals(gameRuleID)) {
```
Pablete1234:
```suggestion
      if (gameRule.getId().equals(gameRuleId)) {
```

core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
```diff
@@ -21,9 +20,18 @@ public void load() {
...
+
+    /* Gets gamerule values from level.dat after being set
+     Maps need doDaylightCycle set in XML for it to be turned on, so doDayLightCycle from level.dat is not saved
+    */
+    for (String gameRule : this.match.getWorld().getGameRules()) {
+      if (!gameRule.equals("doDaylightCycle")) {
+        gameRules.put(gameRule, this.match.getWorld().getGameRuleValue(gameRule));
+      }
+    }
   }
```
Pablete1234: The order of these 2 loops needs to be inverted.

First you should set the gamerules from the world (as fallback) (match.getWorld.getGameRules)
Then override them with the gamerules defined in the XML (this.gamerules)
Lastly, force the doDaylightCycle gamerule to if the XML has timelock on or off. (make sure you define GameRulesModule having a dependency on WorldTimeModule)

There's no need to do the check for doDaylightCycle on each iteration

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -258,31 +258,71 @@ public void handleItemPickup(final PlayerPickupItemEvent event) {
...
+    event
+        .getMatch()
+        .getWorld()
+        .setGameRuleValue(GameRule.DO_DAYLIGHT_CYCLE.getId(), Boolean.toString(false));
   }
 
   @EventHandler
   public void unlockTime(final MatchStartEvent event) {
     // if there is a timelock module and it is off, unlock time
     boolean unlockTime = !event.getMatch().getModule(WorldTimeModule.class).isTimeLocked();
```
Pablete1234, line 296: if you set this as the value for doDaylightCycle directly on the gamerule match module (as suggested in previous comment) the logic for all of this just simplifies down to doing basically the same as you do for fire.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
CoWinkKeyDinkInc: I don't know what you mean by this,

CoWinkKeyDinkInc:
Currently in this PR maps that have doDayLightCycle set to true in the gamerule XML do not get the sun moving. Intercept specifically has this on.

```xml
<gamerules>
    <doDaylightCycle>true</doDaylightCycle>
</gamerules>
```

CoWinkKeyDinkInc:
This PR is in working order now.

Pablete1234 (approved):

Electroid (approved):

PR #882, opened 2021-06-02
Add /scorelimit command

destial:
Since there was a /timelimit command to change the timelimit, I figured a /scorelimit command to change or set a score limit. I found it useful when playing Rage or TDM maps, since I would have to go into the map.xml to change the score limit, and then cycle it again.

scorelimit

I've also added an error message if the map doesn't contain the score module.

error_scorelimit


core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
```diff
@@ -0,0 +1,38 @@
...
+import tc.oc.pgm.util.Audience;
+
+public final class ScoreLimitCommand {
+
+  @Command(
+      aliases = {"scorelimit", "sl"},
+      desc = "Set the score limit",
+      usage = "[number]",
+      help = "Amount must be an integer",
+      flags = "r",
```
Pugzy, line 22: What is the flag r doing?

destial: Actually idk, I just copied from the TimeLimitCommand class since it is similar

core/src/main/java/tc/oc/pgm/command/TimeLimitCommand.java
```diff
@@ -31,26 +32,29 @@ public void timelimit(
...
-            NamedTextColor.YELLOW,
-            clock(duration).color(NamedTextColor.AQUA),
-            result != null ? result.getDescription(match) : translatable("misc.unknown")));
+      audience.sendMessage(
+          translatable(
+              "match.timeLimit.commandOutput",
+              NamedTextColor.YELLOW,
+              clock(duration).color(NamedTextColor.AQUA),
+              result != null ? result.getDescription(match) : translatable("misc.unknown")));
+    } catch (ModuleLoadException e) {
```
Pablete1234: don't use try-catch for control flow. Change needModule for getModule and check for null

core/src/main/java/tc/oc/pgm/command/graph/CommandGraph.java
```diff
@@ -25,28 +25,7 @@
...
-import tc.oc.pgm.command.ModeCommand;
-import tc.oc.pgm.command.ProximityCommand;
-import tc.oc.pgm.command.RestartCommand;
-import tc.oc.pgm.command.SettingCommand;
-import tc.oc.pgm.command.StartCommand;
-import tc.oc.pgm.command.StatsCommand;
-import tc.oc.pgm.command.TeamCommand;
-import tc.oc.pgm.command.TimeLimitCommand;
-import tc.oc.pgm.command.VotingCommand;
+import tc.oc.pgm.command.*;
```
Pablete1234: Avoid wildcard imports, you can change this setting in your IDE

core/src/main/java/tc/oc/pgm/score/MercyRule.java
```diff
@@ -8,7 +8,7 @@
 public class MercyRule {
 
   private final ScoreMatchModule scoreMatchModule;
-  private final int scoreLimit;
```
Pablete1234: Keep this immutable, making this change is opening up to a bunch of bugs

core/src/main/java/tc/oc/pgm/score/MercyRule.java
```diff
@@ -22,6 +22,10 @@ public MercyRule(ScoreMatchModule scoreMatchModule, int scoreLimit, int mercyLim
     calculateLeaders();
   }
 
+  public void setScoreLimit(int scoreLimit) {
+    this.scoreLimit = scoreLimit;
+  }
+
```
Pablete1234:
```suggestion
```

This state is immutable and should remain immutable

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -89,6 +89,14 @@ public int getScoreLimit() {
     return this.config.scoreLimit;
   }
 
+  public void setScoreLimit(int scoreLimit) {
+    if (hasMercyRule()) {
+      mercyRule.setScoreLimit(scoreLimit);
+      return;
+    }
+    this.config.scoreLimit = scoreLimit;
```
Pablete1234: The fact that config isn't immutable is an issue in itself, it shoud always have been, but that aside, this change will mantain itself until the server is restarted, that's bad. 

The "config" represents the values in the XML, they must NOT be modified by pgm during runtime, except when it's to re-parse the XML and generate a new definition (eg: if you modify the xml, or upload a different version of the map), the match-scoped module should never modify the map-scoped value.

Pablete1234 (changes requested):
There are some serious oversights on this implementation, and they must be modified.

What i suggest you do is add a mutable `Integer maxScoreOverride` variable to `ScoreMatchModule`, defaulting to null, and add a method `setMaxScoreOverride` that sets it.
This also allows you to have the `-r` flag, which `r`esets the state back to normal, because all you'll need to do to restablish that state is set the override back to null.
Obviously you'll need to modify a few methods accordingly, like hasScoreLimit and getScoreLimit so that if the override isn't null it returns that.

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Command(
+      aliases = {"scorelimit", "sl"},
+      desc = "Set the score limit",
+      usage = "[number]",
+      help = "Score must be an integer",
+      flags = "r",
+      perms = Permissions.GAMEPLAY)
+  public void scorelimit(
+      Audience audience, Match match, @Nullable Integer amount, @Switch('r') boolean reset) {
```
Pablete1234, line 25: You need to check if the amount is null and reset is false, in which case, it's an invalid use of the command (throw some exception or a warning to the user

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
```diff
@@ -0,0 +1,42 @@
...
+    final ScoreMatchModule score = match.getModule(ScoreMatchModule.class);
+    if (score != null) {
+      if (amount != null) {
+        score.setScoreLimit(amount);
+      } else if (reset) {
+        score.resetScoreLimit();
+      }
```
Pablete1234:
```suggestion
      score.setScoreLimit(reset ? 0 : amount);
```

If you do the earlier check, amount should always be non-null if reset is false. Setting ScoreLimit to 0 should be equivalent to a reset, there's no need for a separate method

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
```diff
@@ -0,0 +1,42 @@
...
+      audience.sendMessage(
+          translatable(
+              "match.scoreLimit.commandOutput",
+              NamedTextColor.YELLOW,
+              Component.text(score.getScoreLimit()).color(NamedTextColor.AQUA)));
```
Pablete1234: This message will get confusing when resetting, it will say "you set the score limit to X", where X can be a null (may even throw a NPE in that case here?), could be 0, could be the result of the current mercy rules (which are not a real limit, but a changing one).

Just have a generic message on reset that says "reset the score limit to default" or something along the lines instead of saying the specific amount

core/src/main/java/tc/oc/pgm/command/TimeLimitCommand.java
```diff
@@ -31,26 +31,29 @@ public void timelimit(
       @Nullable Duration overtime,
       @Nullable Duration maxOvertime,
       @Nullable Duration endOvertime) {
+    final TimeLimitMatchModule time = match.getModule(TimeLimitMatchModule.class);
 
-    final TimeLimitMatchModule time = match.needModule(TimeLimitMatchModule.class);
+    if (time != null) {
```
Pablete1234: prefer short if-returns than long ones, it makes the code more readable than having big blocks of code indented-in:
```
if (time == null) {
  sendWarning(not enabled ...);
  return;
}
```

Then can leave the remaining code un-indented

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -60,10 +61,13 @@ public ScoreMatchModule(Match match, ScoreConfig config, Set<ScoreBox> scoreBoxe
     this.config = config;
     this.scoreBoxes = scoreBoxes;
     this.match.getCompetitors().forEach(competitor -> this.scores.put(competitor, 0.0));
-
     if (this.config.mercyLimit > 0) {
       this.mercyRule = new MercyRule(this, config.scoreLimit, config.mercyLimit);
+      scoreLimit = mercyRule.getScoreLimit();
```
Pablete1234: You cannot do this. Mercy rules change dynamically you cannot save this value, moreover, it just doesn't make sense to save it.

Save a score-override variable, default it to 0 for not-set.

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -50,6 +49,8 @@
 
   private final Match match;
   private final ScoreConfig config;
+  private final int defaultScore;
+  private int scoreLimit;
```
Pablete1234: Move the mutable part of the state lower, don't mix in immutable and mutable properties.

The immutable state in this class goes up to the Set of scorebox, contributions and forward are mutable state

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -72,21 +76,24 @@ public void load() {
   }
 
   public boolean hasScoreLimit() {
-    return this.config.scoreLimit > 0 || hasMercyRule();
+    return scoreLimit > 0 || hasMercyRule();
```
Pablete1234:
```suggestion
    return this.config.scoreLimit > 0 || hasMercyRule() || scoreLimitOverride > 0;
```

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -72,21 +76,24 @@ public void load() {
...
   }
 
   public boolean hasMercyRule() {
     return this.mercyRule != null;
   }
 
   public int getScoreLimit() {
-    checkState(hasScoreLimit());
+    // checkState(hasScoreLimit());
```
Pablete1234: One does not simply comment out preconditions, they exist precisely to avoid bugs in the code, the fact that you do not pass the check means something is wrong, not that you should ignore the check.
```suggestion
    checkState(hasScoreLimit());
```

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -72,21 +76,24 @@ public void load() {
...
+    return scoreLimit;
+  }
 
-    if (hasMercyRule()) {
-      return this.mercyRule.getScoreLimit();
-    }
```
Pablete1234: You need to keep dynamic calls to mercy rule stuff, that changes dynamically thru the match, you cannot hard-code a value to return.

Just add a little `if (scoreLimitOverride > 0) return this.scoreLimitOverride;` before the mercy-rule check

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -72,21 +76,24 @@ public void load() {
...
+  public void resetScoreLimit() {
+    scoreLimit = defaultScore;
+  }
```
Pablete1234:
```suggestion
```
You don't need a method to reset, just set the override to 0

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -72,21 +76,24 @@ public void load() {
...
 
-    return this.config.scoreLimit;
+  public void setScoreLimit(int scoreLimit) {
+    this.scoreLimit = scoreLimit;
   }
```
Pablete1234:
```suggestion
  public void setScoreLimitOverride(int scoreLimitOverride) {
    this.scoreLimitOverride = scoreLimitOverride;
  }
```

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -97,6 +104,10 @@ public double getScore(Competitor competitor) {
     return this.scores.get(competitor);
   }
 
+  public int getDefaultScoreLimit() {
+    return defaultScore;
+  }
+
```
Pablete1234:
```suggestion
```

Don't expose this, it's an implementation detail that shouldn't be known by outside classes

core/src/main/java/tc/oc/pgm/score/ScoreModule.java
```diff
@@ -94,25 +94,24 @@ public ScoreModule parse(MapFactory factory, Logger logger, Document doc)
...
-        config.deathScore =
+        int deathScore =
             XMLUtils.parseNumber(
                 scoreEl.getChild("deaths"), Integer.class, scoreKillsByDefault ? 1 : 0);
-        config.killScore =
+        int killScore =
             XMLUtils.parseNumber(
                 scoreEl.getChild("kills"), Integer.class, scoreKillsByDefault ? 1 : 0);
-
+        config = new ScoreConfig(scoreLimit, deathScore, killScore, mercyLimit);
```
Pablete1234: This is a good change (making ScoreConfig immutable), however this implementation still has issues

```xml
<score>
    <limit>100</limit>
</score>
<score>
    <mercy>3</mercy>
</score>
```

This is completely valid syntax, one iteration of the loop sets the limit to 100, the next iteration should set mercy to 3, and the map should get both properties, however, both current implementation and the modified one carry the same issue, they keep the values of only the last score element (resetting the values of everything else).

Later PGM had fixed this here <https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/score/ScoreModule.java#L98>

While porting the newer implementation is probably a bit overkill, it would be nice to move the 4 ints to outside the loop and update them inside, as well as passing the previous value as default for parseNumber, that way the values incrementally get updated. Then at the end create the ScoreConfig outside the loop.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
destial: Some maps have a score limit and not necessarily 0 points. E.g. KOTH has a set amount of points to obtain.

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
KingOfSquares:
```suggestion
              text(score.getScoreLimit(), NamedTextColor.AQUA));
```
`Component.text` should be imported staticly

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -3,8 +3,7 @@
 import static com.google.common.base.Preconditions.checkState;
 import static net.kyori.adventure.key.Key.key;
 import static net.kyori.adventure.sound.Sound.sound;
-import static net.kyori.adventure.text.Component.text;
-import static net.kyori.adventure.text.Component.translatable;
+import static net.kyori.adventure.text.Component.*;
```
KingOfSquares: No wildcard imports

util/src/main/i18n/templates/match.properties
```diff
@@ -152,6 +152,13 @@ match.timeLimit.result = {1} after {0}
 # {1} = some criteria to win the match
 match.timeLimit.commandOutput = The time limit is {0} with the result {1}
 
+match.timeLimit.notEnabled = The time limit is not available on this map
+
+# {0} = new score limit of the match
+match.scoreLimit.commandOutput = The score limit is now {0}
+
+match.scoreLimit.notEnabled = The score limit is not available on this map
```
KingOfSquares: "**The** score limit" sounds weird to me, maybe something like

`This map does not use score so a score limit can not be set` or just `This map does not use score`?

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
KingOfSquares: The flag r controls the boolean `reset` you have in your method signature

core/src/main/java/tc/oc/pgm/command/TimeLimitCommand.java
KingOfSquares: The time limit module should _always_ be loaded so this is probably not needed? If its not loaded it should be considered a more fatal error, since it by design always loads even though no time limit is specified in the xml.

KingOfSquares (commented):
Great work, glad to see a new contributor 😎 👍

core/src/main/java/tc/oc/pgm/command/ScoreLimitCommand.java
```diff
@@ -0,0 +1,49 @@
...
+    if (amount != null && !reset) {
+      score.setScoreLimitOverride(amount <= 0 ? null : amount);
+    } else if (reset) {
+      score.setScoreLimitOverride(null);
+    } else {
+      audience.sendWarning(
+          translatable(
+              "command.incorrectUsage", NamedTextColor.RED, text("/scorelimit <number> [-r]")));
+      return;
+    }
```
Pablete1234, line 41: This doesn't need 3 branches for the condition, only 2:

```
if (amount == null ? !reset : amount <= 0) {
    sendWarning(...);
    return;
} else {
    setScoreLimitOverride(amount);
}
```

This already enforces that:
 - If you do not set an amount, you must use reset
 - If you set an ammount, it must not be negative or 0

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -72,7 +73,11 @@ public void load() {
   }
 
   public boolean hasScoreLimit() {
-    return this.config.scoreLimit > 0 || hasMercyRule();
+    return this.config.scoreLimit >= 0 || hasMercyRule() || hasScoreLimitOverride();
```
Pablete1234, line 76:
```suggestion
    return this.config.scoreLimit > 0 || hasMercyRule() || hasScoreLimitOverride();
```

core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
```diff
@@ -81,14 +86,19 @@ public boolean hasMercyRule() {
 
   public int getScoreLimit() {
     checkState(hasScoreLimit());
-
     if (hasMercyRule()) {
       return this.mercyRule.getScoreLimit();
     }
-
+    if (hasScoreLimitOverride()) {
```
Pablete1234, line 92: You likely want to check this before you check mercy rules. Otherwise you won't be able to set an override if there's mercy rules defined in the map, as those will define the limit first

core/src/main/java/tc/oc/pgm/score/ScoreModule.java
```diff
@@ -94,25 +95,30 @@ public ScoreModule parse(MapFactory factory, Logger logger, Document doc)
...
-
+      Optional<Integer> scoreLimit = Optional.empty();
+      Optional<Integer> mercyLimit = Optional.empty();
+      Optional<Integer> deathScore = Optional.empty();
+      Optional<Integer> killScore = Optional.empty();
       for (Element scoreEl : scoreElements) {
-        config.scoreLimit = XMLUtils.parseNumber(scoreEl.getChild("limit"), Integer.class, -1);
-        config.mercyLimit = XMLUtils.parseNumber(scoreEl.getChild("mercy"), Integer.class, -1);
+        scoreLimit =
+            Optional.of(XMLUtils.parseNumber(scoreEl.getChild("limit"), Integer.class, -1));
```
Pablete1234, line 105: This is not how optionals work, you're defaulting to -1, which will not be an empty optional.

Additionally you're always overriding the previous value so again you're not making proper use of the optionals, it's not changing anything

core/src/main/java/tc/oc/pgm/score/ScoreModule.java
```diff
@@ -146,7 +152,13 @@ public ScoreModule parse(MapFactory factory, Logger logger, Document doc)
                   region, points, filter, ImmutableMap.copyOf(redeemables), silent));
         }
       }
-      return new ScoreModule(config, scoreBoxFactories.build());
+      return new ScoreModule(
+          new ScoreConfig(
+              scoreLimit.orElse(-1),
+              deathScore.orElse(proto.isOlderThan(MapProtos.DEFAULT_SCORES_TO_ZERO) ? 1 : 0),
+              killScore.orElse(proto.isOlderThan(MapProtos.DEFAULT_SCORES_TO_ZERO) ? 1 : 0),
+              mercyLimit.orElse(-1)),
```
Pablete1234, line 160: all of these "or else" will not trigger unless there was no xml tag for scores at all, otherwise they will be non-empty optionals, which defeats the purpose

Pablete1234 (changes requested):

Electroid:
Closing due to inactivity, feel free to re-open if interested.

PR #883, opened 2021-06-02
Map visibility feature

applenick:
# Map Visibility Feature

Discussed this with @Electroid yesterday. Basically allows for loaded maps to be hidden from public viewing.

Commands:
`/hiddenmaps` - View a list of hidden maps
`/mapvisibility [map]` - Toggle the `hidden` status of the provided map

At the moment hidden maps are only hidden from being visible via `/maps`, however this will be used more widely for other upcoming changes.

## Preview

https://user-images.githubusercontent.com/3377659/120559450-e5fd1f00-c3b5-11eb-8d57-c7f550cf6d8e.mp4

## Suggestions
Could use some suggestions on whether `/hiddenmaps` should be its own dedicated command, or if it should be implemented as a flag under `/maps` for example `/maps -h`.

Also maybe `/mapvisibility` should be `/setvisibility` or `/togglevisibility`? Unsure

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapVisibilityCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapVisibilityCommand.java
@@ -1,0 +1,85 @@
+package tc.oc.pgm.command;
+
+import static net.kyori.adventure.text.Component.text;
+import static net.kyori.adventure.text.Component.translatable;
+
+import app.ashcon.intake.Command;
+import app.ashcon.intake.CommandException;
+import app.ashcon.intake.parametric.annotation.Default;
+import java.util.List;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.event.ClickEvent;
+import net.kyori.adventure.text.event.HoverEvent;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.bukkit.command.CommandSender;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.Permissions;
+import tc.oc.pgm.api.map.MapInfo;
+import tc.oc.pgm.api.map.MapVisibility;
+import tc.oc.pgm.util.Audience;
+import tc.oc.pgm.util.PrettyPaginatedComponentResults;
+import tc.oc.pgm.util.named.MapNameStyle;
+import tc.oc.pgm.util.text.TextFormatter;
+
+public class MapVisibilityCommand {
+
+  @Command(
+      aliases = {"mapvisibility", "mvis"},
+      desc = "Toggle viewing visibility for a map",
+      perms = Permissions.MAPDEV)
+  public void setVisibility(Audience viewer, CommandSender sender, MapInfo map) {
+    MapVisibility visibility = PGM.get().getDatastore().getMapVisibility(map);
+    visibility.setHidden(!visibility.isHidden());
+
+    String key = visibility.isHidden() ? "hidden" : "shown";
+    NamedTextColor color = visibility.isHidden() ? NamedTextColor.RED : NamedTextColor.GREEN;
+
+    viewer.sendMessage(
+        translatable(
+            "map.visibility",
+            NamedTextColor.GRAY,
+            translatable("map.visibility." + key, color),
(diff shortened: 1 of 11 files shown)


Pugzy:
What's the intended purpose of this? What use cases? I can't really think of any off the top of my head so this just seems like unessential code burden.

I'd stick with a command that's specific to maps as setvisibility and togglevisibility are bit generic.

Pablete1234:
I'm not sure exactly what the use of this is,, but, wouldn't it make more sense to just bring back `<phase>` as an XML flag (pahase could be either development or production) instead of storing this to the DB?

That way you could simply hide maps in `<phase>development</phase>` by default, add a flag in `/maps` to filter specific phase (which allows searching all "hidden" (in-development) maps) ?

I don't really get why it'd be something to store in a DB instead of it being dictated by the XML of the map

util/src/main/i18n/templates/map.properties
```diff
@@ -74,6 +74,17 @@ map.cycled = Cycled
 
 map.protectPortal = You may not obstruct this area
 
+# {1} = visibility type (e.g. "hidden")
+# {0} = map name
+map.visibility = Map visibility set to {0} for {1}
+
+map.visibility.hidden = hidden
+
+map.visibility.shown  = shown
```
KingOfSquares, line 83: These strings already exist as
```
setting.visibility.shown
setting.visibility.hidden
```
maybe port to misc?

core/src/main/java/tc/oc/pgm/command/MapVisibilityCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Command(
+      aliases = {"hiddenmaps"},
+      desc = "View a list of hidden maps",
+      perms = Permissions.MAPDEV)
+  public void viewHiddenMaps(Audience viewer, CommandSender sender, @Default("1") int page)
+      throws CommandException {
+    List<MapVisibility> hiddenMaps = PGM.get().getDatastore().getHiddenMaps();
+
+    int resultsPerPage = 8;
```
KingOfSquares, line 53: This looks like its never changed, maybe change it to a static final variable in the class-scope so it can be used in `MapCommand` as well?

core/src/main/java/tc/oc/pgm/command/MapVisibilityCommand.java
```diff
@@ -0,0 +1,85 @@
...
+
+    new PrettyPaginatedComponentResults<MapVisibility>(formattedTitle, resultsPerPage) {
+      @Override
+      public Component format(MapVisibility map, int index) {
+        return text()
+            .append(text((index + 1) + ". "))
+            .append(map.getMap().getStyledName(MapNameStyle.COLOR_WITH_AUTHORS))
+            .clickEvent(ClickEvent.runCommand("/map " + map.getMap().getName()))
+            .hoverEvent(
+                HoverEvent.showText(
```
KingOfSquares, line 76: maybe import static `runCommand` and `showText`?

calcastor:
I'm obviously not involved with PGM development directly, but via map development on OCC. I also see very little, if any, applications for this as it currently stands. 

> I'm not sure exactly what the use of this is,, but, wouldn't it make more sense to just bring back `<phase>` as an XML flag (pahase could be either development or production) instead of storing this to the DB?
> 
> That way you could simply hide maps in `<phase>development</phase>` by default, add a flag in `/maps` to filter specific phase (which allows searching all "hidden" (in-development) maps) ?
> 
> I don't really get why it'd be something to store in a DB instead of it being dictated by the XML of the map

This seems more useful to me as there are a few applications for the `<phase>` module that I can think of and would use. Filtering of these via a `/maps` flag/tag would be great, as well as a line in `/map` indicating the phase, such as there is for the proto of a map. As I've already mentioned though I'm struggling to think of any reason why a map would need to not be visible from `/maps`.

applenick:
Sorry for the lack of updates, I spoke with Pablo the other day and will convert this feature to instead use the `<phase>` tag.

applenick:
Closed in favor of #891

PR #890, opened 2021-06-18
Custom pool cycle times

applenick:
# Custom pool cycle times

This PR allows for map pools to have custom cycle times defined in the `map-pools.yml` 

## Example `map-pools.yml`
```yml
pools:
  tnt:
    type: ordered
    cycle-time: 15s   # The match would cycle after 15 seconds instead of 30 in this example
    players: 1
    maps:
    - cos(tnt)
    ...
```

Also I included a commit which bumps [adventure](https://github.com/KyoriPowered/adventure) to `4.8.1` which resolves an error when trying to build. Not sure if it was fine to combine the two, but it saved some time. Let me know if you want a separate PR for that.

The changes to map pools have been tested and work as intended 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java
@@ -16,6 +16,7 @@ import tc.oc.pgm.events.ListenerScope;
 import tc.oc.pgm.events.PlayerLeaveMatchEvent;
 import tc.oc.pgm.events.PlayerPartyChangeEvent;
 import tc.oc.pgm.restart.RestartManager;
+import tc.oc.pgm.rotation.MapPoolManager;
 
 @ListenerScope(MatchScope.LOADED)
 public class CycleMatchModule implements MatchModule, Listener {
@@ -60,7 +61,8 @@ public class CycleMatchModule implements MatchModule, Listener {
     mapOrder.matchEnded(match);
 
     if (!RestartManager.isQueued()) {
-      Duration duration = PGM.get().getConfiguration().getCycleTime();
+      Duration duration = getCycleTime(mapOrder);
+
       if (!duration.isNegative()) {
         startCountdown(duration);
       }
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java
```diff
@@ -71,4 +73,23 @@ public void onMatchEnd(MatchFinishEvent event) {
...
+      }
+    }
+
+    // Default to the main cycle time when not found
+    if (cycleTime == null || cycleTime.isNegative()) {
+      cycleTime = PGM.get().getConfiguration().getCycleTime();
+    }
+
+    return cycleTime;
+  }
```
Pablete1234: Just move this method over to the MapOrder interface, have it default to `PGM.get().getConfiguration().getCycleTime();` and let implementations override the method.

Doing the if-hell of checking for specifically one implementation is a very bad practice since it does not allow for any other 3rd party implementation except for MapPoolManager to ever implement this feature

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java
applenick: Thanks for the feedback @Pablete1234 ! I've rewritten this to be more flexible for 3rd party implementations 👍 Let me know if you have any other suggestions.

core/src/main/java/tc/oc/pgm/rotation/MapPoolManager.java
```diff
@@ -264,6 +264,14 @@ public void matchEnded(Match match) {
     activeMapPool.matchEnded(match);
   }
 
+  @Override
+  public Duration getCycleTime() {
+    if (activeMapPool != null && !activeMapPool.getCycleTime().isNegative()) {
+      return activeMapPool.getCycleTime();
+    }
+    return PGM.get().getConfiguration().getCycleTime();
```
Pablete1234:
```suggestion
    Duration cycleTime;
    if (activeMapPool != null && !(cycleTime = activeMapPool.getCycleTime()).isNegative()) {
      return cycleTime;
    }
    return PGM.get().getConfiguration().getCycleTime();
```

Just avoiding the extra call to get cycle time

applenick: Thanks @Pablete1234, adjusted 👍

Pablete1234 (approved):

Electroid (approved):

PR #891, opened 2021-06-19
Reintroduce map phase XML tag

applenick:
# Reintroduce map phase XML tag

This PR reintroduces the map phase tag, allowing for map authors to identify whether their map is in a `DEVELOPMENT` or `PRODUCTION` phase. In addition to including the xml tag, the phase will now be displayed under `/map` for those with the debug perm. Also `/maps` can now be filtered by using the `-p [phase]` flag. Maps in the development phase are hidden from `/maps` by default.

### XML Example
```xml
<phase>development</phase> # Identify the map is in a developmental stage
<phase>production</phase> # Identify the map is in a production stage (production is the default value)

```

### /maps and /map Example

Example of `/map` when a map is in the `PRODUCTION` phase
production

Example of using `/map` which now has a `-p [phase]` flag
maps

As always, please let me know if there's any suggestions! 👍 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/map/Phase.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/Phase.java
@@ -13,6 +13,11 @@ public enum Phase {
   }
 
   public static Phase of(String query) {
-    return query.startsWith("dev") ? DEVELOPMENT : PRODUCTION;
+    for (Phase phase : Phase.values()) {
+      if (phase.toString().toLowerCase().startsWith(query.toLowerCase())) {
+        return phase;
+      }
+    }
+    return null;
   }
 }
--- a/core/src/main/java/tc/oc/pgm/api/map/Phase.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/Phase.java
@@ -2,6 +2,7 @@ package tc.oc.pgm.api.map;
 
 import static net.kyori.adventure.text.Component.translatable;
 
+import java.util.Locale;
 import net.kyori.adventure.text.Component;
 
 public enum Phase {
@@ -13,8 +14,9 @@ public enum Phase {
   }
 
   public static Phase of(String query) {
+    query = query.toLowerCase(Locale.ROOT);
     for (Phase phase : Phase.values()) {
-      if (phase.toString().toLowerCase().startsWith(query.toLowerCase())) {
+      if (phase.toString().toLowerCase(Locale.ROOT).startsWith(query)) {
         return phase;
       }
     }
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/api/map/Phase.java
```diff
@@ -0,0 +1,18 @@
...
+public enum Phase {
+  PRODUCTION,
+  DEVELOPMENT;
+
+  public Component toComponent() {
+    return translatable("map.phase." + this.name().toLowerCase());
+  }
+
+  public static Phase of(String query) {
+    return query.startsWith("dev") ? DEVELOPMENT : PRODUCTION;
```
Pablete1234: A bit of a too lenient check, phase "kdnfgojkdfjngjidfn" would match as production.

I'd iterate over Phase.values(), and check for name().toLowerCase().startsWith(query), if none match return a null and make the command show a proper exception about it

core/src/main/java/tc/oc/pgm/command/MapCommand.java
```diff
@@ -81,6 +83,12 @@ public void maps(
       search = search.filter(map -> matchesName(map, name));
     }
 
+    if (phase != null) {
+      search = search.filter(map -> matchesPhase(map, phase));
+    } else {
+      search = search.filter(map -> map.getPhase() == Phase.PRODUCTION);
+    }
```
Pablete1234: Can be simplified to:

```java
Phase phase = phaseStr == null ? Phase.PRODUCTION : Phase.of(query);
if (phase == null) throw ...; // invalid phase passed as parameter
search = search.filter(map -> map.getPhase() == phase);
```

This also removes the matchesPhase method calling Phase.of(query); for each of the maps

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/command/MapCommand.java
applenick: Thanks for the feedback @Pablete1234! Excellent suggestions 👍

core/src/main/java/tc/oc/pgm/api/map/Phase.java
```diff
@@ -13,6 +13,11 @@ public Component toComponent() {
   }
 
   public static Phase of(String query) {
-    return query.startsWith("dev") ? DEVELOPMENT : PRODUCTION;
+    for (Phase phase : Phase.values()) {
+      if (phase.toString().toLowerCase().startsWith(query.toLowerCase())) {
```
Pablete1234:
```suggestion
    query = query.toLowerCase(Locale.ROOT);
    for (Phase phase : Phase.values()) {
      if (phase.toString().toLowerCase(Locale.ROOT).startsWith(query)) {
```

When using toLowerCase always include the locale, otherwise bad things can happen on special languages where rules you may think make sense no longer apply. Also, lowercasing the query can be moved out of the loop

Pablete1234 (changes requested):

Electroid (approved):

PR #901, opened 2021-07-14
Add clear-effects option to ClearItemsKit

CoWinkKeyDinkInc:
This PR adds a method to clear potions and other status effects from a player using a kit. This works just like the `<clear/>` and `<clear-items>` tags. This also adds a new method of parsing the ClearItemsKit which allows more finer configuration. Because of this, it is now possible to clear/keep only the armor slots.

```xml
<kit id="clear-effects" force="true">
      <clearinventory items="false" armor="false" effects="true"/>
</kit>
<!-- legacy -->
<kit id="clear-effects" force="true">
      <clear-effects/>
</kit>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/KitParser.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitParser.java
@@ -168,8 +168,8 @@ public abstract class KitParser {
   }
 
   public ClearItemsKit parseClearItemsKit(Element el) throws InvalidXMLException {
-    Element applyClear = el.getChild("clearinventory");
-    if (applyClear != null) {
+    Element applyClear = el.getChild("clear");
+    if (applyClear != null && applyClear.hasAttributes()) {
       boolean items = XMLUtils.parseBoolean(applyClear.getAttribute("items"), false);
       boolean armor = XMLUtils.parseBoolean(applyClear.getAttribute("armor"), false);
       boolean effects = XMLUtils.parseBoolean(applyClear.getAttribute("effects"), false);
@@ -178,7 +178,6 @@ public abstract class KitParser {
       // legacy
       if ("".equals(el.getChildText("clear"))) return new ClearItemsKit(true, true, false);
       if ("".equals(el.getChildText("clear-items"))) return new ClearItemsKit(true, false, false);
-      if ("".equals(el.getChildText("clear-effects"))) return new ClearItemsKit(false, false, true);
     }
     return null;
   }
--- a/core/src/main/java/tc/oc/pgm/kits/KitParser.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitParser.java
@@ -168,8 +168,18 @@ public abstract class KitParser {
   }
 
   public ClearItemsKit parseClearItemsKit(Element el) throws InvalidXMLException {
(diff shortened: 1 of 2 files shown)


Pablete1234:
Naming is a bit mixed-up here, you have `clearinventory` with no separation, alongside `clear-items` or `clear-effects`. Also, `clearinventory` has more of a connotation of "items in your inventory" not a generic clear for everything.

I'd consider not adding any new clear-effects, as it does not seem to be documented in `docs.oc.tc` so i don't think it was ever added, and updating `clear` to include the attributes for items, armor, and effects.

`<clear/>` <- old method, defaults to clearing items and armor, not effects
`<clear items="false" armor="true" effects="true"/>` <- clears armor and effects

I'd completely remove clear-effects & clearinventory, and deprecate clear-items in favor of `<clear armor="false"/>` in a higher proto (1.4.3)

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -168,8 +168,17 @@ public WalkSpeedKit parseWalkSpeedKit(Element el) throws InvalidXMLException {
...
-    if ("".equals(el.getChildText("clear-items"))) return new ClearItemsKit(false);
+    Element applyClear = el.getChild("clear");
+    if (applyClear != null && applyClear.hasAttributes()) {
+      boolean items = XMLUtils.parseBoolean(applyClear.getAttribute("items"), false);
+      boolean armor = XMLUtils.parseBoolean(applyClear.getAttribute("armor"), false);
+      boolean effects = XMLUtils.parseBoolean(applyClear.getAttribute("effects"), false);
+      return new ClearItemsKit(items, armor, effects);
+    } else {
+      // legacy
+      if ("".equals(el.getChildText("clear"))) return new ClearItemsKit(true, true, false);
```
Pablete1234: It will be extremely counter-intuitive if `<clear/>` clears items, however `<clear effects="false" />` does not, and neither does `<clear id="asdf" />` (the id does absolutely nothing here, but it's still something someone may try to do).

Keep the default items armor and effects consistent, not "legacy", simply make items and armor default to true, and effects default to false.

Also i'm pretty sure right now this isn't properly supported, as it seems like `el` is the parent and you search for a single `clear` child:
```xml
<kit>
    <clear items="true" armor="false" effects="false"/>
    <clear items="false" armor="true" effects="false"/>
</kit>
```
It should probably be supported even if it's a bit stupid

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/kits/KitParser.java
CoWinkKeyDinkInc: This would become inconsistent with how the rest of the kits are applied like the Shield, Double Jump and Fly Kit.

Pablete1234 (approved):

Pablete1234 (approved):

Pablete1234:
@Electroid any reason this isn't merged yet?

CoWinkKeyDinkInc:
Yep this is ready to merge and has backwards compatability.

Electroid (approved):

PR #904, opened 2021-07-25
Apply colors to TextFormatter with size of 1

CoWinkKeyDinkInc:
Discovered by @calcastor, where the colour on the gamemode title does not apply to maps using a single `<gamemode>` tag. 

image

This function is used in a handful of places and I've tested some of them.

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/text/TextFormatter.java
+++ b/util/src/main/java/tc/oc/pgm/util/text/TextFormatter.java
@@ -45,7 +45,7 @@ public final class TextFormatter {
       case 0:
         return empty();
       case 1:
-        return textList.get(0);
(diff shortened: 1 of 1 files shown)


CoWinkKeyDinkInc:
Build failed, but it doesn't look like it's my fault. It compiled properly on my end.

applenick:
> Build failed, but it doesn't look like it's my fault. It compiled properly on my end.

Seems to be a problem with ViaVersion's repo. It most likely worked on your end due to having a copy of the 3.0.2 ViaVersion cached in your local maven repo (from previous builds of PGM).

 I just submitted #905 which once merged should fix the compile issue.

Electroid (approved):

PR #905, opened 2021-07-26
Bump ViaVersion to 4.0.1

applenick:
This PR updates ViaVersion from 3.0.2 -> 4.0.1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,9 +77,9 @@
 
         <!-- Multi-protocol library for future Minecraft versions -->
         <dependency>
-            <groupId>us.myles</groupId>
+            <groupId>com.viaversion</groupId>
             <artifactId>viaversion</artifactId>
-            <version>3.0.2-SNAPSHOT</version>
+            <version>4.0.1</version>
             <scope>provided</scope>
             <optional>true</optional>
             <!-- Fix transitive dependency breaking builds -->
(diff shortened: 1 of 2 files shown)


calcastor:
ViaVersion has a legacy API that redirects any references to `us.myles.ViaVersion` to the new API. Maybe it would be best to keep that to allow servers to use older versions of ViaVersion et al. if necessary, given bugs such as dropped items showing as stone for all 1.7 users with upstream ViaRewind for instance?

applenick:
> ViaVersion has a legacy API that redirects any references to `us.myles.ViaVersion` to the new API. Maybe it would be best to keep that to allow servers to use older versions of ViaVersion et al. if necessary, given bugs such as dropped items showing as stone for all 1.7 users with upstream ViaRewind for instance?

Oh shoot! Is that a thing? 

This was mainly to fix the compile error in #904

applenick:
Thanks @calcastor, found the api-legacy package! That was neat to learn. All fixed now 👍

calcastor:
> > ViaVersion has a legacy API that redirects any references to `us.myles.ViaVersion` to the new API. Maybe it would be best to keep that to allow servers to use older versions of ViaVersion et al. if necessary, given bugs such as dropped items showing as stone for all 1.7 users with upstream ViaRewind for instance?
> 
> Oh shoot! Is that a thing?
> 
> This was mainly to fix the compile error in #904

That is my understanding of what [this](https://github.com/ViaVersion/ViaVersion/blob/8fb51bf3b7d1106c9aafe2344556eb2b82207061/api-legacy/src/main/java/us/myles/ViaVersion/api/Via.java) does at least. I've been modifying the pom.xml to reference the latest version of ViaVersion when I build locally, but haven't been updating ViaUtils and it has seemed to work 100% as intended. I'd assume that updating ViaUtils to not use the legacy API might break it when using older versions of ViaVersion. 

edit: see above I guess I missed a comment when writing this out haha!

PR #908, opened 2021-08-07
Rebrand some filters as query modifiers

KingOfSquares:


Electroid:
This looks good, but could you explain the motivation/what the changes will do?

KingOfSquares:
> This looks good, but could you explain the motivation/what the changes will do?

The filters i changed gave no good indication that they have children, having common interfaces/abstract implementations (`SingleFilterFunction MultiFilterFunction QueryModifier`) makes it easier to know how the filters will behave. 

The next step I want to commit is a `toString` implementation on `QueryModifier` that matches the other two classes to make debugging during xml parsing(or anywhere really) easier for scenarios you want to prevent usage of some filters. An example of a scenario where this would be beneficial is when accepting only dynamic filters for something like a portal trigger. If the user provides an invalid filter inside one of the filters that has children its easy to tell the user what's wrong.

I know a `toString` could easily be implemented on all the filters I changed in this commit, but then it would be unnecessary boilerplate.

Electroid (approved):

PR #910, opened 2021-08-07
annotate TNTRenderMatchModule with ListenerScope

KingOfSquares:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
@@ -17,8 +17,10 @@ import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.match.MatchScope;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.ListenerScope;
 import tc.oc.pgm.util.nms.NMSHacks;
 
+@ListenerScope(value = MatchScope.LOADED)
 public class TNTRenderMatchModule implements MatchModule, Listener {
 
   private static final double MAX_DISTANCE = Math.pow(64d, 2);


core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
```diff
@@ -17,8 +17,10 @@
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.match.MatchScope;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.ListenerScope;
 import tc.oc.pgm.util.nms.NMSHacks;
 
+@ListenerScope(value = MatchScope.LOADED)
```
Electroid, line 23: I think this should actually just be `RUNNING`.

Electroid: But if we do this, we need to clean up `entities` when the match is finished. (we can't guarantee the recurring task will execute one last time)

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
KingOfSquares: Why should this not run before/after a match? e.g seeing where your TNT _would_ have exploded right after a match ends seems reasonable.

Electroid: It's just unnecessary.. a lot of stuff can happen post-cycle, sending a more fake packets to clients for no reason isn't a good idea.

KingOfSquares: Ok, ill move it to `Running` and figure out the cleaning thing

Pugzy: I agree with KingOfSquares on this one having it scoped to `LOADED`. 
TNT still triggers and chains after the match ends so no reason to stop the module on match end.

Pablete1234: I also think it should stay on LOADED, match not running doesn't mean tnt can't still be moving.

Pablete1234 (approved):

Electroid (approved):

PR #836, opened 2021-03-23
Time Adjust 

applenick:
# Time Adjust

New feature suggested by dragonstomper64 👏 

Basically allows for objectives to extend the time limit of a match once completed (core leaked, wool placed, destroyable destroyed) 

## XML 
```xml
<time-adjust>

<!-- When red core is leaked, add 10 mins -->
<adjust id="red-core" time="10m"/> 

<!-- When blue wool is placed, subtract 5 mins -->
<adjust id="blue-wool" time="-5m"/> 

<!-- When green monument is destroyed, add 10 mins and don't broadcast -->
<adjust id="green-monument" time="10m" broadcast="false"/> 

</time-adjust>
```

### Wool
https://youtu.be/6zOz3aXkQa0

### Core
https://youtu.be/eDyttJGii40

### Destroyables
https://youtu.be/xYo0gNiur24

## Screenshots
Example of broadcast when time is increased
increase

Example of broadcast when time is decreased
decrease

## ToDo

Objectives:
- [x] Cores
- [x] Wools
- [x] Destroyables
- [ ] Flags
- [ ] Control Points
- [ ] Score box?

## Feedback
Any feedback on how to improve this would be appreciated 👍 

Signed-off-by: applenick <applenick@users.noreply.github.com>

PR #914, opened 2021-08-22
Dynamic filters!

KingOfSquares:
## Dynamic filters 🔮:tada:

This PR proposes an implementation of the dynamic filter concept originally introduced in ProjectAres.
In addition it also comes bundled with:
- A refactor of the Portal feature to be able to use dynamic filters for triggering (the other use-cases like kits/scoreboxes/structures can be added later)
- Removal of Player, Party and Match `#getQuery` since they now implement `Filterable` (which means that the objects implement the query instead)
- Add a TODO in PlayerQuery about some wrong usage (the fix is not in scope for this PR)

After this PR any feature in PGM can register FilterListeners through `#onChange/onRise/onFall` which will enable the feature to be able to trigger functionality whenever a dynamic filter matches/unmatches its condition.

From the outside, the functionality of dynamic filters is quite similar to how it was in ProjectAres with one exception. Filters now have to override `#getRelevantEvents` and declare which events should be listened to that could change a filters opinion on someone (e.g. if KillStreakFilter is used MatchPlayerDeathEvent should be listened for because that might change a filters opinion on a player). If it does not override the method the filter is regarded as non-dynamic.

On the inside (inside `FilterMatchModule`) there have been changes from how ProjectAres solved it.
Before, `FilterMatchModule` would listen for a set of hardcoded events ([here](https://github.com/OvercastNetwork/ProjectAres/blob/7484cb81fcc33814fe26812684660da4e0dbd671/PGM/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java#L203-L273)) and invalidate players/the match whenever those events where called.
In this PR, the module only listens for events that the filters say are relevant (`Filter#getRelevantEvents()`). If multiple filters need the same event it is only listened for once.

This new approach enables more types of filters to be dynamic, and allows the module to be more precise in _when_ to invalidate someone.

In addition to all the old dynamic filters (excluding the time filter) the following filters are now dynamic:
- kill-streak
- carrying (item)
- holding (item)
- wearing (item)
Also, all filters with children(also query modifiers) are dynamic if their child(ren) are

@Electroid Please remember to credit both me and Pugzy in the final commit

##### The testing of these changes was powered by the [diamyx-pack](https://github.com/Diamyx/diamyx-pack)


core/src/main/java/tc/oc/pgm/api/match/Match.java
```diff
@@ -407,4 +408,35 @@ default boolean finish() {
...
+      result = Stream.concat(result, Stream.of((R) this));
+    }
+    if (Party.class.isAssignableFrom(type)) {
+      result = Stream.concat(result, (Stream<R>) getParties().stream().filter(type::isInstance));
+    }
+    if (type.isAssignableFrom(MatchPlayer.class)) {
+      result = Stream.concat(result, (Stream<? extends R>) getPlayers().stream());
+    }
+    return result;
+  }
```
Electroid: Let's keep this to `MatchImpl`

core/src/main/java/tc/oc/pgm/filters/TimeFilter.java
```diff
@@ -9,7 +9,7 @@
 
   public TimeFilter(Duration duration) {
     this.duration = duration;
-  }
+  } // TODO, time queue
```
Electroid, line 12: Are there time or countdown events it could listen to?

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterDispatcher.java
```diff
@@ -0,0 +1,85 @@
+package tc.oc.pgm.filters.dynamic;
+
+import java.util.function.Consumer;
+import tc.oc.pgm.api.filter.Filter;
+
+public interface FilterDispatcher {
```
Electroid, line 6: I don't think the naming of rise/fall was obvious in the previous implementation. I would think `onUpdate`, `onAllow`, and `onDeny` are more obvious.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -0,0 +1,605 @@
...
+                + " listener="
+                + listener
+                + " filter="
+                + filter);
+    register(scope, filter, true, listener);
+    register(scope, filter, false, listener);
+  }
+
+  /**
+   * {@inheritDoc}
```
Electroid: We don't need to copy the interface docs, we can just remove all of these `inheritDoc` sections.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -0,0 +1,605 @@
...
+  @Override
+  public void tick(Match match, Tick tick) {
+    if (!match.isRunning()) return;
+
+    this.tick();
+  }
+
+  public void tick() {
+    final Set<Filterable<?>> checked = new HashSet<>();
+    for (; ; ) {
```
Electroid: This is bad style, let's change to a `while(true)`

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -0,0 +1,605 @@
...
+        try {
+          this.cachedHandles.put(event, this.createExtractingMethodHandle(event));
+        } catch (NoSuchMethodException | IllegalAccessException e) {
+          this.cachedHandles.remove(event);
+          match
+              .getLogger()
+              .severe("Unable to create MethodHandle extracting Filterable or Player for " + event);
+          e.printStackTrace();
+          continue;
+        }
```
Electroid: Utility function

core/src/main/java/tc/oc/pgm/api/filter/Filter.java
```diff
@@ -1,14 +1,53 @@
...
   QueryResponse query(Query query);
 
+  default boolean response(Query query) {
+    switch (query(query)) {
+      case ALLOW:
+        return true;
+      case DENY:
+        return false;
+      default:
+        throw new UnsupportedOperationException("Filter did not respond to the query");
```
Electroid, line 47: Is there a reason we need to make this throw? (having functions that returns booleans that throw are not nice)

Could this just be `isAllowed` ?

core/src/main/java/tc/oc/pgm/api/party/Party.java
```diff
@@ -132,4 +137,19 @@ default boolean isParticipating() {
   default boolean isObserving() {
     return !this.isParticipating();
   }
+
+  @Override
+  default Optional<? extends Filterable<? super PartyQuery>> filterableParent() {
```
Electroid: Let's do `Nullable` instead of optionals.

core/src/main/java/tc/oc/pgm/api/party/Party.java
```diff
@@ -132,4 +137,19 @@ default boolean isParticipating() {
...
+    return Optional.of(getMatch());
+  }
+
+  @Override
+  default Stream<? extends Filterable<? extends PlayerQuery>> filterableChildren() {
```
Electroid: `Iterable` instead of `Stream`

core/src/main/java/tc/oc/pgm/api/region/Region.java
```diff
@@ -75,4 +77,6 @@
   public Iterator<BlockVector> getBlockVectorIterator();
 
   public Iterable<BlockVector> getBlockVectors();
+
+  public Collection<Class<? extends Event>> getRelevantEvents();
```
Electroid, line 81: Can this be `Iterable`?

Electroid (changes requested):
Very cool to see both of your work on this. I've done a high-level pass, but we'll need some more reviews

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterDispatcher.java
Pablete1234: I disagree on this one, rise and fall are way better names to represent this.

Rise means "onChangeFromDenyToAllow" if you call it onAllow one may think it keeps on repeatedly being called while the filter allows, which is not the case, same for onDeny

core/src/main/java/tc/oc/pgm/api/filter/Filter.java
Pablete1234: https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/Filter.java#L29

This was taken from 1.9+ PGM, i think it works as intended, you should **only** use this shortcut method if you know it **will** give a response, `assertRespondsTo(Class)` method is mentioned in the original but that method is not here.

Also keep in mind for all of the dynamic filter stuff response **has** to be either allow or deny, if it's neither when do you tell it's a raise and when it's a fall? It's essentially an exception for unsupported case, which is justified, i think the javadoc should make it clear that this is for special cases only where you know filter won't abstain, and to use #query instead for general use.

Regarding using just `isAllowed`, that would be hiding bugs under the carpet, if a filter abstains it would be treated as a deny, triggering a fall, and when back to allow it would trigger a raise

core/src/main/java/tc/oc/pgm/api/party/Party.java
Pablete1234: `Iterable` is a pain to work with, just return `Collection` and let the callers do .stream on the collection

core/src/main/java/tc/oc/pgm/api/region/Region.java
Pablete1234: `Iterables` are a pain to work with, you can literally only get an iterator from them, can't know size, can't easily stream (without going thru stream support hoops), and they're generally not any better to return than a collection, when we **know** it's a collection.

I know the intent here is to return the most abstract type not to tie the implementation to any specific type, but Collection is still a very non-specific type (eg: it's not a List or Set), and is way nicer to work with than iterable

core/src/main/java/tc/oc/pgm/filters/CarryingItemFilter.java
```diff
@@ -1,13 +1,33 @@
...
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    return ImmutableList.of(
+        PlayerPickupItemEvent.class,
+        PlayerDropItemEvent.class,
+        ApplyKitEvent.class,
+        InventoryClickEvent.class,
+        InventoryDragEvent.class,
+        PlayerItemBreakEvent.class);
+  }
```
Pablete1234: There's PGM generifying events `ItemTransferEvent` & `PlayerItemTransferEvent` that would trigger under most if not all of these circumstances

core/src/main/java/tc/oc/pgm/filters/FlagStateFilter.java
```diff
@@ -23,6 +27,11 @@ public FlagStateFilter(
     this.state = state;
   }
 
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    return ImmutableList.of(FlagStateChangeEvent.class);
```
Pablete1234: For all of these one-event collections consider using a singleton:
```suggestion
    return Collections.singleton(FlagStateChangeEvent.class);
```
They may be simpler than an immutable list, and they're immutable too

core/src/main/java/tc/oc/pgm/filters/MultiFilterFunction.java
```diff
@@ -33,6 +36,13 @@ public MultiFilterFunction(Iterable<? extends Filter> filters) {
     this.filters = ImmutableList.copyOf(filters);
   }
 
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    List<Class<? extends Event>> allEvents = new ArrayList<>();
+    this.filters.forEach(f -> allEvents.addAll(f.getRelevantEvents()));
+    return ImmutableList.copyOf(allEvents);
```
Pablete1234:
```suggestion
    return this.filters.stream().flatMap(f -> f.getRelevantEvents().stream())
            .collect(Collectors.collectingAndThen(Collectors.toList(), ImmutableList::copyOf));
```
It hurts to see lambdas being used to "fill a collection" outside of the stream itself, when the stream is what should be collected into a list itself

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -0,0 +1,605 @@
...
+
+  // #getActor is mostly for SportPaper (PlayerAction).
+  // #getPlayer covers events which does not provide Entity
+  // as the lower boundary
+  private final String[] entityGetterNames = new String[] {"getPlayer", "getActor", "getActor"};
+  private final MethodType playerMethodType = MethodType.methodType(Player.class);
+  private final MethodType[] entityMethodTypes =
+      new MethodType[] {
+        this.playerMethodType, this.playerMethodType, MethodType.methodType(Entity.class)
+      };
```
Pablete1234: This is a 600 line class where half of it is about doing reflection on events to get player or actor

Outsource all of that logic to somewhere else and keep this to just the filter stuff, whenever you need to get this stuff call the util.

Make that util static and reuse all of the event -> method stuff, you don't need to create that for each match

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -0,0 +1,605 @@
...
+
+    this.tick();
+  }
+
+  public void tick() {
+    final Set<Filterable<?>> checked = new HashSet<>();
+    for (; ; ) {
+      // Collect Filterables that are dirty, and have not already been checked in this tick
+      final Set<Filterable<?>> checking = ImmutableSet.copyOf(Sets.difference(dirtySet, checked));
+      if (checking.isEmpty()) break;
```
Pablete1234:
```suggestion
    final Set<Filterable<?>> checked = new HashSet<>(), checking;
    while (!(checking = ImmutableSet.copyOf(Sets.difference(dirtySet, checked))).isEmpty()) {
```
It may not look the cleanest, but at least it's not an infinite loop and the exit condition is clear
Could probably put `ImmutableSet.copyOf(Sets.difference(dirtySet, checked))` in a function and call that for clarity:
`while (!(checking = getMissingChecks()).isEmpty())`

CoWinkKeyDinkInc:
#836 Can be recreated after this PR is merged

core/src/main/java/tc/oc/pgm/api/region/Region.java
Electroid: That's fine with me.

core/src/main/java/tc/oc/pgm/api/filter/Filter.java
Electroid: Sounds good.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterDispatcher.java
KingOfSquares: I have to agree with Pablo here, `Rise` being a separate term is important to understand that it triggers only once when it starts to allow. The explanation for each method is described quite thoroughly in the javadocs anyways.

core/src/main/java/tc/oc/pgm/filters/FlagStateFilter.java
KingOfSquares: `ImmutableList.of(E element)` return a `SingetonImmutableList`, but the native Collection will always be less "fancy". Since we don't need any features of the `ImmutableList` ill go for `Collections.singleton`

core/src/main/java/tc/oc/pgm/filters/TimeFilter.java
KingOfSquares: No, there is no events that mirror time flow in game except the countdown start, end and cancel events(afaik). There are many ways to solve this, but it felt like this PR had enough scope creep already.

core/src/main/java/tc/oc/pgm/api/party/Party.java
```diff
@@ -132,4 +135,20 @@ default boolean isParticipating() {
   default boolean isObserving() {
     return !this.isParticipating();
   }
+
+  @Override
+  @Nullable
+  default Match filterableParent() {
```
Electroid: Let's stick with the `get` style in this repo, that style is a remanent of the previous PGM.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
KingOfSquares: Added bonus of moving out this stuff is that now cached handles persist over different matches.

core/src/main/java/tc/oc/pgm/util/MethodHandleUtils.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return null;
+  }
+
+  public static void removeCachedHandle(Class<? extends Event> event) {
+    cachedHandles.remove(event);
+  }
+
+  @SuppressWarnings("unchecked")
+  public static MethodHandle getHandle(Class<? extends Event> event) {
```
KingOfSquares: Ideally this should be the only entry point, since this class lives independent of all other PGM stuff.

e.g `getHandle` should call `createExtractingMethodHandle` the first time it gets a new event.

Pablete1234 (approved):

PR #915, opened 2021-08-24
Ensure control points using "recovery" and "decay" work as intended

calcastor:
#864 reintroduced decay and recovery rates to control points from 1.9+ PGM, while also introducing the possibility for already captured control points to decay from a captured state. However, ControlPointParser only recognizes and parses `recovery-rate` and `decay-rate`, despite 1.9+ PGM's implementation using `recovery` and `decay`. This PR ensures compatibility with maps previously made for 1.9+ PGM by making sure `recovery` and `decay` are also recognized.

I've also made sure that `owned-decay` can also be used alongside `owned-decay-rate`, to keep wording consistent between all 3 attributes.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
@@ -79,9 +79,9 @@ public abstract class ControlPointParser {
 
     final double decayRate, recoveryRate, ownedDecayRate;
     final Node attrIncremental = Node.fromAttr(elControlPoint, "incremental");
-    final Node attrDecay = Node.fromAttr(elControlPoint, "decay-rate");
-    final Node attrRecovery = Node.fromAttr(elControlPoint, "recovery-rate");
-    final Node attrOwnedDecay = Node.fromAttr(elControlPoint, "owned-decay-rate");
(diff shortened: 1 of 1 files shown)


Pablete1234 (approved):
Nice catch, didn't notice we weren't complying with the 1.9+ pgm standard when we used these names.

For reference, the original names were definetly recovery and decay:
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java#L77

Electroid (approved):

CoWinkKeyDinkInc:
It seems this PR has broken the dropper maps which use `recovery="false"`.

calcastor:
> It seems this PR has broken the dropper maps which use `recovery="false"`.

Correct me if I'm wrong, but while the [1.9+ PGM docs](https://docs.oc.tc/modules/control_points) state that `true` and `false` were acceptable values for `recovery` and `decay`, the code in 1.9+ PGM [seen here](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java#L75) doesn't seem to imply that those ever functioned as acceptable values. In other words, I don't think these maps were using the attribute correctly to begin with.

PR #917, opened 2021-09-07
Ensure doDaylightCycle is set to correct value on match start

calcastor:
#879 ensures that game rules like `doDaylightCycle` and `doFireTick` are only enabled once a match starts, if the map has them enabled. However, it seems that the definition of `doDaylightCycle` from the `level.dat` files of maps unintentionally takes precedence over the default setting of `<timelock>` to off, when `<timelock>`  and `doDaylightCycle`  are both undefined within a map's XML. This PR ensures that the default `<timelock>` setting of `off` again takes precedence over the map world's setting, while also fixing an overlooked albeit small error that ensures `doFireTick` is disabled once a match ends.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
@@ -19,21 +19,23 @@ public class GameRulesMatchModule implements MatchModule {
 
   @Override
   public void load() {
-    // saves and sets gamerules from XML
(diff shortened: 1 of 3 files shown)


Pablete1234:
Double-check that gamerules module has world module as a dependency, otherwise the world time module may silently just be null

Electroid:
> Double-check that gamerules module has world module as a dependency, otherwise the world time module may silently just be null

Yea, a quick check shows that the dependency is not setup properly.

Electroid (approved):

PR #926, opened 2021-10-09
Add Filters to Monument Modes

CoWinkKeyDinkInc:
Call me Mr Monument Modes with how much PRs I make for it :sunglasses: 

This adds a `filter` attribute to `<mode>` which triggers when a dynamic filter is set to or changes to allow.

```xml
<modes>
    <!-- Activates 10 seconds after the condition in mode-filter is allow -->
    <mode id="glass-mode" filter="mode-filter" after="10s" material="glass" name="`eGlass Mode" show-before="10s"/>
    <!-- Activates 20 seconds after match start -->
    <mode id="gold-mode" after="20s" material="gold block" name="`eGold Mode" show-before="10s"/>
</modes>
<filters>
    <all id="mode-filter">
        <objective>red-mon</objective>
    </all>
</filters>
<cores material="obsidian" leak="2">
    <core team="red-team" region="left-core" name="Red Core" modes="gold-mode"/>
    <core team="blue-team" region="right-core" name="Blue Core" modes="gold-mode"/>
</cores>
<destroyables>
    <destroyable id="red-mon" name="Red Monument" owner="red-team" region="mon-red" materials="obsidian" modes="glass-mode"/>
    <destroyable id="blue-mon" name="Blue Monument" owner="blue-team" region="mon-blu" materials="obsidian" modes="glass-mode"/>
</destroyables>
```

Currently a few more small updates need to be made, such as updating the `/modes list`, and other mode commands which as of this post when used will stop the modes from activating. Special thanks for @KingOfSquares and @Pugzy for readding the dynamic filters making this all possible :)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modes/Mode.java
+++ b/core/src/main/java/tc/oc/pgm/modes/Mode.java
@@ -17,8 +17,7 @@ public class Mode extends SelfIdentifyingFeatureDefinition {
   private final MaterialData material;
   private final Duration after;
   private final Filter filter;
-  private final @Nullable
-  String name;
+  private final @Nullable String name;
   private final Component componentName;
   private final Duration showBefore;
   private boolean modeComplete = false;
@@ -28,32 +27,36 @@ public class Mode extends SelfIdentifyingFeatureDefinition {
   }
 
   public Mode(
-          final @Nullable String id,
-          final MaterialData material,
-          final Duration after,
-          final Filter filter,
-          final @Nullable String name,
-          Duration showBefore
-  ) {
+      final @Nullable String id,
+      final MaterialData material,
+      final Duration after,
+      final Filter filter,
+      final @Nullable String name,
+      Duration showBefore) {
     super(id);
     this.material = material;
     this.after = after;
     this.filter = filter;
     this.name = name;
     this.componentName =
-            text(name != null ? name : getPreformattedMaterialName(), NamedTextColor.RED);
+        text(name != null ? name : getPreformattedMaterialName(), NamedTextColor.RED);
     this.showBefore = showBefore;
   }
 
-  public void load(FilterMatchModule fmm, ModeChangeCountdown countdown, CountdownContext countdownContext, Match match) {
+  public void load(
+      FilterMatchModule fmm,
+      ModeChangeCountdown countdown,
+      CountdownContext countdownContext,
+      Match match) {
     // if filter returns ALLOW at any time in the match, start countdown for mode change
-    fmm.onRise(filter, FilterListener -> {
-      if (match.isRunning() && !modeComplete) {
-        countdownContext.start(countdown, countdown.getMode().getAfter());
-        modeComplete = true;
-      }
-    }
-    );
+    fmm.onRise(
+        filter,
+        FilterListener -> {
+          if (match.isRunning() && !modeComplete) {
+            countdownContext.start(countdown, countdown.getMode().getAfter());
+            modeComplete = true;
+          }
+        });
   }
 
   public MaterialData getMaterialData() {
@@ -84,8 +87,7 @@ public class Mode extends SelfIdentifyingFeatureDefinition {
     return this.modeComplete;
   }
 
-  public @Nullable
-  String getName() {
+  public @Nullable String getName() {
     return this.name;
   }
 
--- a/core/src/main/java/tc/oc/pgm/modes/Mode.java
+++ b/core/src/main/java/tc/oc/pgm/modes/Mode.java
@@ -13,7 +13,7 @@ import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 public class Mode extends SelfIdentifyingFeatureDefinition {
   private final MaterialData material;
   private final Duration after;
-  private final Filter filter;
+  private final @Nullable Filter filter;
   private final @Nullable String name;
   private final Component componentName;
   private final Duration showBefore;
@@ -26,7 +26,7 @@ public class Mode extends SelfIdentifyingFeatureDefinition {
       final @Nullable String id,
       final MaterialData material,
       final Duration after,
-      final Filter filter,
+      final @Nullable Filter filter,
       final @Nullable String name,
       Duration showBefore) {
     super(id);
@@ -59,7 +59,7 @@ public class Mode extends SelfIdentifyingFeatureDefinition {
     return this.showBefore;
   }
 
-  public Filter getFilter() {
+  public @Nullable Filter getFilter() {
     return this.filter;
   }
 
--- a/core/src/main/java/tc/oc/pgm/modes/Mode.java
+++ b/core/src/main/java/tc/oc/pgm/modes/Mode.java
@@ -8,6 +8,7 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
 import org.bukkit.material.MaterialData;
 import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.countdowns.CountdownContext;
 import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 import tc.oc.pgm.filters.dynamic.FilterMatchModule;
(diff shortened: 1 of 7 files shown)


core/src/main/java/tc/oc/pgm/modes/Mode.java
```diff
@@ -7,34 +7,58 @@
...
+import tc.oc.pgm.filters.dynamic.FilterMatchModule;
 
 public class Mode extends SelfIdentifyingFeatureDefinition {
   private final MaterialData material;
   private final Duration after;
+  private final Filter filter;
   private final @Nullable String name;
   private final Component componentName;
   private final Duration showBefore;
+  private boolean modeComplete = false;
```
Pablete1234: Feature definitions must be immutable. Try playing the same map twice, it will work the first time, but will not work the second.

`modeComplete` is a bit ambiguous, you may think it's the mode change is complete, but it isn't. It's the countdown having started, which can be misleading. Try to find a better name.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/modes/ObjectiveModesMatchModule.java
```diff
@@ -38,17 +46,26 @@ public ObjectiveModesMatchModule(Match match, List<Mode> modes) {
...
     for (ModeChangeCountdown countdown : this.countdowns) {
-      this.countdownContext.start(countdown, countdown.getMode().getAfter());
+      Query query = new MatchQuery(match.getEvent(), match);
+      if (countdown.getMode().getFilter().query(query).isAllowed()
+          || countdown.getMode().getFilter() == null) {
+        if (!countdown.getMode().isModeComplete()) {
+          this.countdownContext.start(countdown, countdown.getMode().getAfter());
+          countdown.getMode().setModeComplete();
+        }
+      }
```
KingOfSquares: Why is the mode queried manually? If it has a dynamic filter it will be queried by the filter match module

Pablete1234: this is checking if the filter is null or if it matches, right at match start

i think the rationale here is it may not trigger an "on rise" if it was always allowed (See: an ALWAYS filter) on match start

CoWinkKeyDinkInc: I found a more smarter way which makes the default filter null if undefined in XML, so this doesn't have to deal with the dynamic filter which complains about `StaticFilter.ALLOW`.

CoWinkKeyDinkInc:
Here are a few example images of what a player would see when using `/modes list`. Please tell me if I am using the correct checkmark and crossmark, I couldn't find them in the code.

CoWinkKeyDinkInc:
@KingOfSquares On random occasion the server will stop when I try to join because "The Filter did not respond to the query", what do you know could be done to fix this? Could it be because the listener is created and checking before the match is fully loaded?

core/src/main/java/tc/oc/pgm/modes/Mode.java
```diff
@@ -7,34 +7,50 @@
...
   }
 
+  public void load(
+      FilterMatchModule fmm, ModeChangeCountdown countdown, CountdownContext countdownContext) {
+    // if filter returns ALLOW at any time in the match, start countdown for mode change
+    fmm.onRise(
+        filter,
+        FilterListener -> {
+          countdownContext.start(countdown, countdown.getMode().getAfter());
+        });
```
Pablete1234:
```suggestion
        listener -> countdownContext.start(countdown, countdown.getMode().getAfter()));
```
variable names always lowercase, also you can just in-line this one

core/src/main/java/tc/oc/pgm/modes/Mode.java
```diff
@@ -7,34 +7,50 @@
...
+  }
+
```
Pablete1234: get this logic away from here.

This is a **Definition** this must ONLY represent the values/data in the XML, it should hold no mutable state, and this kind of logic also makes no sense here. That is match-scoped and should be part of the match-module that does this logic, or an util if you wish.

Notice this function is literally using nothing from this class except `filter`, which you already have a getter for, all other things used are passed as parameters, and just don't belong in here

Pablete1234 (changes requested):
There's also a general question of how do you mix-and-match dynamic modes and non-dynamic modes.

What if you have glass core mode whenever they destroy the monument, and gold at 15 minutes.

If they get the monument within 15 minutes they switch straight to glass, and then at 15 to gold for the rest of the match?
However if they wait it out, it goes to gold in 15 minutes, and later when they get the monument it goes to glass and stays like that the rest of the match?

I think some thought should be put into those scenarios

core/src/main/java/tc/oc/pgm/modes/Mode.java
CoWinkKeyDinkInc: I have done this based off of the implementation of the dynamic filter in `Portal.java`, which also has this logic in the same location and borrows only one trait in the class. This listener does not change any traits in the definition

CoWinkKeyDinkInc:
I added a `/mode start` command so you can start a countdown for a specific objective mode. You select the mode based on the number in `/mode list`. This command starts a countdown at a certain time period instead of increasing or decreasing the countdown like in `/mode push`. Made in mind for modes with a filter, this works for both regardless of filter.

## Example
`/mode start 2 20m` - Starts the second monument mode in 20 minutes

core/src/main/java/tc/oc/pgm/modes/Mode.java
Pablete1234: After giving it quite a bit of thought and talking it with @Pugzy, i still think it should be part of the match module, not the feature definition

`Portal.java` load method should be moved to the match module or somewhere else that is match-scoped, not map-scoped, but that's a separate issue to be adressed in another pr

CoWinkKeyDinkInc: Sounds reasonable, I did think it was odd that it was in the non match module regardless.

CoWinkKeyDinkInc:
After moving the filter, I still experience the same infrequent issue of the filter not responding to the query when first joining the server. I need @Pugzy or @KingOfSquares to help me out on this one, I'm not sure if I need to load something first or a mistake with the listeners. Other than this issue I would say the PR is ready to go.

util/src/main/i18n/templates/misc.properties
```diff
@@ -120,6 +120,12 @@ misc.or = {0} or {1}
 # do not translate, keep as-is
 misc.snowman = ☃
 
+# do not translate, keep as-is
+misc.checkmark = ✔
+
+# do not translate, keep as-is
+misc.crossmark = ✘
```
Electroid: We don't need to translate these, let's put them as static `TextComponent` constants.

util/src/main/i18n/templates/command.properties
```diff
@@ -66,6 +68,21 @@ command.simplePageHeader = {0} of {1}
...
+command.modesPushed = Affected modes have been pushed
+
+# {0} = mode name (e.g. GOLD, Glass Mode)
+command.selectedModePushed = {0} will start in
+
+command.forwards = forwards
+
+command.backwards = backwards
+
+command.by = by
```
Electroid: This needs more context for translators, same as the above.

Electroid (changes requested):

util/src/main/i18n/templates/misc.properties
CoWinkKeyDinkInc: Where would these go?

core/src/main/java/tc/oc/pgm/modes/Mode.java
```diff
@@ -7,28 +7,32 @@
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
 import org.bukkit.material.MaterialData;
+import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 
 public class Mode extends SelfIdentifyingFeatureDefinition {
   private final MaterialData material;
   private final Duration after;
+  private final Filter filter;
```
Pablete1234:
```suggestion
  private final @Nullable Filter filter;
```

core/src/main/java/tc/oc/pgm/modes/Mode.java
```diff
@@ -7,28 +7,32 @@
...
   public Mode(final MaterialData material, final Duration after, Duration showBefore) {
-    this(null, material, after, null, showBefore);
+    this(null, material, after, null, null, showBefore);
   }
 
   public Mode(
       final @Nullable String id,
       final MaterialData material,
       final Duration after,
+      final Filter filter,
```
Pablete1234:
```suggestion
      final @Nullable Filter filter,
```

core/src/main/java/tc/oc/pgm/modes/Mode.java
```diff
@@ -55,6 +59,10 @@ public Duration getShowBefore() {
     return this.showBefore;
   }
 
+  public Filter getFilter() {
```
Pablete1234:
```suggestion
  public @Nullable Filter getFilter() {
```

Pablete1234 (commented):
I'd add the nullable annotation to make it clear that it can be null, besides that lgtm

Pablete1234 (approved):

Electroid (approved):

PR #931, opened 2021-10-24
Add infinite attribute to blocks

### LINKED ISSUE #781: Infinite Stacks
I think it would be a nice addition to have infine stacks. E.g. could be used in tnt maps supplying infinite tnt.

CoWinkKeyDinkInc:
Fixes #781 as suggested by @RuedigerLP 

## Example
```xml
<kit id="the-kit" force="true">
        <item slot="0">bow</item>
        <item slot="2" amount="64">arrow</item>
        <!-- gives infinite blocks --> 
        <item slot="8" material="bedrock" infinite="true"/>
        <helmet locked="true">chainmail helmet</helmet>
        <boots locked="true">chainmail boots</boots>
        <effect duration="900" amplifier="3">jump_boost</effect>
    </kit>
```

This PR gives affected blocks `Infinite ∞` in the lore, which is unable to be created or edited in survival, to safely hold the trait. The infinity character on it's own does not look so well in game so I prefixed it with infinite to make it more clear. I tried expanding this to potions but it did not seem to work out well.


Pablete1234:
In the current state, this is extremely hacky.

In the xml syntax, adapt pgm so that `amount="oo"` can be used instead of adding a new infinite attribute. It is very common pgm syntax to use "oo" for infinity (eg: region coordinates)

Secondly, giving 1 item is a bad idea for several reasons. For one the user may not notice it's infinite or they have multiple, and may think they have just one. For this purpose either a stack with 64 or 127 (abusing mc a bit) may be a better. Another problem with single item is that client-server desyncs will be frequent, you click twice, clent-side you wasted your block, and you don't see it come back until after your 2nd click, where the block latee pops-up as a server-side place instead of a client-side one. Players fast-bridging will want to kill you.

Giving more than one item should also make it easier as you can just change the amount of items in the stack back to 64 (or 127, or -1, whatever works best).
It also comes with challenges, as the stack can be split in half and you may not want them to be sharing half of the infinite block. You should monitor inventory events and make sure the stack always stays as a whole.

Lastly, do not ever use lore string as a check, we're better than that. Look at how metadata for grenades or itemmods is done, pgm has some nice utils for it, it saves as an internal nbt metadata flag

CoWinkKeyDinkInc:
The PR now does everything smarter like you were saying. Players will see the infinite block as holding -1 items in it, when the block is placed they are given -1 blocks, which resets it back to -1. The other advantage of using -1 blocks is that it cannot be split, stacked, or dragged, so listeners aren't needed to deal with those events.

```xml
<kits>
    <kit id="the-kit" force="true">
        <item slot="0">bow</item>
        <item slot="2" amount="64">arrow</item>
        <item slot="8" material="bedrock" amount="oo"/>
    </kit>
</kits>
```

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -422,7 +429,12 @@ public ItemStack parseItem(Element el, Material type, short damage) throws Inval
       throw new InvalidXMLException("Invalid item/block", el);
     }
 
+    if (amountString != null && amountString.equals("oo") && !itemStack.getType().isBlock()) {
```
```suggestion
    if (amount == -1 && !itemStack.getType().isBlock()) {
```

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -539,6 +551,10 @@ public void parseCustomNBT(Element el, ItemStack itemStack) throws InvalidXMLExc
       ItemTags.PREVENT_SHARING.set(itemStack, true);
     }
 
+    if (el.getAttributeValue("amount") != null && el.getAttributeValue("amount").equals("oo")) {
```
```suggestion
    if (amount == -1) {
```

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -413,7 +413,14 @@ public ItemStack parseItem(Element el, Material type) throws InvalidXMLException
...
   public ItemStack parseItem(Element el, Material type, short damage) throws InvalidXMLException {
-    int amount = XMLUtils.parseNumber(el.getAttribute("amount"), Integer.class, 1);
+    int amount;
+    String amountString = el.getAttributeValue("amount");
+    if (amountString != null && amountString.equals("oo")) {
+      // blocks with -1 items cannot be dropped, stacked or dragged/split
+      amount = -1;
+    } else {
+      amount = XMLUtils.parseNumber(el.getAttribute("amount"), Integer.class, 1);
+    }
```
```suggestion
    int amount = XMLUtils.parseNumber(el.getAttribute("amount"), Integer.class, true, 1);
    if (amount = Integer.MAX_VALUE) amount = -1;
```

core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
```diff
@@ -172,4 +173,15 @@ public void checkItemTransfer(ItemTransferEvent event) {
...
   }
+
+  @EventHandler(priority = EventPriority.LOW, ignoreCancelled = true)
+  public void checkInfiniteBlocks(BlockPlaceEvent event) {
+    if (ItemTags.INFINITE.get(event.getPlayer().getItemInHand()) != null) {
+      ItemStack blockPlacedItem = new ItemStack(event.getBlockPlaced().getType());
+      // infinite block contains -1 items, giving -1 items sets the amount back to -1
+      blockPlacedItem.setAmount(-1);
+      ItemTags.INFINITE.set(blockPlacedItem, true);
+      event.getPlayer().setItemInHand(blockPlacedItem);
```
Pablete1234: This is not properly retaining block metadata. If the material is wool, you will be losing the color of said wool.
You should not be creating new itemstacks, instead, update the existing stack in the player inventory by setting the amount back to what it was.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/kits/KitParser.java
CoWinkKeyDinkInc: I don't understand this solution, parseNumber will return its own XMLDefinition error when it tries to parse `oo`, which is why I've made the separate String variable to parse it.

Pablete1234: `parseNumber` has multiple overloads (same name, different input parameters), if you use the one with an `infinite` parameter and pass in a `true`, it supports parsing `oo` as infinite, and returns Integer.MAX_VALUE for infinite integer (for doubles/floats there's a specific representation of infinity)

The parsing of such is treated here:
https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java#L250


With the suggested alternative, you'd be calling:
https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java#L310
which, will return the default 1 if there is no node,  and otherwise calls:
https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java#L266
which handles the format exception if any, and relies on to:
https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java#L245
which can treat the `oo` given that `infinity` is set to true

core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
```diff
@@ -172,4 +173,12 @@ public void checkItemTransfer(ItemTransferEvent event) {
       event.setCancelled(true);
     }
   }
+
+  @EventHandler(priority = EventPriority.LOW, ignoreCancelled = true)
+  public void checkInfiniteBlocks(BlockPlaceEvent event) {
+    if (ItemTags.INFINITE.get(event.getPlayer().getItemInHand()) != null) {
+      // infinite block contains -1 items, giving -1 items sets the amount back to -1
+      event.getPlayer().getItemInHand().setAmount(-1);
+    }
```
Electroid:
```suggestion
   final ItemStack item = event.getPlayer().getItemInHand();
   if (ItemTags.INFINITE.has(item)) {
      // infinite block contains -1 items, giving -1 items sets the amount back to -1
      item.setAmount(-1);
    }
```

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -422,7 +429,12 @@ public ItemStack parseItem(Element el, Material type, short damage) throws Inval
       throw new InvalidXMLException("Invalid item/block", el);
     }
 
+    if (amount == -1 && !itemStack.getType().isBlock()) {
+      throw new InvalidXMLException("infinity can only be applied to a block material", el);
```
Electroid, line 429: Why can't we support infinite snowballs or enderpearls?

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/kits/KitParser.java
CoWinkKeyDinkInc: I haven't tested them at the time, I'll see if works.

Pablete1234 (approved):

PR #932, opened 2021-10-28
Add dynamic kit rules

Pugzy:
## Dynamic kits 🔮👕🎉  

The PR implements dynamic kits in the same fashion as Overcast Network Project Ares did.

Let's you `give`, `lend`, and `take` a kit based on a filter rising or falling.

```xml
<kits>
    <!-- Kit and Filter References -->
    <lend kit="jumper-kit" filter="carrying-gold" />
    <give kit="regen-kit" filter="holding-redstone" />
    <!-- Inline kit and filters -->
    <take>
        <kit>
            <fly/>
        </kit>
        <filter>
            <flag-carried>flag</flag-carried>
        </filter>
    </take>
    ...
</kits>
```

Not to be confused with the documented but 'deprecated' docs syntax http://docs.oc.tc/modules/kits confirmed as incorrect by Mitch (being the main user of the feature) and Pablo (checking the original commit history and PR to change).

Only a few maps used this on OCN (three total I believe) as it was a relatively new feature so if not implemented like for like the impact would be a couple of Arcade maps 🤷 that haven't been played in years.
 
Open to opinions on a better XML syntax if anyone has any ideas. I had wanted to use an `<apply>` to apply kit rules in the same way a filter is applied to a region but the above seemed like the simplest approach.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
@@ -4,6 +4,7 @@ import static net.kyori.adventure.text.Component.text;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
+import java.util.Set;
 import org.bukkit.Material;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
@@ -21,9 +22,11 @@ import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.module.exception.ModuleLoadException;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.events.ListenerScope;
 import tc.oc.pgm.events.PlayerResetEvent;
+import tc.oc.pgm.filters.dynamic.FilterMatchModule;
 import tc.oc.pgm.kits.tag.Grenade;
 import tc.oc.pgm.kits.tag.ItemTags;
 import tc.oc.pgm.util.event.ItemTransferEvent;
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/kits/KitRule.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public Kit getKit() {
+    return kit;
+  }
+
+  public Filter getFilter() {
+    return filter;
+  }
+
+  public void load(FilterMatchModule fmm) {
```
Electroid: I'm not sure this should exist in a `FeatureDefinition`. It's a detail of the match module.

core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
```diff
@@ -32,10 +35,19 @@
...
+  public KitMatchModule(Match match, Set<KitRule> kitRules) {
     this.match = match;
+    this.kitRules = kitRules;
+  }
+
+  @Override
+  public void load() throws ModuleLoadException {
+    FilterMatchModule fmm = match.needModule(FilterMatchModule.class);
+
+    kitRules.forEach(kitRule -> kitRule.load(fmm));
```
Electroid: For loop.

Electroid (changes requested):

Electroid (approved):

mitchts:
One difference I have noticed between this version of dynamic kits and the previous implementation is that you can't trigger a kit to apply immediately as the match starts. 

Take this filter from Build: Parkour which used to apply a kit immediately as the match began via `<give .../>`

```xml
<all id="build-spawn-kit">
    <not>
        <time>3s</time>
    </not>
    <region id="all-arenas-whole"/>
    <participating/>
    <match-running/>
</all>
```

Altering this filter to allow after 1 second applies the kit to the player as expected, however.

Pablete1234:
I think match-running filter may not exist in this pgm version

PR #933, opened 2021-10-31
Apply item mods to across modules

Pugzy:
### The Issue

Maps that use item mods can cause items (typically arrows and gapples) to be non-stackable.
If the item mod is applied to the items in the player's inventory but not on items picked up or given to them they will not stack.

### Background on Item Mods

Item mods work by checking all items to see if modifications are required, once checked all items are marked with meta (in the current case hidden lore). Items that are not checked and are received by a player are missing this extra data so the items do not match.

### The Solution(s)

There are a few elements fixed/changed to make this function as expected.

- Item mods rules applied to parsed kits inside the kill rewards and block drop modules.
- Adding an extra check for picking up arrows inside item mods (arrow pickup event doesn't exist in this version of Bukkit).
- Allowing the modification of picked up arrows by removing the arrow (faking the pickup) and giving the correct item.
- Applying that same logic to the `ToolRepairMatchModule` so items don't just disappear but look to be picked up (with the animation and world sound unlike before).

Thanks to @Pablete1234 and @PabloPintor for debugging and testing.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsModule.java
+++ b/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsModule.java
@@ -7,6 +7,7 @@ import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.logging.Logger;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.MaterialData;
@@ -44,7 +45,8 @@ public class BlockDropsModule implements MapModule {
   public static class Factory implements MapModuleFactory<BlockDropsModule> {
     @Override
     public Collection<Class<? extends MapModule>> getWeakDependencies() {
-      return ImmutableList.of(KitModule.class, RegionModule.class, FilterModule.class);
+      return ImmutableList.of(
+          KitModule.class, RegionModule.class, FilterModule.class, ItemModifyModule.class);
     }
 
     @Override
@@ -53,6 +55,8 @@ public class BlockDropsModule implements MapModule {
       List<BlockDropsRule> rules = new ArrayList<>();
       FilterParser filterParser = factory.getFilters();
       RegionParser regionParser = factory.getRegions();
+      final Optional<ItemModifyModule> itemModifier =
+          Optional.ofNullable(factory.getModule(ItemModifyModule.class));
 
       for (Element elRule :
           XMLUtils.flattenElements(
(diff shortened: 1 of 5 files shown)


core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsModule.java
```diff
@@ -53,6 +54,8 @@ public BlockDropsModule parse(MapFactory factory, Logger logger, Document doc)
       List<BlockDropsRule> rules = new ArrayList<>();
       FilterParser filterParser = factory.getFilters();
       RegionParser regionParser = factory.getRegions();
+      final Optional<ItemModifyModule> itemModifier =
+          Optional.ofNullable(factory.getModule(ItemModifyModule.class));
```
Pablete1234, line 59: You probably need a dependency for this, or depending on the map you may sometimes see this work or not

core/src/main/java/tc/oc/pgm/killreward/KillRewardModule.java
```diff
@@ -52,6 +53,8 @@ public MatchModule createMatchModule(Match match) {
     public KillRewardModule parse(MapFactory factory, Logger logger, Document doc)
         throws InvalidXMLException {
       ImmutableList.Builder<KillReward> rewards = ImmutableList.builder();
+      final Optional<ItemModifyModule> itemModifier =
+          Optional.ofNullable(factory.getModule(ItemModifyModule.class));
```
Pablete1234, line 57: same here

Pablete1234 (changes requested):

Pablete1234 (approved):

core/src/main/java/tc/oc/pgm/killreward/KillRewardModule.java
```diff
@@ -62,7 +65,9 @@ public KillRewardModule parse(MapFactory factory, Logger logger, Document doc)
               0)) {
         ImmutableList.Builder<ItemStack> items = ImmutableList.builder();
         for (Element itemEl : elKillReward.getChildren("item")) {
-          items.add(factory.getKits().parseItem(itemEl, false));
+          final ItemStack itemStack = factory.getKits().parseItem(itemEl, false);
+          itemModifier.ifPresent(imm -> imm.applyRules(itemStack));
```
Electroid, line 69: If the goal of `ItemModifyModule` is to ensure that _all_ `ItemStack`s are modified to certain rules, I would think that `parseItem` should return the already modified item? The problem here is that the caller has to remember to call `applyRules`, which would be quite easy to forget.

Pablete1234: Original pgm did it like this and i don't know why, but something tells me it's better not to.

I feel there may be use-cases for parseItem that don't require itemmodding

Electroid (approved):

PR #941, opened 2021-11-23
Check objective visibility on displaying mode change messages

CoWinkKeyDinkInc:
This PR fixes an ongoing issue that started strangely happening without any idea how. The mode change message in the chat would be repeated for each objective that changed mode, and would still appear even when the objective's `show` attribute is false. This adds a proper checker to make sure the mode message is only sent out once per mode change, and that it won't send the message if all objectives that use the mode has `show=false`. I first seen this yesterday when the message repeated 8 times on Death DeNile.

From the code this issue should've been a problem for a long time but has seemed to popup about two weeks ago :man_shrugging:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java
@@ -155,11 +155,12 @@ public class CoreMatchModule implements MatchModule, Listener {
     }
   }
 
-  @EventHandler(priority = EventPriority.MONITOR)
+  @EventHandler(priority = EventPriority.HIGHEST)
   public void onObjectiveModeSwitch(final ObjectiveModeChangeEvent event) {
     for (Core core : this.cores) {
       if (core.getModes() == null || core.getModes().contains(event.getMode())) {
         core.replaceBlocks(event.getMode().getMaterialData());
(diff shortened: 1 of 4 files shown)


Pablete1234 (changes requested):
Add a getAffectedObjectives or similar to the event, and use that on the event handler to find if any are non-hidden

Trying to iterate all objectives here is a mess

Alternatively you can add a boolean show field to the event, which the handler sets to true if any core/destroyable changed has show=true

Pablete1234 (changes requested):
https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java#L125

https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/core/CoreMatchModule.java#L155

Edit both of those event handlers, make them HIGHEST priority instead of monitor, and make it so:
`if (core.isVisible()) event.setVisible(true);`, and similarly for monuments

Make ObjectiveModeEvent's `visible` non-final, and add the setVisible setter

Pablete1234 (approved):
Looks much cleaner now, nice work

Electroid (approved):

Issue #939, opened 2021-11-21
Bring back pgm objective filters


Pablete1234:
Current pgm only has support for the `objective` filter, and it is not properly documented over at pgm.dev.

We should bring back all 3 ways of parsing this filter (they all result in reusing the GoalFilter, just with different parameters in the constructor)

What 1.12+ pgm supported was:
**Objective** - Supports `any="true"` or `team="team-id"` - We have this already, although it is not documented.
`<objective>goal-id</objective>` - Matches if the team of the player is the one that completed it
`<objective any="true">goal-id</objective>` - Matches if ANY team completed it
`<objective team="red-team">goal-id</objective>` - Matches if red team completed it

**Completed** - Does NOT support `any="true"` or `team="team-id"`
`<completed>goal-id</completed>` - This is equivalent to `<objective any="true">goal-id</objective>`

**Captured** - Supports `team="team-id"` but does NOT support `any="true"`
`<captured>goal-id</captured>` - Matches if the team of the player is the one that completed it
`<captured team="team-id">goal-id</captured>` - Matches if the specific team completed it

This seems quite verbose and is probably like it is for backwards compatibility reasons, since the any="true" was added much later, and i'm assuming before you were just assumed to use `completed` if you wanted that behavior
They all in-code end up using the same GoalFilter with the same logic, just sometimes with a "hardcoded" team (or no-team, and checks all of them)

<https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/parser/FilterDefinitionParser.java#L324>
<https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/matcher/party/GoalFilter.java>
<https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/operator/TeamFilterAdapter.java>

Footnote 1: in all places where i mentioned *team* i mean competitor, which means it works for FFA if you assume each "player" is a "team", obviously you can't specify a specific team in xml for ffa
Footnote 2: the filter never abstained in old pgm.

CoWinkKeyDinkInc:
I got the first part documented in https://github.com/PGMDev/Website/commit/8fefd129900bbff375339f50875d2bd084cadac1
So essentially, all these you described do the same thing as the objective attributes?

Pablete1234:
yeah, they are the same as objective with the specific attributes

PR #945, opened 2021-12-11
#939 Bring back PGM objective filters

chidddy:
Completes #939
tested with a locally modified [ballot box(for `completed` tag)](https://github.com/OvercastCommunity/CommunityMaps/tree/master/scorebox/standard/ballot_box) and [cherokee (for `captured` tag)](https://github.com/Xerocoles/stratus-maps/tree/master/competitive/cherokee)

while i did only test with control points, considering it's just calling the same `GoalFilter` constructor with set values it should function the same, i was unsure whether to throw an exception for attributes that `completed` and `captured` don't support but decided it realistically wasn't necessary.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
+++ b/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
@@ -432,7 +432,10 @@ public abstract class FilterParser {
         this.factory.getFeatures().createReference(new Node(el), GoalDefinition.class);
 
     Attribute attrTeam = el.getAttribute("team");
-    XMLFeatureReference<TeamFactory> team = attrTeam != null ? this.factory.getFeatures().createReference(new Node(attrTeam), TeamFactory.class) : null;
+    XMLFeatureReference<TeamFactory> team =
+        attrTeam != null
+            ? this.factory.getFeatures().createReference(new Node(attrTeam), TeamFactory.class)
+            : null;
 
     return new GoalFilter(goal, team, false);
   }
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/filters/FilterParser.java
```diff
@@ -418,6 +418,31 @@ public GoalFilter parseGoal(Element el) throws InvalidXMLException {
...
+        this.factory.getFeatures().createReference(new Node(el), GoalDefinition.class);
+
+    Attribute attrTeam = el.getAttribute("team");
+    XMLFeatureReference<TeamFactory> team;
+    if (attrTeam != null) {
+      team = this.factory.getFeatures().createReference(new Node(attrTeam), TeamFactory.class);
+
+    } else {
+      team = null;
+    }
```
KingOfSquares:
```suggestion
    XMLFeatureReference<TeamFactory> team = attrTeam != null ? this.factory.getFeatures().createReference(new Node(attrTeam), TeamFactory.class) : null;
```
Just a personal opinion on format. Good work 👍

chidddy: yea that's a much cleaner way of doing it, thanks for the feedback

chidddy:
a sad reminder that "commit suggestion" does not sign off the commit...

Pablete1234:
a sad reminder that "commit suggestion" does not enforce formatting, and that line is too long to pass so you'll need to run the formatter

KingOfSquares:
Sad GitHub moment 😔

Pablete1234 (approved):

Electroid (approved):

PR #949, opened 2022-01-03
Add method to get list of kits

dentmaged:
Used by https://github.com/bolt-rip/kiteditor

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitMatchModule.java
@@ -4,7 +4,6 @@ import static net.kyori.adventure.text.Component.text;
 
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
-import java.util.Set;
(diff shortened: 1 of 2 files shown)


KingOfSquares:
@dentmaged that project seems to be private. If you want to show usage you need to copy some code or something.

Furthermore, why are you exposing this when you can just fetch it from the `MatchFeatureContext`?

KingOfSquares:
Actually, reading through the source code I see that kits are never put into the context. I think this should be the approach instead of exposing the kits in the way this PR currently does, though.

Electroid:
I think using `MatchFeatureContext` would be the proper fix.

Pablete1234:
You cannot use `MatchFeatureContext` as kits are not stored in it, kits are not a feature to start with.

To add further, where kits ARE added is either `KitContext` for legacy maps or `FeatureDefinitionContext` inside MapFactory, but that is scoped to parse-time, and discarded afterwards

Pablete1234:
@Electroid could you re-review?

PR #950, opened 2022-01-05
Introduce Monostable filters

KingOfSquares:
Allows for time period based filters. 
This PR also makes the existing time filter dynamic.

I would like feedback on the design, old PGM used the filter definitions as `FeatureFactory`s to create the `Reactor` instances, but that is not ideal for this PGM because it does not use injection. 


KingOfSquares:
Seems to have failed since electroids repo is down

core/src/main/java/tc/oc/pgm/api/filter/ReactorBased.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Implemented by filters which can act dynamically but needs more than Bukkit {@link Event}s to
+ * properly invalidate {@link Filterable}s. A filter can theoretically both depend on listening to
+ * some events AND have a reactor, but this is rarely the case.
+ *
+ * <p>Often(But not necessarily always), in addition to using non-event factors when taking into
+ * account when to invalidate something, a filter implementing this will in wrap another dynamic
+ * filter and react based on the changed state of that filter.
+ */
+public interface ReactorBased extends FilterDefinition {
```
Pablete1234: based, based on what?

Call this ReactorFactory or something else, "Based" is a bit of a weird term to use here

core/src/main/java/tc/oc/pgm/filters/Filters.java
```diff
@@ -1,14 +1,42 @@
...
 
-/** Utility methods useful for implementing filters */
+/** Utility methods useful for implementing and working with filters */
 public class Filters {
   private Filters() {}
 
+  public static List<Filter> flattenFilter(Filter filter, List<Filter> result) {
+    result.add(filter);
+
+    // No easier way to check for child filters
```
Pablete1234: Then maybe add one?

Consider just adding a `ParentFilter` interface or similar that just extends Filter, and adds a `getChildren` method
then simply call getChildren here recrusively if the filter is of type ParentFilter, and make sure that all filters (SingleFunctionFilter, MultiFilterFunction, and QueryModifier) all implement ParentFilter too and reply to getChildren with a stream or list of their filters, essentially, the logic you've done here, but with each having their own

You can even go a step further and just have that getChildren function be the recursive one itself, that way if you just call getChildren on the topmost filter you get all the child ones without needing to recurse yourself

core/src/main/java/tc/oc/pgm/filters/MonostableFilter.java
```diff
@@ -0,0 +1,111 @@
...
+  private final Duration duration;
+  final Class<? extends Filterable<?>> scope;
+
+  // Filterables that currently pass the inner filter, mapped to the instants that they expire.
+  // They are not actually removed until the inner filter goes false.
+  final Map<Filterable<?>, Instant> endTimes = new HashMap<>();
+
+  // Has all info about end times, and such we need access to the instance when testing incoming
+  // queries
+  private Reactor reactor;
```
Pablete1234: You can not have this here.

The filter is a map-scoped element, it's like a definition, it has no mutable state, meanwhile, the reactor is a match-scoped element which mutates during the match

You must store the reactor elsewhere (likely the FilterMatchModule), not in the filter

In a situation where 2 matches on the same map were running at once, how would the reactor be reached? it simply wouldn't be possible

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -89,10 +93,22 @@ public FilterMatchModule(Match match) {
...
-  public void onMatchLoad(MatchLoadEvent event) {
+  // See MatchImpl#load
+  public void onMatchLoad() {
+
+    // First, find and create all necessary reactors
+    this.context
+        .getAll(FilterDefinition.class)
+        .forEach(
+            filter -> {
+              for (Filter filt3r : Filters.flattenFilter(filter, new ArrayList<>())) {
```
Pablete1234: Im sorry what. What is filt3r. just name it f, or child

```suggestion
              for (Filter child : Filters.flattenFilter(filter, new ArrayList<>())) {
```

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -89,10 +93,22 @@ public FilterMatchModule(Match match) {
...
-  // We have to listen to this event to guarantee that all other modules have been loaded
-  // and that there is no unresolved xml filter references.
-  @EventHandler
-  public void onMatchLoad(MatchLoadEvent event) {
+  // See MatchImpl#load
+  public void onMatchLoad() {
+
+    // First, find and create all necessary reactors
+    this.context
+        .getAll(FilterDefinition.class)
```
Pablete1234: I'm pretty sure that by getting all Filters, you're already finding the children, at the very least the ones with an id because there's no way those aren't put in the context

By going over each of the filters, and then also checking for children of the filters, you're likely creating reactors multiple times for some filters

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -105,6 +121,9 @@ public void onMatchLoad(MatchLoadEvent event) {
               }
               this.registerListenersFor(filter.getRelevantEvents());
             });
+
+    // These two have to be on separate passes, in case a reactor wants to register a child filter
+    // for dynamic listening
```
Pablete1234: these what? there's nothing following this comment

core/src/main/java/tc/oc/pgm/filters/modifier/QueryModifier.java
```diff
@@ -42,4 +45,14 @@ protected QueryResponse queryTyped(Q query) {
 
     return modifiedQuery == null ? QueryResponse.ABSTAIN : filter.query(modifiedQuery);
   }
+
+  /**
+   * ONLY VISIBLE FOR {@link Filters#flattenFilter(Filter, List)}
+   *
+   * @return the child filter
+   */
+  @ApiStatus.Internal
```
Pablete1234: what's the need to be so restrictive, its' fine if this is visible imo

core/src/main/java/tc/oc/pgm/match/MatchImpl.java
```diff
@@ -811,6 +812,11 @@ public void load() throws ModuleLoadException {
       startTickables(MatchScope.LOADED);
       addParty(observers);
 
+      // This needs to be called BEFORE the match has completed loading
+      // but AFTER every other match module is finished with loading.
+      // This is to ensure that all match modules has registered their dynamic filters.
+      this.needModule(FilterMatchModule.class).onMatchLoad();
```
Pablete1234: this can not exist here.

You must find a way to make it compatible with simply using MatchLoadEvent on lowest or something along the lines
Why can this not happen after loaded is set to true?

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/match/MatchImpl.java
KingOfSquares: The old PGM dynamic system would prevent registering dynamic filters after the match started. Except for general security/avoiding bugs I can't imagine why this would be enforced

core/src/main/java/tc/oc/pgm/filters/modifier/QueryModifier.java
KingOfSquares: The thought was to restrict access such that no-one thinks that this method is public API that returns a Filter that would respond in the same way as this modifier with a set child.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
KingOfSquares: Its referring to the code above the comment.

core/src/main/java/tc/oc/pgm/api/filter/ReactorBased.java
KingOfSquares: I guess it makes more sense in Norwegian. Based as in "basert (på) reaktor".

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
Pablete1234: The normal thing is to have the comment directly above the code it is explaining, not after it, so move it to where it belongs

core/src/main/java/tc/oc/pgm/filters/modifier/QueryModifier.java
Pablete1234: Consider just naming this `getChild` or `getChildFilter` instead, avoids the confusion of people thinking it's directly the same filter

core/src/main/java/tc/oc/pgm/match/MatchImpl.java
Pablete1234: Filters not being modifyiable after match start makes sense, but we probably want to allow them while match is still loading

Worst case scenario make "loaded" a 3-step enum so instead of false and true, you go from UNLOADED to LOADING, then match load event, and after the event you set it to LOADED

For the boolean-based "isLoaded"  check you return false for UNLOADED, and true for LOADING and LOADED (avoid breaking other things in the codebase, keeping the same behavior as now)
Then add another method that returns the actual value of the enum, and use that in the dynamic filter stuff, where you allow them being created in either UNLOADED or LOADING, but not LOADED

Before doing all this tho, make sure to check what even uses that loaded flag to start with because i assume not much will actually use it

core/src/main/java/tc/oc/pgm/filters/Filters.java
KingOfSquares: This method is not necessary because all filters can be found the feature definition context anyways

core/src/main/java/tc/oc/pgm/filters/modifier/QueryModifier.java
KingOfSquares: This method is not necessary because all filters can be found the feature definition context anyways

core/src/main/java/tc/oc/pgm/match/MatchImpl.java
KingOfSquares: Then, what about just allowing registering until `MatchScope.RUNNING`?

KingOfSquares: That makes more sense imo

KingOfSquares:
Ive drafted changes based on the feedback. I have yet to test these changes, so don't merge them yet 🙃

core/src/main/java/tc/oc/pgm/filters/MonostableFilter.java
```diff
@@ -0,0 +1,112 @@
...
+import tc.oc.pgm.filters.dynamic.Filterables;
+
+public class MonostableFilter extends SingleFilterFunction implements ReactorFactory {
+
+  private final Duration duration;
+  final Class<? extends Filterable<?>> scope;
+
+  // Filterables that currently pass the inner filter, mapped to the instants that they expire.
+  // They are not actually removed until the inner filter goes false.
+  final Map<Filterable<?>, Instant> endTimes = new HashMap<>();
```
Pablete1234: Filters are map-scoped and should contain no mutable state in them

Checking the original implementation (https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/matcher/match/MonostableFilter.java) all state was properly contained within the reactor, which is a match-scoped

core/src/main/java/tc/oc/pgm/filters/MonostableFilter.java
```diff
@@ -0,0 +1,112 @@
...
+        Filterables.extractFilterable(matchQuery).getFilterableAncestor(this.scope);
+    if (filterable == null)
+      throw new IllegalArgumentException(
+          "The scope of this filter does not match the query it received");
+
+    final Reactor reactor =
+        matchQuery
+            .getMatch()
+            .needModule(FilterMatchModule.class)
+            .getReactorInstance(MonostableFilter.Reactor.class);
```
Pablete1234: I'm confused by this

Is there ONE reactor for all monostable filters, or one reactor for EACH monostable filter?

If there is one global reactor for all monostable filters, then this makes sense, but "createReactor" documentation is wrong, as it states that createReactor is called for all filters, not just for "one random monostable filter that happens to be the first".

I think what makes most sense is for each monostable to have its own reactor, which stores state for that filter in that match, and instead of `getReactorInstance(Class)` have a proper `getReactorInstance(ReactorFactory)` and pass `this` as parameter

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -66,13 +67,17 @@
 public class FilterMatchModule implements MatchModule, FilterDispatcher, Tickable, Listener {
 
   private final Match match;
+  private final FeatureDefinitionContext context;
   private final List<Class<? extends Event>> listeningFor = new LinkedList<>();
-  private final PriorityQueue<TimeFilter> timeFilterQueue = new PriorityQueue<>();
+
+  private final Map<Class<? extends ReactorFactory.Reactor>, ReactorFactory.Reactor>
+      activeReactors = new HashMap<>();
```
Pablete1234:
```suggestion
  private final Map<ReactorFactory, ReactorFactory.Reactor>activeReactors = new HashMap<>();
```
Each reactor factory has its reactor

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -89,10 +94,26 @@ public FilterMatchModule(Match match) {
...
+
+    // First, find and create all necessary reactors
+    this.context
+        .getAll(FilterDefinition.class)
+        .forEach(
+            filter -> {
+              if (filter instanceof ReactorFactory) {
+                final ReactorFactory.Reactor reactor =
+                    ((ReactorFactory) filter).createReactor(this.match, this);
+                this.activeReactors.put(reactor.getClass(), reactor);
```
Pablete1234:
```suggestion
                this.activeReactors.put((ReactorFactory) filter, reactor);
```

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -497,5 +520,9 @@ public void onFlagStateChange(FlagStateChangeEvent event) {
     this.invalidate(match);
   }
 
+  public <T extends ReactorFactory.Reactor> T getReactorInstance(Class<T> reactorType) {
+    return (T) this.activeReactors.get(reactorType);
+  }
```
Pablete1234:
```suggestion
  public <T extends ReactorFactory.Reactor> T getReactor(ReactorFactory factory) {
    return (T) this.activeReactors.get(factory);
  }
```

Feel free to make ReactorFactory a generic interface, so you can use the generic parameter there

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/filters/MonostableFilter.java
KingOfSquares: Yes, this is something I forgot to move over...

core/src/main/java/tc/oc/pgm/filters/MonostableFilter.java
KingOfSquares: It seems like my quick draft didn't make sense after all.. :cry:

Ill make sure to actually scope them to the filter and not the reactor type..

KingOfSquares:
Currently waiting for #964 
EDIT: Not anymore! Ready for review

core/src/main/java/tc/oc/pgm/api/filter/ReactorFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    private final FilterMatchModule fmm;
+    protected final Match match;
+
+    public Reactor(Match match, FilterMatchModule fmm) {
+      this.match = match;
+      this.fmm = fmm;
+    }
+
+    /** Called at the end of each match for reactors that need cleanup */
+    public void finalizeReactor() {}
```
Pablete1234:
```suggestion
    public void unload() {}
```
this is already a reactor, saying "finalizeReactor" is redundant, it'd be just "finalize", which... means something else so just use unload which is more common around the pgm codebase

core/src/main/java/tc/oc/pgm/filters/ParentFilter.java
```diff
@@ -0,0 +1,11 @@
+package tc.oc.pgm.filters;
+
+import java.util.List;
+import tc.oc.pgm.api.filter.Filter;
+
+/** A filter that has some children filters for some reason. */
```
Pablete1234:
```suggestion
/** A filter that has children filters for any reason. */
```

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -89,10 +104,29 @@ private static class ListenerSet {
...
+    // First, find and create all necessary reactors
+    this.filterContext
+        .getAllUnchecked(ReactorFactory.class)
+        .forEach(
+            filter -> {
+              final ReactorFactory.Reactor reactor = filter.createReactor(this.match, this);
+              this.activeReactors.put(
+                  (ReactorFactory<? extends ReactorFactory.Reactor>) filter, reactor);
+              System.out.println(
+                  "_________________Activated reactor " + filter + "______________________");
```
Pablete1234: ...

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -89,10 +104,29 @@ private static class ListenerSet {
...
+            });
+    this.filterContext // Check for any factories hidden in filter trees
+        .getAllUnchecked(ParentFilter.class)
+        .forEach(this::findAndCreateReactorFactories);
```
Pablete1234: Can't this all be replaced with simply doing findAndCreateReactorFactories with all the Filter.class instances, instead of doing ReactorFactory first and ParentFilter second?
Also, do we know for certain that all filters are registered to the context (regardless of if they have an id or not)?

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -107,9 +141,21 @@ public void onMatchLoad(MatchLoadEvent event) {
...
   }
 
+  private void findAndCreateReactorFactories(Filter filter) {
+    if (filter instanceof ParentFilter) {
+      ((ParentFilter) filter).getChildren().forEach(this::findAndCreateReactorFactories);
+    }
+    if (filter instanceof ReactorFactory) {
+      final ReactorFactory.Reactor reactor =
+          ((ReactorFactory<?>) filter).createReactor(this.match, this);
+      this.activeReactors.put((ReactorFactory<? extends ReactorFactory.Reactor>) filter, reactor);
```
Pablete1234: You should check that a reactor doesn't already exist before creating another. Currently you could be potentially creating it twice if someone did this xml:

```
<filters>
  <time id="after-5min">5m</time>
  <all>
    <filter id="after-5min"/>
  </all>
</filters>
```
First after-5min would get registered for being ReactorFactory, then <all> would get registered for being a ParentFilter, creating the reactor once again.

Also, are references resolved? if a filter references another it probably should be the referenced one which is checked for instance of reactor factory.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -107,9 +141,21 @@ public void onMatchLoad(MatchLoadEvent event) {
...
+    }
+  }
+
   @Override
   public void unload() {
     HandlerList.unregisterAll(this.dummyListener);
+    this.activeReactors.values().forEach(ReactorFactory.Reactor::finalizeReactor);
```
Pablete1234:
```suggestion
    this.activeReactors.values().forEach(ReactorFactory.Reactor::unload);
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
KingOfSquares: 🙃

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
KingOfSquares: Yes, all filters parsed by the filter parsers are added to a context in all cases. If no id is used it just means that you have to iterate over everything like i'm doing here to discover it.

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
KingOfSquares: References should all be resolved at this point, so ill add a case for it. Good catch

KingOfSquares:
I've pushed the response to code review for initial review again, but the code is not yet tested on a server

KingOfSquares:
Ok code has been tested an the changes seem to work as expected

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -107,9 +130,26 @@ public void onMatchLoad(MatchLoadEvent event) {
...
+    }
+    if (filter instanceof XMLFilterReference) {
+      this.findAndCreateReactorFactories(((XMLFilterReference) filter).get());
+    }
+    if (filter instanceof ReactorFactory) {
+      ReactorFactory<? extends ReactorFactory.Reactor> factory =
+          (ReactorFactory<? extends ReactorFactory.Reactor>) filter;
+      if (this.activeReactors.containsKey(factory)) return; // Prevent double registering
+      final ReactorFactory.Reactor reactor = factory.createReactor(this.match, this);
+      this.activeReactors.put(factory, reactor);
```
Pablete1234:
```suggestion
      activeReactors.computeIfAbsent(factory, f -> f.createReactor(match, this));
```

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -107,9 +130,26 @@ public void onMatchLoad(MatchLoadEvent event) {
             });
   }
 
+  private void findAndCreateReactorFactories(Filter filter) {
+    if (filter instanceof ParentFilter) {
+      ((ParentFilter) filter).getChildren().forEach(this::findAndCreateReactorFactories);
+    }
+    if (filter instanceof XMLFilterReference) {
+      this.findAndCreateReactorFactories(((XMLFilterReference) filter).get());
+    }
```
Pablete1234:
```suggestion
    if (filter instanceof XMLFilterReference) {
      this.findAndCreateReactorFactories(((XMLFilterReference) filter).get());
      return;
    }
```
i'd move this up above parent filter, if it's an xml reference just resolve it and return, no need to check for parent filter or reactor factory

Pablete1234 (commented):
just a couple small changes, otherwise lgtm

Pablete1234 (dismissed):

core/src/main/java/tc/oc/pgm/filters/dynamic/FilterMatchModule.java
```diff
@@ -66,13 +71,24 @@
 public class FilterMatchModule implements MatchModule, FilterDispatcher, Tickable, Listener {
 
   private final Match match;
+  private final ContextStore<?> filterContext;
   private final List<Class<? extends Event>> listeningFor = new LinkedList<>();
-  private final PriorityQueue<TimeFilter> timeFilterQueue = new PriorityQueue<>();
+
+  private final Map<ReactorFactory<? extends ReactorFactory.Reactor>, ReactorFactory.Reactor>
```
Pablete1234:
```suggestion
  private final Map<ReactorFactory<?>, ReactorFactory.Reactor>
```
Reactor factory's generic type is already bounded, no need to set a bound here

core/src/main/java/tc/oc/pgm/filters/dynamic/Filterables.java
```diff
@@ -28,4 +31,12 @@ public static Class<? extends Filterable<?>> scope(Filter filter) {
             + filter.getQueryType().getSimpleName()
             + " does not have a filterable scope");
   }
+
+  /** Extract the most specific filterable possible from the given query */
+  public static Filterable<?> extractFilterable(MatchQuery query) {
```
Pablete1234: this method could probably be in MatchQuery interface, and the Filterables.scope could probably just be in the Filter interface

Pablete1234 (approved):

Electroid (approved):

PR #951, opened 2022-01-06
Add bowDamageTaken

calcastor:
This PR distinguishes between damage taken and damage taken by bow, similarly to how damage dealt and bow damage dealt are separated.

The first commit in this PR is from the version of PGM deployed on Stratus (originally https://github.com/ShinyDialga/PGM/tree/stratus, currently https://github.com/calcastor/PGM/tree/stratus). I expanded on the change to add it to PGM's verbose stats menu.

I spoke with @dentmaged about this a few months back and he was pretty sure a change like this would incur API breakage, so that is something to note. I am also unaware of any potential reasoning for this being left out intentionally, beyond speculation.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -107,7 +107,7 @@ public class StatsMatchModule implements MatchModule, Listener {
             + absorptionHearts;
 
     if (damager != null) getPlayerStat(damager).onDamage(realFinalDamage, bow);
-    getPlayerStat(damaged).onDamaged(realFinalDamage);
+    if (damaged != null) getPlayerStat(damaged).onDamaged(realFinalDamage, bow);
   }
 
   @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -107,7 +107,7 @@ public class StatsMatchModule implements MatchModule, Listener {
             + absorptionHearts;
 
     if (damager != null) getPlayerStat(damager).onDamage(realFinalDamage, bow);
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -106,7 +106,7 @@ public void onDamage(EntityDamageByEntityEvent event) {
             + absorptionHearts;
 
     if (damager != null) getPlayerStat(damager).onDamage(realFinalDamage, bow);
-    getPlayerStat(damaged).onDamaged(realFinalDamage);
+    if (damaged != null) getPlayerStat(damaged).onDamaged(realFinalDamage, bow);
```
KingOfSquares: Why this if condition? The condition is checked on line 99 already, so this will always be false.
(Guessing because of old commit)

util/src/main/i18n/templates/match.properties
```diff
@@ -211,4 +212,3 @@ admin.start.unknownState = Match could not be started at this time.
 admin.end.unknownError = Match could not be ended at this time.
 
 admin.setPool.activeCycle = You may not adjust the pool while match is cycling.
-
```
KingOfSquares: Why remove this newline?

KingOfSquares (changes requested):

util/src/main/i18n/templates/match.properties
calcastor: I think this might have been a result of an IDE or text editor setting, I've readded it since that seems preferred.

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
calcastor: This was part of the original commit yes, I've removed it seeing as it's unnecessary.

Electroid:
Is this actually a breaking change? It doesn't appear so?

calcastor:
> Is this actually a breaking change? It doesn't appear so?

I spoke with @dentmaged again and he wasn't sure why he thought so originally. I'm not sure if that's still the case.

Pablete1234 (approved):

Electroid (approved):

PR #952, opened 2022-01-07
Cache locales to avoid allocations

Pablete1234:
`parseLocale` creates a new locale, and that internally calls intern for al lthe strings. it takes a significant hit in translating stuff and is the majority of the cpu time taken for scoreboard rendering.

This is a profile where you can see the overall time to render the scoreboard, and highlighted in red how long parsing locales takes, most of it being String#intern called on the locale allocation. The 2 columns correspond to `translateLegacy` and `renderGoal` calls respectively

img


Electroid (dismissed):

Electroid:
@Pablete1234 LGTM. Mind fixing the formatting?

Pablete1234:
had ran the formatter on core instead of in the root PGM, and this was in util :sweat_smile: , should be good to go now

Electroid (approved):

PR #964, opened 2022-02-03
Refactor FeatureDefinitionContext to extend ContextStore

KingOfSquares:
In preparation for finalizing #950 I decided to split out this part into its own PR. 

The idea behind the change is to be able to depend on the general `ContextStore` in places where you need some `FeatureDefinition` that may be in its own context on legacy maps(proto 1.3), e.g `FilterContext` or in the general `FeatureDefinitionContext` on newer maps(1.4 proto). Then the correct context can be passed without having to instance check for different types of contexts or other hackery.

I also decided to update naming in the `ContextStore`, as it is now used in newer protos where `id` is the proper term.(in contrast to `name`)

Also other small changes...

Note: The `Map` used to store feature definitions is now a `TreeMap`(Same as `ContextStore` has been using until now), which technically could reduce performance but its so small that I think its negligible.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/features/FeatureDefinitionContext.java
+++ b/core/src/main/java/tc/oc/pgm/features/FeatureDefinitionContext.java
@@ -1,21 +1,26 @@
 package tc.oc.pgm.features;
 
 import com.google.common.collect.HashMultimap;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.SetMultimap;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nullable;
 import org.jdom2.Element;
 import tc.oc.pgm.api.feature.FeatureDefinition;
 import tc.oc.pgm.api.feature.FeatureReference;
 import tc.oc.pgm.api.feature.FeatureValidation;
+import tc.oc.pgm.util.collection.ContextStore;
 import tc.oc.pgm.util.xml.InvalidXMLException;
 import tc.oc.pgm.util.xml.Node;
 
-public class FeatureDefinitionContext {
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/controlpoint/ControlPointModule.java
```diff
@@ -82,7 +82,7 @@ public ControlPointModule parse(MapFactory factory, Logger logger, Document doc)
           XMLUtils.flattenElements(doc.getRootElement(), "control-points", "control-point")) {
         ControlPointDefinition definition =
             ControlPointParser.parseControlPoint(factory, elControlPoint, false, serialNumber);
-        factory.getFeatures().addFeature(elControlPoint, definition);
+        factory.getFeatures().add(elControlPoint, definition);
```
Pugzy: Is there a reason for changing this to `add()` from `addFeature()`? 

Seems to just make the file changes include lots of one-liner tweaks.

KingOfSquares: To conform to `ContextStore` method names

Pablete1234: keep `addFeature`, add a new method `add` that calls addFeature, or make addFeature use add

Electroid: Let's go with Pab's suggestion so we don't accidentally make breaking changes.

Electroid: We can remove it in a future release.

KingOfSquares: Good points! PGM is also an API after all, I'll change it as Pablo suggested

KingOfSquares:
In an attempt to move this faster the PR has been slimmed down drastically to only include necessary changes for the code to work

Pablete1234 (approved):

Electroid (approved):
Looks good.

PR #966, opened 2022-02-04
Prevent double queueing of setnext maps in shuffled rotations

alexsosnovsky:
There has been a long-standing bug where if you setnext a map during a shuffled rotation (RandomMapPool), the setnexted map plays twice. This was because the shuffled pool added the map to its queue, on top of the map already being handled by the overriderMap variable in MapPoolManager. This pull request removes the queue addition, so that the setnexted map simply gets handled normally by the overriding map setup.


CoWinkKeyDinkInc:
You forgot to sign off your commit

alexsosnovsky:
> You forgot to sign off your commit

Yeah I'll take care of that sometime tomorrow.

PR #968, opened 2022-02-18
Expose more kit information

Pablete1234:
Exposes if the kit is being forced or not on the kit apply event, as well as allowing forced flag to be changed

Exposes what a clear kit is actually clearing


KingOfSquares:
The `setForce` addition feels like something that could lead to unexpected behaviour if any plugin could edit it without server owners knowing. Thoughts? Maybe it could be logged per-match at low level or something?

Pablete1234:
I don't think a plugin being able to change the `force` flag is *unexpected* behavior, i think it adds flexibility although i generally do not think it should be used

Electroid (approved):

PR #982, opened 2022-04-01
Paper 1.18

cswhite2000:
I was looking into cross version support for PGM https://github.com/PGMDev/PGM/issues/852 and it seemed to me like a first step towards that direction would be making a version of PGM that runs on 1.18. This has mostly done that.

As it is right now, the changes on my branch runs PGM on paper 1.18.2, although with some features missing.

I so far haven't implemented anything using NMS or Packets, so most of the things that rely on that are not yet implemented for 1.18.2.

The following have not been implemented:

- The tab list, besides the header and the footer
- Scoreboard team packets - these are important
- 1.7 flag beams
- TNTRenderMatchModule
- Player freezing - no fake entities without packets or nms
- Player death animation
- Portals on (some or most) maps since they rely on sending a teleport relative packet

I haven't found anything else that doesn't work.

Much of the changes are because of the deprecation / disabling of MaterialData. So all usages had to be switched to either using Material or BlockData.

Commented out code are things I haven't changed to work for 1.18.

I did add a dependency on [XSeries](https://github.com/CryptoMorin/XSeries) since it helps a lot with parsing old items / blocks from XMLs.

This draft pull request is to show progress with 1.18 version support and to provide a reference for what changes would need to be made to make a cross compatible version of PGM. This should not be merged as is.

Signed-off-by: cswhite2000 <18whitechristop@gmail.com>


core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
```diff
@@ -52,7 +51,7 @@ public static ControlPointDefinition parseControlPoint(
     Filter visualMaterials;
     List<Filter> filters = filterParser.parseFiltersProperty(elControlPoint, "visual-materials");
     if (filters.isEmpty()) {
-      visualMaterials = VISUAL_MATERIALS;
+      visualMaterials = new ColorableMaterialFilter();
```
KingOfSquares, line 54: Should probably be instanced

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -155,15 +139,15 @@ public void onEnable() {
...
+      mapLibrary.loadNewMaps(false).get(45, TimeUnit.SECONDS);
     } catch (ExecutionException | InterruptedException | TimeoutException e) {
       e.printStackTrace();
     }
 
     if (!mapLibrary.getMaps().hasNext()) {
       PGMConfig.registerRemoteMapSource(mapSourceFactories, PGMConfig.DEFAULT_REMOTE_REPO);
       try {
-        mapLibrary.loadNewMaps(false).get(30, TimeUnit.SECONDS);
+        mapLibrary.loadNewMaps(false).get(45, TimeUnit.SECONDS);
```
KingOfSquares, line 150: Why the increased timeout limit here?

Electroid:
Thanks for taking this on @cswhite2000! If PGM can run on both 1.8 and 1.18, so many communities would benefit.

Though, I guess that's the million dollar question, right? Is it feasible to support both? Based on what you've done so far -- which is amazing! -- what do you think?

cswhite2000:
I think it could be done.

Essentially we'd have another repo with a general api for doing everything that is different between versions and then have implementations for each version, that are loaded dynamically based off of which version the server is running. The adapters for non 1.8 versions would not use anything craftbukkit or nms realted, at least not anytime soon. 

Then we'd have pgm change all the version specific calls to use the wrappers from the compatability api.

We could also just make it a maven module instead of a seperate repo. Not sure which would be better.

Electroid:
I think using a module would be fine. Separate repositories makes it difficult to make changes.

benrobson:
Hey CS,
More than happy to use my current version Network as a test space for this, we're currently using TGM.

KingOfSquares (commented):
1.18 hype

core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
cswhite2000: Yeah, it probably should be.

core/src/main/java/tc/oc/pgm/PGMPlugin.java
cswhite2000: Parsing legacy materials in 1.18 is not an efficient process which makes the map loading phase take longer. 30 seconds was too short for the 1000 map repo I was testing it against.

Issue #498, opened 2020-05-22
Separate license of API and implementation


Electroid:
Similar with the website, the soon-to-be-released API is new and can be put under a non-restrictive license to allow developers to use it freely. This fixes concerns with "dynamic linking" where other plugins (say a moderation plugin) access some high-level PGM concepts, but should not need to be limited to the APGL license.

PR #983, opened 2022-04-06
Refactor api out of core

cswhite2000:
This is refactoring api out of core for [feat/1.18](https://github.com/PGMDev/PGM/tree/feat/1.18).

This does have util depending on api. I'm not sure if that is a problem.

The next step is taking usages of the bukkit api (at least the things different between 1.8 and 1.8) out of api. This will be handled in a future pr.

As a note this pr is for merging into [feat/1.18](https://github.com/PGMDev/PGM/tree/feat/1.18) not dev.

Signed-off-by: cswhite2000 <18whitechristop@gmail.com>


api/src/main/java/tc/oc/pgm/api/StringUtils.java
```diff
@@ -1,4 +1,4 @@
-package tc.oc.pgm.util;
+package tc.oc.pgm.api;
```
cswhite2000, line 2: StringUtils shouldn't be in api, but its used in SettingsValue.

api/src/main/java/tc/oc/pgm/api/xml/Node.java
```diff
@@ -1,4 +1,4 @@
-package tc.oc.pgm.util.xml;
+package tc.oc.pgm.api.xml;
```
cswhite2000, line 2: I don't really think that the `tc.oc.pgm.api.xml` implementations belong in API, but a fair number of the API methods use them as parameters.

api/src/main/java/tc/oc/pgm/api/blockdrops/BlockDrops.java
```diff
@@ -1,4 +1,4 @@
-package tc.oc.pgm.blockdrops;
+package tc.oc.pgm.api.blockdrops;
```
cswhite2000, line 2: I don't think this should be in API, but its used elsewhere.

cswhite2000 (commented):
I've commented on a few things I thought might benefit from a second look.

api/src/main/java/tc/oc/pgm/api/StringUtils.java
KingOfSquares: If we just move the static method in SettingsValue(`#search`) it should be fine? The static method in itself also looks like a method that should be in util.

KingOfSquares: https://github.com/PGMDev/PGM/blob/148a9d39ec9085468befe9c1844f4afa45b49862/api/src/main/java/tc/oc/pgm/api/setting/SettingValue.java#L97-L109 for reference

api/src/main/java/tc/oc/pgm/api/xml/Node.java
KingOfSquares: Now that the api is going to be split out it would make more sense to just separate the xml parsing stuff out to its own module.(would make it easier for development of 3rd party tools etc...) But it should probably be postponed until this whole cross-version refactor is done to not clutter things too much?

KingOfSquares:
Also, maybe #498 would be cool to have in this?

api/src/main/java/tc/oc/pgm/api/xml/Node.java
Electroid: Let's defer this to another phase, but yes, we should figure out how to remove this from `api`.

PR #524, opened 2020-06-04
Add names to posts. Add support for multiple posts for one flag.

mrcookieunderscore13:
Fix for #359 

I added a name attribute to posts. When the flag is respawning, if a name is specified, it will say which post the flag will respawn to.

Also, I added support for a flag to have multiple posts. Each post can still have multiple blocks, so all previous xmls should still work. Also, the flag element now has an attribute called sequential. If sequential="true", when the flag respawns, it will cycle through the posts; if false, it will randomly choose a post to respawn to. When a flag is respawning, it will first choose a post, and then choose one of the post's points.

Signed-off-by: mrcookie_13 batbmrcookie@gmail.com

PR #984, opened 2022-04-12
Rewrite flag posts to support more configurations

Pablete1234:
Fixes a few issues introduced in earlier changes to posts (#524 and #528), and adds a proper implementation ground for filtered posts (#759), main fixes:
 - Flag no longer needs to hold a list of posts, this is undesirable as what post is used next may change with a net. This can create some nasty side-effects (eg: a flag which is captured on a net that sets specific post, later returning to a previous post in the list instead of staying in the net-defined post).
 - Flag no longer needs to keep track of sequential posts and other convoluted logic
 - Removed the introduced direct dependency of "Flag" to "Post predeterminedPost"
 - Removed the behavior change of flags showing next spawn on pickup by default.
 - Removed introduced mutability in both Post and FlagDefinition. Those classes should & must stay immutable.
 - and a few others...
 
 Changes:
  - Flags will (again) require a *unique* post, as they always should have
  - Posts can have children posts (named), and those children posts can be filtered and traversed either sequentially or randomly
  - The old Post is now PostDefinition (the map-level feature) and Post is now a match-time feature, which holds the required state to know which post definition to use at each time.
  - A Post may rely directly on a SinglePost, or in a CompositePost (which contains several single posts inside)
  
  
Example XMLs:
Simple flag with one post
```xml
<flag ...>
    <post>x,y,z</post>
</flag>
```

Flag with 3 posts, will randomly use one of them. The first post defined must contain the flag on match load, it is the fallback post.
```xml
<flag ... >
    <post>
        <post name="Center">x,y,z</post>
        <post name="East">x,y,z</post>
        <post name="West">x,y,z</post>
    </post>
</flag>
```

Flag with 3 posts, center is the default and start position, but the flag won't respawn on it after match start due to the filter.
In this case fallback was specified, but otherwise it would just pick the first post.

If all filters disallow, the fallback WILL be used even if its filter denies.
Because of sequential = true, the flag will always bounce between east and west (center isn't used)
```xml
<flag ... >
    <post sequential="true" fallback="center-post">
        <post id="center-post" name="Center" respawn-filter="never">x,y,z</post>
        <post name="East">x,y,z</post>
        <post name="West">x,y,z</post>
    </post>
</flag>
```

The same as above, but the post is defined separately, and referenced as a flag attribute instead of embedded.
```xml
<flags>
    <post id="flag-post" sequential="true">
        <post name="Center" respawn-filter="never">x,y,z</post>
        <post name="East">x,y,z</post>
        <post name="West">x,y,z</post>
    </post>
    <flag post="flag-post"/>
</flags>
 ```

 Note: this still requires testing.
 Note 2: while this PR implements most of the functionality in #759, it does not implement the last-post filter. If it still is required it can be added in a future PR.
 
 Special thanks to @calcastor for testing the feature
 
## Edit:
Also renamed Net to NetDefinition, now all 3 flag-related-definitions are suffixed with Definition, while Flag and Post refer to the match-time objects. Also added a specific deprecation notice that points to the documentation as well as this issue.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
+++ b/core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
@@ -7,8 +7,14 @@ import tc.oc.pgm.goals.Goal;
  * have filters that respond specifically to goal queries, but currently we do not.
  */
 public class GoalQuery extends MatchQuery {
+  private final Goal<?> goal;
 
-  public GoalQuery(Goal goal) {
+  public GoalQuery(Goal<?> goal) {
     super(null, goal.getMatch());
+    this.goal = goal;
+  }
+
+  public Goal<?> getGoal() {
+    return goal;
   }
 }
--- a/core/src/main/java/tc/oc/pgm/flag/FlagParser.java
+++ b/core/src/main/java/tc/oc/pgm/flag/FlagParser.java
@@ -5,6 +5,7 @@ import com.google.common.collect.ImmutableSet;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.Nullable;
 import net.kyori.adventure.text.Component;
 import org.bukkit.DyeColor;
@@ -36,6 +37,7 @@ public class FlagParser {
   private final FilterParser filterParser;
   private final PointParser pointParser;
 
+  private final AtomicInteger postIdSerial = new AtomicInteger(1);
   private final List<PostDefinition> posts = new ArrayList<>();
   private final List<Net> nets = new ArrayList<>();
   private final List<FlagDefinition> flags = new ArrayList<>();
@@ -68,6 +70,8 @@ public class FlagParser {
 
   private CompositePost parseCompositePost(Element el) throws InvalidXMLException {
     String id = el.getAttributeValue("id");
+    if (id == null) id = PostDefinition.makeDefaultId(null, postIdSerial);
+
     boolean sequential = XMLUtils.parseBoolean(el.getAttribute("sequential"), false);
 
     ImmutableList.Builder<SinglePost> chBuilder = ImmutableList.builder();
@@ -97,6 +101,7 @@ public class FlagParser {
   private SinglePost parseSinglePost(Element el) throws InvalidXMLException {
     String id = el.getAttributeValue("id");
     @Nullable String name = el.getAttributeValue("name");
+    if (id == null) id = PostDefinition.makeDefaultId(name, postIdSerial);
 
     FeatureReference<TeamFactory> owner =
         factory.getFeatures().createReference(Node.fromAttr(el, "owner"), TeamFactory.class, null);
--- a/core/src/main/java/tc/oc/pgm/flag/FlagParser.java
+++ b/core/src/main/java/tc/oc/pgm/flag/FlagParser.java
@@ -17,12 +17,16 @@ import tc.oc.pgm.api.map.factory.MapFactory;
 import tc.oc.pgm.api.region.Region;
 import tc.oc.pgm.filters.FilterParser;
 import tc.oc.pgm.filters.StaticFilter;
+import tc.oc.pgm.flag.post.CompositePost;
+import tc.oc.pgm.flag.post.PostDefinition;
+import tc.oc.pgm.flag.post.SinglePost;
 import tc.oc.pgm.goals.ProximityMetric;
 import tc.oc.pgm.kits.Kit;
 import tc.oc.pgm.points.PointParser;
 import tc.oc.pgm.points.PointProvider;
 import tc.oc.pgm.points.PointProviderAttributes;
 import tc.oc.pgm.teams.TeamFactory;
+import tc.oc.pgm.util.xml.InheritingElement;
 import tc.oc.pgm.util.xml.InvalidXMLException;
 import tc.oc.pgm.util.xml.Node;
 import tc.oc.pgm.util.xml.XMLUtils;
@@ -32,7 +36,7 @@ public class FlagParser {
   private final FilterParser filterParser;
   private final PointParser pointParser;
 
-  private final List<Post> posts = new ArrayList<>();
+  private final List<PostDefinition> posts = new ArrayList<>();
   private final List<Net> nets = new ArrayList<>();
   private final List<FlagDefinition> flags = new ArrayList<>();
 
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

File: core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java [hunk 1]
```diff
@@ -7,8 +7,14 @@
  * have filters that respond specifically to goal queries, but currently we do not.
  */
 public class GoalQuery extends MatchQuery {
```
KingOfSquares, line 9: Now that this is used we should probably include a matching interface in `tc.oc.pgm.api.filter.query` as every(?) other query seems to have their own counterpart.

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
```diff
@@ -7,8 +7,14 @@
...
+  private final Goal<?> goal;
 
-  public GoalQuery(Goal goal) {
+  public GoalQuery(Goal<?> goal) {
```
KingOfSquares, line 12: Would not `<? extends GoalDefinition>` be "more" correct?

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
```diff
@@ -218,10 +254,9 @@ public FlagDefinition parseFlag(Element el) throws InvalidXMLException {
     Kit dropKit = factory.getKits().parseKitProperty(el, "drop-kit", null);
     Kit carryKit = factory.getKits().parseKitProperty(el, "carry-kit", null);
     boolean multiCarrier = XMLUtils.parseBoolean(el.getAttribute("shared"), false);
-    boolean sequential = XMLUtils.parseBoolean(el.getAttribute("sequential"), false);
```
KingOfSquares, line 221: I agree this change moving the logic to the post is good , but this seems to break the use of sequential on flag elements as it was introduced in #524 ([PGM v0.9](https://github.com/PGMDev/PGM/releases/tag/v0.9)) Any plans to make this new implementation backwards compatible in some way?

core/src/main/java/tc/oc/pgm/flag/post/SinglePost.java
```diff
@@ -0,0 +1,182 @@
...
+      respawnSpeed; // Makes respawn time proportional to distance, flag "moves" back at this m/s
+  private final ImmutableList<PointProvider> returnPoints; // Spawn points for the flag
+  private final boolean
+      sequential; // Search for spawn points sequentially, see equivalent field in SpawnInfo
+  private final boolean permanent; // Flag enters Completed state when at this post
+  private final double pointsPerSecond; // Points awarded while any flag is at this post
+  private final Filter pickupFilter; // Filter players who can pickup a flag at this post
+  private final Filter respawnFilter; // Filter if a flag can respawn to this post
+  private final @Nullable String
+      postName; // The post's name, shown in chat when the flag is respawning
```
KingOfSquares: Only visual: Maybe just have the comments above the variables instead of having an indent between the variable type and the variable name.

```java
//Comment
private final Variable name
```
instead of 
```java
private final Variable
    name //Comment
```

KingOfSquares (commented):
Net should probably also become NetDefinition in similarly to how Post is handled in this PR? But I guess this PR is full enough as-is

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
Pablete1234: I agree, but it is out of scope for this pr

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
Pablete1234: Goal itself is already bounded to goal definition, pretty sure that is enforced here anyways
Regardless it simply is a way not to raise a warning

core/src/main/java/tc/oc/pgm/flag/post/SinglePost.java
Pablete1234: It'd look better if it was just not having a line-jump, but sadly the formatter enforces that sooo...
i may move them all to above the var

KingOfSquares: Yea looks like formatter jank…
Moving them all sounds like a good solution

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
Pablete1234: I did consider backwards compatibility, but ended up deciding it was simply not worth the trouble (it only affects new maps which can be edited easily, legacy maps still work just fine).
If you want multiple posts you will have to update the map to put them all inside a wrapping post element, where you can define sequential, and other future behaviors for "composite posts" if we decide we want such things. Keeping this support for flag -> multiple posts will simply create us more trouble down the line.
I already talked about it with @calcastor who is the main person here handling map changes and he seems to be fine with this breaking change

KingOfSquares: Sure but mame only handles the maps for OCC(/Stratus?) and not the entire PGM ecosystem(even though its probably most of it)? It would be nice with a warning(Maybe a XML exception?, could refer to this PR in a comment or something) at least if you intend to introduce this breaking change, since the behavior would be unusual compared to the past handling of breaking changes in the XML standard in PGM.

I do agree it is not worth the trouble to properly introduce backwards compatibility in this case though.

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
KingOfSquares: Oh yeah. Even though Intellisense says it returns "capture of ?" instead of "capture of ? extends GoalDefinition" it seems to work as expected. I guess it would only be for clarity when reading the source at this point.

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
calcastor: I think some sort of XML exception referring to this PR specific could be helpful, and perhaps a specific note in a future v0.14's release notes. I'm able to adapt maps to new format on OCC and Stratus. Warzone's public maps repo has several of the same versions of maps from OCC's public maps repo so I can also PR those and/or speak with their map development staff about the changes. That covers as much ground as I think we can. Regardless I'd consider this breakage worth resolving the issues unintentionally brought forward with previous PRs, while maintaining the intended behaviours that are important for KotF as a gamemode.

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
Pablete1234: image

Goal's generic already supports only stuff that extend GoalDefinition, so it's just wouldn't be allowed to be anything else regardless

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
Pablete1234: I've implemented a specific warning message for deprecation

calcastor:
I've tested this with a map that requires some of these changes as well as updated versions of all relevant maps loaded to OCC/BOLT private servers with no issues.

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
```diff
@@ -50,29 +56,81 @@ private void checkDeprecatedFilter(Element el) throws InvalidXMLException {
     }
   }
 
-  public Post parsePost(Element el) throws InvalidXMLException {
+  private void checkDeprecatedMultiPost(Element el) throws InvalidXMLException {
+    if (el.getChildren("post").size() <= 1) return;
+    throw new InvalidXMLException(
```
Electroid, line 61: Why can we not support this anymore?

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
```diff
@@ -50,29 +56,81 @@ private void checkDeprecatedFilter(Element el) throws InvalidXMLException {
...
+    // This should never happen, since it should've parsed single post instead.
+    if (children.isEmpty())
+      throw new InvalidXMLException("Expected children posts, but found none", el);
+
+    SinglePost fallback = null;
+    Node fallbackAttr = Node.fromAttr(el, "fallback");
+    if (fallbackAttr != null) {
+      fallback = factory.getFeatures().get(fallbackAttr.getValue(), SinglePost.class);
+      if (fallback == null) {
+        throw new InvalidXMLException(
```
Electroid, line 99: Don't we have a more generic way to error when a feature definition is missing?

core/src/main/java/tc/oc/pgm/flag/post/SequentialPostResolver.java
```diff
@@ -0,0 +1,44 @@
...
+  @Override
+  public SinglePost get() {
+    if (currentIdx == -1) return fallback;
+    return posts.get(currentIdx);
+  }
+
+  @Override
+  public SinglePost peekNext(Flag flag) {
+    for (int offset = 1; offset < posts.size(); offset++) {
+      SinglePost p = posts.get((currentIdx + offset) % posts.size());
```
Electroid:
```suggestion
      final SinglePost post = posts.get((currentIdx + offset) % posts.size());
```
Style.

core/src/main/java/tc/oc/pgm/flag/post/SequentialPostResolver.java
```diff
@@ -0,0 +1,44 @@
...
+      if (p.getRespawnFilter().query(new GoalQuery(flag)).isAllowed()) return p;
+    }
+    return fallback;
+  }
+
+  @Override
+  public SinglePost getNext(Flag flag) {
+    for (int offset = 1; offset < posts.size(); offset++) {
+      SinglePost p = posts.get((currentIdx + offset) % posts.size());
```
Electroid:
```suggestion
      final SinglePost post = posts.get((currentIdx + offset) % posts.size());
```

core/src/main/java/tc/oc/pgm/flag/post/SinglePost.java
```diff
@@ -0,0 +1,187 @@
...
+import tc.oc.pgm.api.feature.FeatureReference;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.flag.Flag;
+import tc.oc.pgm.points.AngleProvider;
+import tc.oc.pgm.points.PointProvider;
+import tc.oc.pgm.points.PointProviderLocation;
+import tc.oc.pgm.teams.TeamFactory;
+
+public class SinglePost extends PostDefinition {
```
Electroid, line 20: Why not call this just `Post`?

Electroid (commented):
Overall, looks good. A few comments and questions

core/src/main/java/tc/oc/pgm/flag/post/SinglePost.java
Pablete1234: Post is the match-time Post, this is the map-time "PostDefinition" (for a single post)

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
Pablete1234: The docs have always stated that the flag should have *one* post element (or attribute), it was a mistake to allow multiple posts to be defined within a flag (in #524), and it forces things like `sequential=true"` to be defined in the flag, which is far from ideal. The flag is not sequential, the *post* is what cycles sequentially.

Old maps will still work, it affects maps using multiple posts created within the timeframe between 524 and this pr, and @calcastor who is heavily involved in map creation has already updated the syntax for the maps affected (see comments in this pr).

core/src/main/java/tc/oc/pgm/flag/FlagParser.java
Pablete1234: Unsure, this is mostly a copy from how the flag searches for an inner post a few lines lower on: https://github.com/PGMDev/PGM/pull/984/files/1a5371e3417272a84a471042970a2db90720dc4f#diff-7aa0581d0bce176dc53a9eadaca6d5a541de970854921dbe6b302cd51f542807R195

Electroid (approved):

PR #986, opened 2022-04-13
Expand visibility controls for objectives

applenick:
# Expand visibility controls for objectives

This PR was inspired by a set of changes @Pablete1234 provided from his previous work back on the old PGM. Introduced are several new attributes that can be used when defining objectives in XML. These new show flags allow map authors to precisely control where and how objectives are displayed within a match. For example, you'll now be able to hide if a wool is shown on the scoreboard, or if a core broadcasts messages or not. This will hopefully provide mapmakers with greater flexibility to customize the user experience per-map. 

## Show Flags

| attribute         | description     |  value |
|--------------|-----------|------------|
|`show-messages` | Broadcast messages related to objective | true/false|
|`show-effects`  | Play sounds, fireworks, and effects | true/false |
|`show-info`  | Display objective under commands (/match) |true/false
|`show-sidebar` | Display objective on scoreboard | true/false
|`stats`  | Whether objective counts as a win condition | true/false

The existing `show` attribute will now function as the default value for all the above flags, which is still by default `true`. 

These show flags work for the following objective types:
- Cores
- Monuments
- Wools
- Flags
- Control Points

## XML Syntax
```xml
<!-- Hide a wool from the scoreboard -->
<wool show-sidebar="false" team="blue" color="purple">
    <block location="-1,75,97"/>
</wool>

<!-- Hide core break/leak broadcast messages -->
<core show-messages="false" team="red" region="red-core"/>

<!-- Prevent control point from playing sounds / fireworks -->
<control-point show-effects="false" name="Center" capture-time="20s"/>

<!-- Hide a monument from /match and the scoreboard (combining multiple show flags) -->
<destroyable show-info="false" show-sidebar="false" owner="red" region="red-monument"/>

<!-- Define using `show` to set a default value, then overriding a single type (effects) -->
<core show="false" show-effects="true" team="red" region="red-core"/>
```

## Feedback
As per usual, these changes have been tested and should all work as intended. If you have any suggestions/requests let me know and I'd be glad to look further into it 👍 


core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
```diff
@@ -0,0 +1,45 @@
...
+    SHOW_SIDEBAR("show-sidebar"),
+    STATS("stats");
+
+    private final String name;
+
+    ShowFlag(String name) {
+      this.name = name;
+    }
+  }
+
```
KingOfSquares: Should probably be `EnumSet` instead of `Set` everywhere in this class

core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
```diff
@@ -86,7 +85,7 @@ public String toString() {
         + ", craftable="
         + this.craftable
         + ", visible="
-        + this.visible
+        + this.getShowOptions()
```
KingOfSquares, line 88: If I understand this code right this won’t show anything useful right? Should probably just forward the toString to the underlying enum set.

KingOfSquares (commented):
Looking at XML parsing in general the ShowOptions class could probably just be a Set parsed in XMLUtils? I like the simplicity of the implementation though, so if you decide to keep it as-is I have a couple of suggestions.

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
KingOfSquares: Ops I clicked the wrong lines here, I blame mobile!!!

Pablete1234: I think using the Set interface is ok, any set will work so there's no need to restrict it

However, the parse implementation *should* be initializing an enum set instead of a `new HashSet<>()`
The constructor is private so the only way to create these is thru parse anyways, so that will accomplish the same goal

core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
Pablete1234: ShowOptions just requires a proper toString method for this to work

KingOfSquares: Yes that's what I tried to say hehe

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
KingOfSquares: True! As long as the initializer is changed it should work fine

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
```diff
@@ -0,0 +1,45 @@
...
+import tc.oc.pgm.util.xml.XMLUtils;
+
+public class ShowOptions {
+
+  public enum ShowFlag {
+    SHOW_MESSAGES("show-messages"),
+    SHOW_EFFECTS("show-effects"),
+    SHOW_INFO("show-info"),
+    SHOW_SIDEBAR("show-sidebar"),
+    STATS("stats");
```
Pugzy: I don't think "stats" is a good indication of what this is actually doing.

Doesn't `required` already do what this is trying to do?

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
applenick: Excellent, will have the parse implementation initialize as an enum set 👍

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
Pablete1234: Stats was for *actual* stats back on octc, so, if it would add to the objectives in your profile and reward raindrops.

With this PR what it (should) become is a fallback for required, so if required is set that takes precedence, but if it's unset then it falls-back to using whatever the stats show-flag says.
The flag will stay regardless as part of the API so if other integrating plugins want to make use of the flag they can, even if pgm doesn't do anything with it (because it doesn't handle stats, raindrops, or anything alike)

applenick: Haha, was just writing this as a reply but you beat me to it @Pablete1234 👍 
Will get those changes out in a few hours.

core/src/main/java/tc/oc/pgm/goals/Goal.java
```diff
@@ -38,13 +38,10 @@
...
-   * Returns true if the goal acts "normally". Normal behavior is defined when the goal is visible
-   * via mediums such as a {@link net.kyori.adventure.bossbar.BossBar}, the Scoreboard, and chat. If
-   * a call to this method returns false, this goal will not show up anywhere.
-   *
-   * <p>In most cases, this should simply delegate to {@link GoalDefinition#isVisible()}
+   * Returns true if the goal has the provided show flag enabled. Show flags define the goal
+   * behavior, making it visible only via certain mediums such as the scoreboard and chat.
    */
-  boolean isVisible();
+  boolean hasShowFlag(ShowOptions.ShowFlag flag);
```
Electroid:
```suggestion
  boolean hasOption(ObjectiveOption option);
```

Let's use "option" so we don't confuse folks with actual flag objectives.

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
```diff
@@ -0,0 +1,45 @@
+package tc.oc.pgm.goals;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.jdom2.Element;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.XMLUtils;
+
+public class ShowOptions {
```
Electroid, line 9: I think this should just be a top-level `enum`.

```java
public enum ObjectiveOption {
   SHOW_MESSAGES("show-messages")
   // ...
}
```

Parsing code should exist exclusively in `XMLUtils`.

Electroid (changes requested):
Looks great. Requested a few minor changes

core/src/main/java/tc/oc/pgm/goals/ShowOptions.java
applenick: > Parsing code should exist exclusively in `XMLUtils`.

@Electroid How should the implementation be handled for moving parsing to  `XMLUtils`. Should `ObjectiveOption` be moved as well so we can reference it directly and have the method be something like:
```java
public static Set<ObjectiveOption> parseObjectiveOptions(Element el, boolean def) {
// ...
}
```
Or were you thinking of a more generic parsing method? I feel like since this is a pretty specific use case it's better off residing inside of `ObjectiveOptions`, similar to how we already have a `KitParser` or `ControlPointParser`. If you want I can extract the static parse method to a separate `ObjectiveOptionParser` class, but seems unnecessary. 

Any suggestions would be appreciated 👍

Pablete1234: > I think this should just be a top-level `enum`.

The thing is "ShowOptions" is not an enum, it's a set of show flags. and it makes sense for them to be wrapped rather than having to pass around an `ImmutableSet<ObjectiveOption>`

Also, i have one major issue with naming it "objective option" and that is that it's an over-generalization, this is specifically for "show-*" toggles, not for any random boolean you want to put on an objective, otherwise we'd shove half of the objective xml into here

Maybe we can have it be called `ShowOption` if you want to avoid confusion with actual flag objectives

Electroid (dismissed):
LGTM for now, improvement is good enough.

core/src/main/java/tc/oc/pgm/goals/ObjectiveOption.java
```diff
@@ -0,0 +1,19 @@
+package tc.oc.pgm.goals;
+
+public enum ObjectiveOption {
```
Pablete1234:
```suggestion
public enum ShowOption {
```

Pablete1234 (changes requested):
`ObjectiveOption` / `ObjectiveOptions` / `hasOption` are too generic of a name. What is "option" supposed to mean, anything you put on the xml is an "option" (the region? the material? the owning team?) but that isn't what this is about.

These are show-flags, makes sense @Electroid don't wanna call them "flags" for the confusion with the Flag objective but this needs to represent what it's about

core/src/main/java/tc/oc/pgm/goals/ObjectiveOption.java
applenick: Agreed! I went ahead and renamed `ObjectiveOption` to `ShowOption` 👍

Pablete1234 (dismissed):

Electroid (dismissed):

PR #987, opened 2022-04-14
Add GoalQuery to PGM api

KingOfSquares:

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/filter/query/GoalQuery.java
+++ b/core/src/main/java/tc/oc/pgm/api/filter/query/GoalQuery.java
@@ -3,6 +3,6 @@ package tc.oc.pgm.api.filter.query;
 import tc.oc.pgm.goals.Goal;
 import tc.oc.pgm.goals.GoalDefinition;
 
-public interface GoalQuery extends MatchQuery {
-  Goal<? extends GoalDefinition> getGoal();
+public interface GoalQuery<T extends GoalDefinition> extends MatchQuery {
+  Goal<T> getGoal();
 }
--- a/core/src/main/java/tc/oc/pgm/api/filter/query/GoalQuery.java
+++ b/core/src/main/java/tc/oc/pgm/api/filter/query/GoalQuery.java
@@ -1,0 +1,8 @@
+package tc.oc.pgm.api.filter.query;
+
+import tc.oc.pgm.goals.Goal;
+import tc.oc.pgm.goals.GoalDefinition;
+
+public interface GoalQuery extends MatchQuery {
+  Goal<? extends GoalDefinition> getGoal();
+}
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/api/filter/query/GoalQuery.java
```diff
@@ -0,0 +1,8 @@
+package tc.oc.pgm.api.filter.query;
+
+import tc.oc.pgm.goals.Goal;
+import tc.oc.pgm.goals.GoalDefinition;
+
+public interface GoalQuery extends MatchQuery {
+  Goal<? extends GoalDefinition> getGoal();
```
Pablete1234:
```suggestion
public interface GoalQuery<T extends GoalDefinition> extends MatchQuery {
  Goal<T> getGoal();
```

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
```diff
@@ -1,14 +1,18 @@
...
 import tc.oc.pgm.goals.Goal;
+import tc.oc.pgm.goals.GoalDefinition;
 
-/**
- * Currently, the only thing this class does is derive the Match from a Goal. In the future, we may
- * have filters that respond specifically to goal queries, but currently we do not.
- */
-public class GoalQuery extends MatchQuery {
+public class GoalQuery extends MatchQuery implements tc.oc.pgm.api.filter.query.GoalQuery {
+  private final Goal<? extends GoalDefinition> goal;
```
Pablete1234:
```suggestion
public class GoalQuery<T extends GoalDefinition> extends MatchQuery implements tc.oc.pgm.api.filter.query.GoalQuery<T> {
  private final Goal<T> goal;
```

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
```diff
@@ -1,14 +1,18 @@
...
 
-  public GoalQuery(Goal goal) {
+  public GoalQuery(Goal<? extends GoalDefinition> goal) {
```
Pablete1234:
```suggestion
  public GoalQuery(Goal<T> goal) {
```

core/src/main/java/tc/oc/pgm/filters/query/GoalQuery.java
```diff
@@ -1,14 +1,18 @@
...
     super(null, goal.getMatch());
+    this.goal = goal;
+  }
+
+  @Override
+  public Goal<? extends GoalDefinition> getGoal() {
```
Pablete1234:
```suggestion
  public Goal<T> getGoal() {
```

Pablete1234 (changes requested):
You should use probably be using generics as they are meant to be used, not just as wildcards everywhere

KingOfSquares:
> You should use probably be using generics as they are meant to be used, not just as wildcards everywhere

Good point.

Pablete1234 (approved):

Electroid (approved):

PR #995, opened 2022-04-22
Display shared cores & destroyables as shared

### LINKED ISSUE #967: Make Cores and Monuments appear as global objectives when shared by multiple attackers
Currently when monuments and cores are shared by multiple attackers, they will be displayed for every attacking team in the scoreboard resulting in situations like the screenshot below.
unknown

This results in a confusing messy scoreboard when attempting to make Bedwars style maps and other similar maps that use monuments or cores with more than two teams. The ideal fix for this would be to instead make monuments and cores global objectives when shared by multiple attackers, like is the case currently with flags and control points, which would result in them only appearing once per objective on the scoreboard instead of once per attacking team.

Pablete1234:
Fixes #967 

Additionally, it fixes shared objectives not showing in /match correctly (showing under each team instead), shows them in a non-arbitrary order, and prevents observers getting a % completion for destroyables with a single block.


Electroid (approved):
Nice! 🚀

PR #999, opened 2022-04-24
Implement players filter

Pablete1234:
Re-introduces the <players> filter, which matches if the inner filter matches for a certain range of players

Example:
```xml
<players min="3" max="5">
  <crouching/>
</players>
```
Will match if there are between 3 and 5 players crouching. 

The filter defaults to always, meaning you can also use the following syntax to match if there's less than 5 players on the match:
```xml
<players max="5"/>
```

Max defaults to infinite, min defaults to 1.
It also has `participants` (defaults to true) and `observers` (defaults to false) for if participating players and observing players respectively should be included.

Note: this has not yet been tested
Note for @KingOfSquares, this will probably require extra work and a reactor to work, but that's reliant on monostable filters PR. Currently a player change party event invalidates the player, but not the current match, so i assume it won't actually make this filter re-compute when a player switches team, meaning it won't function dynamically as expected.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/filter/Filter.java
+++ b/core/src/main/java/tc/oc/pgm/api/filter/Filter.java
@@ -32,6 +32,10 @@ public interface Filter {
     return ImmutableList.of();
   }
 
+  default boolean isDynamic() {
+    return !getRelevantEvents().isEmpty();
+  }
+
   /** Least-derived query type that this filter might not abstain from */
   Class<? extends Query> getQueryType();
 
--- a/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
+++ b/core/src/main/java/tc/oc/pgm/filters/FilterParser.java
@@ -616,4 +616,18 @@ public abstract class FilterParser {
   public PlayerBlockQueryModifier parsePlayerFilter(Element el) throws InvalidXMLException {
     return new PlayerBlockQueryModifier(parseChild(el));
   }
+
+  @MethodParser("players")
+  public PlayerCountFilter parsePlayerCountFilter(Element el) throws InvalidXMLException {
+    Filter child =
+        el.getChildren().isEmpty()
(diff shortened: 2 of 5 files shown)


KingOfSquares:
I agree that it makes sense to make this filter have its own reactor instead of just always invalidating the match every time the PlayerPartyChangeEvent is thrown.(Like FilterMatchModule currently does for FlagStateChangeEvent)

Maybe an idea to make this filter amount of players on a unique team? Such that teams can have stuff change for them only, like a secret path only opening if X amount of players on your team is standing near a "trigger".

Pablete1234:
> Maybe an idea to make this filter amount of players on a unique team? Such that teams can have stuff change for them only, like a secret path only opening if X amount of players on your team is standing near a "trigger".

For that, you should just use a inner filter for team:
```
<players>
    <all>
        <region id="some-region-id"/>
        <team>red-team</team>
    </all>
</players>
```

The players filter itself does not need a team

KingOfSquares:
> > Maybe an idea to make this filter amount of players on a unique team? Such that teams can have stuff change for them only, like a secret path only opening if X amount of players on your team is standing near a "trigger".
> 
> For that, you should just use a inner filter for team:
> 
> ```
> <players>
>     <team>red-team</team>
> </players>
> ```
> 
> The players filter itself does not need a team

Ah of course, that is a better solution!

Pablete1234:
Regarding the reactor, just look at this:
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/filters/matcher/match/PlayerCountFilter.java

Essentially the inner filter is registered for onChange of match players, and that adds/removes from a set of players who are currently matching.

When reactors are a thing, such method is more efficient. Currently i'm smartly short-circuiting so that it evaluates the least amount possibly necessary to guarantee the range, but just being able to react to the child filter changes on the player would be even better for this to be used dynamically

Electroid (approved):

PR #1000, opened 2022-04-24
Force location queries to check center of block

Pugzy:
This makes a `LocationQuery` use the center of a block when checking if the player is inside a region. Region checks use the PlayerCoarseMoveEvent which happens when a player enters or leaves a whole block, when combined with a  `contains` check using the player's actual coordinates results in counting them as being in a region below them when they're not.

This was noticed when shifting/standing above a scorebox/portal that shared the same region.  Take the below example:

```xml
<cuboid id="blue-scorebox" min="1825,7,-1361" max="1828,9,-1364"/>
```
If a player was stood on top of the block at the max of this region (y = 9.0 exactly) then the portal's `contains` check matches but the scoreboxes `enters` check does not. Portals (dynamic)(and other region `contains` LocationQuery checks) are currently using the players location to detect if they are in the region which triggers the portal when they're not inside the region causing them to be prematurely teleport and not receive points.

This can be seen here: https://discord.com/channels/86514356862320640/172382420954251264/962731072234328214

Whilst tinkering with the above checking through the stack of events and methods the below changes were also identified.

~~- Replace the usage of `PlayerQuery` (without location) in the `RegionMatchModule` with the player itself as the player is already a query (due to the dynamic changes). This allows the `PlayerQuery` to be changed to be a `PlayerWithLocationQuery` used only for checking locations that are not the player's current location (only used in flag drop locations atm).~~

~~- The above change also means that the `event` is passed with the `rfa` check and `query` as it is no longer contained in the `PlayerQuery` wrapper (so needs to be passed so it can be cancelled if needed).~~

- Removes a duplicate call to get the PGM player inside the `FilterMatchModule`.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/filter/query/LocationQuery.java
+++ b/core/src/main/java/tc/oc/pgm/api/filter/query/LocationQuery.java
@@ -1,7 +1,12 @@
 package tc.oc.pgm.api.filter.query;
 
 import org.bukkit.Location;
+import tc.oc.pgm.util.block.BlockVectors;
 
 public interface LocationQuery extends MatchQuery {
   Location getLocation();
+
+  default Location getBlockCenter() {
+    return BlockVectors.center(getLocation());
+  }
 }
--- a/core/src/main/java/tc/oc/pgm/api/region/Region.java
+++ b/core/src/main/java/tc/oc/pgm/api/region/Region.java
@@ -11,6 +11,7 @@ import org.bukkit.event.Event;
 import org.bukkit.util.BlockVector;
 import org.bukkit.util.Vector;
 import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.filter.query.LocationQuery;
 import tc.oc.pgm.regions.Bounds;
 
 /** Represents an arbitrary region in a Bukkit world. */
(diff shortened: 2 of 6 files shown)


core/src/main/java/tc/oc/pgm/regions/RegionMatchModule.java
```diff
@@ -128,7 +125,6 @@ public void applyEffects(final PlayerCoarseMoveEvent event) {
 
     Vector from = event.getBlockFrom().toVector();
     Vector to = event.getBlockTo().toVector();
-    Query query = new tc.oc.pgm.filters.query.PlayerQuery(event, player);
```
Pablete1234: By getting rid of this, and using just the player instead of the query, you're breaking any filter that relies on the event, (eg: cause filters)

core/src/main/java/tc/oc/pgm/regions/RegionMatchModule.java
```diff
@@ -413,30 +409,34 @@ private static Material getHangingType(Hanging hanging) {
...
    *
    * @return false if the query abstained, otherwise true
    */
-  protected static boolean processQuery(RegionFilterApplication rfa, Query query) {
+  protected static boolean processQuery(Event event, RegionFilterApplication rfa, Query query) {
     if (rfa.filter == null) {
       return false;
     }
 
     switch (rfa.filter.query(query)) {
```
Pablete1234: The query being used does not have the event, meaning, again, cause filters will not work

Pablete1234 (changes requested):
Keep the block center changes, undo the player query changes. The event is needed in the filters

Pablete1234 (approved):

Electroid (approved):

PR #1002, opened 2022-05-02
Add splash potions entity to Spawner module

CoWinkKeyDinkInc:
This PR adds falling potions (splash potions) as a droppable entity that can be configured with the XML spawner. These are automatically used when using `<effect> <effects> <potion> <potions>`. Some work needs to be done on it first, like fixing the positions that the potions spawn, flame effects, potion colour and so on.

```xml
<spawners>
    <!-- drops a slowness splash potion -->
    <spawner spawn-region="gold-apple" player-region="gold-apple" delay="5s">
        <effect duration="10" amplifier="1">slowness</effect>
    </spawner>
    <!-- drops a splash potion that contains both speed and jumpboost -->
    <spawner spawn-region="arrows" player-region="arrows" delay="5s">
        <effect duration="10" amplifier="3">speed</effect>
        <effect duration="10" amplifier="6">jump_boost</effect>
    </spawner>
</spawners>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/spawner/Spawner.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/Spawner.java
@@ -1,5 +1,6 @@
 package tc.oc.pgm.spawner;
 
+import java.util.Objects;
 import org.bukkit.Effect;
 import org.bukkit.Location;
 import org.bukkit.event.EventHandler;
@@ -36,7 +37,6 @@ public class Spawner implements Listener, Tickable {
   public Spawner(SpawnerDefinition definition, Match match) {
     this.definition = definition;
     this.match = match;
-
     this.lastTick = match.getTick().tick;
     this.players = new OnlinePlayerMapAdapter<>(PGM.get());
     calculateDelay();
@@ -51,7 +51,6 @@ public class Spawner implements Listener, Tickable {
             definition.spawnRegion.getRandom(match.getRandom()).toLocation(match.getWorld());
         spawnable.spawn(location, match);
         match.getWorld().spigot().playEffect(location, Effect.FLAME, 0, 0, 0, 0.15f, 0, 0, 40, 64);
-
         spawnedEntities = spawnedEntities + spawnable.getSpawnCount();
       }
       calculateDelay();
@@ -102,18 +101,18 @@ public class Spawner implements Listener, Tickable {
       return;
     }
 
-    int entitySpawnerID = -1;
-    int targetSpawnerID = -1;
+    String entitySpawnerId = "";
+    String targetSpawnerId = "";
     if (event.getEntity().hasMetadata(METADATA_KEY)) {
-      entitySpawnerID =
-          MetadataUtils.getMetadata(event.getEntity(), METADATA_KEY, PGM.get()).asInt();
+      entitySpawnerId =
+          MetadataUtils.getMetadata(event.getEntity(), METADATA_KEY, PGM.get()).toString();
     }
     if (event.getTarget().hasMetadata(METADATA_KEY)) {
-      targetSpawnerID =
-          MetadataUtils.getMetadata(event.getTarget(), METADATA_KEY, PGM.get()).asInt();
+      targetSpawnerId =
+          MetadataUtils.getMetadata(event.getTarget(), METADATA_KEY, PGM.get()).toString();
     }
     // Cancel the merge if the items are from different PGM spawners
-    if (entitySpawnerID != targetSpawnerID) {
+    if (!entitySpawnerId.equals(targetSpawnerId)) {
       event.setCancelled(true);
       return;
     }
@@ -121,8 +120,8 @@ public class Spawner implements Listener, Tickable {
 
   private void handleEntityRemoveEvent(Metadatable metadatable, int amount) {
     if (metadatable.hasMetadata(METADATA_KEY)) {
-      if (MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()).asInt()
-          == definition.numericID) {
+      if (Objects.equals(
+          MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()), definition.numericId)) {
         spawnedEntities -= amount;
         spawnedEntities = Math.max(0, spawnedEntities);
       }
--- a/core/src/main/java/tc/oc/pgm/spawner/Spawner.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/Spawner.java
@@ -81,47 +81,23 @@ public class Spawner implements Listener, Tickable {
 
   @EventHandler(priority = EventPriority.HIGHEST)
   public void onItemMerge(ItemMergeEvent event) {
-    boolean entityTracked = false;
-    boolean targetTracked = false;
-    if (event.getEntity().hasMetadata(METADATA_KEY)) {
-      entityTracked = true;
-    }
-    if (event.getTarget().hasMetadata(METADATA_KEY)) {
-      targetTracked = true;
-    }
-
-    // Do nothing if neither item is from a PGM Spawner
-    if (!entityTracked && !targetTracked) {
-      return;
-    }
-
-    // Cancel the merge if only 1 of the items is from a PGM Spawner
-    if ((entityTracked && !targetTracked) || (!entityTracked && targetTracked)) {
-      event.setCancelled(true);
-      return;
-    }
-
-    String entitySpawnerId = "";
-    String targetSpawnerId = "";
-    if (event.getEntity().hasMetadata(METADATA_KEY)) {
-      entitySpawnerId =
+    boolean entityTracked = event.getEntity().hasMetadata(METADATA_KEY);
+    boolean targetTracked = event.getTarget().hasMetadata(METADATA_KEY);
+    if (!entityTracked && !targetTracked) return; // None affected
+    if (entityTracked && targetTracked) {
+      String entitySpawnerId =
           MetadataUtils.getMetadata(event.getEntity(), METADATA_KEY, PGM.get()).toString();
-    }
-    if (event.getTarget().hasMetadata(METADATA_KEY)) {
-      targetSpawnerId =
+      String targetSpawnerId =
           MetadataUtils.getMetadata(event.getTarget(), METADATA_KEY, PGM.get()).toString();
+      if (entitySpawnerId.equals(targetSpawnerId)) return; // Same spawner, allow merge
     }
-    // Cancel the merge if the items are from different PGM spawners
-    if (!entitySpawnerId.equals(targetSpawnerId)) {
-      event.setCancelled(true);
-      return;
-    }
+    event.setCancelled(true);
   }
 
   private void handleEntityRemoveEvent(Metadatable metadatable, int amount) {
     if (metadatable.hasMetadata(METADATA_KEY)) {
       if (Objects.equals(
-          MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()), definition.numericId)) {
+          MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()), definition.getId())) {
         spawnedEntities -= amount;
         spawnedEntities = Math.max(0, spawnedEntities);
       }
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
@@ -10,16 +10,16 @@ import tc.oc.pgm.api.region.Region;
 @FeatureInfo(name = "spawner")
 public class SpawnerDefinition implements FeatureDefinition {
 
+  public final String numericId;
   public final Region spawnRegion;
   public final Region playerRegion;
   public final int maxEntities;
   public final Duration minDelay, maxDelay, delay;
   public final List<Spawnable> objects;
   public final Filter playerFilter;
-  public final int numericID;
 
   public SpawnerDefinition(
-      int numericID,
+      String numericId,
       List<Spawnable> objects,
       Region spawnRegion,
       Region playerRegion,
@@ -28,7 +28,7 @@ public class SpawnerDefinition implements FeatureDefinition {
       Duration minDelay,
       Duration maxDelay,
       int maxEntities) {
-    this.numericID = numericID;
+    this.numericId = numericId;
     this.spawnRegion = spawnRegion;
     this.playerRegion = playerRegion;
     this.maxEntities = maxEntities;
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
@@ -19,6 +19,7 @@ public class SpawnerDefinition implements FeatureDefinition {
   public final int numericID;
 
   public SpawnerDefinition(
+      int numericID,
       List<Spawnable> objects,
       Region spawnRegion,
       Region playerRegion,
@@ -26,8 +27,8 @@ public class SpawnerDefinition implements FeatureDefinition {
       Duration delay,
       Duration minDelay,
       Duration maxDelay,
-      int maxEntities,
-      int numericID) {
+      int maxEntities) {
+    this.numericID = numericID;
     this.spawnRegion = spawnRegion;
     this.playerRegion = playerRegion;
     this.maxEntities = maxEntities;
@@ -36,6 +37,5 @@ public class SpawnerDefinition implements FeatureDefinition {
     this.delay = delay;
     this.objects = objects;
     this.playerFilter = playerFilter;
-    this.numericID = numericID;
   }
 }
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
@@ -2,6 +2,8 @@ package tc.oc.pgm.spawner;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.annotation.Nullable;
 import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.region.Region;
@@ -35,5 +37,17 @@ public class SpawnerDefinition extends SelfIdentifyingFeatureDefinition {
     this.delay = delay;
     this.objects = objects;
     this.playerFilter = playerFilter;
+  }
+
+  @Override
+  protected String getDefaultId() {
+    return super.makeDefaultId();
+  }
+
+  public static String makeDefaultId(@Nullable String name, AtomicInteger serial) {
+    return "--"
+        + makeTypeName(SpawnerDefinition.class)
+        + "-"
+        + (name != null ? makeId(name) : serial.getAndIncrement());
   }
 }
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
@@ -2,15 +2,13 @@ package tc.oc.pgm.spawner;
 
 import java.time.Duration;
 import java.util.List;
-import tc.oc.pgm.api.feature.FeatureDefinition;
 import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.region.Region;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 
 @FeatureInfo(name = "spawner")
-public class SpawnerDefinition implements FeatureDefinition {
-
-  public final String numericId;
+public class SpawnerDefinition extends SelfIdentifyingFeatureDefinition {
   public final Region spawnRegion;
   public final Region playerRegion;
   public final int maxEntities;
@@ -19,7 +17,7 @@ public class SpawnerDefinition implements FeatureDefinition {
   public final Filter playerFilter;
 
   public SpawnerDefinition(
-      String numericId,
+      String id,
       List<Spawnable> objects,
       Region spawnRegion,
       Region playerRegion,
@@ -28,7 +26,7 @@ public class SpawnerDefinition implements FeatureDefinition {
       Duration minDelay,
       Duration maxDelay,
       int maxEntities) {
-    this.numericId = numericId;
+    super(id);
     this.spawnRegion = spawnRegion;
     this.playerRegion = playerRegion;
     this.maxEntities = maxEntities;
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
@@ -89,11 +89,14 @@ public class SpawnerModule implements MapModule {
           objects.add(item);
         }
 
+        List<PotionEffect> thrownPotion = new ArrayList<>();
+        // All listed effects for a spawner will be included in one splash potion
         for (Element spawnable :
             XMLUtils.getChildren(element, "potion", "potions", "effect", "effects")) {
           PotionEffect potionEffect = XMLUtils.parsePotionEffect(spawnable);
-          List<PotionEffect> thrownPotion = new ArrayList<>();
           thrownPotion.add(potionEffect);
+        }
+        if (!thrownPotion.isEmpty()) {
           SpawnablePotion potion = new SpawnablePotion(thrownPotion, numericID);
           objects.add(potion);
         }
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
@@ -97,10 +97,10 @@ public class SpawnerModule implements MapModule {
           Duration duration = null;
           Integer amplifier = null;
           if (spawnable.getAttribute("duration") != null) {
-            duration = XMLUtils.parseSecondDuration(Node.fromAttr(spawnable), "duration");
+            duration = XMLUtils.parseSecondDuration(Node.fromAttr(spawnable, "duration"));
           }
           if (spawnable.getAttribute("amplifier") != null) {
-            amplifier = XMLUtils.parseNumber(spawnable, Integer.class);
+            amplifier = XMLUtils.parseNumber(spawnable.getAttribute("amplifier"), Integer.class) - 1;
           }
           for (Element effectEl : XMLUtils.getChildren(spawnable, "effect")) {
             Attribute presetDuration = effectEl.getAttribute("duration");
@@ -110,16 +110,16 @@ public class SpawnerModule implements MapModule {
               thrownPotion.add(XMLUtils.parsePotionEffect(effectEl));
             } else if (duration != null) {
               // if <effect> mentions duration and does not mention amplifier
-              if (presetAmplifier == null) {
+              if (presetDuration != null && presetAmplifier == null) {
                 PotionEffect potionEffect =
                     new PotionEffect(
                         presetPotion.getType(),
-                        (int) TimeUtils.toTicks(duration),
+                        (int) TimeUtils.toTicks(XMLUtils.parseSecondDuration(Node.fromAttr(effectEl, "duration"))),
                         presetPotion.getAmplifier());
                 thrownPotion.add(potionEffect);
               }
               // if <effect> does not have duration and has amplifier
-              else if (presetDuration == null) {
+              else if (presetDuration == null && presetAmplifier != null) {
                 PotionEffect potionEffect =
                     new PotionEffect(
                         presetPotion.getType(),
@@ -128,7 +128,7 @@ public class SpawnerModule implements MapModule {
                 thrownPotion.add(potionEffect);
               }
               // if <effect> has neither attributes
-              else if (amplifier != null) {
+              else if (presetDuration == null && amplifier != null) {
                 PotionEffect potionEffect =
                     new PotionEffect(
                         presetPotion.getType(), (int) TimeUtils.toTicks(duration), amplifier);
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
@@ -99,22 +99,22 @@ public class SpawnerModule implements MapModule {
           if (potionChildren.isEmpty()) {
             throw new InvalidXMLException("Expected child effects, but found none", spawnerEl);
           }
-          int potionName = 0;
+          int damageValue = 0;
           if (potionEl.getAttribute("damage") != null) {
-            potionName = XMLUtils.parseNumber(potionEl.getAttribute("damage"), Integer.class, 0);
+            damageValue = XMLUtils.parseNumber(potionEl.getAttribute("damage"), Integer.class, 0);
           } else {
             for (PotionEffect potionEffect : potionChildren) {
               // PotionType lists "true" potions, PotionEffectType "potionEffect.getType()" lists
               // all possible status effects (ie wither, blindness, etc)
               // Use the first listed PotionType for potion color
               if (PotionType.getByEffect(potionEffect.getType()) != null) {
-                potionName = PotionType.getByEffect(potionEffect.getType()).getDamageValue();
+                damageValue = PotionType.getByEffect(potionEffect.getType()).getDamageValue();
                 break;
               }
             }
           }
           objects.add(
-              new SpawnablePotion(potionChildren, potionName, "spawner-" + numericId.get()));
+              new SpawnablePotion(potionChildren, damageValue, "spawner-" + numericId.get()));
         }
 
         SpawnerDefinition spawnerDefinition =
--- a/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
@@ -54,7 +54,7 @@ public class SpawnerModule implements MapModule {
       RegionParser regionParser = factory.getRegions();
       KitParser kitParser = factory.getKits();
       FilterParser filterParser = factory.getFilters();
-      AtomicInteger numericId = new AtomicInteger(0);
+      AtomicInteger postIdSerial = new AtomicInteger(1);
 
       for (Element spawnerEl :
           XMLUtils.flattenElements(doc.getRootElement(), "spawners", "spawner")) {
@@ -65,9 +65,7 @@ public class SpawnerModule implements MapModule {
         Attribute minDelayAttr = spawnerEl.getAttribute("min-delay");
         Attribute maxDelayAttr = spawnerEl.getAttribute("max-delay");
 
-        if (spawnerEl.getAttributeValue("id") == null) {
-          id = "spawner-" + numericId.getAndIncrement();
-        }
+        if (id == null) id = SpawnerDefinition.makeDefaultId("spawner", postIdSerial);
 
         if ((minDelayAttr != null || maxDelayAttr != null) && delayAttr != null) {
           throw new InvalidXMLException(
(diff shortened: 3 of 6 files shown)


core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -87,6 +89,18 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
           objects.add(item);
         }
 
+        List<PotionEffect> thrownPotion = new ArrayList<>();
+        // All listed effects for a spawner will be included in one splash potion
+        for (Element spawnable :
+            XMLUtils.getChildren(element, "potion", "potions", "effect", "effects")) {
+          PotionEffect potionEffect = XMLUtils.parsePotionEffect(spawnable);
+          thrownPotion.add(potionEffect);
```
TheMolkaPL:
```suggestion
          thrownPotion.add(XMLUtils.parsePotionEffect(spawnable));
```

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -87,6 +89,18 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+        }
+        if (!thrownPotion.isEmpty()) {
+          SpawnablePotion potion = new SpawnablePotion(thrownPotion, numericID);
+          objects.add(potion);
```
TheMolkaPL:
```suggestion
          objects.add(new SpawnablePotion(thrownPotion, numericID));
```

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,44 @@
...
+import org.bukkit.metadata.FixedMetadataValue;
+import org.bukkit.potion.PotionEffect;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.spawner.Spawnable;
+import tc.oc.pgm.spawner.Spawner;
+
+public class SpawnablePotion implements Spawnable {
+  private ItemStack potionItem;
+  private String METADATA_VALUE;
```
TheMolkaPL:
```suggestion
  private String metadataValue;
```

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public SpawnablePotion(List<PotionEffect> potion, int spawnerID) {
```
TheMolkaPL:
```suggestion
  public SpawnablePotion(List<PotionEffect> potion, int spawnerId) {
```

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,44 @@
...
+          new PotionEffect((effect.getType()), effect.getDuration(), effect.getAmplifier()), false);
+    }
+    potionItem.setItemMeta(potionMeta);
+    this.potionItem = potionItem;
+  }
+
+  @Override
+  public void spawn(Location location, Match match) {
+    ThrownPotion thrownPotion = location.getWorld().spawn(location, ThrownPotion.class);
+    thrownPotion.setItem(potionItem);
```
TheMolkaPL: I would clone the `potionItem` here.

TheMolkaPL (changes requested):

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
CoWinkKeyDinkInc: `METADATA_VALUE` doesn't change, and is also capitalized in SpawnableItem, perhaps we should make both the metadata and ItemStack as a final?

TheMolkaPL: Even when its `final` its should still be camelCase. Only static final things should be UPPER_CASE.

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
CoWinkKeyDinkInc: I don't see what benefits there is to making a new copy of the potion item. Could you explain?

CoWinkKeyDinkInc:
The potion effects are now nested inside `<potion>` and can inherit attributes from a higher element, which follows XML conventions seen in other modules. Thanks to @Pablete1234 for the suggestion.

```xml
<spawners>
    <spawner spawn-region="gold-apple" player-region="gold-apple" delay="5s">
        <potion duration="10">
            <effect amplifier="1">slowness</effect>
        </potion>
    </spawner>
    <spawner spawn-region="arrows" player-region="arrows" delay="5s">
        <potion duration="10" amplifier="3">
            <effect>speed</effect>
            <effect>jump_boost</effect>
        </potion>
    </spawner>
</spawners>
```

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -50,7 +54,7 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
       KitParser kitParser = factory.getKits();
       FilterParser filterParser = factory.getFilters();
 
-      int numericID = 0;
+      int numericId = 0;
```
Pablete1234: Make this an atomic integer and move it outside the local-scope of the method as a final variable
Then when wanting to use it simply use getandincrement
That avoids that, potentially, if this method was called multiple times it would re-use the same ids for different spawners

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -83,10 +87,55 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+          // not present
+          Duration duration = null;
+          Integer amplifier = null;
+          if (potionEl.getAttribute("duration") != null) {
+            duration = XMLUtils.parseSecondDuration(Node.fromAttr(potionEl, "duration"));
+          }
+          if (potionEl.getAttribute("amplifier") != null) {
+            amplifier = XMLUtils.parseNumber(potionEl.getAttribute("amplifier"), Integer.class);
+          }
+          for (Element effectEl : XMLUtils.getChildren(potionEl, "effect")) {
```
Pablete1234: There's XMLUtils to flatten children, or you can use `InheritingElement` directly (see: https://github.com/PGMDev/PGM/pull/984/files#diff-7aa0581d0bce176dc53a9eadaca6d5a541de970854921dbe6b302cd51f542807R87)
Regardless, you should not be needing to do such a complex "search" of parent-and-child re-searching all the same elements everywhere

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -83,10 +87,55 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+              }
+              // if <effect> does not have duration and has amplifier
+              else if (presetDuration == null && presetAmplifier != null) {
+                addPotionEffect(thrownPotion, presetPotion, duration, presetAmplifier);
+              }
+              // if <effect> has neither attributes
+              else if (presetDuration == null && amplifier != null) {
+                addPotionEffect(
+                    thrownPotion, presetPotion, duration, potionEl.getAttribute("amplifier"));
+              }
```
Pablete1234: Is there no XMLUtil parser for an effect? what does parsing effect kits use? maybe promote that to an XMLUtil and re-use it in both kits and here, at the end of the day, you're parsing *an effect* and should work the same.

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -83,10 +87,55 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+            } else {
+              throw new InvalidXMLException(
+                  "<effect> must have a duration assigned to it by a duration attribute or <potion>",
+                  effectEl);
+            }
```
Pablete1234: If you simply use `Node.fromRequiredAttr(effectEl, "duration")`, pgm will take care of throwing the exception for you, and you don't even need this if-hell parsing

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -97,13 +146,27 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+        Duration duration,
+        Attribute amplifier)
+        throws InvalidXMLException {
+      PotionEffect potionEffect =
+          new PotionEffect(
+              potion.getType(),
+              (int) TimeUtils.toTicks(duration),
+              XMLUtils.parseNumber(amplifier, Integer.class) - 1);
+      thrownPotion.add(potionEffect);
+    }
```
Pablete1234: this method should not exist, because it shouldn't be needed to call from 4 places, you need to remove branches in your if-else hell above so that this is just a one-time use

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnableItem.java
```diff
@@ -11,18 +11,18 @@
 
 public class SpawnableItem implements Spawnable {
 
-  private ItemStack stack;
-  private String METADATA_VALUE;
+  private final ItemStack stack;
+  private final String metadataValue;
```
Pablete1234:
```suggestion
  private final String spawnerId;
```
Just rename it to something better

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnableItem.java
```diff
@@ -11,18 +11,18 @@
...
 
-  public SpawnableItem(ItemStack stack, int spawnerID) {
+  public SpawnableItem(ItemStack stack, int spawnerId) {
```
Pablete1234:
```suggestion
  public SpawnableItem(ItemStack stack, String spawnerId) {
```
Pass in a string directly, and use the actual id of the spawner specified in xml if one is available (otherwise, use an autogenerated one like "spanwer-0", "spanwer-1" etc)

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
Pablete1234: Prevents mutability and it's generally good practice. If the potion item was to be modified (eg: the amount goes from 1 to 0) then forever here you're going to have a potion item with amount 0, and this perssists matches because it's map-scoped. The whole functionality of this class should be immutable, so ensure immutability by returning copies.

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,44 @@
...
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.spawner.Spawnable;
+import tc.oc.pgm.spawner.Spawner;
+
+public class SpawnablePotion implements Spawnable {
+  private final ItemStack potionItem;
+  private final String metadataValue;
+
+  public SpawnablePotion(List<PotionEffect> potion, int spawnerId) {
```
Pablete1234:
```suggestion
  public SpawnablePotion(List<PotionEffect> potion, String spawnerId) {
```
same as above

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,44 @@
...
+import org.bukkit.metadata.FixedMetadataValue;
+import org.bukkit.potion.PotionEffect;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.spawner.Spawnable;
+import tc.oc.pgm.spawner.Spawner;
+
+public class SpawnablePotion implements Spawnable {
+  private final ItemStack potionItem;
+  private final String metadataValue;
```
Pablete1234:
```suggestion
  private final String spawnerId;
```
same as above

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public SpawnablePotion(List<PotionEffect> potion, int spawnerId) {
+    this.metadataValue = Integer.toString(spawnerId);
+    ItemStack potionItem = new ItemStack(Material.POTION);
+    PotionMeta potionMeta = (PotionMeta) potionItem.getItemMeta();
+    for (PotionEffect effect : potion) {
+      potionMeta.addCustomEffect(
+          new PotionEffect((effect.getType()), effect.getDuration(), effect.getAmplifier()), false);
```
Pablete1234: Why can't you directly add the potion effect?
```suggestion
      potionMeta.addCustomEffect(effect, false);
```

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -83,10 +87,55 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
             XMLUtils.getChildren(
                 element, "item")) { // TODO Add more types of spawnables once entity parser is built
```
Pablete1234:
```suggestion
            XMLUtils.getChildren(element, "item")) {
```
This todo becomes outdated i guess :)

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -97,13 +146,27 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
                 minDelay,
                 maxDelay,
                 maxEntities,
-                numericID);
+                numericId);
```
Pablete1234: id should be the first param for spawner definition, and should come from a player-defined attribute if one is available, otherwise, use an autogenerated one
spawner definition, if it doesn't already, should extend SelfIdentifyingFeatureDefinition
You can take inspiration for the autogenerated id stuff from "PostIdSerial" and `PostDefinition.makeDefaultId` https://github.com/PGMDev/PGM/pull/984/files#diff-7aa0581d0bce176dc53a9eadaca6d5a541de970854921dbe6b302cd51f542807R81 (keep in mind that this example creates an ID from a NAME, which, neither composite posts nor spawners have names, which is why first param is always null)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
```diff
@@ -19,15 +19,16 @@ public class SpawnerDefinition implements FeatureDefinition {
   public final int numericID;
 
   public SpawnerDefinition(
+      int numericID,
```
Pablete1234:
```suggestion
      String id,
```

Make SpawnerDefinition extend SelfIdentifyingFeatureDefinition, and pass this id as param to super constructor

CoWinkKeyDinkInc:
The ThrownPotion entity now appears correctly by using NMS hacks, since its appearance can't be changed after spawning. The potion particles appear off centre (should be in centre of red wool) as you can see in the screenshot, but this is pretty minor. Currently the metadata isn't set to the splash potion entity, but other than that it's complete imo.

KingOfSquares:
Does the NMS hack work on Spigot as well?

core/src/main/java/tc/oc/pgm/spawner/SpawnerDefinition.java
```diff
@@ -10,24 +10,25 @@
...
   public final Region spawnRegion;
   public final Region playerRegion;
   public final int maxEntities;
   public final Duration minDelay, maxDelay, delay;
   public final List<Spawnable> objects;
   public final Filter playerFilter;
-  public final int numericID;
 
   public SpawnerDefinition(
+      String numericId,
```
Pablete1234:
```suggestion
      String id,
```
tf does numeric id mean, this is plain and simply, id.

extend SelfIdentifyingFeatureDefintion and pass it up to the parent, do not store it in SpawnerDefintion.

core/src/main/java/tc/oc/pgm/spawner/Spawner.java
```diff
@@ -102,27 +101,27 @@ public void onItemMerge(ItemMergeEvent event) {
...
       return;
     }
   }
 
   private void handleEntityRemoveEvent(Metadatable metadatable, int amount) {
     if (metadatable.hasMetadata(METADATA_KEY)) {
-      if (MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()).asInt()
-          == definition.numericID) {
+      if (Objects.equals(
+          MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()), definition.numericId)) {
```
Pablete1234:
```suggestion
          MetadataUtils.getMetadata(metadatable, METADATA_KEY, PGM.get()), definition.getId())) {
```

core/src/main/java/tc/oc/pgm/spawner/Spawner.java
```diff
@@ -102,27 +101,27 @@ public void onItemMerge(ItemMergeEvent event) {
...
-          MetadataUtils.getMetadata(event.getTarget(), METADATA_KEY, PGM.get()).asInt();
+      targetSpawnerId =
+          MetadataUtils.getMetadata(event.getTarget(), METADATA_KEY, PGM.get()).toString();
     }
     // Cancel the merge if the items are from different PGM spawners
-    if (entitySpawnerID != targetSpawnerID) {
+    if (!entitySpawnerId.equals(targetSpawnerId)) {
       event.setCancelled(true);
       return;
     }
```
Pablete1234: Whole logic of this event handler seems rather superfluous, including the code above. Can be simplified:
```
    boolean entityTracked = event.getEntity().hasMetadata(METADATA_KEY);
    boolean targetTracked = event.getTarget().hasMetadata(METADATA_KEY);
    if (!entityTracked && !targetTracked) return; // None affected
    if (entityTracked && targetTracked) {
      String entitySpawnerId =
          MetadataUtils.getMetadata(event.getEntity(), METADATA_KEY, PGM.get()).toString();
      String targetSpawnerId =
          MetadataUtils.getMetadata(event.getTarget(), METADATA_KEY, PGM.get()).toString();
      if (entitySpawnerId.equals(targetSpawnerId)) return; // Same spawner, allow merge
    }
    event.setCancelled(true);
```

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,82 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
-                element, "item")) { // TODO Add more types of spawnables once entity parser is built
-          ItemStack stack = kitParser.parseItem(spawnable, false);
-          SpawnableItem item = new SpawnableItem(stack, numericID);
+        for (Element itemEl : XMLUtils.getChildren(spawnerEl, "item")) {
+          ItemStack stack = kitParser.parseItem(itemEl, false);
+          SpawnableItem item = new SpawnableItem(stack, "spawner-" + numericId.get());
           objects.add(item);
         }
 
+        ImmutableList.Builder<PotionEffect> chBuilder = ImmutableList.builder();
```
Pablete1234: This is outside the inner loop for<potion>, that means, the effects are gonna carry over:

```xml
<spanwers>
  <spawner>
    <potion id="potion-1">
      <effect>A</effect>
    </potion>
    <potion id="potion-2">
      <effect>B</effect>
    </potion>
  </spawner>
</spanwers>
```
In this example, potion 1 would have effect A, and potion 2 would have both effect A and B

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,82 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+              // all possible status effects (ie wither, blindness, etc)
+              // Use the first listed PotionType for potion color
+              if (PotionType.getByEffect(potionEffect.getType()) != null) {
+                damageValue = PotionType.getByEffect(potionEffect.getType()).getDamageValue();
+                break;
+              }
+            }
+          }
+          objects.add(
+              new SpawnablePotion(potionChildren, damageValue, "spawner-" + numericId.get()));
```
Pablete1234:
```suggestion
              new SpawnablePotion(potionChildren, damageValue, id));
```
A string id should be created beforehand, taken from the XML "id" attribute if one is available, only if null do you put an autogenerated id

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,82 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
 
         List<Spawnable> objects = new ArrayList<>();
-        for (Element spawnable :
-            XMLUtils.getChildren(
-                element, "item")) { // TODO Add more types of spawnables once entity parser is built
-          ItemStack stack = kitParser.parseItem(spawnable, false);
-          SpawnableItem item = new SpawnableItem(stack, numericID);
+        for (Element itemEl : XMLUtils.getChildren(spawnerEl, "item")) {
+          ItemStack stack = kitParser.parseItem(itemEl, false);
+          SpawnableItem item = new SpawnableItem(stack, "spawner-" + numericId.get());
```
Pablete1234: id string should be pre-generated, ONCE, not in every place where it is used

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,82 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+              }
+            }
+          }
+          objects.add(
+              new SpawnablePotion(potionChildren, damageValue, "spawner-" + numericId.get()));
+        }
+
         SpawnerDefinition spawnerDefinition =
             new SpawnerDefinition(
+                "spawner-" + numericId.getAndIncrement(),
```
Pablete1234: the id string has to be created beforehand, not in each place

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+    potionItem.setItemMeta(potionMeta);
+    this.potionItem = potionItem;
+  }
+
+  @Override
+  public void spawn(Location location, Match match) {
+    NMSHacks.EntityPotion entityPotion = new NMSHacks.EntityPotion(location, potionItem);
+    // TODO set metadata when necessary
+    ((CraftWorld) location.getWorld()).getHandle().addEntity(entityPotion);
```
Pablete1234: The whole point of NMS hacks is not to do these nms hacks in-line.
if you're needing to import anything from craft-bukkit, that's where it has to be moved to nms hacks (notice you're importing `org.bukkit.craftbukkit.v1_8_R3.CraftWorld;` here)
Create a `NMSHacks.spawnThrownPotion(location, item)` or something like that

Pablete1234 (changes requested):

CoWinkKeyDinkInc:
> Does the NMS hack work on Spigot as well?

The entity does not spawn using Spigot 1.8.8. It returns this error:
```
[20:04:31 ERROR]: [MatchImpl] Could not tick tc.oc.pgm.spawner.Spawner@5713494e
java.lang.NoClassDefFoundError: org/bukkit/attribute/AttributeModifier
        at java.lang.Class.getDeclaringClass0(Native Method) ~[?:1.8.0_332]
        at java.lang.Class.getDeclaringClass(Class.java:1235) ~[?:1.8.0_332]
        at java.lang.Class.getEnclosingClass(Class.java:1277) ~[?:1.8.0_332]
        at java.lang.Class.getSimpleBinaryName(Class.java:1443) ~[?:1.8.0_332]
        at java.lang.Class.getSimpleName(Class.java:1309) ~[?:1.8.0_332]
        at org.bukkit.craftbukkit.v1_8_R3.SpigotTimings.getEntityTimings(SpigotTimings.java:91) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.Entity.<init>(Entity.java:120) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.EntityProjectile.<init>(EntityProjectile.java:45) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.EntityPotion.<init>(EntityPotion.java:31) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at tc.oc.pgm.util.nms.NMSHacks$EntityPotion.<init>(NMSHacks.java:1332) ~[PGM.jar:?]
        at tc.oc.pgm.spawner.objects.SpawnablePotion.spawn(SpawnablePotion.java:32) ~[PGM.jar:?]
        at tc.oc.pgm.spawner.Spawner.tick(Spawner.java:52) ~[PGM.jar:?]
        at tc.oc.pgm.match.MatchImpl$TickableTask.run(MatchImpl.java:721) [PGM.jar:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_332]
        at tc.oc.pgm.util.concurrent.TaskExecutorService$Task.run(TaskExecutorService.java:228) [PGM.jar:?]
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftTask.run(CraftTask.java:71) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftScheduler.mainThreadHeartbeat(CraftScheduler.java:350) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.MinecraftServer.B(MinecraftServer.java:723) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.DedicatedServer.B(DedicatedServer.java:374) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.MinecraftServer.A(MinecraftServer.java:654) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at net.minecraft.server.v1_8_R3.MinecraftServer.run(MinecraftServer.java:557) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
        at java.lang.Thread.run(Thread.java:750) [?:1.8.0_332]
```
@KingOfSquares

KingOfSquares:
> > Does the NMS hack work on Spigot as well?
> 
> The entity does not spawn using Spigot 1.8.8. It returns this error:
> 
> ```
> [20:04:31 ERROR]: [MatchImpl] Could not tick tc.oc.pgm.spawner.Spawner@5713494e
> java.lang.NoClassDefFoundError: org/bukkit/attribute/AttributeModifier
>         at java.lang.Class.getDeclaringClass0(Native Method) ~[?:1.8.0_332]
>         at java.lang.Class.getDeclaringClass(Class.java:1235) ~[?:1.8.0_332]
>         at java.lang.Class.getEnclosingClass(Class.java:1277) ~[?:1.8.0_332]
>         at java.lang.Class.getSimpleBinaryName(Class.java:1443) ~[?:1.8.0_332]
>         at java.lang.Class.getSimpleName(Class.java:1309) ~[?:1.8.0_332]
>         at org.bukkit.craftbukkit.v1_8_R3.SpigotTimings.getEntityTimings(SpigotTimings.java:91) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.Entity.<init>(Entity.java:120) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.EntityProjectile.<init>(EntityProjectile.java:45) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.EntityPotion.<init>(EntityPotion.java:31) ~[spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at tc.oc.pgm.util.nms.NMSHacks$EntityPotion.<init>(NMSHacks.java:1332) ~[PGM.jar:?]
>         at tc.oc.pgm.spawner.objects.SpawnablePotion.spawn(SpawnablePotion.java:32) ~[PGM.jar:?]
>         at tc.oc.pgm.spawner.Spawner.tick(Spawner.java:52) ~[PGM.jar:?]
>         at tc.oc.pgm.match.MatchImpl$TickableTask.run(MatchImpl.java:721) [PGM.jar:?]
>         at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_332]
>         at tc.oc.pgm.util.concurrent.TaskExecutorService$Task.run(TaskExecutorService.java:228) [PGM.jar:?]
>         at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftTask.run(CraftTask.java:71) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftScheduler.mainThreadHeartbeat(CraftScheduler.java:350) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.MinecraftServer.B(MinecraftServer.java:723) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.DedicatedServer.B(DedicatedServer.java:374) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.MinecraftServer.A(MinecraftServer.java:654) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at net.minecraft.server.v1_8_R3.MinecraftServer.run(MinecraftServer.java:557) [spigot-1.8.8.jar:git-Spigot-21fe707-741a1bd]
>         at java.lang.Thread.run(Thread.java:750) [?:1.8.0_332]
> ```
> 
> @KingOfSquares

Then you should adjust the NMSHack using the SportPaper check so it uses a Spigot compatible method. Not a real worry since its only potion colors right?

CoWinkKeyDinkInc:
Luckily it looks like it was just a building error on my end, using the PGM compiled by the GitHub Builds shows it spawning the potion properly without any errors in Spigot.

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,86 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
-        Attribute maxDelayAttr = element.getAttribute("max-delay");
+        String id = spawnerEl.getAttributeValue("id");
+        Region spawnRegion = regionParser.parseRequiredRegionProperty(spawnerEl, "spawn-region");
+        Region playerRegion = regionParser.parseRequiredRegionProperty(spawnerEl, "player-region");
+        Attribute delayAttr = spawnerEl.getAttribute("delay");
+        Attribute minDelayAttr = spawnerEl.getAttribute("min-delay");
+        Attribute maxDelayAttr = spawnerEl.getAttribute("max-delay");
+
+        if (spawnerEl.getAttributeValue("id") == null) {
+          id = "spawner-" + numericId.getAndIncrement();
```
Pablete1234: move the generation of this id to SpawnerDefinition itself, check PostDefinition for an example
Also use pgm's default fallback generation, iirc it goes a bit like `--(type)-(name)` in this case type would be spawner and name just the id (like posts) so: `--spawner-1` and such
It generally will generate less collisions on accident

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,86 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
-          ItemStack stack = kitParser.parseItem(spawnable, false);
-          SpawnableItem item = new SpawnableItem(stack, numericID);
+        for (Element itemEl : XMLUtils.getChildren(spawnerEl, "item")) {
+          ItemStack stack = kitParser.parseItem(itemEl, false);
+          SpawnableItem item = new SpawnableItem(stack, id);
           objects.add(item);
         }
 
+        for (Element potionEl : XMLUtils.getChildren(spawnerEl, "potion")) {
+          ImmutableList.Builder<PotionEffect> chBuilder = ImmutableList.builder();
```
Pablete1234:
```suggestion
          ImmutableList.Builder<PotionEffect> effectsBuilder = ImmutableList.builder();
```

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,86 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
+          for (Element potionChild : potionEl.getChildren("effect")) {
+            chBuilder.add(XMLUtils.parsePotionEffect(new InheritingElement(potionChild)));
+          }
+          ImmutableList<PotionEffect> potionChildren = chBuilder.build();
```
Pablete1234:
```suggestion
          ImmutableList<PotionEffect> effects = chBuilder.build();
```

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,86 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
                 playerRegion,
                 playerFilter,
                 delay,
                 minDelay,
                 maxDelay,
-                maxEntities,
-                numericID);
-        factory.getFeatures().addFeature(element, spawnerDefinition);
+                maxEntities);
+        factory.getFeatures().addFeature(spawnerEl, spawnerDefinition);
```
Pablete1234, line 133: i *think* you may no longer need the EL here, since the definition is self identifying now, but double-check (try to just pass spawnerDefinition)

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,40 @@
...
+      potionMeta.addCustomEffect(effect, false);
+    }
+    potionItem.setItemMeta(potionMeta);
+    this.potionItem = potionItem;
+  }
+
+  @Override
+  public void spawn(Location location, Match match) {
+    new NMSHacks.EntityPotion(location, potionItem).spawn();
+    // TODO set metadata when necessary
```
Pablete1234: self-explainatory

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnablePotion.java
```diff
@@ -0,0 +1,40 @@
...
+  private final ItemStack potionItem;
+  private final String spawnerId;
+
+  public SpawnablePotion(List<PotionEffect> potion, int damageValue, String spawnerId) {
+    this.spawnerId = spawnerId;
+    // Potion "name" determines potion color
+    ItemStack potionItem = new Potion(damageValue).splash().toItemStack(1);
+    PotionMeta potionMeta = (PotionMeta) potionItem.getItemMeta();
+    for (PotionEffect effect : potion) {
+      // overwrite = false
```
Pablete1234:
```suggestion
```

Pablete1234 (changes requested):
Looking pretty good in general, just a few more tweaks

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
CoWinkKeyDinkInc: The element is still required.

Pablete1234 (approved):

Electroid (approved):

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,84 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
       RegionParser regionParser = factory.getRegions();
       KitParser kitParser = factory.getKits();
       FilterParser filterParser = factory.getFilters();
+      AtomicInteger postIdSerial = new AtomicInteger(1);
```
Pablete1234, line 57: spawnerIdSerial

core/src/main/java/tc/oc/pgm/spawner/SpawnerModule.java
```diff
@@ -49,58 +54,84 @@ public SpawnerModule parse(MapFactory factory, Logger logger, Document doc)
...
-        Attribute minDelayAttr = element.getAttribute("min-delay");
-        Attribute maxDelayAttr = element.getAttribute("max-delay");
+        String id = spawnerEl.getAttributeValue("id");
+        Region spawnRegion = regionParser.parseRequiredRegionProperty(spawnerEl, "spawn-region");
+        Region playerRegion = regionParser.parseRequiredRegionProperty(spawnerEl, "player-region");
+        Attribute delayAttr = spawnerEl.getAttribute("delay");
+        Attribute minDelayAttr = spawnerEl.getAttribute("min-delay");
+        Attribute maxDelayAttr = spawnerEl.getAttribute("max-delay");
+
+        if (id == null) id = SpawnerDefinition.makeDefaultId("spawner", postIdSerial);
```
Pablete1234, line 68: `if (id == null) id = SpawnerDefinition.makeDefaultId(null, spawnerIdSerial);`

PR #1003, opened 2022-05-12
Return all definitions instead of only definitions with ids from FeatureDefinitionContext

KingOfSquares:
Title, 

before this PR calling this method for FeatureDefinitionContext would only return definitions registered with an id, resulting in a loss of stuff like this:

```xml
        <filter>
            <all>
                <time>2m</time>
                <team>hiders</team>
            </all>
        </filter>
```

Should probably be merged before #950


KingOfSquares:
Merged into #950

PR #1005, opened 2022-05-22
Cleanup filters & implement missing (players, rank & countdown)

Pablete1234:
This PR cleans up filters in PGM and orders them in different folders similar to how legacy pgm (1.12+) ordered them. Makes them much easier to work with instead of having a single package with 100 files.

### Filter changes
 - TypedFilter is now an interface (with TypedFilter.Impl being a class), this allows using TypedFilter alongside a SingleFilterFunction.
 - TypedFilter is now forced to return a boolean, cleans up the QueryResult.fromBoolean used in every filter. If you implement TypedFilter you MUST return to the query. If you want to abstain, use a WeakTypedFilter instead.
 - Implemented TeamFilterAdapter, goal filter uses this, and now score and rank filters support it to (eg: you can `<score team="red">5</score>` to check if red team has a score of 5)
 - Added helper methods in MatchQuery to get a module or reactor

### Reimplemented filters
 - Player count filter:
   - `<players min="3" max="5"><crouching/></players>` allows when between 3 and 5 in the match are crouching
 - Rank filter:
   - `<rank min="4" max="7">` allows for competitors ranked 4 to 7 (mostly useful for FFA, but can also help distinguish winning and losing team)
 - Countdown filter (without bossbar):
   -  `<countdown duration="5s"><crouching/></countdown>` allows 5 seconds after you start crouching.

Note: this has not yet been tested


Pablete1234:
@KingOfSquares i'd like if you could take a look at this too given how much you've worked on dynamic filters

core/src/main/java/tc/oc/pgm/api/filter/query/MatchQuery.java
```diff
@@ -12,4 +16,16 @@ default Filterable<?> extractFilterable() {
...
     return this.getMatch();
   }
+
+  default <T extends MatchModule> T moduleRequire(Class<T> cls) {
+    return getMatch().needModule(cls);
+  }
+
+  default <T extends MatchModule> Optional<T> moduleOptional(Class<T> cls) {
+    return Optional.ofNullable(getMatch().getModule(cls));
+  }
```
KingOfSquares, line 26: Why does this need to be different than using the underlying methods?

core/src/main/java/tc/oc/pgm/filters/FilterDispatcher.java
```diff
@@ -1,13 +1,15 @@
...
 import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.filter.FilterListener;
 
 public interface FilterDispatcher {
 
   /**
    * Register the given {@link FilterListener} to receive notifications of changes to the response
-   * of the given {@link Filter} to any {@link Filterable} assignable to the given scope.
+   * of the given {@link Filter} to any {@link tc.oc.pgm.filters.Filterable} assignable to the given
+   * scope.
```
KingOfSquares: Seems like this class got a bunch of fully qualified imports when the class was moved, maybe just import it to reduce unnecessary changes?

core/src/main/java/tc/oc/pgm/filters/matcher/WeakTypedFilter.java
```diff
@@ -0,0 +1,25 @@
...
+ * A filter that NEVER responds to queries outside of {@link #queryType()}, and SOMETIMES responds
+ * to queries extending {@link #queryType()}.
+ *
+ * <p>Queries of the latter type are passed to {@link #queryTyped(Query)}.
+ *
+ * <p>The runtime type of {@link Q} is detected automatically if it is specified by a subclass.
+ */
+public interface WeakTypedFilter<Q extends Query> extends FilterDefinition {
+
+  Class<? extends Q> queryType();
```
KingOfSquares, line 16: Why did you rename this method? Should probably just stay with the getter style like the rest of the codebase

KingOfSquares (commented):
A few general points:

- I don't see the point of preserving TyperFilter as a class in addition to the interface? Why not just let filters implement the interface?
- Seems like the oc.tc.pgm.restart package has gotten some weird diffs by mistake?
- While I also like optionals, the general style of PGM seems to be more `@Nullable`, (I've gotten feedback to not use optionals before) 

I'm a real fan of the changes though, cleaner and smarter 😎 👍

core/src/main/java/tc/oc/pgm/filters/matcher/WeakTypedFilter.java
Pablete1234: I've been streamlining our codebase to fit 1.12's style and i feel like this one made some sense, after all, it's not *getting* a property (like, a mutable state) it's just the query type of this weakly typed filter

core/src/main/java/tc/oc/pgm/api/filter/query/MatchQuery.java
Pablete1234: it's convenient

Pablete1234:
> I don't see the point of preserving TyperFilter as a class in addition to the interface? Why not just let filters implement the interface?

I believe in 1.12 pgm's codebase this was a thing with how references were handled, if you extend Filter.Impl it would have the correct equals/hashcode while if you didn't it'd be incorrect? I'm not exactly sure, i don't even know that we have an actual main abstract filter that all others extend from, they used to just extend typed filter, should probably extend just filter now (a main abstract class) but did not want to bother with it.

> Seems like the oc.tc.pgm.restart package has gotten some weird diffs by mistake?

Maybe line-ends changed? i'm not sure why they'd be included here

> While I also like optionals, the general style of PGM seems to be more @Nullable, (I've gotten feedback to not use optionals before)

1.12's pgm used them around and for some things they make it easier to work with things via .map etc than having to do null-checks everywhere. I think we should use them where they make things easier

core/src/main/java/tc/oc/pgm/api/filter/query/MatchQuery.java
KingOfSquares: Oh right this will be accessible through the actual match object as well, I thought it was weird having separate behaviors for the query and the match but it makes sense now.

core/src/main/java/tc/oc/pgm/filters/matcher/WeakTypedFilter.java
KingOfSquares: Its just every time I do that I've been told to keep it to this old PGM style or to prevent breaking changes 🙃

KingOfSquares:
> > I don't see the point of preserving TyperFilter as a class in addition to the interface? Why not just let filters implement the interface?
> 
> I believe in 1.12 pgm's codebase this was a thing with how references were handled, if you extend Filter.Impl it would have the correct equals/hashcode while if you didn't it'd be incorrect? I'm not exactly sure, i don't even know that we have an actual main abstract filter that all others extend from, they used to just extend typed filter, should probably extend just filter now (a main abstract class) but did not want to bother with it.

I always thought it was a thing because of the injection 🤔

core/src/main/java/tc/oc/pgm/filters/matcher/WeakTypedFilter.java
Pablete1234: On the API? understandable, but the point is this method has been *removed* from the api (Filter.java no longer has it), and is now only existing for *some* filters which want *convenience methods* for their type, as an internal thing.

Electroid (approved):

PR #1008, opened 2022-05-31
Update dependencies and resolve new deprecated method usage

calcastor:
Update ViaVersion to 4.2.1
Update JDOM to 2.0.6.1
Update Adventure libraries to ~~4.10.1~~ 4.11.0 and platform to 4.1.0
Update FastBoard to 1.2.1
Update JUnit Jupiter to 5.8.2
Update JGit to 5.13.0, the last version supporting Java 8

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blitz/BlitzMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/blitz/BlitzMatchModule.java
@@ -123,7 +123,7 @@ public class BlitzMatchModule implements MatchModule, Listener {
                           lives == 1 ? "misc.life" : "misc.lives",
                           NamedTextColor.AQUA,
                           text(lives))),
-                  Title.Times.of(Duration.ZERO, fromTicks(60), fromTicks(20))));
+                  Title.Times.times(Duration.ZERO, fromTicks(60), fromTicks(20))));
     }
   }
 
--- a/core/src/main/java/tc/oc/pgm/command/MatchCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MatchCommand.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
 import net.kyori.adventure.text.Component;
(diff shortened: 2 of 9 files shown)


core/src/main/java/tc/oc/pgm/blitz/BlitzMatchModule.java
```diff
@@ -123,7 +123,7 @@ public void handleSpawn(final ParticipantSpawnEvent event) {
                           lives == 1 ? "misc.life" : "misc.lives",
                           NamedTextColor.AQUA,
                           text(lives))),
-                  Title.Times.of(Duration.ZERO, fromTicks(60), fromTicks(20))));
+                  Title.Times.times(Duration.ZERO, fromTicks(60), fromTicks(20))));
```
Electroid, line 126: Were these breaking changes? Even if we update, I'd rather continue to use the deprecated methods to minimize the amount of breakage too shading among other things.

KingOfSquares: Both this and the GlobalTranslator method is deprecated for removal, but will stay until 5.0.0
https://github.com/KyoriPowered/adventure/blob/15a2939c7566d8eabc7008eb233ff176a824cfe2/api/src/main/java/net/kyori/adventure/title/Title.java#L125
https://github.com/KyoriPowered/adventure/blob/15a2939c7566d8eabc7008eb233ff176a824cfe2/api/src/main/java/net/kyori/adventure/translation/GlobalTranslator.java#L63

Electroid: Let's keep them for now then.

calcastor: I've tested all changes in the PR and this specifically does not appear to be a breaking change. The titles appear as configured. However I am uncertain how to test the switch to `GlobalTranslator.translator()` fully, though there appears to be no issues with translations showing correctly when a client is using a language other than English in my testing.

Electroid (approved):

PR #956, opened 2022-01-17
Remove legacy aliases from pool command

Pablete1234:
The legacy /rot or /rotation commands have long been deprecated since the introduction of pools, the most common type of pool has not been a rotation for a very long time now and i think it's time to move on from legacy naming and fully embrace pools

PR #1010, opened 2022-06-05
Rework map pools & add map vote picker

Pablete1234:
Reworks a few aspects in map pools, and makes the voted pool use a `MapVotePicker` to decide what maps go on the vote, as well as handling the score to weight conversion via a custom formula defined in the config. 
This extra step allows for arbitrary modifications to score for extra functionality such as:
  - lowering weights for gamemodes that already got representation on the vote (eg: if first map is CTW, CTW is less likely to appear in the following picks)
  - Current online playercount to map size ratio (prefer playing maps of proper size over wrongly-sized ones).
  - Allow tweaking via config values all of the internal dials that affect this, so they can be fine-tuned in production


core/src/main/java/tc/oc/pgm/rotation/pools/VotingPool.java
```diff
@@ -1,48 +1,42 @@
...
 public class VotingPool extends MapPool {
 
-  // Number of maps in the vote, unless not enough maps in pool
-  public static final int MAX_VOTE_OPTIONS = 5;
-  // Number of maps required for a custom vote (/vote)
-  public static final int MIN_CUSTOM_VOTE_OPTIONS = 2;
-  // If maps were single voted, it would avg to this default
-  public static final double DEFAULT_WEIGHT = 1d / MAX_VOTE_OPTIONS;
+  // Arbitrary default of 1 in 5 players liking each map
+  public static final double DEFAULT_SCORE = 0.2;
```
Electroid, line 21: What are the advantages/disadvatages of having this score be more pessimistic (closer to 0) or optimistic (closer to 1)?

Pablete1234: At server start, all maps are just as likely, so there isn't a difference, however, after a vote, the scores of the maps that were voted are set to their real vote score.
What that means, is if after a vote a map gets 25% of the votes and ends up at 0.25, how will that compare against all other maps still sitting on defaults?
If the default is 0.8: the map with 25% vote is "super bad" in relation to everything else, so it's super unlikely to appear in a future vote
If the default is 0.1: the map with 25% vote is "super good" in relation to everything else, so it's super likely to appear in a future vote

From what i've seen a 0.2 is pretty typical, because there's 5 options to vote for (and it seems like players voting for multiple options is pretty rare, most people seem to not even know that's possible). So maps that get less than expectancy of 1/5th are underperforming and maps above 1/5th are overperforming, and so, the 0.2 score. This does not represent a change over the previous, its just written differently

TLDR: it sets the "break-even" point for maps being voted. Maps being voted above default get more likely, maps under default get less likely.

core/src/main/java/tc/oc/pgm/command/MapPoolCommand.java
```diff
@@ -43,7 +43,7 @@ public final class MapPoolCommand {
   private static final DecimalFormat SCORE_FORMAT = new DecimalFormat("00.00%");
 
   @Command(
-      aliases = {"pool", "rotation", "rot"},
```
calcastor, line 46: While these changes are going to improve voted pools overall, at the very least I think this specific change should be split off of this PR if not dropped altogether. The (non-verbal, yet still valid) feedback on #956 when this change was originally made was overwhelmingly against the change. Though I personally prefer using voted pools in my position as a map developer on OCC, there are servers in the greater PGM community that have elected to not use them and to keep static rotations. These may be pools by definition, but this does not discount the use of the word "rotation" to refer to them as dates back to OCN.

Pablete1234: I believe it is time to drop the outdated terms, they've been in the codebase for legacy purposes for a long while, #956 was reverted waiting for a release to be made before such a breaking change, a release was cut just yesterday so we're good to go on that department

core/src/main/resources/map-pools.yml
```diff
@@ -26,6 +26,112 @@ pools:
     # How long should should a map cycle last?
     cycle-time: "15s"
 
+
+    # Voted pools support modifiers which come in the form of formulas (does not affect any other type of pool).
```
Electroid: I worry this is complicated and verbose, I'd rather not invent our own formula format/parser. What if we used JavaScript instead? You define a `JS` file with a function, then PGM invokes it with parameters.

https://stackoverflow.com/a/22856530

Then you can use JS Math to define your function, and it's much easier to follow and read. Also, it reduces our maintenance burden.

Electroid (changes requested):

core/src/main/resources/map-pools.yml
Pablete1234: I am very well aware about how complicated this turned, and it is far worse than i anticipated it'd get. However, i'm not convinced that running arbitrary code (eg: js) will be a better solution, for a multitude of reasons (ie: it brings its own set of problems):
 - Ensuring the (arbitrary) code doesn't mess with any functionality anywhere. Running js via nashorn allows interacting with java objects and we would NOT want that happening.
 - This needs run fast, as it'll run for *every map* in the pool *5 times* every vote. This config system, although not yet tested for performance, compiles at runtime and is essentially chained lambdas, so i anticipate it should run pretty fast. Meanwhile the JS may even be getting compiled & evaluated every time it's called.
  - I'm not sure what the current state of JS engines in java is. I know nashorn was deprecated for removal and removed in jdk 12 (the stack overflow you point at is 8 years old). GraalVM has its own alternative but for portability purposes i'd not want to tie pgm to a specfic jvm vendor (especially since graalvm has worked poorly memory-wise with pgm in the past.
  
Regarding maintainability, i may be a bit naive but i believe this is flexible enough that we're unlikely to need to maintain this for quite a few years, as all values (and even diff formulas) can be changed all via config. That's the reason it turned into this, it gives maximum flexibility and control **in the config file** so that the code doesn't need to be changed to tweak weights and what each thing affects score in what way.

Electroid (changes requested):

core/src/main/resources/map-pools.yml
Electroid: What if we used something more lightweight instead? The dependency below is only 25kb.

https://stackoverflow.com/questions/4681959/algebra-equation-parser-for-java
https://lallafa.objecthunter.net/exp4j/

Pablete1234: Something like exp4j seems way more viable for this (it wasn't something i found when searching for stuff like this), looking into it, it seems like it already has built in most of the functions we have, the only one we have is bound(val, min, max) which we'd have to implement as a custom one

I'm not sure about how performance will be but it does parse and tokenize ahead of time, then execution is simply iterating over all tokens and evaluating, so it should be viable

Electroid: Let's try it out, rest of your refactoring looks good.

core/src/main/java/tc/oc/pgm/rotation/vote/Formula.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+import java.util.function.ToDoubleFunction;
+import net.objecthunter.exp4j.Expression;
+import net.objecthunter.exp4j.ExpressionBuilder;
+import net.objecthunter.exp4j.function.Function;
+import org.bukkit.configuration.InvalidConfigurationException;
+
+/** Implementation of different formulas configurable & composable via config sections. */
+public interface Formula extends ToDoubleFunction<Formula.Context> {
```
Electroid: Could we move this to `util` and make it a `final class`?

core/src/main/resources/map-pools.yml
```diff
@@ -26,6 +26,28 @@ pools:
...
+    #
+    # On top of that PGM implements:
+    # - bound(val, min, max)
+    #
+    # Available variables:
+    #  - score: the maps' score, determined by percentage of players who voted for it, or a default
+    #  - same_gamemode: amount of other maps of the same gamemode already picked for the vote
+    #  - mapsize: the size for this map in total players (sum of all team counts)
+    #  - players: amount of active players online (eg: obs count as half)
+    modifier: >-
```
Electroid, line 51: Should this be called `formula`?

core/src/main/resources/map-pools.yml
```diff
@@ -26,6 +26,28 @@ pools:
...
+    # This formula is parsed by exp4j library, it's quite flexible and supports many built-in functions.
+    #
+    # On top of that PGM implements:
+    # - bound(val, min, max)
+    #
+    # Available variables:
+    #  - score: the maps' score, determined by percentage of players who voted for it, or a default
+    #  - same_gamemode: amount of other maps of the same gamemode already picked for the vote
+    #  - mapsize: the size for this map in total players (sum of all team counts)
+    #  - players: amount of active players online (eg: obs count as half)
```
Electroid, line 41: Could we add a note: "DO NOT MODIFY UNLESS YOU KNOW WHAT YOU'RE DOING."

Electroid (dismissed):
Approved, but minor changed requested.

core/src/main/java/tc/oc/pgm/rotation/vote/Formula.java
Pablete1234: If we were to move it to util, it'd have to be not as part of specifically map voting but a more generic wrapper for Expression from exp4j

core/src/main/resources/map-pools.yml
Pablete1234: I think modifier is more accurate, it's not just a *formula* it's a *modifier* applied to the score to get what the weight for the map should be

core/src/main/resources/map-pools.yml
Pablete1234: Could maybe list a couple simpler alternatives for non-playercount-dependant modifier, but yeah

Electroid (approved):
Thanks for landing these improvements! Excited 🎉

PR #1012, opened 2022-06-08
Fix race condition loading core/destroyable modes

calcastor:
This PR restores statelessness to CoreModule and DestroyableModule factories to resolve a race condition when loading core and monument maps with modes. The race condition resulted in loading errors for these maps due to mismatching of modes between different maps. This issue is most easily observed on servers with a large amount of maps, such as OCC, where well over 1000 maps are loaded at one time.

Debugged and fixed with the help of @Pablete1234

### DIFF
--- a/core/src/main/java/tc/oc/pgm/core/CoreModule.java
+++ b/core/src/main/java/tc/oc/pgm/core/CoreModule.java
@@ -70,7 +70,6 @@ public class CoreModule implements MapModule {
   }
 
   public static class Factory implements MapModuleFactory<CoreModule> {
-    private MapFactory factory;
 
     @Override
     public Collection<Class<? extends MapModule>> getWeakDependencies() {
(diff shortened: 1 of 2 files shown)


Pablete1234 (approved):

Pablete1234:
To explain a bit more what happened here, it's that pgm module factories must be stateless, and when adding a past change (to allow selecting which modules each core/destroyable should use), state was added.
When entering on the `parse` function, it set a local variable, that was accessed later (from a function to search a mode), the issue here is these `parse` functions can be called concurrently, so what would happen is map A starts parsing, sets its own context, then (before map a finishes) map B starts parsing, sets its own context. Now when map A reaches to search for the modes, it gets modes from map B, and if map B doesn't have them, you get errors about invalid xml (missing mode).

This change fixes it

Electroid (approved):

Issue #1001, opened 2022-04-28
Kit-stretching ideas (actions)


Pablete1234:
This is an issue for ideas that could be implemented into pgm in the form of kits, even though they really stretch the definition of what *kit* is supposed to mean. The new terminology they'll adopt is *triggers*, which can be found over at #1016 .

Keep in mind, that not all of these must be viable to implement, or make sense to implement. It's just a list of anything that *could maybe be looked at*, some may become impractical or non viable upon closer inspection or when trying to look at details of implementation.

The ideas/brainstorming for future kits:
 - [x] variable-related kits (see #881)
 - [x] `text` - sends a text message to the player #900
 - [x] `sound` - plays a certain sound for the player #1216
 - [ ] `pickup-flag`, `drop-flag` - makes the player pick-up or drop the flag #930
 - [ ] `match-end` - ends the match, optionally:  tie, player wins, specific team wins, or default to a victory condition (like time-limit)
 - [x] `timelimit` - add or subtract a certain amount of time from the match time limit
 - [x] `score` - award score, (could even replace scoreboxes?)
 - [x] `teleport` - teleport the player, (could even replace portals?)
 - [ ] `capture`, `uncapture` - modify control points' ownership #942
 
 Feel free to suggest more ideas for kits-that-aren't-really-kits

KingOfSquares:
Didn't avicus use something called executors or something? Could be a viable alternative to stuff not belonging under the kit term.

lefckntea:
can i make tnt tag (flag tag?) with pickup-flag and drop-flag? or does it need more stuff like the transfer flag when hitting someone idea ruediger had in the issue mentioned 

they‘d also needa go boom then for the tnt tag mechanic

Issue #881, opened 2021-06-01
Variables match module proposal


Pablete1234:
This is something that pgm has lacked for long, essentially arbitrary variables that can be set and modified/kept track of, they should allow for more complex gamemodes to be made easier, often hidden hills are used for a "boolean bit" that can be set by a player or similar, this is however far from ideal.

There are many things that are intentionally overlooked for simplicity on this *first* proposal, the idea is for this to be expanded & modified based on comments and ideas from others.

Initial syntax proposal example:
```xml
<variables>
    <var id="progress"      scope="match"  initial-value="0"/>
    <var id="team-score"    scope="team"   initial-value="0"/>
    <!-- if the variable should be kept if they rejoin. always, same-team, never -->
    <var id="player-score"  scope="player" initial-value="0" persist="always"/>
    <var id="temp"          scope="match"  initial-value="0"/>
</variables>

<filters>
    <!-- Maybe a better spawn unlocks after a team gets 1000 score? -->
    <variable id="unlock-spawn" var="team-score" range="[1000, oo)" />
</filter>


<kits>
    <kit id="award-progress">
        <!-- Add 5 to the overall match progress -->
        <let var="progress" operation="add" amount="5"/>
        <!-- Calculate a random value between 20 and 100 -->
        <let var="temp" operation="random" amount="80"/>
        <let var="temp" operation="add"    amount="20"/>
        <!-- Add up the progress for both player & team -->
        <let var="team-score"   operation="add" amount="temp"/>
        <let var="player-score" operation="add" amount="temp"/>
    </kit>
</kits>
````

This would make it so the map has 3 main variables, one that is the same for everyone for overall progress, one for team progress, and one for player progress.

The variables can be modified by using kits, which i think is the best "framework" that PGM has to make actions happen (when filters activate, or when you enter a region), although admiteddly currently PGM has no dynamic filters so dynamic kits don't quite give us as much flexibility as that will give us.

You can use these variables in as many filters as you want, maybe enabling faster (or slower) spawns, giving different kits to players that progress, or whatever else. It would also be interesting to allow other places in the codebase to take variables as parameters, so you could for example make respawn time directly dependant on a variable, as this current approach would require you to setup as many respawns as you possibly want, each with a filter for a different score.

Operations should be flexible to be able to be able to implement new ones in the future if they prove useful, currently the main set would be `add`, `subtract`, `multiply`, `divide`, `mod`, `set`, `min`, `max`, `random`.

[...]

PR #1016, opened 2022-06-16
Implement actions and triggers!

Pablete1234:
What are actions? Actions are the next generation of *kits*. Kits in fact, are a type of action.

Why not just use kits? Kits are always player-scoped, this is not ideal for all features, there's alot of interoperability between kits and actions so this shouldn't be an issue (kits can have player-scoped actions, and kits themselves are a type of player-scoped action), and actions can be scoped-out or scoped-in.

This is mainly a stepping stone towards #881, it is also laying down the ground for #1001, as those will no longer be kits, and will just be actions instead.

Currently the only implemented actions are action (an action with other actions inside), message, switch-scope, and kit

```xml
  <kits>
    <kit id="spawn">
      <item slot="0" unbreakable="true" material="stone sword"/>
      <action>
        <message text="You were given a kit!"/>
      </action>
    </kit>
  </kits>
  <actions>
    <action id="do-stuff" scope="player">
      <message text="Sent a chat message!"/>
      <switch-scope outer="player" inner="team">
        <message text="All players on the team are getting this one!"/>
        <switch-scope outer="team" inner="player">
          <kit id="spawn"/>
        </switch-scope>
      </switch-scope>
      <kit>
        <item>diamond</item>
      </kit>
    </action>
    <message id="standalone-text" text="This is a standalone text trigger"/>
    <trigger filter="some-dynamic-filter" action="do-stuff" scope="player"/>
    <trigger filter="another-dynamic-filter" action="standalone-text" scope="player"/>
  </actions>
```

Update: 
triggers have been renamed to actions
when has been renamed to trigger

It makes more sense that "actions" are "triggered" by triggers. The triggers listen for dynamic filters. this was feedback from @Pugzy


core/src/main/java/tc/oc/pgm/trigger/modifier/TriggerModifiers.java
```diff
@@ -0,0 +1,51 @@
...
+import com.google.common.collect.ImmutableTable;
+import com.google.common.collect.Table;
+import java.util.function.Function;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.party.Party;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.trigger.Trigger;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+
+public interface TriggerModifiers {
```
KingOfSquares: Any specific reason you went with the current design instead of a design similar to how Filterables are done? (Decendants/Anscestors).
Maybe the Filterable API could be used as-is, with the outer/inner relation maybe just being saved in some other way like an Enum instead which maps to different usages of the Filterable API.

core/src/main/java/tc/oc/pgm/trigger/TriggerParser.java
```diff
@@ -0,0 +1,165 @@
...
+
+  protected Method getParserFor(Element el) {
+    return methodParsers.get(el.getName().toLowerCase());
+  }
+
+  private <T> TriggerDefinition<T> parseDynamic(Element el) throws InvalidXMLException {
+    Method parser = getParserFor(el);
+    if (parser != null) {
+      try {
+        //noinspection unchecked
```
KingOfSquares: Maybe just a `@SuppressWarnings("unchecked")` instead?

Pablete1234: that needs to be on a method, class, or field level, a return X can't have that, so i'd need to make it for the whole method

core/src/main/java/tc/oc/pgm/trigger/modifier/TriggerModifiers.java
Pablete1234: I'm not exactly sure how the whole filterable API works, but an important point of triggerables is the ability to switch scopes and to be able to specifically know the scope something is running in. This will be especially relevant for variable-related triggers

core/src/main/java/tc/oc/pgm/trigger/TriggerParser.java
KingOfSquares: Yeah my thought was the `praseDynamic` method

Pablete1234:
Implemented bound inference, so specifying the scope of a trigger isn't required if it can be infered from the parent, the biggest effect is switch-scope only requires `inner`, as long as the outer can be derived.

Example where it can be infered:
```xml
  <actions>
    <actionid="trigger-id" scope="player">
      <switch-scope inner="match">
        <message text="This is match scoped! Converted from player to match"/>
      </switch-scope>
    </action>
  </actions>
```

Example where it can NOT be inferred (outer is required here, or else it won't work):
```xml
  <action>
    <switch-scope id="scope-switch" outer="player" inner="match">
      <message text="This is match scoped! Converted from player to match"/>
    </switch-scope>
    <action id="trigger-id" scope="player">
      <action id="scope-switch"/> <!-- This references the trigger defined above -->
    </action>
  </actions>
```

Additionally, now the Kit interface fully implements Trigger<MatchPlayer>, because at the end, kits are just a trigger that is scoped to a player. This means you can just define a kit inside a trigger instead of being forced to reference one.

```xml
  <action>
    <kit id="a-kit">
      <item>diamond</item>
    </kit>
    <trigger filter="some-filter" action="a-kit" scope="player">
  </action>
```
In this situation, the `trigger` is working exactly equally as a `<lend kit="a-kit" filter="some-filter"/>` would.

Also, Actions now support triggering and untriggering, in the same way removable kits work.

core/src/main/java/tc/oc/pgm/trigger/TriggerMatchModule.java
```diff
@@ -0,0 +1,37 @@
...
+      setupTrigger(triggerRule, fmm);
+    }
+  }
+
+  private <T extends Filterable<?>> void setupTrigger(TriggerRule<T> rule, FilterMatchModule fmm) {
+    fmm.onChange(
+        rule.getScope(),
+        rule.getFilter(),
+        (t, val) -> {
+          if (val) rule.getTrigger().trigger(t);
```
KingOfSquares: What about `f` or `filterable` instead of `t` and `r` or `response` instead of `val`

core/src/main/java/tc/oc/pgm/trigger/triggers/AbstractTrigger.java
```diff
@@ -0,0 +1,20 @@
...
+    this.cls = cls;
+  }
+
+  @Override
+  public Class<S> getScope() {
+    return cls;
+  }
+
+  @Override
+  public void untrigger(S s) {}
```
KingOfSquares: Why not default both `trigger` and `untrigger`? It could make sense to keep the possibility open of a trigger only reacting to a filter falling instead of internally having to invert the filter response to use `trigger`

Pablete1234: The point of a default is just because *most triggers* will not have an `untrigger`, so you don't force them to add a method.
It is still perfectly possible for a trigger to implement untrigger and leave trigger empty, but usually you'd want them to implement trigger

core/src/main/java/tc/oc/pgm/action/Trigger.java
```diff
@@ -22,7 +22,7 @@ public Filter getFilter() {
     return filter;
   }
 
-  public Trigger<? super T> getTrigger() {
-    return trigger;
+  public Action<? super T> getTrigger() {
```
KingOfSquares: `getAction`?

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -730,15 +730,15 @@ public MaxHealthKit parseMaxHealthKit(Element parent) throws InvalidXMLException
     return new MaxHealthKit(maxHealth);
   }
 
-  public TriggerKit parseTriggerKit(Element parent) throws InvalidXMLException {
-    if (parent.getChildren("trigger").isEmpty()) return null;
+  public ActionKit parseTriggerKit(Element parent) throws InvalidXMLException {
```
KingOfSquares: `parseActionKit`?

core/src/main/java/tc/oc/pgm/trigger/triggers/AbstractTrigger.java
KingOfSquares: I agree using `trigger` should be the primary way of creating an Action, but I can't see a reason to just default implement `untrigger` preemptively?

Electroid (approved):

PR #1020, opened 2022-07-04
Reintroduce Loot Module

CoWinkKeyDinkInc:
Return of the Minecraft Hunger Games!

This PR reintroduces the Loot module which distributed items with or without probability to containers. This utilizes brand new code. This work was requested on behalf of @TBG1000 and the other fine folks at @warzone

# New additions
 - IDs are no longer required (self identifying like SpawnerDefinition)
 - `<fill>` can be placed inside of `<loot>`, since this is the only ID association seen so far 

# TODO
- [ ] Implement the Dynamic Filter `refill-trigger`
- [ ] Implement `weight` probability
- [ ] Fix `refill-interval`
- [ ] Make count range work
- [x] Fix `<any>`

# Examples
```xml
<lootables>
    <loot id="stuff">
        <!-- Always include these items -->
        <item material="sponge"/>
        <item material="stained_clay" damage="1"/>
        <!-- Include if filter matches opener of the container -->
        <maybe filter="only-red">
            <item material="stained clay" damage="14" amount="64"/>
            <item material="leather helmet" color="b3312c"/>
        </maybe>
        <!-- Choose one element at random -->
        <any>
            <item material="stone" damage="1"/>
            <item material="stone" damage="2"/>
            <item material="stone" damage="3"/>
        </any>
    </loot>
    <fill loot="stuff" refill-interval="10s" refill-clear="true"/>
</lootables>
```
```xml
<lootables>
    <loot>
        <item material="sponge"/>
        <item material="stained_clay" damage="1"/>
        <!-- new feature -->
        <fill refill-interval="10s" refill-clear="true"/>
    </loot>
</lootables>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Modules.java
+++ b/core/src/main/java/tc/oc/pgm/api/Modules.java
@@ -56,6 +56,8 @@ import tc.oc.pgm.killreward.KillRewardMatchModule;
 import tc.oc.pgm.killreward.KillRewardModule;
 import tc.oc.pgm.kits.KitMatchModule;
 import tc.oc.pgm.kits.KitModule;
+import tc.oc.pgm.loot.LootMatchModule;
+import tc.oc.pgm.loot.LootModule;
 import tc.oc.pgm.modes.ObjectiveModesMatchModule;
 import tc.oc.pgm.modes.ObjectiveModesModule;
 import tc.oc.pgm.modules.ArrowRemovalMatchModule;
@@ -250,6 +252,7 @@ public interface Modules {
     register(
         WorldBorderModule.class, WorldBorderMatchModule.class, new WorldBorderModule.Factory());
     register(SpawnerModule.class, SpawnerMatchModule.class, new SpawnerModule.Factory());
+    register(LootModule.class, LootMatchModule.class, new LootModule.Factory());
 
     // MapModules that are also MatchModules
     register(WorldTimeModule.class, WorldTimeModule.class, new WorldTimeModule.Factory());
--- a/core/src/main/java/tc/oc/pgm/blockdrops/BlockDrops.java
+++ b/core/src/main/java/tc/oc/pgm/blockdrops/BlockDrops.java
@@ -5,10 +5,12 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.material.MaterialData;
+import tc.oc.pgm.kits.Kit;
 
 /** The result of breaking a block */
 public class BlockDrops {
   public final ImmutableMap<ItemStack, Double> items; // probability -> item
+  public final Kit kit;
   public final int experience;
   public final @Nullable MaterialData replacement;
   public final @Nullable Float fallChance;
@@ -17,12 +19,14 @@ public class BlockDrops {
 
   public BlockDrops(
       Map<ItemStack, Double> items,
+      Kit kit,
       int experience,
       @Nullable MaterialData replacement,
       @Nullable Float fallChance,
       @Nullable Float landChance,
       @Nullable Double fallSpeed) {
     this.items = ImmutableMap.copyOf(items);
+    this.kit = kit;
     this.experience = experience;
     this.replacement = replacement;
     this.fallChance = fallChance;
@@ -37,6 +41,8 @@ public class BlockDrops {
         + this.replacement
         + " items.size="
         + this.items.size()
+        + " kit="
+        + this.kit
         + " experience="
         + this.experience
         + " fallChance="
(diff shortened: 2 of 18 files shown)


CoWinkKeyDinkInc:
After taking a closer look at maps that have used `<lootables>` in their XML like [`skywars.xml`](https://github.com/MCResourcePile/overcast-includes/blob/master/gamemodes/skywars.xml) and [`survival-games.xml`](https://github.com/MCResourcePile/overcast-includes/blob/master/gamemodes/survival-games.xml), I've made some changes to let `<any>`, `<maybe>` and `<option>` be nested inside each other. The examples in the XML are quite long but here is a snippet:
```xml
<loot id="stuff">
        <maybe filter="pre-5">
            <any count="2..3">
                <item material="stone sword"/>
                <item material="wood sword"/>
            </any>
        </maybe>
        <any count="0..3">
            <item material="iron sword"/>
            <item material="bow"/>
            <item material="iron helmet"/>
            <any>
                <item material="arrow" amount="4"/>
                <item material="arrow" amount="6"/>
            </any>
...
```

Need some help implementing things like `refill-trigger`, weight probability, etc. Especially `<cache>` which is a lot.

This XML is an example of everything working at the moment:
```xml
<lootables>
    <loot id="stuff">
        <!-- Always include these items -->
        <item material="sponge"/>
        <item material="stained_clay" damage="1"/>
        <!-- Include if filter matches opener of the container -->
        <maybe filter="only-red">
            <item material="stained clay" damage="14" amount="64"/>
            <item material="leather helmet" color="b3312c"/>
        </maybe>
        <maybe filter="only-blue">
            <item material="stained clay" damage="11" amount="64"/>
            <item material="leather helmet" color="334CB2"/>
        </maybe>
        <!-- Choose one element at random -->
        <any>
            <item material="stone" damage="0"/>
            <item material="stone" damage="1"/>
            <item material="stone" damage="2"/>
            <item material="stone" damage="3"/>
            <item material="stone" damage="4"/>
            <item material="stone" damage="5"/>
            <item material="stone" damage="6"/>
        </any>
        <!-- Choose three elements at random -->
        <any count="3">
            <item material="wood" damage="0"/>
            <item material="wood" damage="1"/>
            <item material="wood" damage="2"/>
            <item material="wood" damage="3"/>
            <item material="wood" damage="4"/>
            <item material="wood" damage="5"/>
        </any>

    </loot>
    <fill loot="stuff" refill-interval="10s" refill-clear="true"/>
</lootables>
```

Pablete1234 (changes requested):
Please implement this as it used to be implemented in original PGM. A complete reimplementation of the feature will both destroy backwards compatibility in subtle ways you won't even know and will cause more trouble having to re-implement all the logic (and NAMING) that already was figured out.



https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/loot/FillListener.java#L50

The module should take in a List of `Filler`s and a list of `Cache`s. (list of loot should not even be passed in, neither should anyItems, options, anyChildren, or any other property)

The [`Filler`](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/loot/Filler.java) has loot, filter, refill trigger, refill interval, and refill clear.

The [`Loot`](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/loot/Loot.java) has an item stack composition.

The [`Composition`](https://github.com/OvercastNetwork/ProjectAres/tree/master/PGM/src/main/java/tc/oc/pgm/compose) and [`CompositionParser`](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/compose/CompositionParser.java) have an `elements` method that returns a stream of elements (for lootables, that is, items) based on a filter query, and internally are implemented as all of the none, maybe, all, any and option stuff, which can be nested within eachother endlessly but it is of no concern to lootables. Lootables will just ask for elements with a query and will get a response (a stream of items), and will put them in the chest

Finally, the [`Cache`](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/loot/Cache.java) simply has a region, and a filter.

All the logic for lootables is governed by the [`FillListener`](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/loot/FillListener.java) which in our case will just be part of `LootablesMatchModule`, but such loot match module does not have to contain any kind of logic for treating any, all, maybe, etc.
Also the logic for bossbars (LootCountdown) does not exist and should not exist. Lootables do not create countdowns, countdowns are a different PGM feature (which we do not have yet) but it is NOT this PR's problem to solve, create a different PR to reimplement countdowns.

CoWinkKeyDinkInc:
Unfortunately this is far beyond my knowledge. I'm really sorry to disappoint everyone here.

PR #1021, opened 2022-07-05
Update dynamic filters immediately on sneak

OhPointFive:
Currently, if you sneak, dynamic filters don't update when players sneak immediately — only after they've moved a block. Dynamic filters happening immediately when you press shift may be desirable — perhaps a glass spawn platform where players can press shift to activate a portal to drop down where they want.

This PR makes it so that when a player presses shift, dynamic filters are updated immediately.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/PlayerMovementFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/PlayerMovementFilter.java
@@ -1,8 +1,9 @@
 package tc.oc.pgm.filters;
 
+import com.google.common.collect.ImmutableList;
 import java.util.Collection;
-import java.util.Collections;
 import org.bukkit.event.Event;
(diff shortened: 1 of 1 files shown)


KingOfSquares:
Thanks for the PR, it seems like you did not sign off your commit though, which is required for this repo. Are you able to correct that?

OhPointFive:
> Thanks for the PR, it seems like you did not sign off your commit though, which is required for this repo. Are you able to correct that?

Oh sorry, I had heard this was no longer required. Last time I tried to set that up I ran into some issues but I'll give it another go.

Edit: I think I have signed it now, let me know if I didn't do it right.

Pablete1234 (approved):
~~As simon pointed out you'll need to signoff the commit, but otherwise~~ lgtm

Electroid (approved):

PR #1022, opened 2022-07-05
Add score changes on dynamic filters

OhPointFive:
This allows changing scores based on dynamic filters for teams like so:

```xml
<score>
    <on team="some-team-filter" score="2"/>
</score>
```

and for the team a player belongs to like so:

```xml
<score>
    <on player="some-player-filter" score="5"/>
</score>
```

I'm not too familiar with the naming conventions so I imagine my choices will not be the best. Let me know if they should be renamed.

Also, I did this rather on a whim, so I didn't realize it might make more sense as part of #1016. I thought I'd submit this just in case we want it in the score module too.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/score/ScoreMatchModule.java
@@ -30,6 +30,7 @@ import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.module.exception.ModuleLoadException;
 import tc.oc.pgm.api.party.Competitor;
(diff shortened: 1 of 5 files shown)


Pablete1234:
I believe this one should be part of action/triggers, should be an action that can run in the context of either a player or a team. You can use q scope optional arg or derive it from context, similar in concept to how scope-switch gets its outer

OhPointFive:
I agree with Pablo here — I'm not sure if I'm supposed to close this myself or wait for someone else to look at it.

Electroid:
@OhPointFive Actions and triggers have been merged, so you can rebase this branch and change it to use it.

Pablete1234:
On-top of actions and triggers, variables were also merged.

See my comment on #1047 about how i think this could be implemented in a more flexible way that allows arbitrary score calculations. I'm going to close this as it's been stale for a couple months, feel free to open a new PR using actions & variables, or ask how to go about it if you want to take on it yourself.

Issue #976, opened 2022-03-19
Implement contested-decay attribute for control points


lefckntea:
If i have decay on a hill, it decays once "nobody dominates" the hill (documentation)
There should be a difference between the that state where no one is on the hill and where both/multiple teams are on the point because gameplay wise they're very different and it wouldnt make sense for the hill to decay when it's contested.

See my map Sunstroke played in last conquest tournament 
if it's somehow possible with pgm magic then sure let me knowbut the xml is probably already hacky enough

lefckntea:
add like a decay-rule similar to capture-rule idk man

Pablete1234:
My proposal is to add `contested-rate` which defaults to `decay-rate` (for backwards compatibility reasons), and is used whenever the point is not being dominated by a particular team, but is being contested by multiple teams.

At the moment a control point goes at different rates if:
 - A team is capturing (normal rate), eg: Neutral -> Captured
 - A team is restoring their ownership (at recovery-rate), eg: Owned (50% neutral) -> Owned
 - Naturally restoring ownership (at decay-rate), eg: Owned (50% neutral) -> Owned
 - Decaying towards neutral cause no one is on the point (at owned-decay-rate), eg: Owned -> Neutral
 
 The main difference between recovery-rate and decay-rate is simply that the former occurs when a team dominates (they have the players on point) while the later occurs when no team dominates, it be either because no one is on the point, or because capture-rule says no one dominates (eg, exclusive rule and multiple teams, or no team has a majority or a lead).
 
 This `contested-rate` would fill the gap of further separating decaying because no one is on point, vs decaying because there's a conflict between multiple teams capturing (respecting the exclusive/majority/lead rules)

PR #1023, opened 2022-07-06
Add contested-rate to control points

OhPointFive:
As per the proposal in #976, this PR adds contested-rate to control points. This allows hills to decay at a different rate when they are contested than when they are simply unoccupied.

I only implemented it for the `exclusive` capture-rule, since I don't think it's clear how it would work with the other capture rules.

For example:
```xml
<control-points decay-rate="1.25" recovery-rate="1" contested-rate="0" owned-decay-rate="23" capture-time="23s" show-progress="true" neutral-state="true" points="0" required="false">
    <control-point id="hill" name="Hill" capture="hill-region" progress="hill-region"/>
</control-points>
```
(from a modified version of Moonshine)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
@@ -311,6 +311,7 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
       }
     }
 
+    boolean contested = false;
     int lead = 0;
     if (leader != null) {
       lead = playerCounts.get(leader);
@@ -319,6 +320,7 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
       switch (this.definition.getCaptureCondition()) {
         case EXCLUSIVE:
           if (defenderCount > 0) {
+            contested = true;
             lead = 0;
           }
           break;
@@ -336,9 +338,9 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
     }
 
     if (lead > 0) {
-      this.dominateAndFireEvents(leader, calculateDominateTime(lead, duration));
+      this.dominateAndFireEvents(leader, calculateDominateTime(lead, duration), contested);
     } else {
-      this.dominateAndFireEvents(null, duration);
+      this.dominateAndFireEvents(null, duration, contested);
     }
   }
 
@@ -347,12 +349,13 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
    * team can be null, which means no team is dominating the point, which can cause the state to
    * change in some configurations.
    */
-  private void dominateAndFireEvents(@Nullable Competitor dominantTeam, Duration dominantTime) {
+  private void dominateAndFireEvents(
+      @Nullable Competitor dominantTeam, Duration dominantTime, boolean contested) {
     Duration oldCapturingTime = this.capturingTime;
     Competitor oldCapturingTeam = this.capturingTeam;
     Competitor oldControllingTeam = this.controllingTeam;
 
-    this.dominate(dominantTeam, dominantTime);
+    this.dominate(dominantTeam, dominantTime, contested);
 
     if (oldCapturingTeam != this.capturingTeam || !oldCapturingTime.equals(this.capturingTime)) {
       this.match.callEvent(new CapturingTimeChangeEvent(this.match, this));
@@ -404,7 +407,7 @@ public class ControlPoint extends SimpleGoal<ControlPointDefinition>
    * <p>If there is no neutral state, then the point is always either being captured by a specific
    * team, or not being captured at all.
    */
-  private void dominate(Competitor dominantTeam, Duration dominantTime) {
+  private void dominate(Competitor dominantTeam, Duration dominantTime, boolean contested) {
     if (!this.capturable || !TimeUtils.isLongerThan(dominantTime, Duration.ZERO)) {
       return;
     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
```diff
@@ -130,6 +132,12 @@ public static ControlPointDefinition parseControlPoint(
             "capture rule",
             ControlPointDefinition.CaptureCondition.EXCLUSIVE);
 
+    if (attrContested != null
+        && captureCondition != ControlPointDefinition.CaptureCondition.EXCLUSIVE) {
+      throw new InvalidXMLException(
+          "Contested rate may only apply to exclusive capture rule.", attrContested);
+    }
```
Pablete1234: Why is this a limitation? I see it has been implemented in a way that would not support anything except exclusive, however, i see no reason why it should be this way.
On both MAJORITY and LEAD there are cases where contesting can occur, some examples:
 - MAJORITY or LEAD:  Team A has 2 players, team B has 2 players
 - MAJORITY: Team A has 5 players, Team B has 3 players, team C has 2 players

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -311,6 +311,7 @@ protected void tickCapture(Duration duration) {
       }
     }
 
+    boolean contested = false;
     int lead = 0;
```
Pablete1234: I fail to see how `lead` and `contested` represent any different state in this whole block of code.
A lead of > 0, means there's a leader team, so, the point is *not* contested, and that already takes into account the rules required to make it work.
There is one exception to this, and that is when there's no one on the point at all, so lead is naturally 0 in `playerCounts.get(leader)`.

But that essentially means, contested is always equal to `playerCounts.get(leader) > 0  && lead <= 0` (evaluated *after* the switch statement), or alternatively simply, `leader != null  && lead <= 0` because leader is only set if *some player* is dominating.

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -336,9 +338,9 @@ protected void tickCapture(Duration duration) {
     }
 
     if (lead > 0) {
-      this.dominateAndFireEvents(leader, calculateDominateTime(lead, duration));
+      this.dominateAndFireEvents(leader, calculateDominateTime(lead, duration), contested);
     } else {
-      this.dominateAndFireEvents(null, duration);
+      this.dominateAndFireEvents(null, duration, contested);
     }
```
```java
    if (lead > 0) {
      this.dominateAndFireEvents(leader, calculateDominateTime(lead, duration), false);
    } else {
      this.dominateAndFireEvents(null, duration, leader != null);
    }
```
If there's a lead > 0, the point is *not* contested, so just use false.
If there's not a lead, that means the point is contested *if* anyone is on it (meaning, there'd be some leader team)

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -404,7 +407,7 @@ private void dominateAndFireEvents(@Nullable Competitor dominantTeam, Duration d
    * <p>If there is no neutral state, then the point is always either being captured by a specific
    * team, or not being captured at all.
    */
-  private void dominate(Competitor dominantTeam, Duration dominantTime) {
+  private void dominate(Competitor dominantTeam, Duration dominantTime, boolean contested) {
```
Pablete1234, line 408: I'd introduce an illegal argument exception thrown if dominant team is not null and contested is true, as that should never happen. I'd also double-check that it is in fact never happening

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -414,10 +417,13 @@ private void dominate(Competitor dominantTeam, Duration dominantTime) {
       // Point is owned and must go through the neutral state before another team can capture it
       if (dominantTeam == this.controllingTeam) {
         // owner is recovering the point
-        recover(dominantTeam, dominantTime);
+        recover(dominantTeam, dominantTime, contested);
       } else if (dominantTeam != null) {
         // non-owner is uncapturing the point
-        uncapture(dominantTeam, dominantTime);
+        uncapture(dominantTeam, dominantTime, contested);
```
```suggestion
        uncapture(dominantTeam, dominantTime);
```
We can statically deduce that contested must always be false here, if there's a dominant team there is not a contesting situation occuring. Uncapture can just call back on dominate with remaining time passing in a false

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -414,10 +417,13 @@ private void dominate(Competitor dominantTeam, Duration dominantTime) {
       // Point is owned and must go through the neutral state before another team can capture it
       if (dominantTeam == this.controllingTeam) {
         // owner is recovering the point
-        recover(dominantTeam, dominantTime);
+        recover(dominantTeam, dominantTime, contested);
```
Pablete1234: For this one i'm not totally sure if contested *can* be removed, because specifically for recovering towards a neutral state both dominant and controlling would be null, and it could be coming from a situation where the point is contested or not contested. 
However, i also believe in no point would it ever make a difference, because in that situation the hill would simply fall-back to neutral, and after reaching neutral, it would in no case be able to fall to a contested scenario.

So i'm pretty confident that contested can be removed here too, and simply recurse with a `false` because there is no situation where that can make a difference.

core/src/main/java/tc/oc/pgm/controlpoint/ControlPoint.java
```diff
@@ -432,7 +438,10 @@ private void dominate(Competitor dominantTeam, Duration dominantTime) {
         capture(dominantTime);
       } else if (dominantTeam != null) {
         // non-capturing team is dominate, so regress capturing team's progress
-        recover(dominantTeam, dominantTime);
+        recover(dominantTeam, dominantTime, contested);
```
Pablete1234: Here is the other use for recover, here we do know for sure it serves no purpose

Pablete1234 (changes requested):
There's a few things that can be simplified away, and removing them will also add support for lead and majority rules.

OhPointFive:
Mame beat me to fixing everything here (thanks mame!). Thanks for catching all these ways to improve the logic — I believe all the changes you requested are done. I'll re-request a review once it's been tested a bit.

Pablete1234 (approved):
Code-wise looks good now, but of course double check that it all works as intended in tests

Electroid (approved):
Nice, small improvement! Thanks for working on it 🎉

PR #1029, opened 2022-07-10
Structures

KingOfSquares:
put this draft up to get feedback on design and naming(please help)

~~Not yet ready for merge since I want to test more(structures are scary)~~


core/src/main/java/tc/oc/pgm/util/BudgetWorldEdit.java
```diff
@@ -0,0 +1,89 @@
...
+import org.bukkit.util.Vector;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.region.Region;
+import tc.oc.pgm.snapshot.SnapshotMatchModule;
+
+/**
+ * Utils to save, remove and paste blocks in some {@link Region} in some {@link Match} using the
+ * {@link SnapshotMatchModule} as memory.
+ */
+public class BudgetWorldEdit {
```
applenick: Fantastic name 😆

core/src/main/java/tc/oc/pgm/api/module/ModuleGraph.java
```diff
@@ -79,18 +83,22 @@ protected F getFactory(Class<? extends M> key, @Nullable Class<? extends M> requ
...
       }
-      throw new ModuleLoadException(
-          key,
-          "Required by "
-              + requiredBy.getSimpleName()
-              + " but not registered in "
-              + getClass().getSimpleName());
+      factory =
+          dependencyOnlyFactories.get(
+              key); // Only check this if a module is required as a dependency for some other module
```
Pablete1234:
```suggestion
      // Only check this if a module is required as a dependency for some other module
      factory = dependencyOnlyFactories.get(key);
```

core/src/main/java/tc/oc/pgm/map/MapFactoryImpl.java
```diff
@@ -57,7 +57,8 @@ public class MapFactoryImpl extends ModuleGraph<MapModule, MapModuleFactory<? ex
   private FeatureDefinitionContext features;
 
   public MapFactoryImpl(Logger logger, MapSource source) {
-    super(Modules.MAP); // Do not copy to avoid N copies of the factories
+    super(
+        Modules.MAP, Modules.MAP_DEPENDENCY_ONLY); // Do not copy to avoid N copies of the factories
```
Pablete1234:
```suggestion
    super(Modules.MAP, Modules.MAP_DEPENDENCY_ONLY); // Don't copy, avoid N factory copies
```
let's try to avoid formatter sending this to the next line

core/src/main/java/tc/oc/pgm/snapshot/SnapshotMatchModule.java
```diff
@@ -75,6 +73,38 @@ public MaterialData getOriginalMaterial(int x, int y, int z) {
...
 
+  /**
+   * Get the original material data for a {@code region}.
+   *
+   * @param region the region to get block states from
+   * @param materialFilter filters which blocks that will be included in the returned list
+   */
+  public List<MaterialDataWithLocation> getOriginalMaterialData(
+      Region region, Predicate<Material> materialFilter) {
+    List<MaterialDataWithLocation> materialData = new LinkedList<>();
```
Pablete1234: Linked lists are significantly worse memory-wise than an array list, furthermore, this whole method looks bad for performance if anyone decides to use a well-sized region.

Consider returning a Stream or Iterable of material data with location instead, which does not require allocating all at once and instead allows the caller to request "step by step"

core/src/main/java/tc/oc/pgm/snapshot/SnapshotMatchModule.java
```diff
@@ -75,6 +73,38 @@ public MaterialData getOriginalMaterial(int x, int y, int z) {
...
+      if (chunkVector == null
+          || z >> 4 != chunkVector.getChunkZ()
+          || x >> 4 != chunkVector.getChunkX()) {
+        chunkVector = ChunkVector.ofBlock(blockVector);
+        snapshot = chunkSnapshots.get(chunkVector);
+      }
+      BlockVector chunkPos = chunkVector.worldToChunk(x, y, z);
+      MaterialData data = snapshot.getMaterialData(chunkPos);
+      if (!materialFilter.test(data.getItemType())) continue;
+      materialData.add(new MaterialDataWithLocation(data, x, y, z));
```
Pablete1234: For performance reasons i'd reuse this return object, in the iterable, so every time you call .next() you just mutate the object with new vector (location) and data, and return that

core/src/main/java/tc/oc/pgm/structure/Dynamic.java
```diff
@@ -0,0 +1,69 @@
+package tc.oc.pgm.structure;
+
+import org.bukkit.util.Vector;
+import tc.oc.pgm.api.feature.Feature;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.util.BudgetWorldEdit;
+
+public class Dynamic implements Feature<DynamicDefinition> {
```
Pablete1234:
```suggestion
public class DynamicStructure implements Feature<DynamicDefinition> {
```

core/src/main/java/tc/oc/pgm/structure/Dynamic.java
```diff
@@ -0,0 +1,69 @@
...
+
+  private final Match match;
+  final StructureDefinition structureDefinition;
+  final Vector offset;
```
Pablete1234:
```suggestion
  private final StructureDefinition structureDefinition;
  private final Vector offset;
```

make them accessible via getter if they're needed elsewhere

core/src/main/java/tc/oc/pgm/structure/Dynamic.java
```diff
@@ -0,0 +1,69 @@
...
+  private final DynamicDefinition dynamicDefinition;
+
+  // Since the passive filter can skip placing the structure,
+  // we need to keep track of whether its placed or not if we
+  // want to avoid unnecessary clears.
+  boolean placed;
```
Pablete1234:
```suggestion
  private boolean placed;
```
same thing here, if others want to check this, do so via getter

core/src/main/java/tc/oc/pgm/util/BudgetWorldEdit.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * Saves blocks to the memory
+   *
+   * @param region the region to save
+   * @param includeAir if air should be considered as a part of the structure
+   * @param clearAfterSave whether to clear the blocks inside this region after saving them
+   * @param match the match to save blocks in
+   */
+  public static void saveBlocks(
```
Pablete1234: I believe these methods should be available within SnapshotMatchModule, not as static methods in an util.

They clearly interact with a match and the snapshot match module specifically, so anyone calling should have a dependency on snapshot match module to begin with, and be able to call directly. It also feels generally weird to call a static method to "save", like, save blocks to where? If it's calling snapshot match module directly, it's much clearer what actually is going on.

The implementation of the methods doesn't necessarily need to be in snapshot match module directly, but maybe re-routed thru it instead.

So SnapshotMatchModule#saveBlocks(region, includeAir, clearAfterSave) internally calls:
`this.blockManipulator.saveBlocks(region, includeAir, clearAfterSave)`
such blockManipulator being an instance of this BudgetWorldEdit, created within SnapshotMatchModule's constructor and saved as a local property. It prevents needing to pass around a match too.

core/src/main/java/tc/oc/pgm/util/MaterialDataWithLocation.java
```diff
@@ -0,0 +1,18 @@
+package tc.oc.pgm.util;
+
+import org.bukkit.material.MaterialData;
+
+/** Simple util class to store {@link MaterialData} for a specific location. */
+public class MaterialDataWithLocation {
+  public final MaterialData data;
+  public final int x;
+  public final int y;
+  public final int z;
```
Pablete1234: maybe use a BlockVector instead of x y z, means you don't need to unpack it elsewhere

core/src/main/java/tc/oc/pgm/util/MaterialDataWithLocation.java
```diff
@@ -0,0 +1,18 @@
+package tc.oc.pgm.util;
+
+import org.bukkit.material.MaterialData;
+
+/** Simple util class to store {@link MaterialData} for a specific location. */
+public class MaterialDataWithLocation {
```
Pablete1234: maybe LocatedMaterialData would be shorter and simpler?

core/src/main/java/tc/oc/pgm/structure/DynamicDefinition.java
```diff
@@ -0,0 +1,83 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import org.bukkit.util.Vector;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.feature.FeatureInfo;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
+
+@FeatureInfo(name = "dynamic")
+public class DynamicDefinition extends SelfIdentifyingFeatureDefinition {
```
Pablete1234:
```suggestion
@FeatureInfo(name = "dynamic-structure")
public class DynamicStructureDefinition extends SelfIdentifyingFeatureDefinition {
```
using dynamic as a global name is a bad idea, easy to confuse with dynamic filters

core/src/main/java/tc/oc/pgm/structure/StructureMatchModule.java
```diff
@@ -0,0 +1,54 @@
...
+  public void load() throws ModuleLoadException {
+    List<Dynamic> dynamics =
+        dynamicDefinitions.stream().map(d -> new Dynamic(d, match)).collect(Collectors.toList());
+
+    FilterMatchModule fmm = match.needModule(FilterMatchModule.class);
+    for (Dynamic dynamic : dynamics) {
+      match.getFeatureContext().add(dynamic);
+      final DynamicDefinition dynamicDef = dynamic.getDefinition();
+      fmm.onChange(
+          MatchPlayer.class, // TODO make this scope adjustable
```
Pablete1234: ?

core/src/main/java/tc/oc/pgm/structure/StructureModule.java
```diff
@@ -0,0 +1,115 @@
...
+    public StructureModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      if (factory.getProto().isOlderThan(MapProtos.FILTER_FEATURES)) return null; // TODO
+
+      final Map<String, StructureDefinition> structures = new HashMap<>();
+      for (Element elStruct :
+          XMLUtils.flattenElements(doc.getRootElement(), "structures", "structure")) {
+        final StructureDefinition definition =
+            new StructureDefinition(
+                elStruct.getAttribute("id").getValue(), // TODO Default?
```
Pablete1234: If someone doesn't set this, this is a NPE.

`XMLUtils.getRequiredAttribute("id").getValue()` if you want it required, otherwise, handle it somehow

core/src/main/java/tc/oc/pgm/structure/StructureModule.java
```diff
@@ -0,0 +1,115 @@
...
+    return ImmutableList.of(FilterMatchModule.class, SnapshotMatchModule.class);
+  }
+
+  public static class Factory implements MapModuleFactory<StructureModule> {
+
+    @Nullable
+    @Override
+    public StructureModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      if (factory.getProto().isOlderThan(MapProtos.FILTER_FEATURES)) return null; // TODO
```
Pablete1234: todo what exactly?

core/src/main/java/tc/oc/pgm/structure/DynamicScheduler.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.PriorityQueue;
+import java.util.Queue;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchScope;
+
+/**
+ * We need this to make sure that dynamics are placed and cleared in definition order, and that
+ * clears happen before placements.
+ */
+public class DynamicScheduler {
```
Pablete1234: given how little work the scheduler does, and how little work the structure match module does, i'd consider just joining the logic of both together

core/src/main/java/tc/oc/pgm/structure/Dynamic.java
```diff
@@ -0,0 +1,69 @@
...
+
+  Dynamic(DynamicDefinition dynamicDefinition, Match match) {
+    this.match = match;
+    this.dynamicDefinition = dynamicDefinition;
+    this.structureDefinition = this.dynamicDefinition.getStructureDefinition();
+    this.offset = this.findOffset();
```
Pablete1234: Given how StructureDefinition has a non-final field, i think you could also move this offset into Dynamic and lazy-load it. As long as it doesn't allow for any arbitrary change after creation (eg: no setter) and it's just a result of lazy-loading, it's fine if definitions have an *slight* amount of mutable state

core/src/main/java/tc/oc/pgm/structure/DynamicScheduler.java
```diff
@@ -0,0 +1,55 @@
...
+  void queueClear(Dynamic dynamic) {
+    placeQueue.remove(dynamic);
+    clearQueue.add(dynamic);
+    schedule();
+  }
+
+  private void schedule() {
+    match.getExecutor(MatchScope.LOADED).submit(this::process);
+    // match.getScheduler(MatchScope.LOADED)
+    //  .debounceTask(this::process);
```
Pablete1234: ?

core/src/main/java/tc/oc/pgm/structure/DynamicScheduler.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public void process() {
+    while (!clearQueue.isEmpty()) {
+      final Dynamic dynamic = clearQueue.poll();
+      dynamic.clear();
```
Pablete1234:
```suggestion
      clearQueue.poll().clear();
```

core/src/main/java/tc/oc/pgm/structure/DynamicScheduler.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    while (!placeQueue.isEmpty()) {
+      Dynamic dynamic = placeQueue.poll();
+      dynamic.place();
```
Pablete1234:
```suggestion
      placeQueue.poll().place();
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/structure/StructureModule.java
KingOfSquares: idk I think that's leftover

core/src/main/java/tc/oc/pgm/structure/DynamicScheduler.java
KingOfSquares: forgot to remove

core/src/main/java/tc/oc/pgm/util/MaterialDataWithLocation.java
KingOfSquares: Good idea, when I look at the methods I use it makes more sense to keep the Vector than to store the coordinate integers

core/src/main/java/tc/oc/pgm/util/MaterialDataWithLocation.java
KingOfSquares: I mean its shorter but less obvious, if you think less characters is good I'll change it but its just a class used in this one case

core/src/main/java/tc/oc/pgm/structure/StructureMatchModule.java
KingOfSquares: this was while we talked about the scope attribute, I guess it makes more sense to add action stuff for structures at this point?

core/src/main/java/tc/oc/pgm/util/BudgetWorldEdit.java
Electroid: Agreed.

KingOfSquares:
btw, this branch is blocked until I finish lootables(commissioned work)

core/src/main/java/tc/oc/pgm/structure/StructureMatchModule.java
Pablete1234: I think it may simply be absurd to listen to match player for a structure change, this should be listening on a match scope. One player passes the filter and it places? then another stops matching and it removes? what is this?

It should be using a match scope, if the filter you provide is not able to respond to a match then it simply cannot be used there, this is somewhat related to #1050 , where objective modes were listening for onChange without a scope at all (meaning, it'd use `Filterable.class` as a scope) which broke things, as you could use a filter that wouldn't react well to all filterables (eg: it required a player to check something, and couldn't work with just a match query)

Pablete1234 (dismissed):

Electroid (approved):
Thanks for working on this! Excited to bring it back!

PR #1030, opened 2022-07-10
Add kill-entities action

OhPointFive:
This adds a `kill-entities` action. You can use it like so:

```xml
<actions>
    <kill-entities id="some-id" filter="some-filter"/>
    <trigger filter="some-trigger" trigger="some-id" scope="match"/>
</actions>
```

This is being used for Moonshine, and likely similar CTP maps


Electroid:
What if we just called this `<kill>`?
Also, what's the behavior when the entity is a participant, an observer? Can we test it works properly?

calcastor:
Currently seems to break entity tracking for spawners. This is currently being used on a map that has spawners defined with `max-entities="1"`. If filtering to kill dropped items, it seems that if there is an item spawned at the spawner and that item gets cleared, the count for `max-entities` does not change, so the spawner will not function.

Pablete1234:
Consider adding a `players` true/false to know if players (participants) may be included, although impl-wise it should just be a wrapping filter:

```xml
<actions>
  <kill id="kill-everything" players="true"/>
  <kill id="kill-players" players="true" filter="player-entities"/>
  <kill id="kill-all-non-player" filter="player-entities"/>
  <kill id="kill-items" filter="item-entities"/>
</actions>
<filters>
  <participating id="player-entities"/>
  <entity id="item-entities">item</entity>
</filters>
```

If players="true" is included, then the filter is kept as-is, and if no filter is passed it defaults to `allow` (always)

If players="false" (or not set at all), then the filter is wrapped around not-player-and-(your filter) like this:
`<kill id="kill-items" filter="item-entities"/>` 
becomes:
```xml
<kill id="kill-items" filter="no-player-and-item-entities"/>
<and id="no-player-and-item-entities">
  <not><entity>player</entity></not>
  <filter id="item-entites/>
</and>
```
This transformation is not to be taken literally! (obv it won't have an id, and it won't be part of the XML at all, it's just that you'll create an and filter with those 2 other filters inside)

Regarding the killing of non-interacting players (eg: observers, or already dead players), the behavior should be double-checked to make sure they remain unaffected, and it doesn't add to their death stats.

Electroid:
I’d rather not have a separate boolean, what about this:

```xml
<kill filter=“everything”/>
<kill player-filter=“red-team”/>
```

Observers are always exempt, even if included in the filter. If both player-filter and filter are defined, both must pass.

Pablete1234:
If you just set filter, then players are not affected
If you use player-filter, then players can be affected (but won't be if they're not matched by the other filter too)

This behavior is a bit confusing over simply `players="true/false"` which tells if players are affected

OhPointFive:
> Currently seems to break entity tracking for spawners. This is currently being used on a map that has spawners defined with `max-entities="1"`. If filtering to kill dropped items, it seems that if there is an item spawned at the spawner and that item gets cleared, the count for `max-entities` does not change, so the spawner will not function.

~~It seems that Entity#remove() doesn't fire any events for removing the item so there's no way to detect that this happens. I'm not sure the best workaround for this — should I make an event when an entity is manually removed? Change items despawn times to immediate? Somehow notify spawners? @ electroid & pablo for what you guys would prefer.~~

Edit: I found a workaround for now, not sure if it's the best one.

Also I'll add the player filtering however you guys want, just let me know when it's decided.

core/src/main/java/tc/oc/pgm/events/ItemRemovedByActionEvent.java
```diff
@@ -0,0 +1,28 @@
+package tc.oc.pgm.events;
+
+import org.bukkit.entity.Item;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+public class ItemRemovedByActionEvent extends Event {
+  private Item entity;
+
+  public ItemRemovedByActionEvent(Item entity) {
```
Pablete1234: You should not be implementing your own event for this.

Either find a method to despawn the item naturally, or just call a bukkit item despawn event.

core/src/main/java/tc/oc/pgm/action/actions/KillEntitiesAction.java
```diff
@@ -0,0 +1,36 @@
...
+import org.bukkit.entity.Item;
+import org.bukkit.entity.Player;
+import org.bukkit.event.entity.ItemDespawnEvent;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.filters.query.EntityQuery;
+
+public class KillEntitiesAction extends AbstractAction<Match> {
+
+  Filter filter;
```
Pablete1234:
```suggestion
  private final Filter filter;
```

Pablete1234 (changes requested):

Pablete1234 (dismissed):

Pablete1234:
Rebased to fix merge conflicts.

@Electroid could you approve and merge this PR? It's been ready for a couple weeks now

Electroid (approved):
Hoping we would go with `<kill>`, but this is good too. Thanks for working on this 🎉

PR #1032, opened 2022-07-13
Implement spread-teammates for spawns

calcastor:
There are some map concepts where to ability to set `spread="true"` for team spawns is desirable. This PR permits that functionality as a separate attribute titled `spread-teammates`.


Pablete1234 (changes requested):
This behavior is working as intended, ie: spread=true means to not spawn you close to enemy players.

If you're looking for *also* not spawning you close to teammates, you need another xml property (defaulting to original behavior).

Maybe something like spread-teammates=true/false (default to false, which is original behavior)

Pablete1234 (approved):

Electroid (approved):

PR #1038, opened 2022-07-27
Interactive Shops 🛒

### LINKED ISSUE #463: Custom villager trades
Vanilla villager shops get crowded once you have more then a few items and updating villagers over version-changes and map updates can be a pain. To provide a better overview of the shop items i'd suggest that a shop is divided into sub-shops that are displayed in an inventory with items as icons (like with the classes module), those sub-shops would then open a regular villager shop with the trades. Maybe every mob (maybe even every entity, would be funny if ur team mates are a shop xD) could be used as a shop and shops could be controlled by filters.

What would also be cool is if u could spawn in a shop with a spawn-egg, that would be useful, you could sell better shops or take your shop a strategic position.

Maybe it could work like this you spawn the mobs that are the shops with the spawner module and then with something like and entity-mods module one could add shops to a certain kind of entity.

applenick:
# Interactive Shops

Time for a brand new PGM feature, interactive shops! 🤯 For years map authors have had to rely on the traditional method of villager trading, setting up custom trades can be tiresome and non-intuitive for those without the experience. With this feature we bring a sleek and simple way to add shop functionality to a map. 

Additionally, this feature will benefit immensely from the pending includes PR #1028. Once both are merged, many new possibilities unlock such as standardized shop menus across similar gamemodes. 

## Shops
Shops are composed of a few different elements, at the highest level we have the shop itself. The only attribute you can customize is its `name`, the name of a shop must be unique and will be displayed in the title of the GUI menu. Within each shop section you can define a number of categories and then each category can hold up to 28 items. 

### XML Example:
```xml
<shops>
    <!-- Our first example is of a generic item shop -->
    <shop id="item-shop" name="`bItems">
     <!-- You can have an multiple categories, each category will hold up to 28 items -->
      <category id="weapons" name="`cWeapons" material="stone sword"> 

        <!-- Free item (you can exclude price if desired) -->
        <item material="wooden sword" amount="1" price="0"/>
  
        <!-- Currency & Price in same line -->
        <item material="stone sword" amount="1" price="3" currency="emerald" />

        <!-- Payment as sub-element but only using a single currency -->
        <item name="Cool TNT" material="tnt" amount="5">
            <payment price="5" currency="gunpowder"/>
        </item>

        <!-- Multiple Payment requirements-->
        <item material="diamond sword" amount="1">

            <payment price="10" currency="diamond"/>
            <payment price="20" currency="stick"/>
        </item>

      </category>
      <category id="food" name="`aFood" material="apple">
        <item material="apple" name="Tasty Apples" amount="100" price="10" currency="gold ingot"/>
      </category>
    </shop>

    <!-- You can define as many <shops> as needed, here's a second example -->
    <shop id="MagicShop" name="`5Magic Shop">
      <category id="magic" name="`dWands">
        <!-- Item where the kit is defined. Icon will render as a stick but kit can give anything -->
        <item name="Magic Wand" material="stick" price="100" currency="blaze powder" kit="magic-kit">
      </category>
    </shop>
</shops>
```

## Shopkeepers
A shopkeeper is an entity that spawns at match start. These entities wil not die and are unable to be moved. Clicking a shopkeeper will open their related GUI. Each shopkeeper may only be associated with a single shop type, but each shop type can have an unlimited number of shopkeepers. Plus, there's no limitations in place to how many players can have a shop open at the same time.

### XML Example:
```xml
<!-- There's no limit to how many shopkeepers can be in each match, as long as each <shop> has at least 1 keeper -->
<shopkeepers>
    <!-- A shop keeper for the "Items" shop (defaults to villager) -->
    <shopkeeper name="`5Purple `7Item Shop" shop="Items">
       <point yaw="0" pitch="10">-88.5,13,-103.5</point>
    </shopkeeper>

    <!-- You can define a mob attribute to set the entity type -->
    <shopkeeper name="`4Red `dMagic Shop" mob="Witch" shop="MagicShop" />
      <point>-90.5,54,-100.5</point>
    </shopkeeper>
</shopkeepers>
```

## Screenshots

General layout of interactive shops. You'll notice how categories are along the top, purchasable items below. 

Clicking a category will change the purchasable item selection

Hovering over an item will display whether you can purchase the item or not

Example of item with custom name and lore

## Feedback
As always I've tested this extensively and should work as intended. Though with a feature this flexible, I'm sure there's likely to be situations I've not thought of. Please let me know if there's any feedback and I'll be happy to see what can be improved.  

Also while not exactly custom villager trades, I believe this PR would be enough to close #463 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/MultiTradeMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/MultiTradeMatchModule.java
@@ -24,9 +24,10 @@ public class MultiTradeMatchModule implements MatchModule, Listener {
     this.logger = PGM.get().getGameLogger();
   }
 
-  @EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
+  @EventHandler(priority = EventPriority.NORMAL)
   public void onInteract(PlayerInteractEntityEvent event) {
     if (event.getRightClicked() instanceof Villager) {
+      if (event.isCancelled()) return;
       // Fallback to once-at-a-time trading if multi trade does not work
       if (ok) {
         event.setCancelled(true);
--- a/core/src/main/java/tc/oc/pgm/modules/MultiTradeMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/MultiTradeMatchModule.java
@@ -24,10 +24,9 @@ public class MultiTradeMatchModule implements MatchModule, Listener {
     this.logger = PGM.get().getGameLogger();
   }
 
-  @EventHandler(priority = EventPriority.NORMAL)
+  @EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
   public void onInteract(PlayerInteractEntityEvent event) {
     if (event.getRightClicked() instanceof Villager) {
-      if (event.isCancelled()) return;
       // Fallback to once-at-a-time trading if multi trade does not work
       if (ok) {
         event.setCancelled(true);
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -14,6 +14,7 @@ import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.shops.menu.Category;
 import tc.oc.pgm.shops.menu.Icon;
+import tc.oc.pgm.shops.menu.Payment;
 
 @FeatureInfo(name = "shops")
 public class Shop implements FeatureDefinition {
@@ -44,9 +45,9 @@ public class Shop implements FeatureDefinition {
 
   public boolean canPurchase(Icon icon, MatchPlayer buyer) {
     if (!buyer.getMatch().isRunning() || !buyer.isParticipating()) return false;
-    if (icon.getPrice() < 1) return true;
+    if (icon.isFree()) return true;
 
-    return buyer.getInventory().containsAtLeast(new ItemStack(icon.getCurrency()), icon.getPrice());
+    return icon.getPayments().stream().allMatch(p -> p.hasPayment(buyer.getInventory()));
   }
 
   public boolean purchase(Icon icon, MatchPlayer buyer) {
@@ -59,20 +60,21 @@ public class Shop implements FeatureDefinition {
       return false;
     }
 
-    // Remove items from inventory when item is not free
-    if (icon.getPrice() > 1) {
+    if (!icon.isFree()) {
       PlayerInventory inventory = buyer.getInventory();
-      int remaining = icon.getPrice();
-      for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
-        ItemStack item = inventory.getItem(slot);
-        if (item == null || item.getType() != icon.getCurrency()) continue;
-        if (item.getAmount() > remaining) {
-          item.setAmount(item.getAmount() - remaining);
-          inventory.setItem(slot, item);
-          remaining = 0;
-        } else {
-          inventory.setItem(slot, null);
-          remaining -= item.getAmount();
+      for (Payment payment : icon.getPayments()) {
+        int remaining = payment.getPrice();
+        for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
+          ItemStack item = inventory.getItem(slot);
+          if (item == null || item.getType() != payment.getCurrency()) continue;
+          if (item.getAmount() > remaining) {
+            item.setAmount(item.getAmount() - remaining);
+            inventory.setItem(slot, item);
+            remaining = 0;
+          } else {
+            inventory.setItem(slot, null);
+            remaining -= item.getAmount();
+          }
         }
       }
     }
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -21,12 +21,18 @@ public class Shop implements FeatureDefinition {
 
   private static Sound PURCHASE_SOUND = sound(key("fire.ignite"), Sound.Source.MASTER, 1f, 1.4f);
 
+  private final String id;
   private final String name;
   private final ImmutableList<Category> categories;
 
-  public Shop(String name, List<Category> categories) {
-    this.name = name;
+  public Shop(String id, String name, List<Category> categories) {
+    this.id = id;
+    this.name = name != null ? name : id;
     this.categories = ImmutableList.copyOf(categories);
+  }
+
+  public String getId() {
+    return id;
   }
 
   public String getName() {
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -9,30 +9,23 @@ import java.util.List;
 import net.kyori.adventure.sound.Sound;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
-import tc.oc.pgm.api.feature.FeatureDefinition;
-import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 import tc.oc.pgm.shops.menu.Category;
 import tc.oc.pgm.shops.menu.Icon;
 import tc.oc.pgm.shops.menu.Payment;
 
-@FeatureInfo(name = "shops")
-public class Shop implements FeatureDefinition {
+public class Shop extends SelfIdentifyingFeatureDefinition {
 
   private static Sound PURCHASE_SOUND = sound(key("fire.ignite"), Sound.Source.MASTER, 1f, 1.4f);
 
-  private final String id;
   private final String name;
   private final List<Category> categories;
 
   public Shop(String id, String name, List<Category> categories) {
-    this.id = id;
+    super(id);
     this.name = name != null ? name : id;
     this.categories = ImmutableList.copyOf(categories);
-  }
-
-  public String getId() {
-    return id;
   }
 
   public String getName() {
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -4,12 +4,15 @@ import static net.kyori.adventure.key.Key.key;
 import static net.kyori.adventure.sound.Sound.sound;
 import static net.kyori.adventure.text.Component.translatable;
 
-import com.google.common.collect.ImmutableList;
 import java.util.List;
-import net.kyori.adventure.sound.Sound;
+
 import org.bukkit.Material;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
+
+import com.google.common.collect.ImmutableList;
+
+import net.kyori.adventure.sound.Sound;
 import tc.oc.pgm.api.feature.FeatureDefinition;
 import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.player.MatchPlayer;
@@ -67,9 +70,20 @@ public class Shop implements FeatureDefinition {
 
     // Remove items from inventory
     PlayerInventory inventory = buyer.getInventory();
-    for (int i = 0; i < icon.getPrice(); i++) {
-      inventory.removeItem(new ItemStack(icon.getCurrency()));
+    int remaining = icon.getPrice();
+    for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
+      ItemStack item = inventory.getItem(slot);
+      if (item == null || item.getType() != icon.getCurrency()) continue;
+      if (item.getAmount() > remaining) {
+        item.setAmount(item.getAmount() - remaining);
+        inventory.setItem(slot, item);
+        remaining = 0;
+      } else {
+        inventory.setItem(slot, null);
+        remaining -= item.getAmount();
+      }
     }
+
     buyer.getBukkit().updateInventory();
     buyer.applyKit(icon.getKit(), true);
     buyer.playSound(PURCHASE_SOUND);
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -80,7 +80,7 @@ public class Shop extends SelfIdentifyingFeatureDefinition {
     }
 
     buyer.getBukkit().updateInventory();
-    buyer.applyKit(icon.getKit(), true);
+    icon.getAction().trigger(buyer);
     buyer.playSound(PURCHASE_SOUND);
     return true;
   }
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -66,7 +66,7 @@ public class Shop extends SelfIdentifyingFeatureDefinition {
         int remaining = payment.getPrice();
         for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
           ItemStack item = inventory.getItem(slot);
-          if (item == null || item.getType() != payment.getCurrency()) continue;
+          if (item == null || !payment.matches(item)) continue;
           if (item.getAmount() > remaining) {
             item.setAmount(item.getAmount() - remaining);
             inventory.setItem(slot, item);
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -1,0 +1,72 @@
+package tc.oc.pgm.shops;
+
+import static net.kyori.adventure.key.Key.key;
+import static net.kyori.adventure.sound.Sound.sound;
+import static net.kyori.adventure.text.Component.translatable;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import net.kyori.adventure.sound.Sound;
+import org.bukkit.Material;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.PlayerInventory;
+import tc.oc.pgm.api.feature.FeatureDefinition;
+import tc.oc.pgm.api.feature.FeatureInfo;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.shops.menu.Category;
+import tc.oc.pgm.shops.menu.Icon;
+
+@FeatureInfo(name = "shops")
+public class Shop implements FeatureDefinition {
+
+  private static Sound PURCHASE_SOUND = sound(key("fire.ignite"), Sound.Source.MASTER, 1f, 1.4f);
+
+  private final String name;
+  private final ImmutableList<Category> categories;
+
+  public Shop(String name, List<Category> categories) {
+    this.name = name;
+    this.categories = ImmutableList.copyOf(categories);
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public ImmutableList<Category> getCategories() {
+    return categories;
+  }
+
+  public boolean canPurchase(Icon icon, MatchPlayer buyer) {
+    Material currency = icon.getCurrency();
+    int price = icon.getPrice();
+
+    if (buyer.getMatch().isRunning() && buyer.isParticipating()) {
+      PlayerInventory inventory = buyer.getInventory();
+      return inventory.containsAtLeast(new ItemStack(currency), price);
+    }
+
+    return false;
+  }
+
+  public boolean purchase(Icon icon, MatchPlayer buyer) {
+    if (!canPurchase(icon, buyer)) {
+      if (!buyer.getMatch().isRunning()) {
+        buyer.sendWarning(translatable("match.error.noMatch"));
+      } else {
+        buyer.sendWarning(translatable("shop.currency.insufficient"));
+      }
+      return false;
+    }
+
+    // Remove items from inventory
+    PlayerInventory inventory = buyer.getInventory();
+    for (int i = 0; i < icon.getPrice(); i++) {
+      inventory.removeItem(new ItemStack(icon.getCurrency()));
+    }
+    buyer.getBukkit().updateInventory();
+    buyer.applyKit(icon.getKit(), true);
+    buyer.playSound(PURCHASE_SOUND);
+    return true;
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -23,7 +23,7 @@ public class Shop implements FeatureDefinition {
 
   private final String id;
   private final String name;
-  private final ImmutableList<Category> categories;
+  private final List<Category> categories;
 
   public Shop(String id, String name, List<Category> categories) {
     this.id = id;
@@ -39,7 +39,7 @@ public class Shop implements FeatureDefinition {
     return name;
   }
 
-  public ImmutableList<Category> getCategories() {
+  public List<Category> getCategories() {
     return categories;
   }
 
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -76,6 +76,11 @@ public class Shop extends SelfIdentifyingFeatureDefinition {
             remaining -= item.getAmount();
           }
         }
+
+        // Should never happen, canPurchase checks for payment being available in the inventory
+        if (remaining > 0) {
+          throw new IllegalStateException("Player couldn't pay for their purchase.");
+        }
       }
     }
 
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -4,15 +4,12 @@ import static net.kyori.adventure.key.Key.key;
 import static net.kyori.adventure.sound.Sound.sound;
 import static net.kyori.adventure.text.Component.translatable;
 
+import com.google.common.collect.ImmutableList;
 import java.util.List;
-
+import net.kyori.adventure.sound.Sound;
 import org.bukkit.Material;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
-
-import com.google.common.collect.ImmutableList;
-
-import net.kyori.adventure.sound.Sound;
 import tc.oc.pgm.api.feature.FeatureDefinition;
 import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.player.MatchPlayer;
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -7,7 +7,6 @@ import static net.kyori.adventure.text.Component.translatable;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 import net.kyori.adventure.sound.Sound;
-import org.bukkit.Material;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
 import tc.oc.pgm.api.feature.FeatureDefinition;
@@ -44,15 +43,10 @@ public class Shop implements FeatureDefinition {
   }
 
   public boolean canPurchase(Icon icon, MatchPlayer buyer) {
-    Material currency = icon.getCurrency();
-    int price = icon.getPrice();
+    if (!buyer.getMatch().isRunning() || !buyer.isParticipating()) return false;
+    if (icon.getPrice() < 1) return true;
 
-    if (buyer.getMatch().isRunning() && buyer.isParticipating()) {
-      PlayerInventory inventory = buyer.getInventory();
-      return inventory.containsAtLeast(new ItemStack(currency), price);
-    }
-
-    return false;
+    return buyer.getInventory().containsAtLeast(new ItemStack(icon.getCurrency()), icon.getPrice());
   }
 
   public boolean purchase(Icon icon, MatchPlayer buyer) {
@@ -65,19 +59,21 @@ public class Shop implements FeatureDefinition {
       return false;
     }
 
-    // Remove items from inventory
-    PlayerInventory inventory = buyer.getInventory();
-    int remaining = icon.getPrice();
-    for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
-      ItemStack item = inventory.getItem(slot);
-      if (item == null || item.getType() != icon.getCurrency()) continue;
-      if (item.getAmount() > remaining) {
-        item.setAmount(item.getAmount() - remaining);
-        inventory.setItem(slot, item);
-        remaining = 0;
-      } else {
-        inventory.setItem(slot, null);
-        remaining -= item.getAmount();
+    // Remove items from inventory when item is not free
+    if (icon.getPrice() > 1) {
+      PlayerInventory inventory = buyer.getInventory();
+      int remaining = icon.getPrice();
+      for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
+        ItemStack item = inventory.getItem(slot);
+        if (item == null || item.getType() != icon.getCurrency()) continue;
+        if (item.getAmount() > remaining) {
+          item.setAmount(item.getAmount() - remaining);
+          inventory.setItem(slot, item);
+          remaining = 0;
+        } else {
+          inventory.setItem(slot, null);
+          remaining -= item.getAmount();
+        }
       }
     }
 
--- a/core/src/main/java/tc/oc/pgm/shops/Shop.java
+++ b/core/src/main/java/tc/oc/pgm/shops/Shop.java
@@ -6,6 +6,7 @@ import static net.kyori.adventure.text.Component.translatable;
 
 import com.google.common.collect.ImmutableList;
 import java.util.List;
+import java.util.stream.Collectors;
 import net.kyori.adventure.sound.Sound;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.PlayerInventory;
@@ -20,7 +21,7 @@ public class Shop extends SelfIdentifyingFeatureDefinition {
   private static Sound PURCHASE_SOUND = sound(key("fire.ignite"), Sound.Source.MASTER, 1f, 1.4f);
 
   private final String name;
-  private final List<Category> categories;
+  private final ImmutableList<Category> categories;
 
   public Shop(String id, String name, List<Category> categories) {
     super(id);
@@ -34,6 +35,12 @@ public class Shop extends SelfIdentifyingFeatureDefinition {
 
   public List<Category> getCategories() {
     return categories;
+  }
+
+  public List<Category> getVisibleCategories(MatchPlayer player) {
+    return categories.stream()
+        .filter(c -> c.getFilter().query(player).isAllowed())
+        .collect(Collectors.toList());
   }
 
   public boolean canPurchase(Icon icon, MatchPlayer buyer) {
--- a/core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
+++ b/core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
@@ -2,9 +2,6 @@ package tc.oc.pgm.shops;
 
 import static tc.oc.pgm.util.bukkit.BukkitUtils.colorize;
 
-import com.google.common.collect.ImmutableList;
-import java.util.List;
-import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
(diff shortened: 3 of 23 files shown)


core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -0,0 +1,72 @@
...
+        buyer.sendWarning(translatable("shop.currency.insufficient"));
+      }
+      return false;
+    }
+
+    // Remove items from inventory
+    PlayerInventory inventory = buyer.getInventory();
+    for (int i = 0; i < icon.getPrice(); i++) {
+      inventory.removeItem(new ItemStack(icon.getCurrency()));
+    }
```
Pablete1234: seems rather inneficient, since it's an n ^ 2. if someone has to pay 100 for something (and they happen to have the item on their last slot, bottom-right corner, but above hotbar) you're gonna iterate their entire inventory (36 slots) searching for things to remove 100 times.

Instead, remove whole item stacks at once if needed, or subtract the amount directly, something along the lines of:
```suggestion
    int remaining = icon.getPrice();
    for (int slot = 0; slot < inventory.size() && remaining > 0; slot++) {
      ItemStack item = inventory.get(slot);
      if (item == null || item.getType() != icon.getCurrency()) continue;
      if (item.getAmount() > remaining) {
        item.setAmount(item.getAmount() - remaining);
        inventory.setSlot(slot, item);
        remaining = 0;
      } else {
        inventory.setSlot(slot, null);
        remaining -= item.getAmount();
      }
    }
    if (remaining > 0) // WHOOPSIE?
      // consider paying-back the player and following up with an insufficient currency error
```

core/src/main/java/tc/oc/pgm/shops/ShopMatchModule.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @EventHandler
+  public void onShopKeeperAltDamage(VehicleDamageEvent event) {
+    if (isKeeper(event.getVehicle())) {
+      event.setCancelled(true);
+    }
+  }
+
+  @EventHandler
+  public void onShopKeeperAltEnter(VehicleEnterEvent event) {
```
Pablete1234, line 82: You can't move a shopkeeper with vehicles, got it

But what about pushing them with pistons?

Is there no entity move event or similar? may be more robust to listen for that and just teleport the entity back if that happens, but i'm not sure if it's a thing,

(i see you use a freeze entity method, so chances are it may not get moved server-side, 1.7 clients may still see it move because noAi may not exist for them, may be worth looking into)

core/src/main/java/tc/oc/pgm/shops/ShopMatchModule.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @EventHandler
+  public void onShopKeeperAltMove(VehicleEntityCollisionEvent event) {
+    if (isKeeper(event.getVehicle())) {
+      event.setCancelled(true);
+    }
+  }
+
+  private Shop getShopFromEntity(Entity entity) {
```
Pablete1234, line 103:
```suggestion
  private @Nullable Shop getShopFromEntity(Entity entity) {
```
wanna get those nice IDE hints when trying to do potentially nullable operations

core/src/main/java/tc/oc/pgm/shops/ShopMatchModule.java
```diff
@@ -0,0 +1,110 @@
...
+    if (isKeeper(entity)) {
+      MetadataValue meta = entity.getMetadata(ShopKeeper.METADATA_KEY, PGM.get());
+      if (meta.asString() != null) {
+        Optional<Shop> shop =
+            shops.stream().filter(s -> s.getName().equalsIgnoreCase(meta.asString())).findAny();
```
Pablete1234: We do not match by name over here, we use **id**s for that

By the way, if you properly register shops by their id, you need no equals ignore case stuff, simply get it by id

`return shops.get(meta.asString());`

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+import tc.oc.pgm.util.material.Materials;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.XMLUtils;
+
+public class ShopModule implements MapModule {
+
+  private static final Collection<MapTag> TAGS =
+      ImmutableList.of(new MapTag("shops", "Shops", false, true));
+
+  private final Set<Shop> shops = Sets.newHashSet();
```
Pablete1234: Turn it into an id-indexed map
```suggestion
  private final Map<String, Shop> shops;
```

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+  private final Set<ShopKeeper> shopKeepers = Sets.newHashSet();
```
Pablete1234: Make both of these immutable, and passed in the constructor, we don't want these to change after being created by the factory

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  public static class Factory implements MapModuleFactory<ShopModule> {
+
+    @Override
+    public ShopModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      ShopModule module = new ShopModule();
+      KitParser kitParser = factory.getKits();
+      SetMultimap<String, KeeperData> locations = parseShopKeepers(doc.getRootElement(), logger);
```
Pablete1234: I'd say parse shop keepers after you have parsed shops, that way you don't need any of the context "resolution"

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+
+      for (Element shop : XMLUtils.flattenElements(doc.getRootElement(), "shops")) {
+        Attribute shopName = XMLUtils.getRequiredAttribute(shop, "name");
```
Pablete1234:
```suggestion
        String id = XMLUtils.getRequiredAttribute(shop, "id").getValue();
        String name = shop.getAttributeValue("name");
```
Make name optional, id required

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+        List<Category> categories = Lists.newArrayList();
+
+        for (Element category : XMLUtils.getChildren(shop, "category")) {
+          Attribute categoryName = XMLUtils.getRequiredAttribute(category, "name");
+          Attribute categoryMaterial = XMLUtils.getRequiredAttribute(category, "material");
+          Material material = parseMaterial(categoryMaterial, category);
```
Pablete1234:
```suggestion
          String categoryName = XMLUtils.getRequiredAttribute(category, "name").getValue();
          Material material = XMLUtils.parseMaterial(Node.fromRequiredAttr(category, "material"));
```

I'd consider just parsing category as an item, just like icons. That allows adding lore, or other extra stuff with little effort. If you need the "name" for identification purposes, use an id for the category (which you may auto-generate), not a name.

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+          List<Icon> icons = parseIcons(category, kitParser, logger);
+
+          categories.add(new Category(categoryName.getValue(), material, icons));
+        }
+
+        if (categories.isEmpty()) {
+          throw new InvalidXMLException("At least one <category> is required per shop", shop);
+        }
+
+        Shop shopInstance = new Shop(shopName.getValue(), categories);
```
Pablete1234:
```suggestion
        Shop shopInstance = new Shop(shopName, categories);
```
It's redundant to "parse" as an Attribute then have to do getValue everywhere, simply parse to String directly, and use the String

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+            .map(
+                data ->
+                    new ShopKeeper(
+                        data.getName(),
+                        data.getLocation(),
+                        data.getYaw(),
+                        data.getPitch(),
+                        data.getType(),
+                        shopInstance))
+            .forEach(keeper -> module.shopKeepers.add(keeper));
```
Pablete1234: Shops should not be in charge of shopkeepers.

You may have defined a shop with no shop keepers, and that's (while useless) fine
Imagine an import defines 2 or 3 shop templates you *may* or *may not* use, there's no need to enforce a keeper for each.

Additionally, as i suggested a bit earlier, i'd just parse shopkeepers after shops, so you ensure first you have the list of all available shops, then each shop keeper is guaranteed to have the shop available (if no shop by the id exists, then its an XML error)

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    public Float getPitch() {
+      return pitch;
+    }
+
+    public Class<? extends Entity> getType() {
+      return type;
+    }
+  }
```
Pablete1234:
```suggestion
```
Unneeded if you parse things in the right order. Even if you wanted to keep the current order, you should be using XMLFeatureReference so that you can create the shopkeeper *with the shop* and the shop just gets resolved later. This kind of data-holder-in-factory is just too dirty

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+
+  private static Material parseMaterial(Attribute attribute, Element el)
+      throws InvalidXMLException {
+    if (attribute == null) return null;
+    Material type = Materials.parseMaterial(attribute.getValue());
+    if (type == null || (type == Material.AIR)) {
+      throw new InvalidXMLException("Invalid material type '" + attribute.getValue() + "'", el);
+    }
+    return type;
+  }
```
Pablete1234:
```suggestion
```
Unneeded, there's a util for this in XMLUtils already

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+    Integer price =
+        XMLUtils.parseNumber(XMLUtils.getRequiredAttribute(icon, "price"), Integer.class);
+    ItemStack item = kits.parseItem(icon, false);
+
+    Kit kit = null;
+    Attribute kitAttr = icon.getAttribute("kit");
+    if (kitAttr != null) {
+      kit = kits.parseKitProperty(icon, "kit", null);
+    } else {
+      kit = new ItemKit(Maps.newHashMap(), Lists.newArrayList(item));
```
Pablete1234: I'd suggest you just parse the kit as a player-scoped Action instead, and in this case, create a dummy action that is to just give the player the item. (Keep in mind, a kit is still a valid action, so kits can still be used just the same)

A kit will "intelligently merge" with current inventory, and that means if you already have the item it won't be given again, as well as a bunch of other side effects. An example of this is kill rewards, which also just give the item not a kit

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  private static SetMultimap<String, KeeperData> parseShopKeepers(Element root, Logger logger)
+      throws InvalidXMLException {
+    SetMultimap<String, KeeperData> shopkeepers = HashMultimap.create();
+
+    for (Element shopkeeper : XMLUtils.flattenElements(root, "shopkeepers")) {
+      Attribute locationAttr = XMLUtils.getRequiredAttribute(shopkeeper, "location");
+      Attribute yawAttr = XMLUtils.getRequiredAttribute(shopkeeper, "yaw");
+      Attribute pitchAttr = XMLUtils.getRequiredAttribute(shopkeeper, "pitch");
```
Pablete1234: This is essentially what a `PointProvider` gives, look into `PointParser` and just parse this as a child, similar to how spawns do it, it will save you the head-ache of having to parse location, yaw, pitch yourself, and also provide more flexibility: eg angle (which faces the location provided) or spawning the shop in a "random position within the region" capability. Usually, this will just be a <point>x,y,z</point> tho. Another example that uses this is flag posts, which also use a point provider (and can spawn in a whole region, or a union of points, but generally just use a single point)

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,217 @@
...
+      Attribute typeAttr = XMLUtils.getRequiredAttribute(shopkeeper, "type");
```
Pablete1234:
```suggestion
      Shop shop = shops.get(XMLUtils.getRequiredAttribute(shopkeeper, "shop").getValue());
```
if it's null, throw an xml error. Shops should be a map. Keep in mind you must not use state within the Factory class, so you need to pass the Map<String, Shop> it thru as param to the function

core/src/main/java/tc/oc/pgm/shops/menu/ShopMenu.java
```diff
@@ -0,0 +1,202 @@
...
+          getViewer().getBukkit().updateInventory();
+        });
+  }
+
+  private ClickableItem getPageItem(Player player, int page, boolean next) {
+    return ClickableItem.of(
+        getPageIcon(page, next),
+        c -> {
+          getInventory().open(player, page);
+          this.highlightSlot = 1;
```
Pablete1234: this looks arbitrary? is there pagination here? i thought the max was 28 items because there is no pagination, only category -> items

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/shops/ShopMatchModule.java
applenick: Getting the shopkeeper entity to stay still was tricky! The cool thing is we're able to use entities such as minecarts for shops. So that's why the related vehicle events were included. The NMS freeze hack was used due to being unable to find any entity move event last time I checked. There's a `VehicleMoveEvent` but the event is not cancelable, so opted to cancel the collision event. Open to suggestions if you know of a better method to handle this, but after testing everything seemed normal for a 1.18 client. 

Good point about the 1.7 client difference, will have to test that and see what happens.

applenick:
ToDo: (so we don't forget)
- [x] support free items
- [x] complex payment
- [x] `filter` attribute to change item visibility

Complex payment example:
```xml
<item ...>
  <payment currency="a" price="1"/>
  <payment currency="b" price="1"/>
</item>
```

Electroid:
> ToDo: (so we don't forget)
> 
> * `filter` attribute to change item visibility
> * support free items
> * complex payments
> 
> Complex payment example:
> 
> ```
> <item ...>
>   <payment currency="a" price="1"/>
>   <payment currency="b" price="1"/>
> </item>
> ```

Should payments be defined in filter style?

```xml
<!-- 1 Gold -->
<item ...>
   <payment currency="gold"/>
</item>

<!-- 1 Gold AND 1 Iron -->
<item ...>
   <all>
     <payment currency="gold"/>
     <payment currency="iron"/>
   </all>
</item>

<!-- 1 Gold AND 1 Iron, OR 2 Emerald -->
<item ...>
   <any>
     <all>
       <payment currency="gold"/>
       <payment currency="iron"/>
     </all>
     <payment currency="emerald" price="2"/>
   </any>
</item>
```

applenick:
Really cool idea! Once concern though is finding a clean format for the cost lore. 

At the moment cost looks like this:


Here's my proposal for how the format would look with multiple currency requirements:
```
Cost:
✔ 3 Emerald
X 10 Diamond 
```
We'll use the symbols to let users know if they meet the required conditions. Though I'm not sure the best way to format it for the third example you gave. Any suggestions?

Pablete1234:
> Should payments be defined in filter style?
> ```xml
> 
> <item ...>
>    <any>
>      <all>
>        <payment currency="gold"/>
>        <payment currency="iron"/>
>      </all>
>      <payment currency="emerald" price="2"/>
>    </any>
> </item>
> ```
I thought about it but I think this adds alot more complexity than it solves, we cannot straight rely on filters for this functionality because we need to actually go and remove the items, if you have multiple ways of paying up which is the preferred?
If you want <any> functionality you can have the item listed twice, one sold for emerald, the other sold for gold+iron, the user can buy one or the other

applenick:
Here's what a free item looks like:
Screen Shot 2022-08-03 at 1 25 49 AM

XML:
```xml
<item material="wood" amount="1" price="0"/>
```
When the price of an item is below 1, then `currency` is no longer considered a required attribute. I'm thinking perhaps we should have the parsing default price to 0, so that in an instance where price is missing then we can just assume the item is free.

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -0,0 +1,87 @@
...
+import org.bukkit.inventory.PlayerInventory;
+import tc.oc.pgm.api.feature.FeatureDefinition;
+import tc.oc.pgm.api.feature.FeatureInfo;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.shops.menu.Category;
+import tc.oc.pgm.shops.menu.Icon;
+import tc.oc.pgm.shops.menu.Payment;
+
+@FeatureInfo(name = "shops")
+public class Shop implements FeatureDefinition {
```
Pablete1234: use SelfIdentifyingFeatureDefinition and pass the id as super(id) in the constructor

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -0,0 +1,87 @@
...
+          if (item == null || item.getType() != payment.getCurrency()) continue;
+          if (item.getAmount() > remaining) {
+            item.setAmount(item.getAmount() - remaining);
+            inventory.setItem(slot, item);
+            remaining = 0;
+          } else {
+            inventory.setItem(slot, null);
+            remaining -= item.getAmount();
+          }
+        }
```
Pablete1234, line 78: i'd handle what happens if remaning is > 0 in here, although in theory impossible, i can see it happening.
If you define 2 payments of the same material, it will check (independently) each one to see if you have it, but then try to subtract both and potentially fail.

This should be some kind of pgm exception.

core/src/main/java/tc/oc/pgm/shops/ShopMatchModule.java
```diff
@@ -0,0 +1,108 @@
...
+import tc.oc.pgm.api.match.MatchModule;
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.events.ListenerScope;
+import tc.oc.pgm.shops.menu.ShopMenu;
+
+@ListenerScope(MatchScope.RUNNING)
+public class ShopMatchModule implements MatchModule, Listener {
+
+  private Match match;
```
Pablete1234:
```suggestion
  private final Match match;
```

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,191 @@
...
+        if (categories.isEmpty()) {
+          throw new InvalidXMLException("At least one <category> is required per shop", shop);
+        }
+
+        Shop shopInstance = new Shop(shopId, shopName, categories);
+        factory.getFeatures().addFeature(shop, shopInstance);
+        shops.put(shopId, shopInstance);
+      }
+
+      keepers = parseShopKeepers(shops, doc.getRootElement(), pointParser, logger);
```
Pablete1234: given this is used in exactly one location, you're better off just inlining the function.

Also, keepers is defined to a sets.newHashSet() while this completely overrides it, making the original initialization unneeded

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,191 @@
...
+
+      if (location.isEmpty()) {
+        throw new InvalidXMLException("shopkeeper must have a location defined", shopkeeper);
+      }
+
+      if (!shops.containsKey(shopId)) {
+        throw new InvalidXMLException(
+            "No shop with id '" + shopId + "' could be found", shopkeeper);
+      }
+      Shop shop = shops.get(shopId);
```
Pablete1234:
```suggestion
      Shop shop = shops.get(shopId);
      if (shop == null) {
        throw new InvalidXMLException(
            "No shop with id '" + shopId + "' could be found", shopkeeper);
      }
```
no need to lookup the map twice

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,191 @@
...
+  private static List<Icon> parseIcons(Element category, KitParser kits, Logger logger)
+      throws InvalidXMLException {
+    List<Icon> icons = Lists.newArrayList();
+    for (Element icon : XMLUtils.getChildren(category, "item")) {
+      icons.add(parseIcon(icon, kits, logger));
+    }
+
+    if (icons.size() > Category.MAX_ICONS) {
+      throw new InvalidXMLException(
+          "Categories may only contain " + Category.MAX_ICONS + " icons", category);
```
Pablete1234:
```suggestion
          "Categories may only contain up to " + Category.MAX_ICONS + " icons", category);
```

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    if (icons.isEmpty()) {
+      throw new InvalidXMLException("At least one <item> is required per category", category);
```
Pablete1234: This is inconsistent with the above exception, one is calling them simply "icons" while the other uses "<item>". Decide on one and use it in both

core/src/main/java/tc/oc/pgm/modules/MultiTradeMatchModule.java
```diff
@@ -24,9 +24,10 @@ public MultiTradeMatchModule(Match match) {
     this.logger = PGM.get().getGameLogger();
   }
 
-  @EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
+  @EventHandler(priority = EventPriority.NORMAL)
   public void onInteract(PlayerInteractEntityEvent event) {
     if (event.getRightClicked() instanceof Villager) {
+      if (event.isCancelled()) return;
```
Electroid: What's the reasoning for this change?

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -0,0 +1,87 @@
...
+import tc.oc.pgm.shops.menu.Payment;
+
+@FeatureInfo(name = "shops")
+public class Shop implements FeatureDefinition {
+
+  private static Sound PURCHASE_SOUND = sound(key("fire.ignite"), Sound.Source.MASTER, 1f, 1.4f);
+
+  private final String id;
+  private final String name;
+  private final ImmutableList<Category> categories;
```
Electroid, line 24:
```suggestion
  private final List<Category> categories;
```

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public Shop(String id, String name, List<Category> categories) {
```
Electroid, line 26:
```suggestion
  public Shop(String id, String name, Collection<Category> categories) {
```

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public String getId() {
+    return id;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public ImmutableList<Category> getCategories() {
```
Electroid:
```suggestion
  public Collection<Category> getCategories() {
```

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
```diff
@@ -0,0 +1,65 @@
...
+import org.bukkit.entity.Entity;
+import org.bukkit.metadata.FixedMetadataValue;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.points.PointProvider;
+import tc.oc.pgm.util.nms.NMSHacks;
+
+public class ShopKeeper {
+
+  public static final String METADATA_KEY = "SHOP_KEEPER";
```
Electroid: Can this be `private`?

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private final String name;
+  private final ImmutableList<PointProvider> location;
```
Electroid: `
```suggestion
  private final Collection<PointProvider> location;
```

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
```diff
@@ -0,0 +1,65 @@
...
+  private final Class<? extends Entity> type;
+  private final Shop shop;
+
+  public ShopKeeper(
+      @Nullable String name,
+      ImmutableList<PointProvider> location,
```
Electroid:
```suggestion
      Collection<PointProvider> location,
```

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
```diff
@@ -0,0 +1,65 @@
...
+      Shop shop) {
+    this.name = name;
+    this.location = location;
+    this.type = type;
+    this.shop = shop;
+  }
+
+  @Nullable
+  public Location getLocation(Match match) {
+    // TODO: support multiple shopkeeper locations?
```
Electroid: I would think we should support this? Let's say I'm implementing this for Rift; I'd have 4 shop keepers, each would be on both sides.

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
```diff
@@ -0,0 +1,65 @@
...
+  public Class<? extends Entity> getType() {
+    return type;
+  }
+
+  public String getName() {
+    return name == null || name.isEmpty() ? ChatColor.GRAY + getShop().getId() : colorize(name);
+  }
+
+  public void spawn(Match match) {
+    if (getLocation(match) == null) return;
```
Electroid: I'd throw an `IllegalArgumentException` here.

Electroid (changes requested):
Nice work! Some minor nits, also want @Pablete1234 to take another pass.

core/src/main/java/tc/oc/pgm/modules/MultiTradeMatchModule.java
applenick: Reason behind this is to ensure that regular villager trades don't open when the player clicks an interactive shop keeper. 

~~I believe I mistakenly left the `event.isCancelled` in here, since the removal of `ignoreCancelled` should be all that's needed. Good catch!~~ Hm wait, I need to investigate this further 🤔

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
Pablete1234: ^ (see previous comment about this)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
Pablete1234: That is simply 2 shop keepers, for the same shop, i don't think it's relevant

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
Pablete1234: a location can be null and if it happens we should know somehow, but not illegal
make a map exception so it gets logged and mapdevs/devs see it in-game

core/src/main/java/tc/oc/pgm/shops/Shop.java
Pablete1234: Categories are really a list, you need them indexed. it doesn't make sense to generify them to collection

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
Pablete1234: I would try to simply change this to directly PointProvider, not a collection. Multiple shouldn't be supported anyways

core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
applenick: See my commit from the other day, I looked into multiple shopkeeper locations before I saw your comment and thought it was pretty useful. 

For example you can do something like this:
```xml
    <shopkeeper name="`5Purple `7Item Shop" mob="Villager" shop="items">
      <point yaw="-99" pitch="-22">-88.5,14,-103.5</point>
      <point yaw="90" pitch="-22">100.5,16,-103.5</point>
      <point yaw="180" pitch="-22">10,18,-10.5</point>
      <point yaw="0" pitch="-22">105,20,200.5</point>
    </shopkeeper>
```
and it'll spawn the same shop keeper at all those locations. While smaller maps may not find it useful, might be nice for larger scale ones.

core/src/main/java/tc/oc/pgm/shops/Shop.java
applenick: Very good point! Should the logic try to reverse itself or is throwing an exception all that should be done?

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
applenick: Could you elaborate a bit more on this when you have a chance? Still new to Actions, would this be simply parsing with the `ActionParser` or something else?

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -23,7 +23,7 @@ public class Shop implements FeatureDefinition {
 
   private final String id;
   private final String name;
-  private final ImmutableList<Category> categories;
+  private final List<Category> categories;
```
Pablete1234: Why no longer immutable?
It being a definition, i'd ensure this stays immutable

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
Pablete1234: Instead of parsing as a kit, parse as an action, the action parser probably either has, or should have a similar to "parseKitProperty"

Regarding making your own to just give items, yeah, you just create a class which implements Action, and define the method to "trigger"

core/src/main/java/tc/oc/pgm/shops/Shop.java
applenick: Whoops, my mistake! Must not have reverted from when I started to adjust based on Electroid's suggestions to make it generic. Latest commit reverts this.

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
applenick: Ah fantastic! Thanks for explaining, makes sense 👍 I'll see if I can get that done soon

applenick: @Pablete1234 there's `ActionParser#parseKitTrigger`, though believe it won't parse the same way as `parseKitProperty` does. Would it be fine to define a method like `parseKitPropertyTrigger` inside `ActionParser` or is there a different approach you'd suggest? 

As for the dummy action, would something like this work?
```java
public class GiveItemAction extends AbstractAction<MatchPlayer> {

  private final ItemStack item;

  public GiveItemAction(ItemStack item) {
    super(MatchPlayer.class);
    this.item = item;
  }

  @Override
  public void trigger(MatchPlayer player) {
    HashMap<Integer, ItemStack> overflow = player.getPlayer().getInventory().addItem(item);

    // Drop extra items at feet
    if (!overflow.isEmpty()) {
       // will be translatable in production
      player.sendWarning(text("Purchase could not fit in your inventory, so it's been dropped at your feet"));
      World world = player.getWorld();
      overflow.values().forEach(stack -> world.dropItemNaturally(player.getLocation(), stack));
    }
  }
}
```

applenick:
## New feature: customizing currency 💵 

### XML:
```xml
<!-- New `color` attribute changes currency text color -->
<item material="gold chestplate" amount="1" price="3" currency="emerald" color="green"/>

<!-- Usage with individual payments -->
<item material="gold chestplate" amount="1">
  <payment currency="emerald" price="1" color="green"/>
  <!-- Different currency can have different colors-->
  <payment currency="diamond" price="2" color="aqua"/>
</item>


<!-- New custom currency uses a child <item> under <payment> -->
<item material="diamond sword" amount="1">
  <payment price="1">
      </item>
  </payment>
</item>
```

core/src/main/java/tc/oc/pgm/shops/Shop.java
Pablete1234: Honestly up to you, in theory it shouldn't ever happen so an exception may be enough

But i do know one situation where this may happen, which is if payment is "1 diamond" and "3 diamonds". I know it's stupid, but it *can* happen, and if you have 3 diamonds in your inventory it will say you can buy, but you won't be able to pay up.

Another potential avenue for trouble is if you request the payment to be: "3 diamonds" (just material) and "1 super-special diamond" (custom item with name/lore or whatever).

You may have 4 diamonds in total in your inventory, but if the normal diamonds are searched for first (and they find the super-special-diamond and take that as payment first) then you won't be able to complete the payment.

I think all of those solve themselves if you simply disallow the same material being twice in the list of payments for an item, which may be a valid restraint imo.

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
Pablete1234: `ActionParser#parseReference` is what you're looking for, it will return an Action of the bound you ask for:

`Action<MatchPlayer> action = actionParser.parseReference(Node.forAttribute("action"), null, MatchPlayer.class)`
Keep in mind, this action **can** be a kit, because kits implement `Action<MatchPlayer>` so it all ties up together nicely.

Regarding that sample action, yeah it looks good, i'd actually make the "drop-overflow" an optional parameter in the constructor, and expose it as a newly available action:
```xml
<give drop-overflow="false">
    <item ...>
    <item ...>
</give-item>
```
This allows for giving items to a player but without following a kit's logic to giving items (where they won't give more if you already have the item)
It may be out of the scope of this but it'd be nice to have

core/src/main/java/tc/oc/pgm/shops/ShopModule.java
```diff
@@ -0,0 +1,235 @@
...
+                player.sendWarning(
+                    text(
+                        "Purchase could not fit in your inventory, so it's been dropped at your feet"));
+                World world = player.getWorld();
+                overflow
+                    .values()
+                    .forEach(stack -> world.dropItemNaturally(player.getLocation(), stack));
+              }
+            }
+          };
```
Pablete1234: Given that we now have #1049 , this could be using an item kit with repair-tools, deduct-tools, and deduct-items all set to false.
That still doesn't help with what-if you don't have space in your inventory to buy the item, but *maybe* that should be a separate thing? (eg: warning the player their inventory is full if they try to buy something they don't have space for?)

Pablete1234:
Pushed changes to hopefully finish-up the pr:

### Shop keepers
Shop keepers must use a single point provider, if you want multiple locations simply use multiple shop keepers:
```xml
<shopkeepers>
    <shopkeepers name="`5Purple `7Item Shop" mob="Villager" shop="items">
        <shopkeeper yaw="-99" pitch="-22">-88.5,14,-103.5</shopkeeper>
        <shopkeeper yaw="90" pitch="-22">100.5,16,-103.5</shopkeeper>
        <shopkeeper yaw="180" pitch="-22">10,18,-10.5</shopkeeper>
        <shopkeeper yaw="0" pitch="-22">105,20,200.5</shopkeeper>
    </shopkeepers>
    
    <shopkeeper name="Other shop" shop="items" yaw="-99" pitch="-22">-88.5,14,-103.5</shopkeeper>
    <shopkeeper name="Yet another shop" shop="items">
        <cylinder yaw="-99" pitch="-22" base="-88.5,14,-103.5" radius="5" height="1"/>
    </shopkeeper>
</shopkeepers>
```
All of the shopkeepers are valid, they can have the point be themselves (directly a location defined within them, and optionally pitch & yaw) or have a sub-element point or other type of region within them. They must however have a single location provider, multiple inner points/regions are not allowed.


### Kit overflow handling
Historically pgm just did nothing if the items in the kit overflew and did not fit the inventory. Now kits support dropping those items in the ground, and/or sending the player a message when that happens (you don't need both, you can *just* send a message without dropping, or you can drop without a message too). You can also combine this with #1049 to ensure the items are given even if you already have them on your inventory.
```xml
<kit id="spawn" drop-overflow="true" overflow-warning="Woopsie, your inventory is full">
  <item .../>
</kit>
<kit id="spawn" drop-overflow="true">
  <item .../>
  <overflow-warning>{"translate": "shop.purchase.overflow"}</overflow-warning>
</kit>
```
Using the translate message `shop.purchase.overflow` will result in a `⚠ Purchase could not fit in your inventory, so it's been dropped at your feet` message, which is what the shop will utilize by default when buying single-items (non-kit/action selling).


## Note:
These changes have **NOT** been fully tested, someone needs to validate they work and potentially fix any derived issues

Pablete1234 (dismissed):

Pablete1234:
Update:
 - Tested and confirmed shops and shopkeeper changes are working as intended
 - Changed it so shopkeepers will be spawned on cycle, not on match start. Had to make it so chunk is loaded before spawning.
 - Made it so observers clicking shopkeepers don't get the default vanilla villager shop.
 
 It's been tested and I believe it's ready to merge, @Electroid

Electroid (approved):

PR #1043, opened 2022-08-18
Add filters for dead and alive states

Pugzy:
Add filter for checking if a player is alive or dead.

Dynamic so that it's rise can be used to apply a kit.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/matcher/player/PlayerStateFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/matcher/player/PlayerStateFilter.java
@@ -4,9 +4,11 @@ import com.google.common.collect.ImmutableList;
 import java.util.Collection;
 import org.bukkit.event.Event;
 import tc.oc.pgm.api.filter.query.PlayerQuery;
+import tc.oc.pgm.api.match.event.MatchPhaseChangeEvent;
 import tc.oc.pgm.api.player.MatchPlayer;
-import tc.oc.pgm.api.player.event.MatchPlayerDeathEvent;
-import tc.oc.pgm.spawns.events.PlayerSpawnEvent;
+import tc.oc.pgm.events.PlayerPartyChangeEvent;
+import tc.oc.pgm.spawns.events.ParticipantDespawnEvent;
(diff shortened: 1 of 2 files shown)


Pablete1234:
Personally i believe we should use `<interacting/>` instead of "alive", otherwise someone joining obs and re-joining should technically always have been alive, but you'd still want to re-give the kit once they join.

Currently the `<alive/>` filter is going to say obs aren't alive, which can be a bit misleading (same for dead, obs are neither dead nor alive)

core/src/main/java/tc/oc/pgm/filters/matcher/player/PlayerStateFilter.java
```diff
@@ -0,0 +1,31 @@
...
+
+  private final boolean alive;
+
+  public PlayerStateFilter(boolean alive) {
+    this.alive = alive;
+  }
+
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    return ImmutableList.of(PlayerSpawnEvent.class, MatchPlayerDeathEvent.class);
```
Pablete1234: I'd use `ParticipantSpawnEvent` and `ParticipantDespawnEvent`, they won't work for player going from dead to obs tho, which is why i'd consider just making it `<interacting/>` instead which should be fully covered by those 2 events

Pablete1234:
@Electroid i'd like your input/opinion here, how do you feel about alive/dead naming where observers are never considered neither dead nor alive (they will always deny the query)?

I think "interacting" may be a better name for "alive and on a team" (being more direct than just "alive" where one may think obs are alive, when they wouldn't be considered alive). Regardless it's not something good documentation can't solve (making it clear that depsite it being called "alive" observers are never considered alive)

Electroid:
@Pablete1234 Naming of alive and dead seems fine, though I would think observers would always abstain from both queries.

Pablete1234:
Abstaining isn't an option for dynamic filters, must always allow or deny to the query as long as it's the correct type of query (ie: MatchPlayer)

Pablete1234 (approved):

Electroid (approved):

PR #1044, opened 2022-08-19
Lootables

KingOfSquares:
Commissioned work by @TBG1000(Im sorry this took forever)

This PR Re-introduce lootables with hopefully 100% legacy compat(testing has yielded promising results) ~~and proposes a new proto version(omg!) that would change some lootable behavior.~~

~~In old PGM every filler would fill any matching container on the first inventory open without any trigger, in addition as long as a container was not opened only 1 "fill" could be stored and used at inventory open per container.~~

~~The new proposed proto version does not trigger fillers with a `refill-trigger` unless specified, and in addition if a filler with a `refill-trigger` triggers multiple times it will be saved up and filled as many times as it was triggered when the inventory is opened.~~(This idea was discarded from this PR at some point)


core/src/main/java/tc/oc/pgm/kits/Slot.java
```diff
@@ -60,10 +67,9 @@ public boolean isAuto() {
     return index < 0;
   }
 
-  // If we support more inventory types, we would probably widen the type accepted by this method
```
Pablete1234, line 63: Widen the type does not mean make the whole class generic

Simply change `HumanEntity` to `InventoryHolder` and call it a day, having to change the whole codebase to make Slot a generic class is unneeded

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,348 @@
...
+              .map(block -> ((InventoryHolder) block).getInventory()),
+          cache
+              .region()
+              .getChunkPositions()
+              .flatMap(cp -> Stream.of(cp.getChunk(match.getWorld()).getEntities()))
+              .filter(cache.region()::contains)
+              .filter(LootableMatchModule::isFillable)
+              .filter(entity -> !cache.filter().query(new EntityQuery(entity)).isDenied())
+              .map(entity -> ((InventoryHolder) entity).getInventory()));
+    }
```
Pablete1234:
```suggestion
      return cache
              .region()
              .getChunkPositions()
              .map(cp -> cp.getChunk(match.getWorld()))
              .flatMap(ch -> Stream.concat(
                  ch.getTileEntities().stream().map(BlockQuery::new),
                  ch.getEntities().stream().map(EntityQuery::new)))
              .filter(cache.region()::contains)
              .filter(LootableMatchModule::isFillable)
              .filter(q -> !cache.filter().query(q).isDenied())
              .map(q -> q.getInventory()),
    }
```

This can likely be generified in a way where BlockQuery and EntityQuery both implement some InventoryQuery so that you can perform the `getInventory` at the end, everything else just needs to support a query as parameter instead of a block-or-entity, which shouldn't be hard i assume

core/src/main/java/tc/oc/pgm/kits/Slot.java
KingOfSquares: The idea is that the different Slot subclasses can then require their own subclass of InventoryHolder, it would be weird for the Player class to accept InventoryHolder for example.

Pablete1234: I get the reason, but it's simply too much overhead (in terms of complexity) that provides little benefit

core/src/main/java/tc/oc/pgm/kits/Slot.java
```diff
@@ -116,8 +128,8 @@ protected Range<Integer> getAutoIndexRange() {
      * Convert a {@link PlayerInventory} slot index to a {@link Slot} object. Returns null if the
      * index is out of range.
      */
-    public static @Nullable Slot forIndex(int index) {
-      return byIndex.get(index);
+    public static @Nullable Player forIndex(int index) {
+      return (Player) byIndex.get(Player.class, index);
     }
```
Pablete1234: Why is the return type of slot relevant to the caller? i think these classes should essentially be irrelevant to the caller "as long as it works when using the slot" which, it should

```suggestion
    public static @Nullable Slot forIndex(int index) {
      return byIndex.get(Player.class, index);
    }
```

core/src/main/java/tc/oc/pgm/kits/Slot.java
```diff
@@ -197,6 +241,16 @@ protected Range<Integer> getAutoIndexRange() {
...
   }
 
+  public static Slot forInventoryIndex(Inventory inv, int index) {
+    for (Map.Entry<Class<? extends Inventory>, Class<? extends Slot>> entry :
+        byInventoryType.entrySet()) {
+      if (entry.getKey().isAssignableFrom(inv.getClass())) {
+        return byIndex.get(entry.getValue(), index);
+      }
+    }
+    throw new IllegalStateException("Weird inventory type " + inv);
```
Pablete1234:
```suggestion
    throw new UnsupportedOperationException("Unknown inventory type " + inv);
```

core/src/main/java/tc/oc/pgm/loot/Cache.java
```diff
@@ -0,0 +1,26 @@
...
+    this.filter = filter;
+  }
+
+  Region region() {
+    return this.region;
+  }
+
+  Filter filter() {
+    return this.filter;
+  }
```
Pablete1234:
```suggestion
  public Region region() {
    return this.region;
  }

  public Filter filter() {
    return this.filter;
  }
```

core/src/main/java/tc/oc/pgm/loot/FillableAndFiller.java
```diff
@@ -0,0 +1,39 @@
+package tc.oc.pgm.loot;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Objects;
+
+public class FillableAndFiller {
+
+  private final LootableMatchModule.Fillable fillable;
+  private final FillerDefinition filler;
```
Pablete1234: You may be better off using a Pair or immutable pair from guava or apache commons if either of those are available, otherwise it's fine to stay

core/src/main/java/tc/oc/pgm/loot/FillerDefinition.java
```diff
@@ -0,0 +1,116 @@
...
+    this.filter = filter;
+    this.refillTrigger = refillTrigger;
+    this.refillInterval = refillInterval;
+    this.clearBeforeRefill = clearBeforeRefill;
+    this.usesRefillTrigger = this.refillTrigger != null;
+    this.usesRefillInterval = this.refillInterval != TimeUtils.INFINITE_DURATION;
+  }
+
+  /** Items to fill with */
+  Composition<ItemStack> getLoot() {
```
Pablete1234:
```suggestion
  public Composition<ItemStack> getLoot() {
```
make all of the methods here public. I assume you copied them from legacy pgm where they were interfaces, which made them public by default

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,337 @@
...
+                  filler.getRefillTrigger(),
+                  m -> this.triggeredFillerInvalidator.accept(filler));
+        });
+  }
+
+  /**
+   * Return a predicate that applies a Filter to the given InventoryHolder, or null if the
+   * InventoryHolder is not something that we should be filling.
+   */
+  private static @Nullable Predicate<Filter> passesFilter(InventoryHolder holder) {
```
Pablete1234:
```suggestion
  private static @Nullable Predicate<Filter> getFilterPredicate(InventoryHolder holder) {
```
the wording `passesFilter` implies you're checking the filter, when you clearly aren't. You're just creating a predicate to see if it passes the filter

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+  public void onInventoryOpen(InventoryOpenEvent event) {
+    final MatchPlayer opener = this.match.getParticipant(event.getActor());
+    if (opener == null) return;
+
+    final Inventory inventory = event.getInventory();
+    final Predicate<Filter> passesFilter = passesFilter(inventory.getHolder());
+    if (passesFilter == null) return;
```
Pablete1234:
```suggestion
    final Predicate<Filter> filterPredicate = passesFilter(inventory.getHolder());
    if (filterPredicate == null) return;
```

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,337 @@
...
+      if (passesFilter.test(cache.region()) && passesFilter.test(cache.filter())) {
+        fillables.add(new FillableCache(cache));
+      }
+    }
+    // If the inventory is not in any Cache, just fill it directly
+    if (fillables.isEmpty()) {
+      fillables.add(new FillableInventory(inventory));
+    }
+
+    fillables.forEach(fillable -> fillable.fill(opener, fillers));
```
Pablete1234:
```suggestion
    // Find all Caches that the opened inventory is part of
    boolean foundCache = false;
    for (Cache cache : caches) {
      if (passesFilter.test(cache.region()) && passesFilter.test(cache.filter())) {
        foundCache = true;
        new FillableCache(cache).fill(opener, fillers);
      }
    }
    // If the inventory is not in any Cache, just fill it directly
    if (!foundCache) {
      new FillableInventory(inventory).fill(opener, fillers);
    }
```
Any reason why you'd need this the other way around? It does introduce a behavior change, mainly, that the tests for the filter are no longer done all before starting to fill, but it prevents needing the list.

If the order of operations is relevant, then keep it as is

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,337 @@
...
+
+  abstract class Fillable {
+
+    abstract Stream<Inventory> inventories();
+
+    void fill(MatchPlayer opener, List<FillerDefinition> fillers) {
+      // Build a short list of Fillers that are NOT cooling down from a previous fill
+      final List<FillerDefinition> coolFillers =
+          fillers.stream()
+              .filter(filler -> isFillerReady(this, filler))
```
Pablete1234: Given isFillerReady is only used from within Fillable, maybe move it in so you don't need to pass `this` as first parameter?
```suggestion
              .filter(this::isFillerReady)
```
Would allow you to also use a method ref here instead of a lambda

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,337 @@
...
+                                // apply item mods, and put it in the slot.
+                                if (immm != null) immm.applyRules(item);
+                                slots.remove(random.nextInt(slots.size())).putItem(item);
+                              }
+                            });
+                    int newValue =
+                        triggerHistory.row(this).compute(filler, (f, i) -> i == null ? 1 : ++i);
+                    repeat = filler.usesRefillTrigger() && newValue < triggeredFillers.get(filler);
+                  } while (repeat);
+                }
```
Pablete1234: I don't get why this loop is needed

1.12 pgm makes this logic work so that if there's a refill trigger, when it triggers it removes older filled at for that same filler, and doesn't seem to do anything else with it, what am i missing?

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+  }
+
+  /** A wrapper of a slot that belongs to a specified {@link Inventory} */
+  private static class InventorySlot {
+
+    private final Slot slot;
+    private final Inventory inventory;
+
+    private InventorySlot(Slot slot, Inventory inventory) {
```
Pablete1234, line 275: Could add a constructor for inventory + index, and derive slot automatically

util/src/main/java/tc/oc/pgm/util/chunk/ChunkVector.java
```diff
@@ -20,20 +21,24 @@ public static ChunkVector of(Chunk chunk) {
     return new ChunkVector(chunk.getX(), chunk.getZ());
   }
 
-  public static ChunkVector ofBlock(int x, int y, int z) {
+  public static ChunkVector ofChunkVector(int x, int z) {
+    return new ChunkVector(x, z);
+  }
+
+  public static ChunkVector ofBlock(int x, int z) {
```
Pablete1234: If it's `ofBlock`, a block has X, Y and Z. Even if the chunk doesn't use the Z for anything, it's still part of the block, and this function is taking "a block" as input parameter, it's impl detail that the Y is discarded
```suggestion
  public static ChunkVector ofBlock(int x, int y, int z) {
```

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
```diff
@@ -395,15 +398,23 @@ public static <T extends Number & Comparable<T>> T parseNumberInRange(
...
+    Matcher matcher = RANGE_RE.matcher(nodeValue);
     if (!matcher.matches()) {
-      T value = parseNumber(node, node.getValue(), type, true);
+      Matcher legacy = RANGE_LEGACY.matcher(nodeValue);
+      if (legacy.matches()) {
+        String[] split = nodeValue.split("\\.\\.");
+        T lower = parseNumber(node, split[0], type, false);
+        T higher = parseNumber(node, split[1], type, false);
+        return Range.closed(lower, higher);
+      }
```
Pablete1234: I'd be nice if this supported `5..` for "5 or more" and `..5` for "5 or less", currently this requires both sides. Also, because you're using `\\d+`, it does not support `oo` for infinity which is generally accepted by pgm, so i suggest you change to `\\s` like the other range pattern does

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/kits/Slot.java
KingOfSquares: Ah yes, this makes more sense when the inventory holder types were different.

core/src/main/java/tc/oc/pgm/loot/FillerDefinition.java
KingOfSquares: you got me there

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
KingOfSquares: Yes, but as I mentioned in the PR and in the javadocs on this class I changed the behavior for this in the new proto to apply multiple times(multiple rises). Maybe it should be an attribute instead?

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
KingOfSquares: was the "5 or more" thing supported before? As I've understood it its a legacy format.
Ill do the infinity thing

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
Pablete1234: Was this behavior not at all present in 1.12 pgm? If so, can you elaborate as to why we want this change? why is it proto-dependant instead of just an xml flag? does it add extra value that original impl did not have? 

If this behavior is not from 1.12 PGM, next proto should be 1.4.4, as 1.4.3 was used in 1.12 pgm

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
Pablete1234: It probably wasn't, but personally i find the `5..` or `..5` syntax more intuitive than `(oo, 5]` (also, vanilla commands use the `..5` `5..` or `5..10` syntax, eg: `/execute if score player Objective matches 5.. run say hi`), and it really doesn't hurt to have, it won't break behavior as it's simply "one more" way to define it

core/src/main/java/tc/oc/pgm/loot/FillableAndFiller.java
KingOfSquares: Its available through apache in SportPaper. Is that ok to depend on?

Pablete1234: Do spigot/paper also include it? if so, yeah that's fine

KingOfSquares: Not found in paper 1.18. Want me to just make a generic Pair util maybe? Not sure why I just didn't do that

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
KingOfSquares: Sure, but i'm not sure how this affects memory size of each fillable and if its even relevant?

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
KingOfSquares: Not present in project ares pgm.

The more I think about this the more I think the multiple rise/multiple fill thing should be a xml flag. Why I think it should be a thing is because in my opinion it makes sense for it to work especially in scenarios where the trigger is a non-time trigger (e.g completing some objective multiple times) you still want to reward your players multiple times instead of "the amount of times they have opened the lootable chest in between each objective completion"

I still believe proto change makes sense for the explicit fill at match start makes sense. If lootables are used in any other scenarios than maps like SG or Skywars where _all_ chests are used for items throughout the match there is a big probability you don't want it to fill it at match start, but since it is legacy behavior if it is an attribute it has to default to true, which I think does not makes sense

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
Pablete1234: a function does not affect memory size, functions are on the class not on the instance

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
KingOfSquares: If the item filters(e.g. CarryingItemFilter) every get expanded to include entities I can see this change mattering, but it would not change legacy behavior as far as I can see. Think it's worth keeping this behavior for that potential change?

KingOfSquares: Bump, @Pablete1234 ?

core/src/main/java/tc/oc/pgm/kits/ItemKit.java
```diff
@@ -137,10 +137,10 @@ public void apply(MatchPlayer player, boolean force, List<ItemStack> displacedIt
       // Put the remaining kit slotted items into their designated inv slots.
       // If a slot is occupied, add the stack to displacedStacks.
       for (Entry<Slot, ItemStack> kitEntry : event.getSlotItems().entrySet()) {
+        Slot kitSlot = kitEntry.getKey();
         ItemStack kitStack = kitEntry.getValue();
         if (kitStack.getAmount() <= 0) continue;
 
-        Slot kitSlot = kitEntry.getKey();
```
Pablete1234: any reason for this change?

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -284,6 +271,24 @@ public ItemKit parseItemKit(Element el) throws InvalidXMLException {
...
+
+      case "book":
+        item = parseBook(el);
+        break;
+
+      case "head":
+        item = parseHead(el);
+        break;
+    }
+    return item;
```
Pablete1234: this could use returns on each case directly, and include a default for null:
```suggestion
    switch (el.getName()) {
      case "item":
        return parseItem(el, true);
      case "book":
        return parseBook(el);
      case "head":
        return parseHead(el);
      default:
        return null;
    }
```

core/src/main/java/tc/oc/pgm/loot/FillerDefinition.java
```diff
@@ -0,0 +1,94 @@
...
+import tc.oc.pgm.api.feature.FeatureInfo;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.util.compose.Composition;
+
+@FeatureInfo(name = "fill")
+public class FillerDefinition implements FeatureDefinition {
+
+  private final Composition<ItemStack> loot;
+  private final Filter filter;
+  private final @Nullable Filter refillTrigger; // Dynamic
```
Pablete1234: Would this being null be functionally equivalent to simply an always (or a never) filter?
if so, may as well make it non-null and use an always/never for it

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,302 @@
...
+  private final @Nullable ItemModifyMatchModule immm;
+
+  public LootableMatchModule(
+      Logger logger, Match match, List<FillerDefinition> fillers, List<Cache> caches) {
+    this.logger = logger;
+    this.fillers = fillers;
+    this.match = match;
+    this.caches = caches;
+    this.filledAt = new InstantMap<>(new WorldTickClock(match.getWorld()));
+    this.immm = match.getModule(ItemModifyMatchModule.class);
```
Pablete1234, line 61: Consider using moduleOptional here, and instead of having a nullable immm make it an optional. The benefit is you can just imm.ifPresent(...) instead of a null-check, could be more convenient sometimes.

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,302 @@
...
+
+    fillers.forEach(
+        filler -> {
+          if (filler.getRefillTrigger() == null) return;
+          match
+              .needModule(FilterMatchModule.class)
```
Pablete1234: move this outside the for-each and re-use it

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,302 @@
...
+        () ->
+            "Found fillers "
+                + fillers.stream()
+                    .map(FillerDefinition::toString)
+                    .collect(Collectors.joining(", ")));
+
+    // Find all Caches that the opened inventory is part of
+    final List<Fillable> fillables = new ArrayList<>();
+    for (Cache cache : caches) {
+      if (filterPredicate.test(cache.region()) && filterPredicate.test(cache.filter())) {
```
Pablete1234: doesn't testing both separately sound alot like testing an `AllFilter.off(cache.region(), cache.filter())`, if that's the case it may be worth having the caches create one internally and using it here, but this is probably not needed if this is the only place we're checking this

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,302 @@
...
+                // (27 is the standard size for single chests)
+                final List<InventorySlot> slots = new ArrayList<>(inventories.size() * 27);
+                inventories.forEach(
+                    inv -> {
+                      for (int index = 0; index < inv.getSize(); index++) {
+                        if (inv.getItem(index) == null) {
+                          slots.add(new InventorySlot(index, inv));
+                        }
+                      }
+                    });
```
Pablete1234:
```suggestion
                for (Inventory inv : inventories) {
                  for (int index = 0; index < inv.getSize(); index++) {
                    if (inv.getItem(index) == null) 
                      slots.add(new InventorySlot(index, inv));
                  }
                }
```
Sometimes lambdas aren't any shorter, i'd use a traditional loop for this inner-most loop

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
```diff
@@ -0,0 +1,302 @@
...
+                      .forEachOrdered(
+                          item -> {
+                            if (!slots.isEmpty()) {
+                              // For each item, remove a random slot from those remaining,
+                              // apply item mods, and put it in the slot.
+                              if (immm != null) immm.applyRules(item);
+                              slots.remove(random.nextInt(slots.size())).putItem(item);
+                            }
+                          });
+                }
```
Pablete1234:
```suggestion
                // Generate the loot for this Filler
                filler
                    .getLoot()
                    .elements(opener)
                    .limit(slots.size())
                    .map(i -> immm != null ? immm.applyRules(item) : item)
                    .forEachOrdered(i -> slots.remove(random.nextInt(slots.size())).putItem(item)));
```

core/src/main/java/tc/oc/pgm/loot/LootableModule.java
```diff
@@ -0,0 +1,125 @@
...
+public class LootableModule implements MapModule<LootableMatchModule> {
+
+  private static final Function<MapFactory, CompositionParser<ItemStack>> LOOT_PARSER_FACTORY =
+      factory ->
+          new CompositionParser<ItemStack>(factory) {
+            @Override
+            public ItemStack parseUnit(Element element) throws InvalidXMLException {
+              return this.factory.getKits().parseItemStack(element);
+            }
+          };
```
Pablete1234: I feel like this factory wouldn't be needed if the class wasnt annonymous.

Consider simply having an `LootParser extends CompositionParser<ItemStack> {`, and your "factory" can be simply calling: `new LootParser(mapFactory);`, making this Function<...> redundant

core/src/main/java/tc/oc/pgm/loot/LootableModule.java
```diff
@@ -0,0 +1,125 @@
...
+    @Override
+    public LootableModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      List<FillerDefinition> fillers = new ArrayList<>();
+      List<Cache> caches = new ArrayList<>();
+      CompositionParser<ItemStack> lootParser = LOOT_PARSER_FACTORY.apply(factory);
+      Map<String, Composition<ItemStack>> loots = new HashMap<>();
+
+      for (Element loot : XMLUtils.flattenElements(doc.getRootElement(), "lootables", "loot")) {
+        loots.put(loot.getAttributeValue("id"), lootParser.parseElement(loot));
```
Pablete1234: should id be required? if so, make it be
also nothing prevents duplicate ids which could be a common pitfall.

I suggest these be added to the feature context of the map factory, which also ensures there's no duplicate ids of loots overlapping other features (eg: filters)

core/src/main/java/tc/oc/pgm/loot/LootableModule.java
```diff
@@ -0,0 +1,125 @@
...
+        Composition<ItemStack> loot = loots.get(filler.getAttributeValue("loot"));
+        Filter fillableFilter =
+            factory.getFilters().parseFilterProperty(filler, "filter", StaticFilter.ALLOW);
+        Duration refillInterval =
+            XMLUtils.parseDuration(
+                filler.getAttribute("refill-interval"), TimeUtils.INFINITE_DURATION);
+        @Nullable
+        Filter refillTrigger =
+            factory
+                .getFilters()
```
Pablete1234: Consider defining this as a shorthand earlier in the function and reusing it, ie: `FilterParser filters = factory.getFilters()`, as it's used a few times

core/src/main/java/tc/oc/pgm/loot/LootableModule.java
```diff
@@ -0,0 +1,125 @@
...
+                .parseProperty(filler, "refill-trigger", DynamicFilterValidation.MATCH);
```
Pablete1234: as mentioned earlier, could make this non-nullable and simply make the default a StaticFilter.DENY/never

core/src/main/java/tc/oc/pgm/loot/WeightedRandomChooser.java
```diff
@@ -0,0 +1,79 @@
+package tc.oc.pgm.loot;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Random;
+
+public class WeightedRandomChooser<T> {
+
+  private final Map<T, Option> options = new HashMap<>();
```
Pablete1234: Why not use simply `Map<T, Double>` ? doesn't seem like Option adds much value here except for an extra object allocation

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
```diff
@@ -400,50 +403,108 @@ public static <T extends Number & Comparable<T>> Range<T> parseNumericRange(
...
+      } else if (second == null) {
+        lower = first.trim();
+        lowerBound = BoundType.CLOSED;
+        higher = "oo";
+      } else {
+        lower = first.trim();
+        lowerBound = BoundType.CLOSED;
+        higher = second.trim();
+        higherBound = BoundType.CLOSED;
+      }
```
Pablete1234:
```suggestion
      lower = first == null ? "-oo" : first.trim();
      upper = second == null ? "oo" : second.trim();
      lowerBound = BoundType.CLOSED;
      higherBound = BoundType.CLOSED;
```
Currently you're assuming that just because first is null, second won't be, both could be null if you just type `..` for an unbounded range of all numbers. Additionally this simplifies the logic

util/src/test/java/tc/oc/pgm/util/xml/ParseRangeTest.java
```diff
@@ -0,0 +1,102 @@
...
+            Pair.of("  1  ..5    ", Range.closed(1D, 5D)),
+            Pair.of("1..", Range.atLeast(1D)),
+            Pair.of("..5", Range.atMost(5D)),
+            Pair.of("-oo..1", Range.atMost(1D)),
+            Pair.of("1..oo", Range.atLeast(1D)),
+            Pair.of("-oo..oo", Range.all()),
+            Pair.of("-oo ..", Range.all()),
+            Pair.of(" ..oo", Range.all()),
+            Pair.of(".5..5", Range.closed(0.5D, 5D)),
+            Pair.of("0.5  ..5 ", Range.closed(0.5D, 5D)));
```
Pablete1234: could probably add `..` for a test-case of `Range.all()`

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/kits/ItemKit.java
KingOfSquares: no

core/src/main/java/tc/oc/pgm/loot/LootableMatchModule.java
KingOfSquares: I mean its only used a single place in this class, feels like unnecessary object allocation/logic for a simple line looking a little more elegant?

core/src/main/java/tc/oc/pgm/loot/FillerDefinition.java
KingOfSquares: If this is null it never registers a dynamic filter listener(See LootableMatchModule constructor), so to prevent unnecessary registering you would have to do an `#equal` check instead.

lefckntea:
i don't really get how the current lootables idea works but would be cool to have the option to make the chest disappear when it's fully looted

Pablete1234:
@lefckntea i see that as a different feature where you could specify a filter (ie: region or regions) where empty chests get removed, but is completely outside the scope of lootables themselves

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
```diff
@@ -198,6 +198,11 @@ private <F extends Filterable<?>> void register(
           "Filter " + filter + " doesn't respond to " + scope.getSimpleName() + " scope.");
     }
 
+    if (filter.getRelevantEvents().isEmpty()) {
+      // In some cases filters can be dynamic but doesn't need event invalidation
+      return;
+    }
+
```
Pablete1234: A filter with no events may still have listeners, particularily, monostable filters.

We should avoid listening for *event* changes on these, but the listeners should probably still be registered.

Pablete1234 (changes requested):
Left an important note, besides that i believe this needs to be rebased

core/src/main/java/tc/oc/pgm/loot/LootableModule.java
```diff
@@ -0,0 +1,121 @@
...
+import tc.oc.pgm.itemmeta.ItemModifyMatchModule;
+import tc.oc.pgm.util.TimeUtils;
+import tc.oc.pgm.util.compose.Composition;
+import tc.oc.pgm.util.compose.CompositionParser;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.XMLUtils;
+
+public class LootableModule implements MapModule<LootableMatchModule> {
+
+  private final java.util.List<FillerDefinition> fillers;
```
Pablete1234:
```suggestion
  private final List<FillerDefinition> fillers;
```

Pablete1234 (approved):

PR #1045, opened 2022-08-20
EnderChest support 👻 

applenick:
# EnderChest support

Hooray! 👻 This PR adds proper EnderChest support to PGM. 

## XML Example
```xml

<!-- Add drop-off locations for when player switches teams or leave the match -->
<enderchest>
  <!-- Each drop-off requires a region & filter attribute -->
  <dropoff region="red-spawn" filter="red-only"/>
  <dropoff region="blue-spawn" filter="blue-only"/>
</enderchest>

<!-- Specify a fallback location option for when no drop-off locations are defined (AUTO, KEEP, or DELETE) -->
<enderchest fallback="DELETE"/>
```

If you've got any feedback let me know and I'd be more than happy to make accommodations. As always these changes have been tested and should work as intended :+1:


KingOfSquares:
I believe players still would be blocked from placing ender chests because of the listener in EventFilterMatchModule
https://github.com/PGMDev/PGM/blob/963f02a2974c471f356b782b5e0fd741a77869fc/core/src/main/java/tc/oc/pgm/modules/EventFilterMatchModule.java#L259-L271
Maybe also filter that on the enabled boolean?

EDIT: Or maybe not? Since you are cancelling the event, but then this listener is redundant

Also, with this addition maybe it would also make sense to add ender chest support to kits?

applenick:
@KingOfSquares 
Great point! I believe you're correct, my changes don't currently adjust EventFilterMatchModule blocking the placing/breaking of EnderChests. Personally I don't foresee many instances where map authors would want players to be able to freely place them, so chose to preserve that function while allowing interacting with existing EnderChests to be supported. Since the more common usage will be interacting with strategically placed chests on a map.

Would you suggest allowing them to be placed?

> Also, with this addition maybe it would also make sense to add ender chest support to kits?

Could you elaborate on this?

Thanks!

KingOfSquares:
> @KingOfSquares Great point! I believe you're correct, my changes don't currently adjust EventFilterMatchModule blocking the placing/breaking of EnderChests. Personally I don't foresee many instances where map authors would want players to be able to freely place them, so chose to preserve that function while allowing interacting with existing EnderChests to be supported. Since the more common usage will be interacting with strategically placed chests on a map.
> 
> Would you suggest allowing them to be placed?
> 
> > Also, with this addition maybe it would also make sense to add ender chest support to kits?
> 
> Could you elaborate on this?
> 
> Thanks!

For the kit thing, there is some support already in `Slot.java` for being able to place items in players ender chests through kits. A kit could potentially have an ` <item slot="enderchest.1" material="stone sword"/> `

For placing ender chests: I agree that the ability to being able to place ender chests is not correct in most situations, but since they are now usable I only think its a matter of time before someone requests how to enable placing them(for some reason). Why not default it to false and include the possibility as an attribute? I also believe the error message is a little misleading, `Ender chests are disabled` should probably be something like `Placing ender chests are disabled`

core/src/main/java/tc/oc/pgm/enderchest/EnderChest.java
```diff
@@ -0,0 +1,54 @@
...
+
+  public void open(MatchPlayer player) {
+    if (contents == null) {
+      contents =
+          Bukkit.createInventory(
+              player.getBukkit(),
+              rows * 9,
+              player.getParty().getColor()
+                  + player.getNameLegacy()
+                  + "'s "
```
Pablete1234: This is not how you localize strings, the string should be something like `{0}'s chest` and you use a translation key for that, with the player as parameter for 0. 

Otherwise other languages are not able to translate the possesive, "Pablete1234's chest" is fine in english but "Pablete1234's Cofre" is completely wrong spanish (correct would be "Cofre de Pablete1234"), the way for that to be translatable is to support `Cofre de {0}` as the translatable string

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
```diff
@@ -0,0 +1,94 @@
...
+      if (dropoff.getFilter().query(oldParty).isAllowed()) {
+        chest.drop(dropoff.getRegion().getRandom(match.getRandom()).toLocation(match.getWorld()));
+        dropped = true;
+        break;
+      }
+    }
+
+    if (!dropped) {
+      chest.clear();
+    }
```
Pablete1234: We may need a config for what to do if no dropoff matches the filter, if it should delete the items or if it should just keep them (in case they happen to rejoin the same team)?
Unsure if it would really be necessary in practice tho

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
```diff
@@ -0,0 +1,94 @@
...
+  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+  public void onChestInteract(PlayerInteractEvent event) {
+    MatchPlayer player = match.getParticipant(event.getPlayer());
+    if (event.getClickedBlock() == null) return;
+    if (event.getClickedBlock().getType() != Material.ENDER_CHEST) return;
+    if (player == null) return;
+    if (!enabled) {
+      player.sendWarning(translatable("match.disabled.enderChest"));
+      return;
+    }
```
Pablete1234: Apperently you were always able to open/use enderchests if they were already placed in the world, we shouldn't need to prevent this in any situation (unless someone wants to disable use purposefully, and they can do that with applied regions)

However, we do need to retain legacy behavior where if no `<enderchest/>` tag is included in the XML no placing of enderchest blocks is allowed. This currently lies in `EventFilterMatchModule` and since it is not touched by this PR, i assume placing/breaking is still prevented. Should make that event cancellation dependant on if enderchest module is enabled or not.

Pablete1234: Apperently you were always able to open/use enderchests if they were already placed in the world, we shouldn't need to prevent this in any situation (unless someone wants to disable use purposefully, and they can do that with applied regions)

However, we do need to retain legacy behavior where if no `<enderchest/>` tag is included in the XML no placing of enderchest blocks is allowed. This currently lies in `EventFilterMatchModule` and since it is not touched by this PR, i assume placing/breaking is still prevented. Should make that event cancellation dependant on if enderchest module is enabled or not.

core/src/main/java/tc/oc/pgm/enderchest/EnderChestModule.java
```diff
@@ -0,0 +1,74 @@
...
+      FilterParser filters = factory.getFilters();
+      RegionParser regions = factory.getRegions();
+
+      boolean enabled = false;
+      int rows = 3;
+      List<Dropoff> dropoffs = Lists.newArrayList();
+
+      for (Element enderRootEl : doc.getRootElement().getChildren("enderchest")) {
+
+        dropoffs = Lists.newArrayList();
```
Pablete1234:
```suggestion
```
you've already defined a list prior, don't override it every time you find a new enderchest tag, this should be valid and use *both* dropoffs:

```
<enderchest>
 <dropoff .../>
</enderchest
<enderchest>
 <dropoff .../>
</enderchest
```

core/src/main/java/tc/oc/pgm/enderchest/EnderChestModule.java
```diff
@@ -0,0 +1,74 @@
...
+    public EnderChestModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      FilterParser filters = factory.getFilters();
+      RegionParser regions = factory.getRegions();
+
+      boolean enabled = false;
+      int rows = 3;
+      List<Dropoff> dropoffs = Lists.newArrayList();
+
+      for (Element enderRootEl : doc.getRootElement().getChildren("enderchest")) {
```
Pablete1234: Use the XMLUtils flattening method here, it will allow support for nested classes etc

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
Pablete1234: Apperently you were always able to open/use enderchests if they were already placed in the world, we shouldn't need to prevent this in any situation (unless someone wants to disable use purposefully, and they can do that with applied regions)

However, we do need to retain legacy behavior where if no `<enderchest/>` tag is included in the XML no placing of enderchest blocks is allowed. This currently lies in `EventFilterMatchModule` and since it is not touched by this PR, i assume placing/breaking is still prevented. Should make that event cancellation dependant on if enderchest module is enabled or not.

Pablete1234: Apperently you were always able to open/use enderchests if they were already placed in the world, we shouldn't need to prevent this in any situation (unless someone wants to disable use purposefully, and they can do that with applied regions)

However, we do need to retain legacy behavior where if no `<enderchest/>` tag is included in the XML no placing of enderchest blocks is allowed. This currently lies in `EventFilterMatchModule` and since it is not touched by this PR, i assume placing/breaking is still prevented. Should make that event cancellation dependant on if enderchest module is enabled or not.

core/src/main/java/tc/oc/pgm/enderchest/EnderChestModule.java
```diff
@@ -0,0 +1,74 @@
...
+
+        dropoffs = Lists.newArrayList();
+        for (Element dropoffEl : XMLUtils.getChildren(enderRootEl, "dropoff")) {
+          Region region = regions.parseRegionProperty(dropoffEl, "region");
+          Filter filter = filters.parseFilterProperty(dropoffEl, "filter");
+
+          if (region == null || filter == null)
+            throw new InvalidXMLException("Dropoffs require both a region and filter", dropoffEl);
```
Pablete1234:
```suggestion
          Region region = regions.parseRequiredRegionProperty(dropoffEl, RandomPointsValidation.INSTANCE, "region");
          Filter filter = filters.parseRequiredFilterProperty(dropoffEl, "filter");
```
parse using the requred method directly, also make sure to add the validation for random points validation, you wouldn't want someone passing in `<everywhere/>` as a region here for example

core/src/main/java/tc/oc/pgm/enderchest/EnderChestModule.java
```diff
@@ -0,0 +1,74 @@
...
+
+          dropoffs.add(new Dropoff(region, filter));
+        }
+
+        Attribute rowAttr = XMLUtils.getAttribute(enderRootEl, "rows");
+        if (rowAttr != null) rows = XMLUtils.parseNumber(rowAttr, Integer.class);
+
+        if (rows < 1 || rows > 6)
+          throw new InvalidXMLException("Row amount must be between 1 and 6", enderRootEl);
```
Pablete1234: Stick this somewhere in the factory
`private static final Range<Integer> ROW_RANGE = Range.closed(1, 6);`

Then you can parse this directly with:
```suggestion
        rows = XMLUtils.parseNumberInRange(Node.fromAttr(enderRootEl, "rows"), Integer.class, ROW_RANGE, rows);
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
applenick: @Pablete1234 100% agree about having placement retain its legacy behavior. I'll add in a check to allow placement if the enderchest module is enabled.

> Apparently you were always able to open/use enderchests if they were already placed in the world, we shouldn't need to prevent this in any situation

Imo, being able to open/use enderchests in the vanilla sense is not something PGM should allow. I like my solution in blocking the interaction, but an alternative could be to clear all enderchests manually. Main concern of mine is patching the exploit where users can transfer items between matches, which is possible atm as long as enderchests exist between two maps. What do you believe we should do?

(Also I see I made a mistake in returning when not enabled before canceling the event, I'll fix that if we choose to keep this in)

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
applenick: Mind elaborating more? Config as in an additional attribute that can be defined? Perhaps something like `fallback`: 
```xml
<!-- Fallback would only be used when no drop-offs are found -->

<!-- Keeps enderchest contents upon rejoin to same team -->
<enderchest fallback="keep"/>

<!-- Destroy enderchest contents (defaults to this behavior) -->
<enderchest fallback="delete"/>
```

Pablete1234: Yeah, something like that. Maybe need a 3rd option "auto" which is delete if you have any dropoffs configured, and keep if no dropoff is configured, and that's the actual default.

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
Pablete1234: > Imo, being able to open/use enderchests in the vanilla sense is not something PGM should allow.

It's simply behavior pgm should replicate, but without the bug that sharing items between matches is. In other words, if no `<enderchest/>` is defined, opening an enderchest should open a normal 3-row enderchest that lasts for that match. (what this module does!)

Electroid:
Hey @applenick, thanks for working on this.

I think it would be better to integrate ender chest support into the existing `Kit` and `Slot` system. Also, maybe we already have support?? 

https://github.com/PGMDev/PGM/blob/a3992e87639f52f5d3c7ed6f5d2d9690f6c56d4f/core/src/main/java/tc/oc/pgm/kits/Slot.java#L169

applenick:
> Hey @applenick, thanks for working on this.
> 
> I think it would be better to integrate ender chest support into the existing `Kit` and `Slot` system. Also, maybe we already have support??
> 
> https://github.com/PGMDev/PGM/blob/a3992e87639f52f5d3c7ed6f5d2d9690f6c56d4f/core/src/main/java/tc/oc/pgm/kits/Slot.java#L169

No problem! Excellent point about the kit/slot support, I believe Simon mentioned that too in earlier comments. Perhaps we should just scrap the custom inventory logic and have it rely on the built-in enderchests. That way they'll already support slots like mentioned. Though I believe it's important to keep some of the new functionality added like drop-off locations and the fallback attribute. 

Will look into fixing this up over the weekend 👍

applenick:
Finally had some time and fixed this up 🎉 Switched away from using the custom enderchest implementation we originally had, relying now on the traditional enderchest contents. We do lose the capability to customize enderchest row count, but I believe it's a fair tradeoff for being able to have enderchests remain compatible with the Kit/Slot system.

If there's any additional feedback let me know and I'll be happy to make adjustments 👍

core/src/main/java/tc/oc/pgm/enderchest/EnderChestModule.java
```diff
@@ -0,0 +1,69 @@
...
+                DropoffFallback.class,
+                "fallback",
+                DropoffFallback.AUTO);
+        enabled = true;
+      }
+
+      for (Element dropoffEl :
+          XMLUtils.flattenElements(doc.getRootElement(), "enderchest", "dropoff")) {
+        Region region = regions.parseRequiredRegionProperty(dropoffEl, "region");
+        Filter filter = filters.parseRequiredFilterProperty(dropoffEl, "filter");
```
Pablete1234: We can take advantage of the new refactor that streamlines XMLParsers, also, need to add validation to region, don't want someone slapping an <everywhere/> in there, need a region that supports random points:

```suggestion
        Region region = regions.parseRequiredProperty(dropoffEl, "region", RandomPointsValidation.INSTANCE);
        Filter filter = filters.parseRequiredProperty(dropoffEl, "filter");
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/enderchest/EnderChestModule.java
applenick: Excellent! Made the change 👍

Pablete1234 (approved):

Electroid (approved):
🚀

PR #1046, opened 2022-08-26
Score Minimum Value

Samuel-Roach:
Added changes to reflect the change suggested in https://github.com/PGMDev/PGM/issues/1024#issue-1296570186.

Adds ``mercyLimitMin`` to ScoreConfig to store the minimum score required for a MercyRule to enforce.

Mercy Rule now calculates a ``mercyBaseline`` which will be the maximum of the ``mercyLimitMin`` and ``scoreBaseline + mercyLimit``.

This hasn't been tested, although linting and building works. Sidenote: I could use some help with setting up a local PGM server such that I can test changes.


KingOfSquares:
For testing i recommend the `server` module which starts a server with your changes

Pablete1234:
Given that this new min may only ever have an effect if a mercy rule is defined, how about:

```xml
<score>
  <limit>10</limit>
  <mercy min="5">2</mercy>
</score>
```
This would make the max-score start at 5, go at most up to 10, and be always 2-higher than the runner-up

Samuel-Roach:
Makes sense given the dependency on mercy rule. I'll make the change

core/src/main/java/tc/oc/pgm/score/ScoreModule.java
```diff
@@ -101,6 +101,12 @@ public ScoreModule parse(MapFactory factory, Logger logger, Document doc)
         config.scoreLimit = XMLUtils.parseNumber(scoreEl.getChild("limit"), Integer.class, -1);
         config.mercyLimit = XMLUtils.parseNumber(scoreEl.getChild("mercy"), Integer.class, -1);
 
+        List<Element> mercyElements = scoreEl.getChildren("mercy");
+        for (Element mercyEl : mercyElements) {
+          config.mercyLimitMin =
+              XMLUtils.parseNumber(Node.fromAttr(mercyEl, "min"), Integer.class, -1);
+        }
```
Pablete1234: I suggest using `XMLUtils.getUniqueChild` for this, not iterating a list

Samuel-Roach:
Sorry this took a while. Finally got around to using `getUniqueChild`. Tested and confirmed locally.

Pablete1234 (approved):

Electroid (approved):
Thanks for bringing this over the line! 🎉

PR #1047, opened 2022-08-28
Implement variables!

Pablete1234:
### What are variables about? 
They're a way to store information that you can filter on. This is generally done at the moment by abusing invisible hills that a team can capture to store individual bits of information, that is a bad workaround for the lack of this feature.

### How do variables work?
You can define as many variables as you want in `<variables>`, and each one must have a scope. If a variable is scoped to the player, it means each player has their own value for such variable. The available scopes (as with actions/triggers) are `player`, `team` and `match`.

Variables can be updated by the `<set>` action, which requires a `var`iable to set, and a `formula`. The formula can be any general-purpose math expression, and supports using other variables too. If a scope cannot be automatically derived, you should specify one in the element `<set scope="player" var="t_score" formula="p_score+t_score/>`. 

Note: Only variables which exist in the context can be used, eg: you cannot use a player-scoped variable in a party-scoped `set`, which of the players do you use? (you can however do the opposite, use, or set, a party-scoped variable from player-scoped `set`)

You can then filter if variables have certain values by checking with a filter, that works very similarly to `<score>` filter: `<variable var="var-id">[5, 10]</variable>`

The `/variables [player]` command can be used to check all variables in the current match for a player (it will also show the variables for the team and match that player is in)

### Examples:
```xml
<variables>
  <variable id="t_score" scope="team">
  <variable id="p_score" scope="player" default="5">
</variables>
<actions>
  <action id="score-points" scope="player">
    <set var="t_score" value="t_score+p_score">
    <set var="p_score" value="p_score+5">
  </action>
  <message id="completed" text="The team reached 100 points!"/>
  <trigger filter="region-a" trigger="score-points" scope="player"/>
  <trigger filter="reached-score" trigger="completed" scope="team"/>
</action>
<filters>
  <variable id="reached-score" var="t_score">[100,oo)</variable>
</filters>
```
In this example, whenever a player enters `region-a` they increment team score by their current score, and their current score goes up by 5. So first time they score 5, second time score 10, etc. Upon the team reaching 100+ in score, everyone in the team gets a "The team reached 100 points" message in chat.

```xml
<variables>
  <variable id="next_post" scope="match">
</variables>
<kits>
  <kit id="on-flag-pickup">
    <action filter="red-team">
      <set var="next_post" value="1"/>
    </action>
    <action filter="blue-team">
      <set var="next_post" value="2"/>
    </action>
  </kit>
</kits>
<filters>
  <variable id="next-blue" var="next_post">1</variable>
  <variable id="next-red" var="next_post">2</variable>
</filters>
```
In this example, whenever a player picks up the flag and are given the on-flag-pickup kit, depending on their team they'll set the `next_post` variable to 1 or 2, deciding what post it will spawn next in (the posts are filtered based on `next-blue` and `next-red` filters)

This seems to be working as intended with initial fixups.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -9,12 +9,10 @@ import tc.oc.pgm.action.actions.ActionNode;
 import tc.oc.pgm.action.actions.ChatMessageAction;
 import tc.oc.pgm.action.actions.ScopeSwitchAction;
 import tc.oc.pgm.action.actions.SetVariableAction;
-import tc.oc.pgm.api.feature.FeatureDefinition;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.Filterables;
 import tc.oc.pgm.api.map.factory.MapFactory;
 import tc.oc.pgm.features.FeatureDefinitionContext;
-import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
 import tc.oc.pgm.filters.Filterable;
 import tc.oc.pgm.filters.matcher.StaticFilter;
 import tc.oc.pgm.filters.parse.FilterParser;
@@ -31,15 +29,20 @@ import tc.oc.pgm.variables.VariablesModule;
 public class ActionParser {
 
   private final MapFactory factory;
+  private final FeatureDefinitionContext features;
   private final FilterParser filters;
+  private final VariablesModule variables;
   private final Map<String, Method> methodParsers;
 
   public ActionParser(MapFactory factory) {
     this.factory = factory;
+    this.features = factory.getFeatures();
     this.filters = factory.getFilters();
+    this.variables = factory.needModule(VariablesModule.class);
     this.methodParsers = MethodParsers.getMethodParsersForClass(getClass());
   }
 
+  @SuppressWarnings("unchecked")
   public <B extends Filterable<?>> Action<? super B> parse(Element el, @Nullable Class<B> bound)
       throws InvalidXMLException {
     String id = FeatureDefinitionContext.parseId(el);
@@ -62,8 +65,7 @@ public class ActionParser {
 
     // We don't need to add references, they should already be added by whoever created them.
     if (result instanceof ActionDefinition)
-      //noinspection unchecked
-      factory.getFeatures().addFeature(el, (ActionDefinition<? super B>) result);
+      features.addFeature(el, (ActionDefinition<? super B>) result);
     return result;
   }
 
@@ -82,8 +84,7 @@ public class ActionParser {
 
   public <B> Action<? super B> parseReference(Node node, String id, Class<B> bound)
       throws InvalidXMLException {
-    return factory
-        .getFeatures()
+    return features
         .addReference(new XMLActionReference<>(factory.getFeatures(), node, id, bound));
   }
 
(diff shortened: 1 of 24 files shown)


core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -160,4 +166,29 @@ public Kit parseKitTrigger(Element el, Class<?> scope) throws InvalidXMLExceptio
...
+      throw new InvalidXMLException(
+          "Wrong variable scope for '"
+              + var.getId()
+              + "', expected "
+              + var.getScope().getSimpleName()
+              + " which cannot be found in "
+              + scope.getSimpleName(),
+          el);
+
+    VariablesModule vm = this.factory.needModule(VariablesModule.class);
```
KingOfSquares: Any specific reason this module is just not defined at class scope when the parser is created?

core/src/main/java/tc/oc/pgm/filters/parse/FilterParser.java
```diff
@@ -407,6 +412,18 @@ private <T extends FeatureDefinition> XMLFeatureReference<T> reference(Node node
...
+      throws InvalidXMLException {
+    String id = node.getValueNormalize();
+    T val = this.factory.getFeatures().get(id, cls);
+    if (val == null)
+      throw new InvalidXMLException(
+          "Unknown " + SelfIdentifyingFeatureDefinition.makeTypeName(cls) + " ID '" + id + "'",
+          node);
+    return val;
+  }
+
```
KingOfSquares: Any specific reason this is identical in both parsers but not abstracted out to some `XMLReferenceUtils`?

core/src/main/java/tc/oc/pgm/variables/VariableDefinition.java
```diff
@@ -0,0 +1,30 @@
...
+    return scope;
+  }
+
+  public double getDefault() {
+    return def;
+  }
+
+  public Variable<T> getVariable(Match match) {
+    //noinspection unchecked
+    return (Variable<T>) match.getFeatureContext().get(this.getId());
```
KingOfSquares: I still think this is a weird and unnecessary deviation from the current PGM norms of suppressing warnings. I see no reason to change it?

```suggestion
  @SupressWarnings("unchecked")
  public Variable<T> getVariable(Match match) {
    return (Variable<T>) match.getFeatureContext().get(this.getId());
```

KingOfSquares (commented):
I like how this integrates with dynamic filters, directly invalidating seems the most effective way.

core/src/main/java/tc/oc/pgm/action/ActionParser.java
Pablete1234: Nope, it could probably be moved there

core/src/main/java/tc/oc/pgm/filters/parse/FilterParser.java
Pablete1234: The reason is XMLUtils lie in the `util` package, all of the xml references, FeatureDefinition etc exist only in the `core` package, meaning this sadly cannot be put in utils, where it'd best fit.

Could try to put it in some other xml utils file that resides in core but yeah, it was simpler to copy the one method to both parsers at this time

core/src/main/java/tc/oc/pgm/variables/VariableDefinition.java
Pablete1234: Sure, can move that to the method scope

KingOfSquares (dismissed):
Nice. I had a thought about possibly using variables in message strings like tips and/or alerts, but that seems like a good addition after this system gets some use.

Pablete1234:
> Nice. I had a thought about possibly using variables in message strings like tips and/or alerts, but that seems like a good addition after this system gets some use.

I definitely have some ideas for the future of variables, but i first want to see them rolled out and see what kind of use they get in the real world.

I've considered adding different types of variables, some of them maybe read-only, that allow you to interact more closely with other parts of pgm/minecraft.
An example:

```
<variables>
  <score id="team_score"/>
  <health id="player_health"/>
</variables>
```

Kinda like with filters or regions, there'd be many types of variables.

Score is tied to the actual pgm score of the team, so setting that score variable directly changes the teams' score, and reading from it gets the current team score. This allows you to straight operate with the team's score (this would be scoped to the team).
Health would be tied to the player health itself, so reading it is getting the players' current health (this would be scoped to the player).

Just with those 2 you could potentially make a map where the overall score is how much health the players end with for example, or simply award score based on total hearts remaining.

There is alot of potential, but i want to go step-by-step with them.

Electroid (dismissed):
Congrats, such a long awaited feature for PGM! 🎉

Electroid:
@Pablete1234 if you resolve the conflict, feel free to squash and merge.

Electroid (approved):

PR #1055, opened 2022-09-13
Implement item matchers

Pablete1234:
Change item filters like `holding`, `carrying` and `wearing` with `ItemMatcher`s. Item matchers are a more generic way of matching items, supporting checking ignoring metadata, or amount ranges. 

Prior to this, item matching was always exact meta, always ignoring durability, and "at least X" amount matching, this is the default behavior if no extra flags are specified.

Examples:
```xml
<holding amount="[3,6]" ignore-durability="false" ignore-metadata="true">
  <item material="wool" damage="0"/>
</holding>
```
This will match if you're holding between 3 and 6 (inclusive) white wool, ignoring any other metadata (eg: enchantments, renaming, can-place-on, etc).

```xml
<holding ignore-enchantments="true">
  <item material="wool" amount="5"/>
</holding>
```
This will match at least 5 wool (legacy behavior) of any durability (any color)  (legacy behavior), and matching exact metadata (eg: name, can-place-on, etc), but will ignore enchantments.

The new available properties are: `ignore-durability` (default true), `ignore-metadata` (default false), `ignore-name` (default ignore-meta), `ignore-enchantments` (default to ignore-meta).
If `ignore-metadata` is set to true, `ignore-name` and `ignore-enchantments` cannot be false.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/module/exception/ModuleLoadException.java
+++ b/core/src/main/java/tc/oc/pgm/api/module/exception/ModuleLoadException.java
@@ -17,6 +17,10 @@ public class ModuleLoadException extends RuntimeException {
     return key;
   }
 
+  public String getFullMessage() {
+    return getMessage() + (key != null ? " @ " + key.getSimpleName() : "");
+  }
+
   public ModuleLoadException(Class<? extends Module> key, String message) {
     this(key, message, null);
   }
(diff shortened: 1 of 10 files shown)


Pablete1234:
Update: rebased ontop of dev

Electroid (approved):
Yay! 🎉 LGTM

PR #1057, opened 2022-09-23
Fixes & improvements to action parsing

Pablete1234:
Minor fixes and improvements to action parsing.

Added `untrigger-filter` to `action` which lets you decide when untriggering is allowed (before this, untriggering never happened)
Scope validations are now performed as feature validations, meaning forward-references are better handled
Added validation for scope mismatching, before it would be silently ignored instead.
Fixed a couple places referencing actions as "triggers" (old syntax), the legacy format is still supported. eg:
`<trigger filter="dynamic-filter-id" action="action-id" scope="player"/>`


Electroid (approved):
🤙

PR #1059, opened 2022-09-23
Implement payload (prototype)

Pablete1234:
This is an experimental prototype for payload, not a final PR, however, it can be merged for playtesting.

It piggybacks off of ControlPoints entirely, to the point Payloads are "a type of control point" internally, they just use a different region to capture and display some particles on the world. 0% completion is the start of the track (set with location="x,y,z" in xml) and 100% completion is wherever the track ends.

```xml
<payloads>
  <payload location="x,y,z" radius="3" display-filter="filter" capture-time="1m" decay-rate="0.25" .../>
</payloads>
```
Note: In its current state it only properly supports A/D payload. If both teams can capture it, it *works* but it will go towards the same destination for both teams (going back when the non-dominating team goes in it), which is likely undesired behavior.

It supports *literally* everything that control points support (including absurd ones like a progress region, even tho they're unlikely to be of any use with a payload). So things like filtering players, multiple payloads chained (5-cp style), permanent="true", capture-time, decay-rate, contested-rate, recovery-rate, etc, all work.

On top of the control point stuff, it supports `display-filter` which filters if particles are shown. This is useful when combining multiple payloads to signify if they're active or not (this is likely to get removed, changed or reworked in the future when checkpoints are introduced).

How the payload currently looks like:

https://user-images.githubusercontent.com/11789291/192090427-8c315f9d-0442-40fe-a63a-d3f13d37ff12.mp4

### DIFF
--- a/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
+++ b/core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
@@ -11,12 +11,14 @@ import org.jdom2.Element;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.map.factory.MapFactory;
 import tc.oc.pgm.api.region.Region;
+import tc.oc.pgm.filters.matcher.StaticFilter;
 import tc.oc.pgm.filters.matcher.block.BlockFilter;
 import tc.oc.pgm.filters.operator.AnyFilter;
 import tc.oc.pgm.filters.parse.FilterParser;
 import tc.oc.pgm.goals.ShowOptions;
 import tc.oc.pgm.payload.PayloadDefinition;
 import tc.oc.pgm.regions.BlockBoundedValidation;
+import tc.oc.pgm.regions.EverywhereRegion;
 import tc.oc.pgm.regions.RegionParser;
 import tc.oc.pgm.teams.TeamFactory;
 import tc.oc.pgm.teams.TeamModule;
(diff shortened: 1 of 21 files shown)


core/src/main/java/tc/oc/pgm/controlpoint/ControlPointParser.java
```diff
@@ -31,8 +34,14 @@ public abstract class ControlPointParser {
           new BlockFilter(Material.STAINED_GLASS),
           new BlockFilter(Material.STAINED_GLASS_PANE));
 
+  public enum Type {
+    KING,
```
Electroid:
```suggestion
    HILL,
```

core/src/main/java/tc/oc/pgm/controlpoint/RegionPlayerTracker.java
```diff
@@ -16,20 +16,28 @@
 import tc.oc.pgm.util.event.PlayerCoarseMoveEvent;
 
 /** Tracks which players are on a control point and answers some queries about them */
-public class ControlPointPlayerTracker implements Listener {
-  protected final Match match;
-  protected final Region captureRegion;
-  protected final Set<MatchPlayer> playersOnPoint = Sets.newHashSet();
+public class RegionPlayerTracker implements Listener {
+  private final Match match;
+  private final Set<MatchPlayer> playersOnPoint = Sets.newHashSet();
```
Electroid:
```suggestion
  private final Set<MatchPlayer> playersInRegion = Sets.newHashSet();
```

Electroid (commented):
Design looks good, didn't look at the `track` code yet.

Pablete1234:
The `track` stuff is probably the only somewhat-final thing, everything else is just "in there" to fully piggy-back off of control points, in order to save time and have a minimal implementation for testing/design purposes.

Want to see how well this works gameplay-wise, map-wise, and what other elements would make sense code-wise (eg: checkpoints are a big requirement).

When we have that figured out, the end-game is to split control points into a more generic "ControlableObjective" or similar, and have both control point and payload extend that. The controlable objective will hold most if not all of the logic of a control point, like progression, capture rules, decay, etc, while control point/payload will handle (each in their own way) tracking what players are considered in-the-region, how they display, scoreboard, etc

Pablete1234:
Should be ready to be playtested, even if not final, updated original comment to reflect extra stuff.

Electroid (approved):
To enable this feature, you'll need to put the following in your `config.yml`:

```yaml
experiments:
  payload: true
```

PR #1060, opened 2022-09-24
Optimize scoreboard & allow ratelimit

Pablete1234:
Allows scoreboard being rate-limiting under stress situations, in a similar fashion to tablist, where if the renders take long or if the server runs low on TPS it will debounce for longer periods (up to 1 second). This reuses logic originally used in tablist to debounce (in that case, up to 5s), which has now been put in a RateLimiter util, and reused in both places.

The reason for this change is that specifically for ffa, where sidebar renders player names, it can take a non-trivial amount of text serialization/deserialization, which can be costly. Additionally, names have been changed to use the simpler just-color version of the name, which should be faster than the full version with flairs and hover events etc.


Electroid (approved):
Nice refactor, looks good.

PR #1061, opened 2022-09-25
Add item replace action

Pablete1234:
An action that will find item-stacks on the player inventory matching a certain item matcher, and will replace it with a replacement.
You may keep-amount (meaning you get as much as the new item as you had of the previous), and/or keep-enchants (meaning the enchants on the old item will be applied to the replacement before giving it to the user.

```xml
<replace-item keep-amount="false" keep-enchants="false" ignore-metadata="true" amount="2..5">
  <find material="stone"/>
  <replace material="dirt" amount="8"/>
</replace-item>
```

The item-matcher rules are the same as for `<holding ignore-metadata="true"> <item ...> </holding>` since the item-matchers pr #1055 , but the inner matcher is called "find" instead of "item", so it supports things like matching for amount-ranges. 

~~This PR has not once been ran, do not merge until it is tested.~~

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -3,11 +3,13 @@ package tc.oc.pgm.action;
 import com.google.common.collect.ImmutableList;
 import java.lang.reflect.Method;
 import java.util.Map;
+import org.bukkit.inventory.ItemStack;
 import org.jdom2.Element;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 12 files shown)


Electroid (dismissed):

Pablete1234:
Implemented support for `team-color="true"` on the replace action, and while doing it, also added support for several blocks to use that (in kits, shops, killrewards, and item-replace actions), like wool, carpet or stained glass/clay.

This now has received proper testing and is ready to merge

Electroid (approved):

PR #1069, opened 2022-10-08
Implement blitz join filter

Pablete1234:
A while ago blitz filters were implemented, where they can decide if blitz lives are subtracted or not, this additionally adds support for deciding if players should be prevented from joining the match based on a filter:

```xml
<blitz lives="3" filter="after-1m" join-filter="before-1m"/>
```
In the example, after 1 minute players start losing lives when they die. Joining is allowed during that first minute.

The `join-filter` defaults to always DENY, meaning it will not let anyone join after match start. You may change it to a filter that says when it is allowed to join. 

The PR has been tested and works as intended.


Electroid (approved):
Cool feature, excited.

Issue #835, opened 2021-03-20
Block Drop's trample attribute no longer works


CrazyisCreeps:
I brought it up in the OCC mapdev discord channel but I wanted to make sure it won't be forgotten and open up an issue. As of right now, it doesn't appear that the trample attribute is getting triggered. I suspect it's related to the last change made that affected this attribute, where observers were able to trigger events (https://github.com/PGMDev/PGM/pull/795)

Pablete1234:
I think the issue of them not working is probably related to what was fixed in #1086 
Feel free to re-open if you can still replicate the issue, try to include a map where this happens if that's the case

Issue #796, opened 2021-02-05
Cause filter does not work for trample


PabloPintor:
Event for block trampling generates an event that cannot be filtered with the cause filter, because GeneralizingEvent always gets the original cause (like PlayerCoarseMoveEvent).

https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java#L79
https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/filters/CauseFilter.java#L122

PR #1073, opened 2022-10-09
Add break attribute to Block Drops Rule

CoWinkKeyDinkInc:
This PR adds a `break` attribute to the Block Drops Rules that allows better fine tuning of block drops and fixes an issue described in #835 where only affecting trampled blocks stopped working. You used to be able to add `<cause>trample</cause>` into the filter to do the same effect, but this doesn't work anymore as mentioned in #796 (the filter always returns deny).

```xml
<block-drops>
    <!-- break defaults to true -->
    <!-- any stone blocks stepped on by the player turns into dirt -->
    <rule trample="true" break="false" region="everywhere">
        <filter>
            <all>
                <material>stone</material>
                <!-- <cause>trample</cause> stopped working -->
                <any>
                    <sprinting/>
                    <walking/>
                    <crouching/>
                </any>
            </all>
        </filter>
        <!-- Replaces the affected block -->
        <replacement>dirt</replacement>
        <!-- Drop emerald ore when player steps on stone block -->
        <drops>
            <item material="emerald ore"/>
        </drops>
    </rule>
</block-drops>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsModule.java
+++ b/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsModule.java
@@ -72,6 +72,7 @@ public class BlockDropsModule implements MapModule {
 
         boolean dropOnWrongTool =
             XMLUtils.parseBoolean(Node.fromChildOrAttr(elRule, "wrong-tool", "wrongtool"), false);
+        boolean breakable = XMLUtils.parseBoolean(Node.fromChildOrAttr(elRule, "break"), true);
         boolean punchable = XMLUtils.parseBoolean(Node.fromChildOrAttr(elRule, "punch"), false);
         boolean trample = XMLUtils.parseBoolean(Node.fromChildOrAttr(elRule, "trample"), false);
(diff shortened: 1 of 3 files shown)


Pablete1234:
This seems like a poor hack that won't actually solve anything. If i'm reading it correctly, setting break to false makes it so block transform event (breaking blocks) is ignored unless break is true.
If the aim is to fix cause -> trample being broken, investigate the root issue

CoWinkKeyDinkInc:
Trample does work, its just that it will also affect broken blocks too unless you add `<cause>trample</cause>` into the filter. I am guessing that the cause filter stopped working and that @CrazyisCreeps might have been mistaken thinking that trample is fully broken. This PR isn't a hack. Either way, adding this `break` attribute is a more cleaner solution to only changing blocks when trampled.

Pablete1234:
The solution to `<cause>trample</cause>` being broken, is to investigate that issue and fix it, rather than to add an extra flag that needs to be set to work-around the bug

Pablete1234:
Replaced by #1086

PR #1076, opened 2022-10-10
Implement blocks filter & refactor regions

Pablete1234:
Implements a new filter type, **blocks**.
They allows for complex structures where you let players freely edit the air blocks (or other blocks) but prevent then from breaking the original blocks. PGM will on match load check all blocks that match the filter, and reuse them later to allow/deny queries against the filter.

```xml
<union id="woolroom-regions">
  <cuboid min="-50,0,-110" max="-20,20,-100"/>
  <cuboid min="-50,0,100" max="-20,20,110"/>
</union>

<deny id="deny-woolroom-blocks">
  <blocks region="woolroom-regions">
    <not>
      <any>
        <material>air</material>
        <material>web</material>
      </any>
    </not>
  </blocks>
</deny>

<apply block="deny-woolroom-blocks" region="woolroom-regions"/>
```
In this example, the blocks within the woolroom region that aren't air or cobwebs, are made part of the blocks filter. They will be denied from breaking/replacing those blocks in any way. 

The region is required, the filter may be specified as a `filter` attribute or a child filter (as shown in the example). Filter is not required and will default to a not-air filter.

Keep in mind there can be performance considerations, with the new filter, avoid making unnecessarily big regions, when possible make smaller regions and join them in an union as pgm will handle them better.

The PR has been tested and works as intended.


Electroid (approved):
Nice refactor, looks good!

Issue #66, opened 2019-11-09
Tab completion not working in map commands


ChemistryX:

Electroid:
Still an issue?

Brottweiler:
Yes.

Pugzy:
I have attempted to make the changes to allow this on my branch https://github.com/Pugzy/PGM/commit/ae6d1708de563eaf93885905a3ac5075fcd67b51.

Using the same fancy hyphen character as OCN did to replace spaces.

![image](https://user-images.githubusercontent.com/8608892/69014655-6dc23a80-0984-11ea-96f0-399f9e1f1684.png)

Does anyone recall what OCN would do when tab completing something such as `The Ni` [tab] ... le.
Once I have added the space results are no longer suggested as it is a different argument.

Pablete1234:
If you t ried to tab-in `A Ne` it would tab-in as `A New┈Day`, the vanilla mc client won't replace text already added by the user, and that includes the space, wich will be left unmodified. The issue is, intake will likely not handle this very well had similar issues with map tags where only the first tag will auto-complete, anything after the first space never gets thru any `getSuggestions`

rafibaum:
@Pugzy How far did you get with this?

Pugzy:
@rafibaum got some of it working on https://github.com/Pugzy/PGM/commit/ae6d1708de563eaf93885905a3ac5075fcd67b51 ran into the issue above described by Pablo.

PR #1080, opened 2022-10-14
Migrate to cloud framework

Pablete1234:
This is still WIP (a couple command classes still comented out) but already very functional and usable version of pgm using the cloud command framework (https://github.com/Incendo/cloud).

The main if not only regression is that flags are not (yet) supported in the middle of the command, so they expect to always be at the end. However i've already made a proof-of-concept fix for this in https://github.com/Incendo/cloud/commit/430beb7e759c72b8949db276fe985f44f1c72fea, so i'm hoping a fix can either be upstream soon, or we'll have to run a fork until it gets merged.

The benefits are several, from better autocompletion, to support for "MatchPlayer" as both injection (context/sender) and argument (ie: targeting someone else), cleaner commands, better tab-completion, and greedy strings (like /g, /msg etc) no longer replace quotes or prevent `-f` being typed as it's detected as a flag.

Example of greedy text being auto completed by fuzzy-matching (note how entered text is gold not golden):

https://user-images.githubusercontent.com/11789291/195868600-b991e738-017c-4632-9896-287a2db875b6.mp4


core/src/main/java/tc/oc/pgm/command/CycleCommand.java
```diff
@@ -38,14 +37,14 @@ public void cycle(
...
   }
 
-  @Command(
-      aliases = {"recycle", "rematch"},
-      desc = "Reload (cycle to) the current map",
-      usage = "[seconds]",
-      flags = "f",
-      perms = Permissions.START)
+  @CommandMethod("recycle|rematch [duration]")
+  @CommandDescription("Reload (cycle to) the current map")
```
Electroid, line 41: Always wanted to see if we can make these translatable. Do you think it’s possible with cloud?

Pablete1234: the framework is flexible enough that i'm certain it can be done, but it requires using our own annotation

Pablete1234:
This has moved over to ready for review. The migration is complete, although some things no longer work as they used to. There's an upstream PR for cloud to allow flag parsing in any location (https://github.com/Incendo/cloud/pull/395) that is required for things like `/cycle 0 -f map` to work (currently only `/cycle 0 map -f` works).

Besides that one issue, everything else is smoother than it used to, and map autocompletion has gotten alot of work, now very wild things auto-complete nicely and predictably. Things like `/sn gdIII` *will* result in `Golden Drought III` being selected.

core/src/main/java/tc/oc/pgm/command/CycleCommand.java
Citymonstret: Do you want them to be globally translatable or per sender? There are different approaches depending on what you need :)

Pablete1234: We handle all translations in PGM per-sender, via adventure. We have our own systems for that here. Ideally we'd simply make the descriptions be an adventure component created from a translatable with a key specified in the `@CommandDescription` annotation, then in MinecraftHelp make sure it uses the components

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -338,13 +335,11 @@ public InventoryManager getInventoryManager() {
...
-        config.isCommunityMode() ? new CommunityCommandGraph() : new CommandGraph();
-
-    graph.register(vanishManager);
-    graph.register(ChatDispatcher.get());
-
-    new CommandExecutor(this, graph).register();
+    try {
+      new CommandGraph(this);
+    } catch (Exception e) {
+      getLogger().log(Level.SEVERE, "Exception registering commands", e);
```
Electroid, line 341: Is this fatal? Should we disable PGM if this fails?

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
```diff
@@ -0,0 +1,21 @@
+package tc.oc.pgm.api.integration;
+
+import org.bukkit.entity.Player;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.PGM;
+
+public interface Integration {
+
+  public static boolean isFriend(Player a, Player b) {
```
Electroid, line 9: How do you implement this if it's static?

core/src/main/java/tc/oc/pgm/api/region/Region.java
```diff
@@ -140,8 +138,7 @@ default Iterable<BlockVector> getBlockVectors() {
   }
 
   default Stream<BlockVector> getBlockPositions() {
-    return StreamSupport.stream(
-        Spliterators.spliterator(getBlockVectorIterator(), 0, Spliterator.ORDERED), false);
+    return StreamUtils.of(getBlockVectorIterator());
```
Electroid, line 143: Unrelated diff?

core/src/main/java/tc/oc/pgm/command/ClassCommand.java
```diff
@@ -5,27 +5,41 @@
...
-  @Command(
-      aliases = {"class", "selectclass", "c", "cl"},
-      desc = "Select your class")
-  public void classSelect(Match match, MatchPlayer player, @Maybe @Text String query) {
+  @CommandMethod("class|selectclass|c|cl [class]")
+  @CommandDescription("Select your class")
+  public void classSelect(
+      Match match,
+      MatchPlayer player,
+      @Argument(value = "class", suggestions = "pgm-classes") @FlagYielding String query) {
```
Electroid: Why can't this just be `PlayerClass`?

core/src/main/java/tc/oc/pgm/command/ClassCommand.java
```diff
@@ -101,7 +133,7 @@ public void classList(Match match, MatchPlayer player) {
   private ClassMatchModule getClasses(Match match) {
     final ClassMatchModule classes = match.getModule(ClassMatchModule.class);
     if (classes == null) {
-      throw exception("match.class.notEnabled");
+      throw new CommandExecutionException(exception("match.class.notEnabled"));
```
Electroid: This seems pretty annoying to do every time, is there a way we can wrap this at a common layer?

core/src/main/java/tc/oc/pgm/command/CycleCommand.java
jpenilla: With builders you can set a component in the command meta directly (see MinecraftExtrasMetaKeys), with annotations you probably need a command builder modifier to retrieve component instances by key or something. Can either use the CommandDescription itself to store the intermediary key, or could add your own annotation and meta value

KingOfSquares:
Am i understanding correctly that this closes #66 ?

Pablete1234:
> Am i understanding correctly that this closes #66 ?

it does

core/src/main/java/tc/oc/pgm/command/ClassCommand.java
Pablete1234: it can be if i make a parser for it, but i didn't make one because it's used in a single place

core/src/main/java/tc/oc/pgm/command/ClassCommand.java
Pablete1234: Sadly cloud will only allow exception handling for a few exceptions, (CommandExecutionException being one of them), so we have to use that. 

This is generally not relevant for most places that we use exceptions on, as any exception in the method of the command itself will already be appropriately wrapped, but the issue here is this is called by the suggester too, and tab-completion code will not automatically catch your exceptions and re-throw them wrapped, meaning they reach the exception handler as-is, and exception handler does not recognize them as one of the handle-able ones, so they go to a default handling code-path that we cannot hook into.

core/src/main/java/tc/oc/pgm/api/region/Region.java
Pablete1234: i've added `StreamUtils.of`, and migrated other places using `StreamSupport.stream` to it, since one is a shorthand for the other

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
Pablete1234: Community branch does this already, but essentially like:
```java
public interface Integration {

  FriendIntegration FRIENDS;

  public static boolean isFriend(Player a, Player b) {
    return FRIENDS.isFriend(a, b);
  }

}
```

Those are part of the community branch tho, and it wouldn't make sense to bring all of community into here, just a simple no-op defined to be able to make the rest of the code be explicit about friendships affecting commands (tab-completing nicked names etc)

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Pablete1234: It's not *tenichally* fatal, the plugin *will* work without commands.... just a very poor experience if literally no commands at all are able to run

Pablete1234:
Pushed a few changes, made parsers for PlayerClass & SettingKey. Setting keys will suggest & parse based on the previous argument for setting, so if you `/toggle chat ` it will suggest admin/global/team while if you `/toggle effects ` it suggests off/on.


Additionally since https://github.com/Incendo/cloud/pull/395 doesn't seem to be getting any progress, i'm considering making PGM run a cloud fork until the changes get upstreamed so that this PR doesn't keed to keep waitning.

Electroid (dismissed):
@Pablete1234 If you resolve the conflict, this is good to merge.

PR #1082, opened 2022-10-16
Parse Blitz: Rage Gamemode Automatically

CoWinkKeyDinkInc:
This PR allows PGM to display the Blitz: Rage gamemode automatically without having to use `<gamemode>br</gamemode>` to do so. ~~This also adds a `#blitz-rage` tag to make browsing maps a bit easier. Blitz Rage maps will still be tagged separately with `#blitz` and `#rage`.~~

```xml
<!-- Displays "Blitz: Rage" instead of "Blitz and Rage" on scoreboard -->
<blitz>
    <lives>2</lives>
</blitz>
<rage/>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blitz/BlitzModule.java
+++ b/core/src/main/java/tc/oc/pgm/blitz/BlitzModule.java
@@ -5,7 +5,9 @@ import static tc.oc.pgm.util.Assert.assertNotNull;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Range;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.logging.Logger;
 import org.jdom2.Document;
 import org.jdom2.Element;
@@ -24,12 +26,17 @@ import tc.oc.pgm.util.xml.XMLUtils;
 
 public class BlitzModule implements MapModule {
 
-  private static final Collection<MapTag> TAGS =
-      ImmutableList.of(new MapTag("blitz", "Blitz", true, true));
+  private static final MapTag BLITZ = new MapTag("blitz", "Blitz", true, true);
+  // Sets Blitz into an internal MapTag that doesn't display on scoreboard
+  private static final MapTag BLITZ_IDENTIFIER = new MapTag("blitz", "Blitz", false, true);
+  private static final MapTag BLITZ_RAGE =
+      new MapTag("br", "blitz-rage", "Blitz: Rage", true, true);
   private final BlitzConfig config;
+  private final Collection<MapTag> tags;
 
-  public BlitzModule(BlitzConfig config) {
+  public BlitzModule(BlitzConfig config, Collection<MapTag> tags) {
     this.config = assertNotNull(config);
+    this.tags = tags;
   }
 
   @Override
@@ -39,7 +46,7 @@ public class BlitzModule implements MapModule {
 
   @Override
   public Collection<MapTag> getTags() {
-    return TAGS;
+    return tags;
   }
 
   public static class Factory implements MapModuleFactory<BlitzModule> {
@@ -58,6 +65,7 @@ public class BlitzModule implements MapModule {
       boolean broadcastLives = false;
       Filter filter = null;
       Filter joinFilter = null;
+      Set<MapTag> tags = new HashSet<>();
 
       FilterParser filters = factory.getFilters();
       for (Element blitzEl : blitzElements) {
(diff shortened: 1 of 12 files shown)


CoWinkKeyDinkInc:
`<gamemode>` currently still doesn't add maptags as it was deemed out of scope in #873, so maps that do not have to use  `<gamemode>br</gamemode>` should have this line removed so it can have the `#blitz-rage` tag.

Pablete1234:
I don't believe `#blitz-rage` should be a tag, you can query maps with `/maps -t blitz,rage` already, which will search maps with both tags, there's simply no point in having an unified tag.

core/src/main/java/tc/oc/pgm/blitz/BlitzModule.java
```diff
@@ -67,10 +75,19 @@ public BlitzModule parse(MapFactory factory, Logger logger, Document doc)
                 Node.fromChildOrAttr(blitzEl, "lives"), Integer.class, Range.atLeast(1), 1);
         filter = filters.parseProperty(blitzEl, "filter", StaticFilter.ALLOW);
         joinFilter = filters.parseProperty(blitzEl, "join-filter", StaticFilter.DENY);
+
+        if (doc.getRootElement().getChildren("rage") != null) {
```
Pablete1234: This is not how you search for dependencies' values. If you want a check like this you need to add a dependency from blitz to rage module, and in here get the rage module from the factory context to check if it's there.

core/src/main/java/tc/oc/pgm/blitz/BlitzModule.java
```diff
@@ -24,12 +26,17 @@
 
 public class BlitzModule implements MapModule {
 
-  private static final Collection<MapTag> TAGS =
-      ImmutableList.of(new MapTag("blitz", "Blitz", true, true));
+  private static final MapTag BLITZ = new MapTag("blitz", "Blitz", true, true);
+  // Sets Blitz into an internal MapTag that doesn't display on scoreboard
+  private static final MapTag BLITZ_IDENTIFIER = new MapTag("blitz", "Blitz", false, true);
```
Pablete1234: This just feels like a very dirty workarround for how maptags should work

core/src/main/java/tc/oc/pgm/rage/RageModule.java
```diff
@@ -23,15 +30,23 @@ public MatchModule createMatchModule(Match match) {
...
   public static class Factory implements MapModuleFactory<RageModule> {
     @Override
     public RageModule parse(MapFactory factory, Logger logger, Document doc)
         throws InvalidXMLException {
+      Set<MapTag> tags = new HashSet<>();
       if (doc.getRootElement().getChild("rage") != null) {
-        return new RageModule();
+        // If the blitz module is loaded with Rage, display "Blitz: Rage" on scoreboard instead of
+        // "Blitz and Rage"
+        if (doc.getRootElement().getChild("blitz") != null) {
```
Pablete1234: yeah... both modules relying on eachother sounds like a circular dependency, and that's going to be annoying.

Also, you cannot just check for element because this **can** fail.

A map with the following will match this check, but will not actually have blitz enabled:
```
<blitz>
  <lives>oo</lives>
</blitz>
```

Pablete1234 (changes requested):
I've left it in the other comment, i generally don't agree with the *method* being used to try to fit this as a gamemode (ie: making a new tag that combines both). I think the tag system is fine as is, and they should remain 2 tags.
I do agree that the gamemode should be different, and should be auto-detected to show properly on scoreboard as "Blitz: Rage", just not from tags.

CoWinkKeyDinkInc:
I found a smarter way to do this where it gets the two tags on SidebarMatchModule, so now there isn't a `#blitz-rage` tag.

Pablete1234 (changes requested):
Blitz is primary and rage is auxiliary, you do not need all those lines of logic and iterating the maptags, you should be able to simply check at the veey last line (right before the return) if there's 2 tags, first is blitz and second is rage.

CoWinkKeyDinkInc:
[BlitzModule](https://github.com/PGMDev/PGM/blob/0cce36084b73e6cd1e025f1e5cd38301581b4d64/core/src/main/java/tc/oc/pgm/blitz/BlitzModule.java#L27) and [RageModule](https://github.com/PGMDev/PGM/blob/0cce36084b73e6cd1e025f1e5cd38301581b4d64/core/src/main/java/tc/oc/pgm/rage/RageModule.java#L16) are both auxiliary.
```java
private static final Collection<MapTag> TAGS =      //gm   aux
      ImmutableList.of(new MapTag("blitz", "Blitz", true, true));
```
```java
private static final Collection<MapTag> TAGS =
      ImmutableList.of(new MapTag("rage", "Rage", true, true));
```

core/src/main/java/tc/oc/pgm/scoreboard/SidebarMatchModule.java
```diff
@@ -312,6 +312,14 @@ private Component renderTitle(final Config config, final MapInfo map) {
       }
     }
 
+    // Display "Blitz: Rage" rather than "Blitz and Rage"
+    if (map.getTags().contains(new MapTag("blitz", "Blitz", true, true))
+        && map.getTags().contains(new MapTag("rage", "Rage", true, true))) {
+      games.add(translatable("gamemode.br.name").color(NamedTextColor.AQUA));
+      games.remove(translatable("gamemode.blitz.name").color(NamedTextColor.AQUA));
+      games.remove(translatable("gamemode.rage.name").color(NamedTextColor.AQUA));
+    }
```
Pablete1234: In total there will never be more than 2 games in the list, due to how `games.size() < 2` is the only thing that adds auxiliary tags (otherwise breaks) and primary gamemodes breaks with "generic" if multiple primaries are present. This means you can simply clear the games list and add blitz rage directly. 
Additionally, i think creating dummy objects just so they match what you want to remove, or what you want to "contains"-check is generally bad for what are essentially singleton objects (the map tags) which could very well be matched by identity-check, i think it's better to match by checking the id directly.

```suggestion
    if (games.size() == 2 && Stream.of("blitz", "rage").allMatch(
            id-> map.getTags().stream().anyMatch(mt -> mt.getId().equals(id))) {
      games.clear();
      games.add(translatable("gamemode.br.name").color(NamedTextColor.AQUA));
    }
```

Pablete1234 (changes requested):

Pablete1234 (approved):

Electroid (approved):

PR #1091, opened 2022-11-03
Make Gamemode name strings non-translatable

KingOfSquares:
If you are a translator and have any opinion on this, please share!

Now, my view:

When volunteers translate(which is the case for PGM), they tend to go for metaphrasing(literal translation). This is often the case because the paraphrasing alternative(using different words to convey the same meaning), takes more time and effort to be good. For most of PGMs strings this works great, but for strings that use terms instead of just words it gets problematic. 

To take an example, if I translate the words in the string "Capture the Flag" to Norwegian it becomes something like "Fang Flagget" or "Kapre Flagget". This is "correct", but I'm sure that if people reading this try to translate into their own non-english language they will experience the same weird feeling I do. The feeling that while the translated string is "correct", it does not contain the same meaning behind it.

The string "Capture the Flag" has itself become a term which means more than the words used to write it, and the same logic applies to pretty much all the game mode names/acronyms used in PGM. (As @Cazadorian pointed out to me, this logic can also be compared to the thought behind not translating brand names)

This means that when we are asking our translators to translate these terms what we really want is for them to create terms which make sense in their own language, not just literal translations. Some people manage this, but in most cases its just more confusing for the end user(the player) when they see the literal translations which in their mind have little to none connection to the actual game mode being played.


BGMP (commented):
Thanks for this PR, I knew this would be coming up eventually as either an issue or an attempt to change the way these specific translatable keys are handled.

I agree with the idea of making gamemode name strings non-translatable. As it was pointed out by @KingOfSquares, the idea of having a bunch of strings which, in one way or another, force translators to come up with accurate naming for their own language is just bad overall. The underlying problem I see is that in doing so we not only expect users to learn the gamemode names themselves (which is already difficult, PGM has quite a lot) but to also be aware of their own translation for each gamemode, which in most cases will just come off as odd or will not quite hold the meaning the original naming was intended to hold. This is specially unpleasant for old players.

I would also like to point out that there may be instances where translating might be necessary/could be allowed for specific languages. You could keep the translatable strings on the project and add a Glossary rule for them so that proofreaders are adviced not to translate unless strictly necessary. Just an idea though, removing them also works in my opinion.

Indicardo:
I agree.

CoWinkKeyDinkInc:
As someone who has done most of the work in making gamemode strings translatable, I would agree with hard coding the strings since we haven't come to a decision on how to best translate the more complicated ones I suppose. I think that the generic objectives name "Objectives" should still be kept as a translatable for certain contexts.

core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
```diff
@@ -1,31 +1,37 @@
...
+  CONTROL_THE_POINT("cp", "Control the Point", "CP"),
+  CAPTURE_THE_WOOL("ctw", "Capture the Wool", "CTW"),
+  DESTROY_THE_CORE("dtc", "Destoy the Core", "DTC"),
+  DESTROY_THE_MONUMENT("dtm", "Destroy the Monument", "DTM"),
+  FREE_FOR_ALL("ffa", "Free for All", "FFA"),
+  FLAG_FOOTBALL("ffb", "Flag Football", "FFB"),
+  KING_OF_THE_HILL("koth", "King of the Hill", "KotH"),
+  KING_OF_THE_FLAG("kotf", "King of the Flag", "KotF"),
+  MIXED("mixed", "Mixed", "Mixed"),
+  PAYLOAD("pd", "Payload", "PD"),
```
Pugzy: Can we switch `Payload` from `pd` to `payload` pd isn't a great abbreviation (I know it isn't 100% you).

Electroid (dismissed):
Seems reasonable, given that translating these is a hit-or-miss. Make the change to `payload` and will merge.

Electroid:
Also you'll need to rebase because #1082 was merged.

Electroid (dismissed):

Electroid (approved):

PR #1095, opened 2022-11-05
Multiple winner support in FFA tie

### LINKED ISSUE #195: Multiple winner support in FFA tie
When people tie in FFA (or any gamemode? not sure if it's only ffa or not) it says "Game Over!" instead of "Xerocoles and Notch wins!" for example, like it used to say.

KingOfSquares:
Closes #195 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
@@ -14,7 +14,6 @@ import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
@@ -38,7 +37,6 @@ import tc.oc.pgm.teams.Team;
 import tc.oc.pgm.util.LegacyFormatUtils;
 import tc.oc.pgm.util.named.MapNameStyle;
 import tc.oc.pgm.util.named.NameStyle;
-import tc.oc.pgm.util.named.Named;
 import tc.oc.pgm.util.text.TextFormatter;
 
 public class MatchAnnouncer implements Listener {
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
```diff
@@ -74,17 +77,19 @@ public void onMatchEnd(final MatchFinishEvent event) {
...
-                    ? "broadcast.gameOver.teamWinners"
-                    : "broadcast.gameOver.teamWinner",
-                event.getWinner().getName());
+                plural ? "broadcast.gameOver.teamWinners" : "broadcast.gameOver.teamWinner",
+                TextFormatter.list(
+                    winners.stream().map(Named::getName).collect(Collectors.toList()),
+                    NamedTextColor.WHITE));
 
-        if (event.getWinner() == viewer.getParty()) {
+        if (winners.stream().anyMatch(w -> w == viewer.getParty())) {
```
Pablete1234, line 94:
```suggestion
        if (winners.contains(viewer.getParty())) {
```
Any reason this can't simply be a contains check?

core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
```diff
@@ -102,9 +107,19 @@ public void onMatchEnd(final MatchFinishEvent event) {
...
+      Title.Times titleTimes = Title.Times.of(Duration.ZERO, fromTicks(40), fromTicks(40));
+
+      if (winners.size() > 3)
+        viewer.showTitle(
+            title(
+                translatable(
+                    "broadcast.gameOver.winners", text(winners.size(), NamedTextColor.YELLOW)),
+                subtitle,
+                titleTimes));
+      else viewer.showTitle(title(title, subtitle, titleTimes));
```
Pablete1234: given only the text itself changes, i'd consider inlining the if:

```suggestion
      viewer.showTitle(
          title(winners.size() <= 3 ? title : translatable("broadcast.gameOver.winners", text(winners.size(), NamedTextColor.YELLOW)),
              subtitle,
              titleTimes));
```

unsure on how much the formatter will murder this one tho

util/src/main/i18n/templates/match.properties
```diff
@@ -176,6 +176,9 @@ broadcast.gameOver.teamWinner = {0} wins!
 # {0} = winning team name (e.g. plural names, "Attackers" or "Defenders")
 broadcast.gameOver.teamWinners = {0} win!
 
+# {0} = the amount of winners, always a number (e.g. "5 winners")
+broadcast.gameOver.winners = {0} winners!
```
Electroid: I'd rather it just say "Tied!".

Examples:

`winners.size() == 0`
```
Tied!
```

`winners.size() <= 3` (use [`TextComponent#list`](https://github.com/PGMDev/PGM/blob/96bed42423d7802e74c0b608e8ab1f496885058d/util/src/main/java/tc/oc/pgm/util/text/TextFormatter.java#L41))
```
Tied!
ElectroidFilms and Pablete1234
```

```
Tied!
ElectroidFilms, Pablete1234, and KingOfSquares
```

`winners.size() > 3` (refactor [`TextComponent#list`](https://github.com/PGMDev/PGM/blob/96bed42423d7802e74c0b608e8ab1f496885058d/util/src/main/java/tc/oc/pgm/util/text/TextFormatter.java#L41) to have a `maxSize` option, if over, adds the "and N more" to the end)
```
Tied!
ElectroidFilms, Pablete1234, and 3 more
```

Electroid (changes requested):

util/src/main/i18n/templates/match.properties
Pablete1234: I believe the `and 3 more` to be a bad idea, you either show all the winners or show none, they all tied and so should have the same "recognition", there's no fair way to sort them better-to-worse (or else, you'd just have the highest-sorted as the one true winner)

Electroid: @Pablete1234 Fair enough, so we can leave out the "and more".

@KingOfSquares If it's more than 3 winners, don't show any subtitle.

core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
KingOfSquares: Did it to avoid a cast, do you think its worth it?

Pablete1234: Probably doesn't make a difference, up to you, but i'd leave a comment if you leave the stream in that it's to avoid an unchecked cast when calling contains

KingOfSquares:
Newest changes tested with success!

Electroid (approved):

PR #1105, opened 2022-11-24
Implement renderable components & fix player component

Pablete1234:
Implemented a new `RenderableComponent` interface, which allows components that render themselves (they define a `render` method) for each viewing player.

The first (and main) implementation for it, is `PlayerComponent`, which no longer is a fancy-wrapper over TextComponent, instead, it now returns a PlayerComponent, and will convert to TextComponents when sent to the audiences, based on who the viewer is. That means each viewer may see the name differently (eg: different friendship status, or may be able to see or not see nicks).

Audience now creates a BukkitAudiences with our custom renderer, which will try to render our custom components first, then delegate to the default adventure translating renderer.

Main changes:
 - `RenderableComponent` interface & `ComponentRenderer` (they're generic so we can do more custom components in the future, TeamComponent maybe?)
 - `PlayerComponent` now has a `render(CommandSender)` method, to implement the RenderableComponent interface
 - Adapted `TextTranslations` to translate/translateLegacy using the new renderer
 - Adapted `Audiences` to create a `BukkitAudiences` that uses the new renderer
 - Killed `NameStyle.CONCISE`, it only existed to be FANCY, but without revealing, now that's automatically done by the per-player rendering


Pablete1234:
Made it so we proper rendering logic for MatchPlayerState, meaning the *proper* state is rendered, not the player related to that state. That makes things like these work properly:

My view:
image

A 3rd-party non-friend view:
image

In current upstream PGM, that message would show as just 100%, and with whatever state the current player is (ie: nicked or not, and the current team, or potentially offline), while here each contribution properly shows under what player state (and identity) it occurred under.

Electroid (approved):

PR #1106, opened 2022-11-27
Implement PoC inline filters

Pablete1234:
Implements a proof-of-concept for the inline-filter concept. This is an idea i'd like to toy with, but that i'm still unsure as to how good/bad it will be, as it can simplify xml, but do so at the expense of the learning curve becoming harder.

Inline filters basically mean that anywhere a filter id is requested, you may instead write some expression that (internally) creates a filter.

## What this PR does (inline variable filters)

An example, this is **current** xml, we define a variable, a couple filters, and then use those filters (in this case, we use them in actions, but this works *anywhere* that expects a filter id):
```xml
<variables>
  <variable id="some_variable" scope="match"/>
</variables>
<filters>
    <variable id="is-exactly-5" var="some_variable">5</variable>
    <!-- Note that 5.. is alternative syntax for [5, oo) -->
    <variable id="is-5-or-more" variable="some_variable">5..</variable>
</filters>
<actions>
  <trigger filter="is-exactly-5" scope="match">
    <action><message text="The variable is exactly 5!"/></action>
  </trigger>
  <trigger filter="is-5-or-more" scope="match">
    <action><message text="The variable is 5 or more!"/></action>
  </trigger>
</actions>
```

That, can now (with this PR) be written as:
```xml
<variables>
  <variable id="some_variable" scope="match"/>
</variables>
<actions>
  <trigger filter="some_variable=5" scope="match">
    <action><message text="The variable is exactly 5!"/></action>
  </trigger>
  <trigger filter="some_variable=5.." scope="match">
    <action><message text="The variable is 5 or more!"/></action>
  </trigger>
</actions>
```
The difference? you no longer need to define filters for *each value* you want to test for the variable, they'll get created by PGM automatically when parsing. Additionally you no longer need to come up with extra IDs.

## Other inline filters (for the future)

This is a very specific PR that *just* covers variables, but the idea is to open up discussion for also &, | for all and any, ! for not, and maybe ~ for deny:
`filter="~(filter-a & !(filter-b | filter-c | some_variable=..5))"`
would be equivalent to:
```xml
<deny id="some-id-to-use-elsewhere">
  <all>
    <filter id="filter-a"/>
    <not>
      <any>
        <filter id="filter-b"/>
        <filter id="filter-c"/>
        <variable var="some_variable">..5</variable>
      </any>
    </not>
  </all>
</deny>
```

An example on a *real* map xml:
```xml
<filters>
    <not id="blue-in-wr">
        <any>
            <team id="only-red">red</team>
            <material>chest</material>
        </any>
    </not>
    <not id="red-in-wr">
        <any>
            <team id="only-blue">blue</team>
            <material>chest</material>
        </any>
    </not>
</filters>
<regions>
    <apply block="blue-in-wr" region="blue-wool-rooms"/>
    <apply block="red-in-wr" region="red-wool-rooms"/>
</regions>
```

Could be turned into:
```xml
<filters>
    <team id="only-red">red</team>
    <team id="only-blue">blue</team>
    <material id="chest">chest</material>
</filters>
<regions>
    <apply block="!(only-red | chest)" region="blue-wool-rooms"/>
    <apply block="!(only-blue | chest)" region="red-wool-rooms"/>
</regions>
```

There are however some added challenges to that:
 - On a technical level parsing these is quite a bit of work
 - It may be convenient, or it may be underused and just waste the work that went into it
 - It may increase the learning-curve for newcommers, documenting these well would be very important if we did them


Pablete1234:
I've trying to ask map developers for opinion on these, and the general consensus is that they just cannot give an opinion as they do not know if this will be helpful or not.

I believe these changes are small enough in scope that i can confidently say they won't become a burden to maintain, and the only behavior change that can occur is if someone was previously using a filter id with `word=number`, not a single map in the ocn/occ repositories match this so i consider it a non-breaking change.

So the proposal is to merge these changes, and then wait to see how its adoption takes shape, and then seeing if there's a demand for more inline filter like these or not.

Pablete1234:
@Electroid Could we get this reviewed and merged? I believe the impact code/maintainability-wise to be tiny for this simple implementation, and i believe it is a needed step to be able to test their usefulness in the real world. Even if inline filters never get past this, it will still be a useful small addition to simplify complex xmls

Electroid (approved):

PR #1108, opened 2022-11-28
Make module graph load order consistent

Pablete1234:
Currently PGM has a ModuleGraph to resolve dependencies, however, every so often a required dependency is forgotten by a developer, and then what? What happens is because module load order is arbitrary (hash maps), it may work or not work at random, based on server restarts. What makes it worse, is that because "most" dependencies are right, they will force the "right" module to load first **most** of the times, making it extremely difficult to debug or find the *one* missing link in the chain.

Example: modules B, C, D, E, F, G all rely on module A being loaded first. However, module F has forgotten to declare that dependency (the developer making the module forgot to declare it). If, arbitrarily, any of the modules A, B, C, D, E, or G are loaded *before* F, then the error will be completely hidden away, as any of them will force A to load first, and so, by the time F loads, A will already be present. However, in the one-in-a-thousand time, that F loads before them all, all pgm maps using F will fail to load.

This is extremely frustrating behavior to debug, so what this change does is make it so the order is no longer arbitrary, but instead, is consistent with the order modules are defined/registered in pgm. This means if a bug occurs, it will occur **every** time, and if it doesn't occur, it won't **ever** occur.

Additionally, i've cleaned-up the raw usages of MapModule without generics, and made the module factories immutable.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMPlugin.java
+++ b/core/src/main/java/tc/oc/pgm/PGMPlugin.java
@@ -27,7 +27,6 @@ import org.bukkit.plugin.java.JavaPlugin;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.Config;
 import tc.oc.pgm.api.Datastore;
-import tc.oc.pgm.api.Modules;
 import tc.oc.pgm.api.PGM;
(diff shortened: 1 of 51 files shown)


Electroid (approved):
Thanks for cleaning this up

PR #1110, opened 2022-11-28
Allow fireballs to be precise when shot

Pugzy:
Currently, fireballs have randomness when shot. Setting `precise` to true on the projectile fixes this.

`NMSHacks` needed as setting direction via API method call always has the random applied to it.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
@@ -7,6 +7,7 @@ import java.util.Set;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Entity;
+import org.bukkit.entity.Fireball;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
 import org.bukkit.entity.Projectile;
@@ -40,6 +41,7 @@ import tc.oc.pgm.filters.query.BlockQuery;
 import tc.oc.pgm.filters.query.PlayerBlockQuery;
 import tc.oc.pgm.kits.tag.ItemTags;
 import tc.oc.pgm.util.bukkit.MetadataUtils;
+import tc.oc.pgm.util.nms.NMSHacks;
 
 @ListenerScope(MatchScope.RUNNING)
 public class ProjectileMatchModule implements MatchModule, Listener {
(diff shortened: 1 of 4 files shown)


Electroid:
Should we just make this the default? Hard to imagine that the randomness is intended. (e.g. Not have a config value at all. If it's a fireball, just make it precise)

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -92,6 +94,10 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
           projectile =
               player.launchProjectile(
                   projectileDefinition.projectile.asSubclass(Projectile.class), velocity);
+          if (projectile instanceof Fireball && projectileDefinition.precise) {
+            NMSHacks.setFireballDirection(
+                (Fireball) projectile, player.getEyeLocation().getDirection());
```
Pablete1234:
```suggestion
                (Fireball) projectile, velocity);
```

Cazadorian:
> Should we just make this the default? Hard to imagine that the randomness is intended. (e.g. Not have a config value at all. If it's a fireball, just make it precise)

I think it'd likely be best to have it be configurable as although in most cases mapmakers would want no randomness, there may be some cases where it is actually desirable to keep(i.e to limit the effective range of regular fireballs or to make gun style projectiles that use either type not effectively sniper rifles).

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -69,6 +69,7 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
         Duration coolDown = XMLUtils.parseDuration(projectileElement.getAttribute("cooldown"));
         boolean throwable =
             XMLUtils.parseBoolean(projectileElement.getAttribute("throwable"), true);
+        boolean precise = XMLUtils.parseBoolean(projectileElement.getAttribute("precise"), false);
```
Electroid: Let's make `precise` default to `true`. Map makers can disable it.

calcastor: I think this is appropriate; there are little to no maps that use fireballs and "depend" on them being as janky as they tend to be.

Pablete1234 (approved):

Electroid (approved):

PR #1111, opened 2022-12-11
Support for Community (take 2️⃣) 

applenick:
This PR is a fresh take on the changes required to support [Community](https://github.com/PGMDev/Community)! Thanks to the recently merged PR #1105 by @Pablete1234, we were able to remove many unnecessary changes that had accumulated over the years. 

As a refresher, one of the primary changes within this PR is the further implementation of the `Integration` interface, allowing for third-party plugins, like Community, to seamlessly hook into PGM to provide information about friends, nicknames, vanish status, and more. Additionally this PR removes existing community features like moderation and reports.

View corresponding Community changes [here](https://github.com/PGMDev/Community/pull/35)

I'm opening this as a draft for now. Please let me know what changes I can make or if you have any questions. 


core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -208,9 +208,7 @@ public void onEnable() {
     matchManager = new MatchManagerImpl(logger);
 
     vanishManager =
```
Pablete1234: I don't think PGMPlugin should have a vanish manager at all, i think you should just:
`if (config.isVanishEnabled()) Integration.setVanish(new SimpleVanishManager(...));` and that's it
make it so it self-handles its own lifetime (registers events on creation?). 
Maybe it'd be nice for Integration to call some "disable" method on integrations if they're replaced, so SimpleVanishManager can un-register itself as listener if a replacement is set

core/src/main/java/tc/oc/pgm/api/Permissions.java
```diff
@@ -24,6 +24,7 @@ public interface Permissions {
   String JOIN = ROOT + ".join"; // Allowed to join a match as a participant
   String JOIN_CHOOSE = JOIN + ".choose"; // Can choose which team to join
   String EXTRA_VOTE = JOIN + ".extravote"; // Extra map voting power
+  String VOTE_MULTIPLIER = JOIN + ".votecount"; // Extra map voting multiplier
```
Pablete1234: wait up, why are these 2 under join?? `pgm.join.extravote`?

core/src/main/java/tc/oc/pgm/api/match/Match.java
```diff
@@ -407,6 +407,20 @@ default boolean finish() {
...
+   * @return True if friendly fire is on.
+   */
+  boolean getFriendlyFire();
+
+  /**
+   * Set an override for friendly fire
+   *
+   * @param allow True to allow friendly fire
+   */
+  void setFriendlyFireOverride(boolean allow);
```
Pablete1234:
```suggestion
  void setFriendlyFireOverride(Boolean allow);
```
true or false to set it to true or false, or null to remove the override

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -43,6 +47,8 @@ public enum SettingValue {
   EFFECTS_ON("effects", "on", DyeColor.GREEN), // Display special particle effects
   EFFECTS_OFF("effects", "off", DyeColor.RED), // Don't display special particle effects
 
+  MESSAGE_FRIEND("message", "friends", DyeColor.YELLOW), // Only accept friend direct messages
```
Pablete1234: uh, what?

this should be above, probs in the order all, friends, none

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -15,20 +15,24 @@ public enum SettingValue {
...
-  MESSAGE_ON("message", "on", DyeColor.GREEN), // Always accept direct messages
-  MESSAGE_OFF("message", "off", DyeColor.RED), // Never accept direct messages
+  MESSAGE_ON("message", "all", DyeColor.GREEN), // Always accept direct messages
+  MESSAGE_OFF("message", "none", DyeColor.RED), // Never accept direct messages
 
-  OBSERVERS_ON("observers", "on", DyeColor.GREEN), // Show observers
-  OBSERVERS_OFF("observers", "off", DyeColor.RED), // Hide observers
+  OBSERVERS_ON("observers", "all", DyeColor.GREEN), // Show observers
+  OBSERVERS_OFF("observers", "none", DyeColor.RED), // Hide observers
+  OBSERVERS_FRIEND("observers", "friends", DyeColor.YELLOW), // Only show friend observers
```
Pablete1234: probably all, friends, none too here

core/src/main/java/tc/oc/pgm/command/ListCommand.java
```diff
@@ -85,7 +84,7 @@ private void sendTeamInfo(
   private Component formatNames(Collection<MatchPlayer> players, CommandSender sender) {
     List<Component> names =
         players.stream()
-            .filter(mp -> sender.hasPermission(Permissions.STAFF) || !isVanished(mp.getId()))
+            .filter(mp -> sender.hasPermission(Permissions.STAFF) || !isVanished(mp.getBukkit()))
```
Pablete1234: We have an util for this: `Players.isVisible(sender, mp.getBukkit())`

core/src/main/java/tc/oc/pgm/command/VanishCommand.java
```diff
@@ -0,0 +1,34 @@
...
+      translatable("vanish.deactivate", NamedTextColor.GREEN);
+
+  @CommandMethod("vanish|v")
+  @CommandDescription("Toggle vanish status")
+  @CommandPermission(Permissions.VANISH)
+  public void vanish(MatchPlayer sender, @Flag(value = "silent", aliases = "s") boolean silent) {
+    if (Integration.getNick(sender.getBukkit()) != null) {
+      sender.sendWarning(translatable("vanish.deny.nick"));
+      return;
+    }
```
Pablete1234: I think this should go away, being nicked isn't an excuse to not be able to vanish/unvanish

core/src/main/java/tc/oc/pgm/command/util/CommandGraph.java
```diff
@@ -227,13 +226,8 @@ public void registerCommands() {
     register(new TeamCommand());
     register(new TimeLimitCommand());
     register(new VotingCommand());
+    register(new VanishCommand());
```
Pablete1234: this needs to be conditionally registered based on if vanish is enabled on config

core/src/main/java/tc/oc/pgm/integrations/NoopFriendIntegration.java
```diff
@@ -0,0 +1,12 @@
+package tc.oc.pgm.integrations;
+
+import org.bukkit.entity.Player;
+import tc.oc.pgm.api.integration.FriendIntegration;
+
+public class NoopFriendIntegration implements FriendIntegration {
```
Pablete1234: Personally i'd put the noop implementations directly in the interface, so it's `FriendIntegration.Noop`. They're small enough to just be simpler to maintain there

core/src/main/java/tc/oc/pgm/integrations/SimpleVanishManager.java
```diff
@@ -163,6 +153,7 @@ public void onJoin(PlayerJoinEvent event) {
     MatchPlayer player = matchManager.getPlayer(event.getPlayer());
     if (player == null) return;
     if (player.getParty() instanceof Competitor) return; // Do not vanish players on a team
+    if (!player.getBukkit().hasPermission(Permissions.VANISH)) return; // No perms
```
Pablete1234: if this check is done here, there's another one below which is no longer necessary

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -211,7 +197,8 @@ public void sendDirect(
       @NotNull MatchPlayer sender,
       @Argument("player") MatchPlayer receiver,
       @Argument("message") @Greedy String message) {
-    if (vanish.isVanished(sender.getId())) throw exception("vanish.chat.deny");
+    if (Integration.isVanished(sender.getBukkit())) throw exception("vanish.chat.deny");
+    if (Integration.isVanished(receiver.getBukkit())) throw exception("command.playerNotFound");
```
Pablete1234: This should never happen at all. the MatchPlayer parser itself will prevent this as you cannot tab-complete or have the player as an argument at all, *as long as* you're a player who cannot see the receiver. If it's another staff who *can* see you, then them sending you a message is fine tho.

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -365,12 +356,14 @@ private void asyncSendChat(
...
+      final @NotNull Set<Player> players,
+      final @NotNull Channel channel) {
     final AsyncPlayerChatEvent event =
         new AsyncPlayerChatEvent(false, sender.getBukkit(), message, players);
     event.setFormat(format);
     CHAT_EVENT_CACHE.put(event, true);
     match.callEvent(event);
+    match.callEvent(new ChannelMessageEvent(channel, sender.getBukkit(), message));
 
     if (event.isCancelled()) return;
```
Pablete1234: should probs only send the 2nd one if the 1st isn't cancelled

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -384,7 +377,7 @@ private void throwMuted(MatchPlayer player) {
     if (isMuted(player))
       throw exception(
           "moderation.mute.message",
-          text(muted.getOrDefault(player.getId(), ""), NamedTextColor.AQUA));
+          text(Integration.getMuteReason(player.getBukkit()), NamedTextColor.AQUA));
```
Pablete1234: mute reason is nullable, i'd double-check here and use a default fallback

core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
```diff
@@ -112,18 +113,24 @@ public void onMatchEnd(final MatchFinishEvent event) {
...
     // Thus, we delay sending this prominent message, so it is more likely its in the right locale.
-    player
-        .getMatch()
+    match
         .getExecutor(MatchScope.LOADED)
-        .schedule(() -> sendWelcomeMessage(event.getPlayer()), 500, TimeUnit.MILLISECONDS);
+        .schedule(
+            () -> match.callEvent(new MatchJoinMessageEvent(match, player)),
+            500,
+            TimeUnit.MILLISECONDS);
```
Pablete1234: Probably quite an edge case, but this means you could join a match, it cycles to the new match, and *then* you get the message from the first match. Quite an interesting potential race condition.

core/src/main/java/tc/oc/pgm/rotation/vote/MapPoll.java
```diff
@@ -62,6 +63,7 @@ public class MapPoll {
   static final ItemTag<String> VOTE_BOOK_TAG = ItemTag.newString(VOTE_BOOK_METADATA);
 
   private final WeakReference<Match> match;
+  private VotingBookCreator bookCreator;
```
Pablete1234: i'd make it `private static` and provide a public static setter method (non-null), and avoid the create event all together

core/src/main/java/tc/oc/pgm/rotation/vote/MapPoll.java
```diff
@@ -196,18 +205,27 @@ private MapInfo getMostVotedMap() {
...
-    return uuids.stream()
-        .map(Bukkit::getPlayer)
-        // Count disconnected players as 1, can't test for their perms
-        .mapToInt(p -> p == null || !p.hasPermission(Permissions.EXTRA_VOTE) ? 1 : 2)
-        .sum();
+    return uuids.stream().map(Bukkit::getPlayer).mapToInt(this::calcVoteMultiplier).sum();
+  }
+
+  private int calcVoteMultiplier(Player player) {
+    // Count disconnected players as 1, can't test for their perms
```
Pablete1234: misplaced comment, this should be above the return 1, or you can swap the if around to:
`if (player == null) return 1;` then keep the coment

core/src/main/java/tc/oc/pgm/rotation/vote/MapPoll.java
```diff
@@ -179,12 +179,21 @@ private Component getMapBookComponent(MatchPlayer viewer, MapInfo map) {
...
     if (votes == null) throw exception("map.notFound");
 
-    if (votes.add(player)) return true;
-    votes.remove(player);
-    return false;
+    if (votes.add(player)) {
+      addVote = true;
+    } else {
+      votes.remove(player);
+    }
```
Pablete1234:
```suggestion
    boolean added = votes.add(player);
    if (!added) votes.remove(player);
```

core/src/main/java/tc/oc/pgm/rotation/vote/MapPoll.java
```diff
@@ -196,18 +205,27 @@ private MapInfo getMostVotedMap() {
   }
 
   /**
-   * Count the amount of votes for a set of uuids. Players with the pgm.premium permission get
-   * double votes.
+   * Count the amount of votes for a set of uuids. Players with the
```
Pablete1234: broken comment?

core/src/main/java/tc/oc/pgm/rotation/vote/VotePoolOptions.java
```diff
@@ -3,14 +3,16 @@
...
+import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.map.MapInfo;
 import tc.oc.pgm.rotation.pools.VotingPool;
 
 public class VotePoolOptions {
 
-  // Set of maps to be used in custom vote selection
-  private final Set<MapInfo> customVoteMaps;
+  // Set of maps w/ related data to be used in custom vote selection
+  private final Set<CustomMapEntry> customVoteMaps;
```
Pablete1234: I don't like this one, because it no longer guarantees there aren't any duplicates, potentially change to Set<MapInfo, UUID> instead

core/src/main/java/tc/oc/pgm/tablist/MatchTabManager.java
```diff
@@ -318,4 +319,10 @@ public void onPlayerNameChange(NameDecorationChangeEvent event) {
     TabEntry entry = getPlayerEntryOrNull(Bukkit.getPlayer(event.getUUID()));
     if (entry instanceof DynamicTabEntry) ((DynamicTabEntry) entry).invalidate();
   }
+
+  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+  public void onForceTabRefresh(RefreshPlayerTabEntryEvent event) {
```
Pablete1234: Consider we already have `NameDecorationChangeEvent` this seems to be very similar to this, could just reuse the same one?

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
applenick: Whoops! Fixed. I think when we finish this up & deploy to OCC, we're gonna have to do a full reset of existing settings. Although inconvenient, at least the defaults will be better than everyone's being slightly random.

core/src/main/java/tc/oc/pgm/command/util/CommandGraph.java
applenick: Good point! So when third-party plugins set a vanish integration, how should that be handled? Ideally making it so the command is the responsibility of the plugin owning the vanish integration? Or do we always want to use the `VanishCommand` defined in PGM, and just have a method in the Integration to register/unregister it?

Pablete1234: It would be nice if we could make pgm self-register & self-unregister when a vanish manager is set, but that's kinda complicated when commands are involved.
I'd just keep it as part of the config, if you don't want conflicting commands between pgm's /v and your own plugins', just disable it in the config.

applenick: Gotcha! That makes sense 👍

core/src/main/java/tc/oc/pgm/api/Permissions.java
Electroid: Let's introduce:

* `pgm.vote` - can you vote? (defaults to everyone)
* `pgm.vote.extra` - do you get 1 extra vote? (defaults to `pgm.premium`)
* `pgm.vote.multipler` - I forget what this does

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
```diff
@@ -1,21 +1,64 @@
...
+      new AtomicReference<FriendIntegration>(new NoopFriendIntegration());
+  static final AtomicReference<NickIntegration> NICKS =
+      new AtomicReference<NickIntegration>(new NoopNickIntegration());
+  static final AtomicReference<VanishIntegration> VANISH =
+      new AtomicReference<VanishIntegration>(PGM.get().getVanishManager());
+  static final AtomicReference<PunishmentIntegration> PUNISHMENTS =
+      new AtomicReference<PunishmentIntegration>(new NoopPunishmentIntegration());
+
+  public static void setFriendIntegration(FriendIntegration integration) {
+    FRIENDS.set(integration);
```
Electroid: Make sure to add `requireNonNull` checks here

core/src/main/java/tc/oc/pgm/api/integration/FriendIntegration.java
```diff
@@ -0,0 +1,16 @@
+package tc.oc.pgm.api.integration;
+
+import org.bukkit.entity.Player;
+
+/** FriendIntegration - Determine whether two players are friends * */
+public interface FriendIntegration {
```
Electroid, line 6: Let's annotate these integrations with `@FunctionalInterface`

core/src/main/java/tc/oc/pgm/api/match/Match.java
Electroid: Let's just call this `setFriendlyFire`

core/src/main/java/tc/oc/pgm/api/match/event/MatchJoinMessageEvent.java
```diff
@@ -0,0 +1,40 @@
+package tc.oc.pgm.api.match.event;
+
+import com.google.common.collect.Lists;
+import java.util.List;
+import net.kyori.adventure.text.Component;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.player.MatchPlayer;
+
+public class MatchJoinMessageEvent extends MatchEvent {
```
Electroid: Is there a `MatchJoinEvent`? If so, let's just use that and add the `List<Component> extraMessage` to that.

core/src/main/java/tc/oc/pgm/api/match/event/MatchVoteFinishEvent.java
```diff
@@ -0,0 +1,32 @@
+package tc.oc.pgm.api.match.event;
+
+import org.bukkit.event.HandlerList;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.map.MapInfo;
+import tc.oc.pgm.api.match.Match;
+
+public class MatchVoteFinishEvent extends MatchEvent {
```
Electroid, line 8: Why is this event needed?

core/src/main/java/tc/oc/pgm/api/party/Competitor.java
```diff
@@ -17,4 +17,11 @@ public interface Competitor extends Party {
    * @return a visibility
    */
   NameTagVisibility getNameTagVisibility();
+
+  /**
+   * Sets an override for name tag visibility
+   *
+   * @param visibility A {@link NameTagVisibility}
+   */
+  void setNameTagVisibilityOverride(NameTagVisibility visibility);
```
Electroid: `setNameTagVisibility`

core/src/main/java/tc/oc/pgm/api/player/VanishManager.java
```diff
@@ -1,10 +1,10 @@
 package tc.oc.pgm.api.player;
 
 import java.util.Collection;
-import java.util.UUID;
+import tc.oc.pgm.api.integration.VanishIntegration;
 
 /** A manager that holds information related to {@link MatchPlayer}s who are vanished. */
-public interface VanishManager {
+public interface VanishManager extends VanishIntegration {
```
Electroid: Why does `VanishManager` need to be separate from `VanishIntegration`? Can it just all be put within `VanishIntegration`?

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -15,20 +15,24 @@ public enum SettingValue {
   CHAT_ADMIN("chat", "admin", DyeColor.RED), // Send to all server operators
 
   DEATH_OWN("death", "own", DyeColor.RED), // Only send death messages involving self
-  DEATH_ALL("death", "all", DyeColor.GREEN), // Send all death messages, highlight your own
+  DEATH_ALL("death", "all", DyeColor.ORANGE), // Send all death messages, highlight your own
```
Electroid, line 18: All of these new settings need to be added at the bottom of the enum list. That is why there has been weird bugs with settings being lost. Settings are stored with a bitset which is calculated based on the enum's ordinal. This will change the ordinals for existing settings, unless added at the very end.

core/src/main/java/tc/oc/pgm/command/VanishCommand.java
Electroid: Agreed, we should make this possible.

core/src/main/java/tc/oc/pgm/integrations/SimpleVanishManager.java
```diff
@@ -101,18 +98,23 @@ public boolean setVanished(MatchPlayer player, boolean vanish, boolean quiet) {
       match.setParty(player, match.getDefaultParty());
     }
 
+    // Broadcast join/quit message before adjusting vanish status
+    // so name renders normally
+    if (!quiet) {
+      if (vanish) {
+        JoinLeaveAnnouncer.leave(player, JoinVisibility.NONSTAFF);
```
Electroid: It feels like it would be better if `JoinLeaveAnnouncer` listened to `PlayerVanishEvent` and did this from there?

core/src/main/java/tc/oc/pgm/listeners/JoinLeaveAnnouncer.java
```diff
@@ -0,0 +1,113 @@
...
+    MatchPlayer player = this.mm.getPlayer(event.getPlayer());
+    if (player == null) return;
+
+    if (event.getQuitMessage() != null) {
+      event.setQuitMessage(null);
+      leave(player, JoinVisibility.get(player));
+    }
+  }
+
+  public static void join(MatchPlayer player, JoinVisibility visibility) {
```
Electroid: Instead of exposing `static` methods, it would be better if `JoinLeaveAnnouncer` was able to listen to all the events it needed.

core/src/main/java/tc/oc/pgm/listeners/MatchAnnouncer.java
Electroid: True, though rare. I think it's okay to leave for now.

core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
```diff
@@ -31,30 +30,46 @@ public abstract class MapPool implements MapOrder, Comparable<MapPool> {
...
+        section.getBoolean("enabled"),
+        section.getInt("players"),
+        section.getBoolean("dynamic", true),
+        parseDuration(section.getString("cycle-time", "-1s")),
+        Collections.unmodifiableList(
+            section.getStringList("maps").stream()
+                .map(mapName -> getMap(PGM.get().getMapLibrary(), name, mapName))
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList())));
+  }
```
Electroid: Doing config parsing in a constructor is generally not good. I would refactor this into a separate static method `MapPool.parse(ConfigurationSection)`

util/src/main/i18n/templates/command.properties
```diff
@@ -110,3 +110,12 @@ command.message.blocked = {0} is not accepting messages at this time.
 
 # {0} = map name
 command.maps.hover = Click to view {0}
+
+command.noResults = No results match!
```
Electroid: No results found!

core/src/main/java/tc/oc/pgm/api/integration/FriendIntegration.java
Pablete1234: These are definetly *not* functional interfaces. They're interfaces which just happen to have a single method... for now. Nothing prevents them getting more methods in the future and we shouldnt set an interface that says otherwise

core/src/main/java/tc/oc/pgm/api/match/event/MatchVoteFinishEvent.java
Pablete1234: I assumed its sponsor vote refunds etc (you know.if the sponsored map won

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
Pablete1234: I believe we're all aware, issue is we still need decent ordering and we're willing to just wipe settings once to fix inconsistencies and start from scratch

core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
Pablete1234: Funnily enough, such method already exists, and it decides which of the constructors to call.
The issue is each map pool can have diff stuff read from config

core/src/main/java/tc/oc/pgm/api/player/VanishManager.java
applenick: Whoops! So actually we don't even need the VanishManager interface at all, removed it now.

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
applenick: Mentioned in my reply here: https://github.com/PGMDev/PGM/pull/1111#discussion_r1045294024 

Believe it's a fair trade-off to finally get all the new settings in a proper order 👍

core/src/main/java/tc/oc/pgm/api/match/event/MatchVoteFinishEvent.java
applenick: Correct! Required for Community to know if the sponsor's map vote won. If it's not ideal as a match event, an alternative could be renamed as `MapPollFinishEvent` and we can move it from the api to `rotation.vote.events` like the existing `MapPollVoteEvent`.

applenick:
@Pablete1234 & @Electroid thanks for all the feedback so far! I made the requested changes (minus the issues in discussion still open above). Whenever you two have a chance to review the latest changes it'd be appreciated 👍

core/src/main/java/tc/oc/pgm/api/integration/NickIntegration.java
```diff
@@ -13,4 +13,13 @@ public interface NickIntegration {
...
   String getNick(Player player);
+
+  static NickIntegration Noop = new NoopNickIntegration();
+
+  public class NoopNickIntegration implements NickIntegration {
+    @Override
+    public @Nullable String getNick(Player player) {
+      return null;
+    }
+  }
```
Electroid: I'd prefer not to have an implementation in the interface. I would just define these as anonymous classes in `Integration`.

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/api/setting/SettingValue.java
```diff
@@ -15,20 +15,25 @@ public enum SettingValue {
   CHAT_ADMIN("chat", "admin", DyeColor.RED), // Send to all server operators
 
   DEATH_OWN("death", "own", DyeColor.RED), // Only send death messages involving self
-  DEATH_ALL("death", "all", DyeColor.GREEN), // Send all death messages, highlight your own
+  DEATH_FRIENDS(
+      "death", "friends", DyeColor.GREEN), // Only send death messages involving yourself or friends
+  DEATH_ALL("death", "all", DyeColor.ORANGE), // Send all death messages, highlight your own
```
Pablete1234: There's an inconsistency here, observers/messages/join are all -> friends ->none, while death is own -> friends -> all

Maybe worth trying to keep the same order of either widening or restricting in each toggle? if that's the case, it should be own -> all -> friends for deaths, because the first has to be the default (own)

core/src/main/java/tc/oc/pgm/integrations/SimpleVanishIntegration.java
```diff
@@ -86,6 +82,11 @@ public List<MatchPlayer> getOnlineVanished() {
 
   @Override
   public boolean setVanished(MatchPlayer player, boolean vanish, boolean quiet) {
+    final Match match = player.getMatch();
+
+    // Call vanish event first so name renders existing state properly
+    match.callEvent(new PlayerVanishEvent(player, vanish, quiet));
```
Pablete1234: Well, actually...

Depending on if you're vanishing or un-vanishing, the message should be before or after the operation.

When vanishing, the message should be at the start, so, before you vanish.
When unvanishing, the message should be at the end, so you appear as online, instead of appearing with a strikethru or offline color.

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -381,10 +374,11 @@ private void asyncSendChat(
...
 
   private void throwMuted(MatchPlayer player) {
-    if (isMuted(player))
-      throw exception(
-          "moderation.mute.message",
-          text(muted.getOrDefault(player.getId(), ""), NamedTextColor.AQUA));
+    if (isMuted(player)) {
+      Optional<String> muteReason =
+          Optional.ofNullable(Integration.getMuteReason(player.getBukkit()));
+      throw exception("moderation.mute.message", text(muteReason.orElse(""), NamedTextColor.AQUA));
```
Pablete1234: maybe orElse("No reason provided") or smthg?

core/src/main/java/tc/oc/pgm/listeners/JoinLeaveAnnouncer.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  private boolean canView(MatchPlayer viewer, MatchPlayer target, JoinVisibility visibility) {
+    boolean isStaff = viewer.getBukkit().hasPermission(Permissions.STAFF);
+
+    SettingValue option = viewer.getSettings().getValue(SettingKey.JOIN);
+    boolean allowed =
+        option.equals(SettingValue.JOIN_ON)
+            || areFriends(option, viewer.getBukkit(), target.getBukkit());
```
Pablete1234, line 101: Could `if (!allowed) return false;`, and simplify all the succeeding conditions, as allowed would no longer be relevant

core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
```diff
@@ -31,30 +30,46 @@ public abstract class MapPool implements MapOrder, Comparable<MapPool> {
...
-  private MapInfo getMap(MapLibrary library, String mapName) {
+  private static MapInfo getMap(MapLibrary library, String poolName, String mapName) {
     @Nullable MapInfo map = library.getMap(mapName);
     if (map != null) return map;
     PGM.get()
         .getLogger()
-        .warning("[MapPool] [" + name + "] " + mapName + " not found in map repo. Ignoring...");
+        .warning("[MapPool] [" + poolName + "] " + mapName + " not found in map repo. Ignoring...");
     return null;
   }
```
Pablete1234: I feel it's weird to have a static method requiring passing the (non-static) pool name.

Given the nature of the change in the method, being due to it being used in the constructor, i think it makes more sense for this method to become a `List<MapInfo> buildMapList(List<String> mapNames, String pool)` where it handles the small code-block that maps each string to map info and returns the unmodifiable list.

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
```diff
@@ -1,21 +1,104 @@
...
+import static tc.oc.pgm.util.Assert.assertNotNull;
+
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import org.bukkit.entity.Player;
 import org.jetbrains.annotations.Nullable;
-import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.player.MatchPlayer;
 
 public interface Integration {
```
Electroid: Can we make this:

```java
public final class Integration {
  private Integration() {}
  // ...
}

**Comment by Electroid** [comment]
These changes look good, when it's ready I'll do a release since this will remove some community features.

**Comment by applenick** [comment]
> These changes look good, when it's ready I'll do a release since this will remove some community features.

Thanks @Electroid! Very pleased to hear things are looking good. 

I was doing a bit of testing and found a few small necessary changes that seemed to have gotten lost when switching to this new PR. Mainly related to scores & stats, those should be the last changes required for full Community compatibly.  

Review that last commit when you have a chance and let me know if there's anything else I can do to improve things. I'll get on it asap 👍

File: core/src/main/java/tc/oc/pgm/rotation/vote/events/MapPollVoteEvent.java [hunk 39]
```diff
@@ -0,0 +1,42 @@
+package tc.oc.pgm.rotation.vote.events;
+
+import java.util.UUID;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.map.MapInfo;
+
+public class MapPollVoteEvent extends Event {
```
Electroid: `public class MatchPlayerVoteEvent extends MatchPlayerEvent`

core/src/main/java/tc/oc/pgm/rotation/vote/VotePoolOptions.java
```diff
@@ -32,32 +35,42 @@ public boolean toggleMode() {
...
   }
 
   public Set<MapInfo> getCustomVoteMaps() {
-    return customVoteMaps;
+    return ImmutableSet.copyOf(customVoteMaps.keySet());
   }
 
   public boolean isAdded(MapInfo info) {
-    return customVoteMaps.stream().anyMatch(s -> s.getName().equalsIgnoreCase(info.getName()));
+    return getCustomVoteMaps().contains(info);
```
Pablete1234:
```suggestion
    return customVoteMaps.containsKey(info);
```

core/src/main/java/tc/oc/pgm/rotation/vote/VotePoolOptions.java
```diff
@@ -32,32 +35,42 @@ public boolean toggleMode() {
...
   public boolean removeMap(MapInfo map) {
-    return this.customVoteMaps.remove(map);
+    boolean present = this.customVoteMaps.containsKey(map);
+    this.customVoteMaps.remove(map);
+    return present;
   }
 
   public Set<MapInfo> getCustomVoteMaps() {
-    return customVoteMaps;
+    return ImmutableSet.copyOf(customVoteMaps.keySet());
```
Pablete1234: you can do with an unmodifiableSet that wraps the keys, no need to copy this i believe

core/src/main/java/tc/oc/pgm/rotation/vote/VotePoolOptions.java
```diff
@@ -32,32 +35,42 @@ public boolean toggleMode() {
     return replace;
   }
 
+  public boolean canAddVote() {
+    return customVoteMaps.size() < MapVotePicker.MAX_VOTE_OPTIONS;
+  }
+
   public boolean addVote(MapInfo map) {
```
Pablete1234: this is very confusing naming, you're not **adding a vote** you're **adding a map**

core/src/main/java/tc/oc/pgm/rotation/vote/VotePoolOptions.java
```diff
@@ -32,32 +35,42 @@ public boolean toggleMode() {
...
-    if (customVoteMaps.size() < MapVotePicker.MAX_VOTE_OPTIONS) {
-      this.customVoteMaps.add(map);
+    return addVote(map, null);
+  }
+
+  public boolean addVote(MapInfo map, @Nullable UUID playerId) {
```
Pablete1234: same here, should be addMap

core/src/main/java/tc/oc/pgm/rotation/vote/VotePoolOptions.java
```diff
@@ -32,32 +35,42 @@ public boolean toggleMode() {
     return replace;
   }
 
+  public boolean canAddVote() {
```
Pablete1234: canAddMap

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
```diff
@@ -31,31 +27,54 @@ public abstract class MapPool implements MapOrder, Comparable<MapPool> {
...
 
-  private MapInfo getMap(MapLibrary library, String mapName) {
-    @Nullable MapInfo map = library.getMap(mapName);
-    if (map != null) return map;
-    PGM.get()
-        .getLogger()
-        .warning("[MapPool] [" + name + "] " + mapName + " not found in map repo. Ignoring...");
-    return null;
+  private static List<MapInfo> buildMapList(List<String> mapNames, String poolName) {
+    List<MapInfo> mapList = Lists.newArrayList();
```
Pablete1234:
```suggestion
    List<MapInfo> mapList = new ArrayList(mapNames.size());
```
pre-allocate for mapNames' size, avoids re-sizes as elements are added in, and we know what the final size should be from the start

Pablete1234 (dismissed):
I left a final comment but otherwise LGTM

applenick:
@Electroid and @Pablete1234 thank you so much again for your feedback on this PR so far. Your insights and suggestions have been invaluable in improving this code, and I really appreciate the time and effort you've put into reviewing it.

If you have any further feedback or suggestions, please don't hesitate to let me know. I'm always looking for ways to improve and am open to your input.

I marked the PR as ready now and just rebased it on top of the latest upstream.

core/src/main/java/tc/oc/pgm/rotation/vote/book/VotingBookCreatorImpl.java
```diff
@@ -0,0 +1,38 @@
...
+            voted ? SYMBOL_VOTED : SYMBOL_IGNORE,
+            voted ? NamedTextColor.DARK_GREEN : NamedTextColor.DARK_RED));
+    text.append(text(" ").decoration(TextDecoration.BOLD, !voted));
+    text.append(text(map.getName(), NamedTextColor.GOLD, TextDecoration.BOLD));
+    text.hoverEvent(
+        showText(
+            text(
+                map.getTags().stream().map(MapTag::toString).collect(Collectors.joining(" ")),
+                NamedTextColor.YELLOW)));
+    text.clickEvent(runCommand("/votenext -o " + map.getName())); // Fix 1px symbol diff
```
Pablete1234: this comment belongs in line 28, the space is bolded-or-not-bolded to fix the 1px diff between the 2 symbols

util/src/main/i18n/templates/moderation.properties
```diff
@@ -183,6 +195,8 @@ vanish.activate = You have vanished!
 
 vanish.deactivate = You are now visible!
 
+vanish.deny.nick = You are unable to vanish while using a nick
```
Pablete1234: I believe this is no longer used, as this restriction has been lifted

Pablete1234 (changes requested):
Looking great, just left a couple final touches! Nice to see this (finally) merging

Pablete1234 (approved):

Electroid (approved):

PR #1117, opened 2023-01-28
Cleanup cloud command graph & add player parser

Pablete1234:
Extracts pgm-specific logic from CommandGraph into PGMCommandGraph, so that the class can be re-used easier by other plugins, and also adds a Player parser for any command that is leftover using player and not match player


Electroid (approved):

Electroid:
Looks good to merge, but seems to have some build failures.

Pablete1234:
It's related to a cloud library change, i've updated to 1.8.1 to fix a bug, and that version is only existing on our repo ATM.

PR #1118, opened 2023-01-30
Implement scoreboard filters

Pablete1234:
WIP, has not been tested.

Should allow any objective to simply add a `scoreboard-filter="x"` property, with a dynamic filter that says if it should show or not in the scoreboard. Additionally blitz and score modules also support it, and their filters apply for each competitor, so some may show while others do not


Pablete1234:
This has now been tested and works as intended :tada:

Electroid (approved):

PR #1126, opened 2023-02-03
Fix vanish state and tablist

applenick:
Another fix! This PR fixes `/vanish` incorrectly setting the MatchPlayer vanish state after the `PlayerVanishEvent` is called. Also provides the proper fake skin to the `PlayerTabEntry` as the logic under the NMSHacks method does not take into account the skin magic SportPaper does. We'll need to further address that at a later date to better accommodate multi-version support.

Update: see comments below for new logic, primary change now is the removal of the `vanished` state from MatchPlayer.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MatchCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MatchCommand.java
@@ -167,11 +167,11 @@ public final class MatchCommand {
     }
   }
 
-  private int getNonVanishedCount(Collection<MatchPlayer> players) {
+  private long getNonVanishedCount(Collection<MatchPlayer> players) {
     return players.stream()
         .map(MatchPlayer::getBukkit)
-        .mapToInt(p -> Integration.isVanished(p) ? 0 : 1)
-        .sum();
+        .filter(p -> !Integration.isVanished(p))
+        .count();
   }
 
   private MatchPlayer getMatchPlayer(CommandSender sender, Match match) {
--- a/core/src/main/java/tc/oc/pgm/command/MatchCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MatchCommand.java
@@ -24,6 +24,7 @@ import net.kyori.adventure.text.format.TextColor;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.integration.Integration;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchPhase;
 import tc.oc.pgm.api.party.Competitor;
@@ -81,10 +82,7 @@ public final class MatchCommand {
         msg.append(text(teamName, TextFormatter.convert(team.getColor())))
             .append(
                 text(": ", NamedTextColor.GRAY)
-                    .append(
-                        text(
-                            team.getPlayers().stream().filter(mp -> !mp.isVanished()).count(),
-                            NamedTextColor.WHITE)));
+                    .append(text(getNonVanishedCount(team.getPlayers()), NamedTextColor.WHITE)));
 
         if (team.getMaxPlayers() != Integer.MAX_VALUE) {
           msg.append(text("/" + team.getMaxPlayers(), NamedTextColor.GRAY));
@@ -110,10 +108,7 @@ public final class MatchCommand {
                     TextTranslations.translate("match.info.observers", sender),
                     NamedTextColor.AQUA))
             .append(text(": ", NamedTextColor.GRAY))
-            .append(
-                text(
-                    match.getObservers().stream().filter(mp -> !mp.isVanished()).count(),
-                    NamedTextColor.WHITE))
+            .append(text(getNonVanishedCount(match.getObservers()), NamedTextColor.WHITE))
             .build());
 
     viewer.sendMessage(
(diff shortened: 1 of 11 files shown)


core/src/main/java/tc/oc/pgm/integrations/SimpleVanishIntegration.java
```diff
@@ -84,7 +84,10 @@ public List<MatchPlayer> getOnlineVanished() {
   public boolean setVanished(MatchPlayer player, boolean vanish, boolean quiet) {
     final Match match = player.getMatch();
 
-    // Call vanish event first so name renders existing state properly for leave msg
+    // Set vanish status in match player
+    player.setVanished(vanish);
```
Pablete1234: Won't this order of events mean vanishing will shows as name crossed in leave message?
I think you need to do the same logic as when calling the event, where it's set to true/false at astart/end lf method depending on if its enabling or disabling

applenick: @Pablete1234 Do you mean for non-staff broadcasts or staff-only? With the updated logic:

Here is what it looks like when the viewer is a non-staff player, and the staff member runs `/vanish` twice
Screenshot 2023-02-03 at 2 34 40 PM

Here is what it looks like when the viewer is a staff player, and the other staff member leaves and rejoins the server
Screenshot 2023-02-03 at 2 36 29 PM

So correct in that the name is crossed on the leave message for staff members, but since they're logging out vanished we keep the vanish state right? Personally I don't mind the cross-out on leave messages, helps to further distinguish the player as vanished, but if we did want to remove it we'd probably want to adjust something with the player component? Would that be on the right track or is it something I'm missing?

Pablete1234: I'm assuming what's happening here is the name renderer for non-staff is rendering the name as "online-and-in-obs" when it should probably be rendering it as offline given vanish status is enabled at the point it's rendering it, which sounds like a different bug.
i don't mind the admin-chat showing crossed names, but the thing is it technically shouldn't, because it should pretend you're leaving while still not vanished (and vanish right as you leave), and so, appear without crossing-out, and then un-vanish right as you join, making the message (again) show as if you were never vanished to the eyes of everyone else

applenick: > i don't mind the admin-chat showing crossed names, but the thing is it technically shouldn't, because it should pretend you're leaving while still not vanished (and vanish right as you leave), and so, appear without crossing-out, and then un-vanish right as you join

I'm a bit confused about what the solution for this would be 🤔

If I'm understanding it correctly: So upon quit we should always unvanish the player before it gets to the leave message so it appears normally, but still quietly? Though if they're unvanished wouldn't the normal leave broadcast be called? 

My understanding may be totally incorrect, so any guidance on how to make it work the way you described would be helpful. Thanks!

Pablete1234: > So upon quit we should always unvanish the player before it gets to the leave message so it appears normally, but still quietly?

Upon vanishing, first process the player vanish event and the message (so they "leave" unvanished), and then setVanished to true.

Upon un-vanishing, first setVanished to false, then send the event (so they "join" unvanished).

Optionally you may be able to just get rid of vanished state in match player all together, im not sure what the use of it is atm since the state of vanished or not should come from integration

applenick: I went ahead and removed the additional vanish state inside of MatchPlayer. It's really unnecessary now that we have the integration, great suggestion. Believe that should solve the issue, thanks again! If there's any other changes you'd like made let me know 👍

core/src/main/java/tc/oc/pgm/command/MatchCommand.java
```diff
@@ -172,6 +167,13 @@ public void match(Audience viewer, CommandSender sender, Match match) {
     }
   }
 
+  private int getNonVanishedCount(Collection<MatchPlayer> players) {
+    return players.stream()
+        .map(MatchPlayer::getBukkit)
+        .mapToInt(p -> Integration.isVanished(p) ? 0 : 1)
+        .sum();
```
Pablete1234:
```suggestion
        .filter(Integration::isVanished)
        .count();
```

Pablete1234 (dismissed):
Just a small minor change, otherwise looks good :+1:

core/src/main/java/tc/oc/pgm/command/MatchCommand.java
applenick: Wouldn't this achieve the opposite? Since filtering for just vanished would return the vanished count and not the non-vanished count. The purpose of the method is to calculate the number of players in a team who are _not_ vanished.

```java
  private int getNonVanishedCount(Collection<MatchPlayer> players) {
    return players.stream()
        .map(MatchPlayer::getBukkit)
        .mapToInt(p -> Integration.isVanished(p) ? 0 : 1)
        .sum();
  }
```

An alternative approach that aligns with your suggestion could be:
```java
private int getNonVanishedCount(Collection<MatchPlayer> players) {
      return players.size() - Math.toIntExact(players.stream()
                      .map(MatchPlayer::getBukkit)
                      .filter(Integration::isVanished)
                      .count());
  }
```
Though the `mapToInt` method was a little more clean. Let me know which version should be used and I'll make the change if needed. Thanks again for the feedback!

Pablete1234: I didn't realize it was backwards, then the proper solution would be:

```suggest
        .filter(p -> !Integration.isVanished(p))
        .count();
```

Pablete1234 (approved):

Electroid (approved):

PR #1128, opened 2023-02-04
Prevent /reply revealing vanished players

applenick:
Another fix! This PR prevents `/reply` revealing vanished players 👻 Was discovered due to the new ability to be vanished and nicked at the same time. 

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
@@ -448,10 +448,10 @@ public class ChatDispatcher implements Listener {
     String lastKnownName = targetIdent.getName();
     String currentName = Players.getVisibleName(sender, target.getBukkit());
 
-    boolean reveal = Players.shouldReveal(sender, target.getBukkit());
+    // Ensure the target is visible to the viewing sender
     boolean visible = Players.isVisible(sender, target.getBukkit());
 
-    if ((currentName.equalsIgnoreCase(lastKnownName) && visible) || reveal) {
+    if (currentName.equalsIgnoreCase(lastKnownName) && visible) {
       return target.getId();
     }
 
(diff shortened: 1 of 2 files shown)


Pablete1234:
This fix is not quite properly addressing the root of the issue.

Some people should be able to dm eachother if vanished (eg: staff) and some people shouldnt be able to dm eachother even if not vanished (eg: people who rejoined with a different nick).

Consider instead of just storing the UUID of who last messaged you, save an "identity" of who did (uuid, nick) and allow replies only if you should be able to interact with such identity (nick hasn't changed OR you can see thru the nick (staff or friend), or they're not vanished, OR you can see thru the vanish (staff)). Most of this logic exists in the `Players` util class

Electroid (dismissed):

applenick:
@Pablete1234 When you have the time let me know what you think of the most recent commit. The new logic should achieve what you recommended, however I did hold-off on any change related to allowing staff to message while vanished. Wanted to discuss that further, personally feel since staff have the admin-chat to communicate while vanished there's no need to have additional private conversations. However, if you'd like me to change that then let me know and I'll go ahead and adjust it.

core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java
```diff
@@ -429,4 +430,50 @@ private Component getChatFormat(@Nullable Component prefix, MatchPlayer player,
...
+    MatchPlayer target = manager.getPlayer(targetIdent.getPlayerId());
+
+    // Prevent replying to offline players
+    if (target == null) return null;
+
+    // Compare last known and current name
+    String lastKnownName = targetIdent.getName();
+    String currentName = Players.getVisibleName(sender, target.getBukkit());
+
+    boolean reveal = Players.shouldReveal(sender, target.getBukkit());
```
Pablete1234: i believe reveal here shouldn't be needed, because shouldReveal is a an and of isVisible (checking for vanish) + getVisibleName (checking for if can see the nick), which you're essentially reimplementing here yourself

Pablete1234:
I think the logic makes sense, i've left a small comment for a redundant check but otherwise looks good

applenick:
Excellent! Let me know if you'd like me to make any more changes to this.

Pablete1234 (approved):

Electroid (approved):

PR #1131, opened 2023-02-04
Add consumables

OhPointFive:
This allows kits to be applied when eating some food item. Currently it's fairly limited but should be easy to expand in the future, if we want. Some example XML:

```xml
<kits>
    <kit id="spawn">
        <item slot="1" amount="1" damage="1" consumable="fast-apple">golden apple</item>
    </kit>
    <kit id="fast-apple-kit">
        <potion duration="4" amplifier="10">speed</potion>
    </kit>
</kits>
<consumables>
    <consumable id="fast-apple" action="fast-apple-kit"/>
</consumables>
```
or with defaults explicitly specified:
```xml
<consumable id="fast-apple" action="fast-apple-kit" on="eat" override="true"/>
```

`action` specifies the player-scoped action to apply on consumption. You can also use `kit` instead, but the result is the exact same.

`on` specifies how you use the consumable. `eat` is currently the only option, but later we could add click-to-eat consumables.

`override` specifies if it should override vanilla behaviour. For example, if `override` is `false`, this will also give you the vanilla golden apple effects and hunger restoration, but if it is `true` it will override vanilla behavior meaning the only result will be what's specified in the action/kit.

Much of this code is based on the similar `projectiles` module.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/consumable/ConsumableDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/consumable/ConsumableDefinition.java
@@ -1,0 +1,37 @@
+package tc.oc.pgm.consumable;
+
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.action.Action;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
+
+public class ConsumableDefinition extends SelfIdentifyingFeatureDefinition {
+
+  private final Action<? super MatchPlayer> action;
+  private final ConsumeCause cause;
+  /** If true, replaces vanilla behaviour, otherwise keeps vanilla behaviour */
+  private final boolean preventDefault;
+
+  public ConsumableDefinition(
+      @Nullable String id,
+      Action<? super MatchPlayer> action,
+      ConsumeCause cause,
+      boolean preventDefault) {
+    super(id);
+    this.action = action;
+    this.cause = cause;
+    this.preventDefault = preventDefault;
+  }
+
+  public Action<? super MatchPlayer> getAction() {
+    return action;
+  }
+
+  public ConsumeCause getCause() {
+    return cause;
+  }
+
+  public boolean getPreventDefault() {
+    return preventDefault;
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/consumable/ConsumableDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/consumable/ConsumableDefinition.java
@@ -10,17 +10,17 @@ public class ConsumableDefinition extends SelfIdentifyingFeatureDefinition {
   private final Action<? super MatchPlayer> action;
   private final ConsumeCause cause;
   /** If true, replaces vanilla behaviour, otherwise keeps vanilla behaviour */
-  private final boolean preventDefault;
+  private final boolean override;
 
   public ConsumableDefinition(
       @Nullable String id,
       Action<? super MatchPlayer> action,
       ConsumeCause cause,
-      boolean preventDefault) {
+      boolean override) {
     super(id);
     this.action = action;
     this.cause = cause;
-    this.preventDefault = preventDefault;
+    this.override = override;
   }
 
   public Action<? super MatchPlayer> getAction() {
@@ -31,7 +31,7 @@ public class ConsumableDefinition extends SelfIdentifyingFeatureDefinition {
     return cause;
   }
 
-  public boolean getPreventDefault() {
-    return preventDefault;
+  public boolean getOverride() {
+    return override;
   }
 }
--- a/core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
@@ -10,8 +10,9 @@ import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.player.MatchPlayer;
-import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.kits.tag.ItemTags;
+import tc.oc.pgm.util.MatchPlayers;
+import tc.oc.pgm.util.inventory.InventoryUtils;
 
 public class ConsumableMatchModule implements MatchModule, Listener {
 
@@ -32,16 +33,10 @@ public class ConsumableMatchModule implements MatchModule, Listener {
 
     Player player = event.getPlayer();
     MatchPlayer matchPlayer = match.getPlayer(player);
-    ParticipantState playerState = match.getParticipantState(player);
-    if (matchPlayer == null || playerState == null) return;
+    if (!MatchPlayers.canInteract(matchPlayer)) return;
 
-    if (consumableDefinition.getPreventDefault()) {
-      ItemStack itemInHand = player.getItemInHand();
-      if (itemInHand.getAmount() > 1) {
-        itemInHand.setAmount(itemInHand.getAmount() - 1);
-      } else {
-        player.setItemInHand(null);
-      }
+    if (consumableDefinition.getOverride()) {
+      InventoryUtils.consumeItem(player);
       event.setCancelled(true);
     }
 
--- a/core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
@@ -1,0 +1,63 @@
+package tc.oc.pgm.consumable;
+
+import com.google.common.collect.ImmutableSet;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.player.PlayerItemConsumeEvent;
+import org.bukkit.inventory.ItemStack;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchModule;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.player.ParticipantState;
+import tc.oc.pgm.kits.tag.ItemTags;
+
+public class ConsumableMatchModule implements MatchModule, Listener {
+
+  private final Match match;
+  private final ImmutableSet<ConsumableDefinition> consumableDefinitions;
+
+  public ConsumableMatchModule(
+      Match match, ImmutableSet<ConsumableDefinition> consumableDefinitions) {
+    this.match = match;
+    this.consumableDefinitions = consumableDefinitions;
+  }
+
+  @EventHandler
+  private void onItemConsume(PlayerItemConsumeEvent event) {
+    ConsumableDefinition consumableDefinition = getConsumableDefinition(event.getItem());
+    if (consumableDefinition == null) return;
+    if (consumableDefinition.getCause() != ConsumeCause.EAT) return;
+
+    Player player = event.getPlayer();
+    MatchPlayer matchPlayer = match.getPlayer(player);
+    ParticipantState playerState = match.getParticipantState(player);
+    if (playerState == null) return;
+
+    if (consumableDefinition.getPreventDefault()) {
+      ItemStack itemInHand = player.getItemInHand();
+      if (itemInHand.getAmount() > 1) {
+        itemInHand.setAmount(itemInHand.getAmount() - 1);
+      } else {
+        player.setItemInHand(null);
+      }
+      event.setCancelled(true);
+    }
+
+    consumableDefinition.getAction().trigger(matchPlayer);
+  }
+
+  private @Nullable ConsumableDefinition getConsumableDefinition(ItemStack item) {
+    String consumableId = ItemTags.CONSUMABLE.get(item);
+    if (consumableId != null) {
+      for (ConsumableDefinition consumableDefinition : consumableDefinitions) {
+        if (consumableDefinition.getId().equals(consumableId)) {
+          return consumableDefinition;
+        }
+      }
+    }
+
+    return null;
+  }
+}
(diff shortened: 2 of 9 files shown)


Pablete1234:
`action` in pgm has its own meaning at this time, in fact i'd even suggest instead of using `kit` as an attribute, using `action` for that very purpose, since a kit is a type of player-scoped action.

Maybe could be `apply-on="eat"` or simply `on="eat"` ? Feel free to propose alternative names for that attribute

as for `override` maybe we can take inspiration from html5/js, for events they have a `event.preventDefault` method which prevents a click from continuing to do what it would do by default, we could maybe do `prevent-default="true"` for "overriding" what would otherwise be default behavior

core/src/main/java/tc/oc/pgm/consumable/ConsumableDefinition.java
```diff
@@ -0,0 +1,21 @@
+package tc.oc.pgm.consumable;
+
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
+import tc.oc.pgm.kits.Kit;
+
+public class ConsumableDefinition extends SelfIdentifyingFeatureDefinition {
+
+  protected final Kit kit;
```
Pablete1234: Use an Action<MatchPlayer> instead of a kit, generally speaking we're going to "deprecate" kit within pgm features, in favor of player-scoped actions. All kits are actions so you'll always be able to use a kit anywhere that expects a player-scoped action, but not all actions are kits.

core/src/main/java/tc/oc/pgm/consumable/ConsumableDefinition.java
```diff
@@ -0,0 +1,21 @@
+package tc.oc.pgm.consumable;
+
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.features.SelfIdentifyingFeatureDefinition;
+import tc.oc.pgm.kits.Kit;
+
+public class ConsumableDefinition extends SelfIdentifyingFeatureDefinition {
```
Pablete1234, line 8: generally for definitions prefer private fields with getters over simply protected fields

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public @Nullable ConsumableModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      ConsumableModule consumableModule = new ConsumableModule();
+
+      for (Element consumableElement :
+          XMLUtils.flattenElements(doc.getRootElement(), "consumables", "consumable")) {
+        String id = XMLUtils.getRequiredAttribute(consumableElement, "id").getValue();
+        boolean override = XMLUtils.parseBoolean(consumableElement.getAttribute("override"), true);
+        Kit kit = factory.getKits().parseKitProperty(consumableElement, "kit", KitNode.EMPTY);
```
Pablete1234: You can look at shops module for an up-to-date example on how to parse a player action instead of a kit, generally speaking it will change very little, however make sure that you define the dependency between modules

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
```diff
@@ -0,0 +1,58 @@
...
+                Node.fromAttr(consumableElement, "action"),
+                ConsumeAction.class,
+                "consume action",
+                ConsumeAction.EAT);
+
+        ConsumableDefinition consumableDefinition =
+            new ConsumableDefinition(id, kit, consumeAction, override);
+
+        factory.getFeatures().addFeature(consumableElement, consumableDefinition);
+        consumableModule.consumableDefinitions.add(consumableDefinition);
```
Pablete1234: This is terrible practice, do not mutate the module's state.

Create a `Set<ConsumableDefinition>` for parsing, parse all the definitions and add them to the set, and only at the very end, create the `new ConsumableModule(definitions)`, and if possible (ie: no consumables are defined) skip creating it at all, and just return a null.
Also consider making the consumable module constructor create an immutable copy of the set to avoid anything mutating the state, as that would lead to incorrect behavior (ie: re-cycling to the same maps multiple times ending in differing behavior)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
OhPointFive: This, and some of the other code convention issues, were copied from the projectile implementation, so that should probably be updated at some point too.

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
OhPointFive: Not 100% sure I got the dependencies right but I think I did.

Pablete1234 (dismissed):

Pablete1234:
Looks fine to me, but i'd like some map-dev to give some more feedback regarding XML naming, particularly the `on="eat"` and `prevent-default="true"` attributes. Are they good or does anyone else have better names?
cc @Pugzy @calcastor @Cazadorian

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
```diff
@@ -0,0 +1,63 @@
...
+  @EventHandler
+  private void onItemConsume(PlayerItemConsumeEvent event) {
+    ConsumableDefinition consumableDefinition = getConsumableDefinition(event.getItem());
+    if (consumableDefinition == null) return;
+    if (consumableDefinition.getCause() != ConsumeCause.EAT) return;
+
+    Player player = event.getPlayer();
+    MatchPlayer matchPlayer = match.getPlayer(player);
+    ParticipantState playerState = match.getParticipantState(player);
+    if (playerState == null) return;
```
Pablete1234: why do you want their participant state?

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @EventHandler
+  private void onItemConsume(PlayerItemConsumeEvent event) {
+    ConsumableDefinition consumableDefinition = getConsumableDefinition(event.getItem());
+    if (consumableDefinition == null) return;
+    if (consumableDefinition.getCause() != ConsumeCause.EAT) return;
+
+    Player player = event.getPlayer();
+    MatchPlayer matchPlayer = match.getPlayer(player);
```
Pablete1234, line 35: this can be null, the player may actually not be in the match (yes, i know it sounds absurd, but that *can* technically happen)

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
```diff
@@ -0,0 +1,63 @@
...
+    ParticipantState playerState = match.getParticipantState(player);
+    if (playerState == null) return;
+
+    if (consumableDefinition.getPreventDefault()) {
+      ItemStack itemInHand = player.getItemInHand();
+      if (itemInHand.getAmount() > 1) {
+        itemInHand.setAmount(itemInHand.getAmount() - 1);
+      } else {
+        player.setItemInHand(null);
+      }
```
Pablete1234: I've seen this block of code in a few places, potentially look into moving to some util?

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
```diff
@@ -0,0 +1,88 @@
...
+        throws InvalidXMLException {
+      ActionParser actionParser = new ActionParser(factory);
+
+      Set<ConsumableDefinition> consumableDefinitions = new HashSet<>();
+
+      List<Element> consumableElements =
+          XMLUtils.flattenElements(doc.getRootElement(), "consumables", "consumable");
+      if (consumableElements.size() == 0) return null;
+
+      for (Element consumableElement : consumableElements) {
```
Pablete1234: You could simply iterate the elements as-is, and return null at the very end. It's not like iterating the empty list is going to make a difference, which will be simpler.
On top of that, in java prefer using `.isEmpty()` instead of `.size() == 0` for collections

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
```diff
@@ -0,0 +1,88 @@
...
+        String id = XMLUtils.getRequiredAttribute(consumableElement, "id").getValue();
+        boolean preventDefault =
+            XMLUtils.parseBoolean(consumableElement.getAttribute("prevent-default"), true);
+
+        Action<? super MatchPlayer> action;
+        Node actionNode = Node.fromAttr(consumableElement, "action", "kit");
+        if (actionNode != null) {
+          action = actionParser.parseReference(actionNode, null, MatchPlayer.class);
```
Pablete1234: pretty sure you can get rid of the `null` parameter here, there's an alternative method without it. The parameter-ed option is only a thing for when the node you pass in isn't the id (eg: `<action id="x"/>` may pass the whole action element as the node, and just the x as the id string), but it makes little sense here (or in the shop module that you got this from)

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
```diff
@@ -0,0 +1,88 @@
...
+      if (consumableElements.size() == 0) return null;
+
+      for (Element consumableElement : consumableElements) {
+        String id = XMLUtils.getRequiredAttribute(consumableElement, "id").getValue();
+        boolean preventDefault =
+            XMLUtils.parseBoolean(consumableElement.getAttribute("prevent-default"), true);
+
+        Action<? super MatchPlayer> action;
+        Node actionNode = Node.fromAttr(consumableElement, "action", "kit");
+        if (actionNode != null) {
```
Pablete1234: I wonder, does it even make sense to have a consumable with no action? with the functionality offered in this PR, i feel like this shouldn't be optional as simply preventing default and doing nothing seems kinda odd.
Maybe mapdevs can give more input on this?

If it was to be removed, you can simply change to `Node.fromRequiredAttr` and remove the if

Pablete1234 (changes requested):
Generally looking fine, just some small touch-ups

calcastor:
> Looks fine to me, but i'd like some map-dev to give some more feedback regarding XML naming, particularly the `on="eat"` and `prevent-default="true"` attributes. Are they good or does anyone else have better names? cc @Pugzy @calcastor @Cazadorian

It might make sense to use `override` in place of `prevent-default`; there is precedent for this in crafting module. I'd understand if this could seem unclear though.

As for `on`, that doesn't seem like the ideal name for this attribute but at the same time I can't think of much else that could fit better.

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
OhPointFive: I assumed this was so observers can't interact with consumables, which might be problematic for some actions.

Pablete1234: if you don't want observers to interact (and probably don't want dead people either), check if the player can interact with the match:
`ìf (!MatchPlayers.canInteract(matchPlayer)) return;`

participant state is not what you should use for that

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
OhPointFive: Not sure how utils is structured — should I add this to `util/Players.java`, make a new util, or something else?

Pablete1234: I'd probably stick it in some inventory utils or similar, like `InventoryUtils.consumeItem(player)`

Pablete1234:
i agree that given there's precedent for `override` withink pgm we should stick to that

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
```diff
@@ -0,0 +1,63 @@
...
+  @EventHandler
+  private void onItemConsume(PlayerItemConsumeEvent event) {
+    ConsumableDefinition consumableDefinition = getConsumableDefinition(event.getItem());
+    if (consumableDefinition == null) return;
+    if (consumableDefinition.getCause() != ConsumeCause.EAT) return;
+
+    Player player = event.getPlayer();
+    MatchPlayer matchPlayer = match.getPlayer(player);
+    ParticipantState playerState = match.getParticipantState(player);
+    if (matchPlayer == null || playerState == null) return;
```
Pablete1234:
```suggestion
    MatchPlayer matchPlayer = match.getPlayer(player);
    if (!MatchPlayers.canInteract(matchPlayer)) return;
```
there's no need for a participant state, and you can just check for interaction (the util in MatchPlayers handles the null-check for you)

core/src/main/java/tc/oc/pgm/consumable/ConsumableMatchModule.java
Pablete1234: i double-checked and it seems like `InventoryUtils` already exists, and the items are very related (ie: place stack, or similar), so it'd fit right in as `InventoryUtils.consumeItem(player)`

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/consumable/ConsumableModule.java
```diff
@@ -0,0 +1,79 @@
...
+        Node actionNode = Node.fromRequiredAttr(consumableElement, "action", "kit");
+        Action<? super MatchPlayer> action =
+            actionParser.parseReference(actionNode, MatchPlayer.class);
+
+        ConsumeCause cause =
+            XMLUtils.parseEnum(
+                Node.fromAttr(consumableElement, "on"),
+                ConsumeCause.class,
+                "consume cause",
+                ConsumeCause.EAT);
```
Pablete1234: if this is the default, i'd remove it (simply make the above Node.fromRequireAttr, and this as null or omit the param?)

while it may make sense that *eat* is the default now (since it's the only allowed value), it may become more cannon for the default to be right-click in the future. If we don't set a default now, all maps will be forced to explicitly set it, meaning if in the future (when this feature is more complete) we can make a more informed decision as to what the default should be (if any), we'll be able to set it. If we set it now however, it'll be here forever as changes would make it a breaking change.

Pablete1234 (commented):
Just one final touch-up, otherwise lgtm

Pablete1234:
I've also updated the original comment to update override vs prevent-default and clarify kit/action

Pablete1234 (approved):

Electroid (approved):

PR #1136, opened 2023-02-19
Refactor join logic & add support for squads

Pablete1234:
This Implements a relatively big refactor to the join logic in PGM, which makes it more flexible and easier to maintain. We no longer have different codepaths for join, force-join, team switch kit (which adds a title when joining) etc. They now instead all route thru queryJoin & join, but with a JoinRequest, which decides how this player is allowed to join. 

You may now join a player to join with FORCE perms (even if they don't have those themselves) and the whole join process will respect that, EVEN after the join occurs (ie: if a re-balance occurs, or a priority-kick occurs, a forced player won't be kicked out). Before this behavior would be dependant on the users' permissions instead of on what "flags" it used to join.

Ontop of that, it adds support for Squads. These are a new pgm integration to be handled by 3rd party plugins, where PGM has very little logic to support them (expect Community to have support for them very soon). Players in the same squad will see eachother listed higher in the tablist, and underlined.


core/src/main/java/tc/oc/pgm/doublejump/DoubleJumpMatchModule.java
```diff
@@ -65,6 +65,7 @@ public void disable() {
       iterator.remove();
       this.refreshJump(player);
     }
+    this.jumpers.disable();
```
Pugzy, line 68: Unrelated bug fix?

Pablete1234: Yes, i was looking into using online map adapters, and saw this as an usage which was never disabled (so it's a memory leak)

core/src/main/java/tc/oc/pgm/api/match/event/MatchLoadEndEvent.java
```diff
@@ -0,0 +1,28 @@
...
+import org.bukkit.event.HandlerList;
+import tc.oc.pgm.api.match.Match;
+
+/**
+ * Called after a {@link Match} has completely loaded, and all players have been moved to the new
+ * match.
+ *
+ * @see Match#load()
+ */
+public class MatchLoadEndEvent extends MatchEvent {
```
Electroid: `MatchAfterLoadEvent`?

core/src/main/java/tc/oc/pgm/api/match/Match.java
```diff
@@ -357,6 +358,15 @@ default boolean finish() {
...
   boolean setParty(MatchPlayer player, Party party);
 
+  /**
+   * Set or change the {@link Party} of a {@link MatchPlayer}.
+   *
+   * @param player The {@link MatchPlayer}.
+   * @param party The new {@link Party}.
+   * @return Whether the operation was a success.
+   */
+  boolean setParty(MatchPlayer player, Party party, JoinRequest request);
```
Electroid: Why do we need the `Party` argument, when it should be defined from the `JoinRequest`? Also, comment needs update.

Electroid (dismissed):

core/src/main/java/tc/oc/pgm/api/match/Match.java
Pablete1234: The request will have a null party if you picked to join a random tea, but the Party will not be.

ie: you type `/join`, PGM goes thru the join logic for a null team (random), then ends up deciding you go in red team, calls `setParty(player, redTeam, request(null))`

core/src/main/java/tc/oc/pgm/api/match/event/MatchLoadEndEvent.java
Pablete1234: sure, i'll make the change

Electroid (approved):

Electroid (approved):

PR #1138, opened 2023-02-20
Add arrow pickup prevention

OhPointFive:
This adds a way to filter which players can pick up arrows:
```xml
<modifybowprojectile pickup-filter="some-filter"/>
```

The primary use case that I'm aware of is when `pickup-filter="never"`. 
- We're implementing BuildUHC in PGM, and BuildUHC traditionally doesn't allow arrows to be picked up.
- Many players complain about "too many arrows" — preventing arrow pickups provides another way to control how many arrows are available
- Tea's working on controlling when players have arrows via actions & triggers that give & take arrows. These only work by limiting the player to 1 arrow, which is simpler if they can't pick up arrows.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileModule.java
@@ -13,6 +13,7 @@ import tc.oc.pgm.api.map.MapModule;
 import tc.oc.pgm.api.map.factory.MapFactory;
 import tc.oc.pgm.api.map.factory.MapModuleFactory;
 import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.filters.matcher.StaticFilter;
 import tc.oc.pgm.filters.parse.FilterParser;
 import tc.oc.pgm.util.xml.InvalidXMLException;
 import tc.oc.pgm.util.xml.XMLUtils;
@@ -50,7 +51,7 @@ public class ModifyBowProjectileModule implements MapModule<ModifyBowProjectileM
       Class<? extends Entity> projectile = Arrow.class;
       float velocityMod = 1;
       Set<PotionEffect> potionEffects = new HashSet<>();
-      Filter pickupFilter = null;
+      Filter pickupFilter = StaticFilter.ALLOW;
 
       for (Element parent : doc.getRootElement().getChildren("modifybowprojectile")) {
         Element projectileElement = parent.getChild("projectile");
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileModule.java
```diff
@@ -59,11 +69,16 @@ public ModifyBowProjectileModule parse(MapFactory factory, Logger logger, Docume
           potionEffects.add(XMLUtils.parsePotionEffect(potionElement));
           changed = true;
         }
+
+        for (Element noPickupElement : parent.getChildren("pickup")) {
+          pickupFilter = filters.parseFilterProperty(noPickupElement, "filter");
+          changed = true;
+        }
```
Pablete1234: This parses multiple children, taking just the last one. This is incorrect behavior, either ensure there's a single child or combine the filters (i'd just do the former, it's simpler)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileModule.java
```diff
@@ -59,11 +69,17 @@ public ModifyBowProjectileModule parse(MapFactory factory, Logger logger, Docume
           potionEffects.add(XMLUtils.parsePotionEffect(potionElement));
           changed = true;
         }
+
+        Element noPickupElement = parent.getChild("pickup");
+        if (noPickupElement != null) {
+          pickupFilter = filters.parseFilterProperty(noPickupElement, "filter");
+          changed = true;
+        }
```
Pablete1234: This changes doesn't actually enforce there being a unique child, and it feels like the easier way to handle this would be by making the filter a property of modifybowprojectile, it would make the syntax be:
```xml
<modifybowprojectile pickup-filter="x"/>
```
or
```xml
<modifybowprojectile>
  <pickup-filter>...</pickup-filter>
</modifybowprojectile>
```
which goes in-line with how it's generally done in pgm, since you just want one child why not inline it in the parent.

```suggestion
        Filter pickupFilter = filters.parseFilterProperty(parent, "pickup-filter", StaticFilter.ALLOW);
        changed |= pickupFilter != StaticFilter.ALLOW;
```

This would enforce single-child properly and simplify the syntax

core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileModule.java
```diff
@@ -8,39 +8,49 @@
...
     @Override
     public ModifyBowProjectileModule parse(MapFactory factory, Logger logger, Document doc)
         throws InvalidXMLException {
+      FilterParser filters = factory.getFilters();
+
       boolean changed = false;
       Class<? extends Entity> projectile = Arrow.class;
       float velocityMod = 1;
       Set<PotionEffect> potionEffects = new HashSet<>();
+      Filter pickupFilter = null;
```
Pablete1234:
```suggestion
      Filter pickupFilter = StaticFilter.ALLOW;
```
Generally filters can just default to allow/deny instead of null which can make things easier as you avoid null-checks elsewhere

Pablete1234 (changes requested):

Pablete1234 (approved):

Electroid (approved):

PR #1141, opened 2023-02-22
Re-add methods to avoid 3rd party plugin breakage

Pablete1234:
Re-adds methods that were removed previously due to PGM no longer using them, but that other 3rd party plugins (mainly, Events) may be using. They're re-added as deprecated and should be removed in the future, being kept just as a transition period until next release.


Electroid (approved):

Electroid:
Thanks for fixing this.

PR #1142, opened 2023-02-22
Fix book creator never being overriden

Pablete1234:


core/src/main/java/tc/oc/pgm/rotation/vote/MapPoll.java
```diff
@@ -79,7 +79,7 @@ public MapPoll(Match match, List<MapInfo> maps) {
   }
 
   public static void setVotingBookCreator(VotingBookCreator bookCreator) {
-    bookCreator = assertNotNull(bookCreator);
+    MapPoll.bookCreator = assertNotNull(bookCreator);
```
applenick, line 82: Oh, whoops 😅

Electroid (approved):

PR #1145, opened 2023-02-26
Tablist rendering performance & responsiveness improvements

Pablete1234:
Brings many fixes & improvements around tablist, and cycling performance:
  - Fixed vanishing not updating the tablist layout to hide the player
  - Fixed match teleports per second not being respected after cycle gets to 0 (timed out factory)
  - Changed teleports per second to batch N players per second, instead of one every 1000/N millis
  - Fixed default world border (set to 10.000 blocks) not being enforced, causing performance issues if bypassed.
  - Fixed mapmaker flairs being taken away after cycling to a map of the same author
  - Fixed player rendering caching name flairs even after a decoration change event, fixes mapmaker flairs updating on tablist
 - Tablist will now fully pause updating during cycle (up to 30s), and will issue one full re-render after everyone is teleported, then resume normal operation
 - Tablist now has a priority re-render system, if you switch team your own tablist will be prioritized, others can update later. This increases the perceived responsiveness of tablist.
 - Tablist can now be configured to render in batches of N players, so first priority views are updated, then views are updated picking N from the dirty views and re-scheduling the other ones until eventually no views are left to update.
 - Tablist header & footer can now be rendered without doing layout, meaning the scheduled 1s timer (for current match time) will update in a timely manner regardless of how delayed everything else is, and won't force everything else to be triggered on a 1s cooldown.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/integrations/SimpleVanishIntegration.java
+++ b/core/src/main/java/tc/oc/pgm/integrations/SimpleVanishIntegration.java
@@ -84,27 +84,11 @@ public class SimpleVanishIntegration implements VanishIntegration, Listener {
   public boolean setVanished(MatchPlayer player, boolean vanish, boolean quiet) {
     final Match match = player.getMatch();
 
-    // Call vanish event next so name renders existing state properly for leave msg
-    if (vanish) match.callEvent(new PlayerVanishEvent(player, vanish, quiet));
-
-    // Keep track of the UUID and apply/remove META data, so we can detect vanish status from other
-    // projects (i.e utils)
-    if (vanish) {
-      addVanished(player);
-    } else {
-      removeVanished(player);
-    }
-
-    // Ensure player is an observer
-    if (vanish && player.getParty() instanceof Competitor) {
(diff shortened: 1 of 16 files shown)


core/src/main/java/tc/oc/pgm/integrations/SimpleVanishIntegration.java
```diff
@@ -84,28 +84,12 @@ public List<MatchPlayer> getOnlineVanished() {
...
-    } else {
-      removeVanished(player);
-    }
-
-    // Ensure player is an observer
-    if (vanish && player.getParty() instanceof Competitor) {
-      match.setParty(player, match.getDefaultParty());
-    }
+    // Call vanish event which will vanish/unvanish at appropriate levels
+    match.callEvent(new PlayerVanishEvent(player, vanish, quiet));
```
applenick, line 88: Thanks for taking care of the vanish fix! I was investigating it earlier in the day, excellent solution having the vanish logic extracted out of this method 👍

Electroid (approved):
Awesome 👍

PR #1146, opened 2023-02-26
Switch locked items to use item tags

Pugzy:
Moves the locked item logic to be ItemTags on the item rather than a map of locked slots. This allows you to lock anything and not just items (armor can be locked now too). This is how it was eventually done on OCN and the logic around this has been migrated over.

Tested during the recent Bolt draft event and all seemed well.


Pablete1234 (approved):

Electroid (approved):
Thanks for refactoring, looks great!

PR #1149, opened 2023-03-05
Implement pulse filter

Pablete1234:
Implements a pulse filter, somewhat similar to countdown or after:

```xml
<filters>
  <pulse id="some-pulse" period="1s" duration="0.5s" filter="some-filter"/>

  <pulse id="other-pulse" period="60s" duration="15s">
    <match-running/>
  </pulse>
</filters>
```

This filter is dynamic, useful mainly for triggers and actions, where you can make it so once a filter is met (eg: a player entering a region) instead of getting one activation, you can get repeated activations for specifically certain amounts of time (eg: 15s on, 45s off, total period is 60s).


TheRealPear:
liked & subscribed

Electroid (approved):
Neat. Clicked the bell on the subscribe button.

PR #1150, opened 2023-03-06
Add custom firework designs 🎆 

applenick:
# Custom Firework Designs 🎆 
This PR adds the ability to define custom firework effects when giving a firework item. Alongside the existing custom parsers for books and skulls, `<firework>` will allow for custom designs to be specified in an easy to understand format. 

## XML Definition
Element | Attribute | Description
--- | --- | ---
`<firework>` | `power` | Defines the power of the firework (optional). If not specified, the power defaults to 1.
`<explosion>` | `type` | Defines the type of the effect. Valid values: `ball`, `ball large`, `burst`, `creeper`, `star`.  If not specified, type defaults to `burst`. 
  | `flicker` | Defines whether the effect should flicker (optional). Valid values are `true` and `false`. If not specified, flicker defaults to false.
  | `trail` | Defines whether the effect should have a trail (optional). Valid values are `true` and `false`. If not specified, trail defaults to false.
`<color>` |   | Defines a color of the firework. Colors are specified as a six-digit hex value. At least one color must be defined for each explosion.
`<fade>` |   | Defines a fade color for the firework (optional). Colors are specified as a six-digit hex value. Multiple fade colors can be defined for each explosion. If no fade colors are defined, the explosion will not fade.

See [here](https://pgm.dev/docs/reference/misc/colors) for more info on hex colors

## XML Example
```xml
<firework power="1">
  <!-- Defines an explosion -->
  <explosion type="creeper" flicker="true" trail="true">
    <!-- One or more colors can be defined as hex values -->
    <color>193bcf</color>
    <color>edea18</color>
    
    <!-- Defines a fade color as a hex value (optional) -->
    <fade>e2a31a</fade>
  </explosion>
  
  <!-- Another explosion with different attributes -->
  <explosion type="burst" trail="true">
    <color>b319af</color>

    <!-- Multiple fade colors can be defined -->
    <fade>e2a31a</fade>
    <fade>e2a31a</fade>
  </explosion>
</firework>
```

The resulting firework item:
Firework example
firework-final

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/KitParser.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitParser.java
@@ -21,6 +21,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.regex.Pattern;
 import org.bukkit.Color;
+import org.bukkit.FireworkEffect;
+import org.bukkit.FireworkEffect.Type;
 import org.bukkit.GameMode;
 import org.bukkit.Material;
 import org.bukkit.enchantments.Enchantment;
@@ -28,6 +30,7 @@ import org.bukkit.inventory.ItemFlag;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.meta.BookMeta;
 import org.bukkit.inventory.meta.EnchantmentStorageMeta;
+import org.bukkit.inventory.meta.FireworkMeta;
 import org.bukkit.inventory.meta.ItemMeta;
 import org.bukkit.inventory.meta.LeatherArmorMeta;
 import org.bukkit.inventory.meta.PotionMeta;
@@ -260,6 +263,10 @@ public abstract class KitParser {
         case "head":
           item = parseHead(itemEl);
           break;
+
+        case "firework":
+          item = parseFirework(itemEl);
+          break;
       }
 
       if (item != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -398,6 +405,53 @@ public ItemStack parseHead(Element el) throws InvalidXMLException {
...
+    itemStack.setItemMeta(meta);
+    return itemStack;
+  }
+
+  private List<Color> parseColors(List<Node> nodes) throws InvalidXMLException {
+    List<Color> colors = Lists.newArrayList();
+    for (Node node : nodes) {
+      String rawColor = node.getValue();
+      if (rawColor != null) {
+        if (!rawColor.matches("[a-fA-F0-9]{6}")) {
```
Pablete1234: Do we not have some util to parse these? Leather armor had to parse colors before this

applenick: @Pablete1234 Not that I'm aware of. I looked into how leather armor has its color applied:

https://github.com/PGMDev/PGM/blob/59c1dc9f2719fd85165451bf502f921756b3a8e4/core/src/main/java/tc/oc/pgm/kits/KitParser.java#L514-L524

Would it be beneficial to move the color parsing logic to XMLUtils?  Perhaps something like
```java
public static Color parseHexColor(Node node, Color def) throws InvalidXMLException {
      String rawColor = node.getValue();
      if (rawColor != null) {
        if (!rawColor.matches("[a-fA-F0-9]{6}")) {
          throw new InvalidXMLException("Invalid color format", rawColor);
        }
        return Color.fromRGB(Integer.parseInt(rawColor, 16));
      }
      return def;
  }
  ```

Comment (Pablete1234) [reply core/src/main/java/tc/oc/pgm/kits/KitParser.java]: yeah, given at least 2 places are parsing these now, they should be moved over to an util for sure

File: util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java [hunk 2]
```diff
@@ -1130,4 +1130,15 @@ public static Title.Times parseTitleTimes(Element el, Title.Times def)
 
     return Title.Times.times(fadeIn, stay, fadeOut);
   }
+
+  public static Color parseHexColor(Node node, Color def) throws InvalidXMLException {
```
Pablete1234: I'd follow the pattern other methods around here do, one parser parses only a non-null node (required), then there's an overload for a potentially-nullable with a fallback default.
In this case it seems like the default is never properly used (always white), i'd simply go for parseHexColor and throw an invalid xml exception if either the node or the node's value are null, and avoid having a default fallback

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -398,6 +405,47 @@ public ItemStack parseHead(Element el) throws InvalidXMLException {
...
+              .trail(trail)
+              .build());
+    }
+
+    itemStack.setItemMeta(meta);
+    return itemStack;
+  }
+
+  private List<Color> parseColors(List<Node> nodes) throws InvalidXMLException {
+    List<Color> colors = Lists.newArrayList();
```
```suggestion
    List<Color> colors = new ArrayList(nodes.size());
```
pre-allocating the size makes alot of sense in this scenario

Pablete1234 (changes requested):

Pablete1234 (approved):

Pablete1234 (approved):

Electroid (approved):
🎉

PR #1151, opened 2023-03-06
Fix NPE with vanished player join messages

Pugzy:
This fixes two issues which occur when a player joins the match directly on to a team (Events plugin).

When a player joins directly to a team this code make sure they're not vanished. The vanished being non-silent means a join message is shown. The current order of this event means that the players name component (showing in the fake/unvanish join) message is rendered using a team color which is null so throws an NPE.

- Switches unvanishing to use event `PlayerJoinMatchEvent` (player already has team set at this point).
- Null check on match player component to check team is set.
- Don't try unvanish someone who isn't vanished in the first place.
- Not broadcasting unvanish as the new event location means the join listener can manage the join and message.


Pablete1234 (approved):

Electroid (approved):
Looks good, merging because CI is broken rn.

PR #1157, opened 2023-03-13
Implement filter command

Pablete1234:
Implements a `/filter <filter> [target]` command, allowing map developers inspecting the filters while they're working on them.

image
image


Electroid (approved):
Looks good

Issue #971, opened 2022-03-13
Maps only update xml/region file changes if they are set with /recycle


RuedigerLP:
So if u change something in the xml and dont change the version number it will recognize the change if /recycle is used, but cycling to another map and then cycling to the first will not update changes (unless version numbers are changed).

Electroid:
Fixed in #1161

PR #1161, opened 2023-03-18
Default to new map when same version number is used

Pablete1234:
Updated so that the default when the map version is the same is to keep the new map, instead of the old one.

Previous logic read as `if old.version < new.version -> keep new, otherwise keep old`
The new one reads as `if new.version < old.version -> keep old, otherwise keep new`

The subtle change fixes the wrongly behavior explained in #971

Note: always keeping the new map is not a solution, as a map may be loaded from 2 different sources (multiple repositories defined on pgm with overlapping maps) in which case you want always the higher-versioned map to prevail when pgm loads them all on startup, regardless of map loading order.


Electroid (approved):
Yea I remember this code, thanks for fixing it up.

PR #1169, opened 2023-04-14
Clear crafting grid on clear kit

OhPointFive:
Currently, you can smuggle items past a `<clear>` in a kit by putting them in your 2x2 crafting table with your inventory open. This PR fixes that.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/ClearItemsKit.java
+++ b/core/src/main/java/tc/oc/pgm/kits/ClearItemsKit.java
@@ -1,6 +1,9 @@
 package tc.oc.pgm.kits;
 
 import java.util.List;
+import org.bukkit.event.inventory.InventoryType;
+import org.bukkit.inventory.Inventory;
+import org.bukkit.inventory.InventoryView;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.potion.PotionEffect;
 import tc.oc.pgm.api.player.MatchPlayer;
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/kits/ClearItemsKit.java
```diff
@@ -45,6 +47,13 @@ public void applyPostEvent(MatchPlayer player, boolean force, List<ItemStack> di
     if (this.items) {
       player.getBukkit().getInventory().clear();
       player.getBukkit().getOpenInventory().setCursor(null);
+      InventoryView openInventory = player.getBukkit().getOpenInventory();
+      if (openInventory != null) {
+        Inventory topInventory = openInventory.getTopInventory();
```
Pablete1234, line 55: Is this going to clear a whole chest if you have one open and a clear is applied?

OhPointFive: Sorry for the slow update.

I've modified this so it will only clear crafting grids. I've included full 3x3 crafting benches, since you can't store items in them normally so it would be weird if you could sneak items through one. It should no longer clear chests, though.

I don't have a good map to test this on, but it does successfully clear 2x2 crafting grids. If we want to test this more thoroughly let me know.

Pablete1234 (approved):

Electroid (approved):

Issue #1166, opened 2023-04-07
Support for instant respawning


calcastor:
Gamemodes like Bridges are implementable in PGM (as @zzuf-fer has demonstrated very well on OCC) however has some minor discrepancies compared to the gamemode as it functions on servers such as Hypixel. Perhaps the most prominent of these is the lack of ability to instantly respawn following death. This impedes some strategies important to Bridges gameplay, particularly in 1v1 or 2v2 environments, where voiding and instantly respawning is a defence technique. 

PGM currently enforces a minimum respawn time of 1 seconds and some behaviour relating to respawning "expects" this to function correctly as I understand it, so some level of refactoring is required to support instant respawning.

calcastor:
https://github.com/PGMDev/PGM/pull/1171

PR #1171, opened 2023-04-14
Support fast respawns & fix blitz bugged dead state

Pablete1234:
Allows respawn delay to drop from the current minimum of 1.5s down to 0. This has some implications like rotting animation needs to be reset when leaving dead state. It should additionally fix the current blitz issue where observers will see your corpse forever rotting due to leaving the state before visibility is reset.

This has been tested and seems to have fixed the blitz issue completely, as well as allowing arbitrarily low respawn times. Additionally fixes velocity not being reset for players (which caused instant respawns to conserve velocity after death)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
@@ -25,7 +25,6 @@ import org.bukkit.event.inventory.InventoryClickEvent;
 import org.jetbrains.annotations.Nullable;
 import org.spigotmc.event.player.PlayerSpawnLocationEvent;
 import tc.oc.pgm.api.PGM;
-import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.query.Query;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
@@ -77,8 +76,7 @@ public class SpawnMatchModule implements MatchModule, Listener, Tickable {
 
   public RespawnOptions getRespawnOptions(Query query) {
     return module.respawnOptions.stream()
-        .filter(
-            respawnOption -> respawnOption.filter.query(query).equals(Filter.QueryResponse.ALLOW))
+        .filter(respawn -> !respawn.filter.query(query).isDenied())
         .findFirst()
         .orElseThrow(() -> new IllegalStateException("No respawn option could be used"));
   }
--- a/core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
@@ -25,6 +25,7 @@ import org.bukkit.event.inventory.InventoryClickEvent;
 import org.jetbrains.annotations.Nullable;
 import org.spigotmc.event.player.PlayerSpawnLocationEvent;
 import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.query.Query;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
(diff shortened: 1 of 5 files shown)


KingOfSquares:
Closes #1166

core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
```diff
@@ -28,7 +28,7 @@
 public class SpawnModule implements MapModule<SpawnMatchModule> {
 
   public static final Duration DEFAULT_RESPAWN_DELAY = Duration.ofMillis(2000);
-  public static final Duration MINIMUM_RESPAWN_DELAY = Duration.ofMillis(1500);
+  public static final Duration MINIMUM_RESPAWN_DELAY = Duration.ofMillis(0);
```
KingOfSquares:
```suggestion
  public static final Duration MINIMUM_RESPAWN_DELAY = Duration.ZERO
```

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
```diff
@@ -77,8 +76,7 @@ public Match getMatch() {
 
   public RespawnOptions getRespawnOptions(Query query) {
     return module.respawnOptions.stream()
-        .filter(
-            respawnOption -> respawnOption.filter.query(query).equals(Filter.QueryResponse.ALLOW))
+        .filter(respawn -> !respawn.filter.query(query).isDenied())
```
KingOfSquares: This seems to now allow filters abstaining, I can't think of any reason right now why that would be bad for future xmls but it does potentially change behaviour on existing maps?

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
```diff
@@ -80,30 +76,36 @@ public void leaveState(List<Event> events) {
...
+
     super.leaveState(events);
   }
 
   @Override
   public void tick() {
     long age = age();
 
-    if (!kitted && ticksUntilRespawn() <= 0) {
+    if (!kitted && ticksUntilRespawn() <= 0 && age >= 20) {
```
KingOfSquares: Maybe dub this internal 20 tick limit for the death kit to some type of "quick respawn" in code?
Something like
```java
private static final long QUICK_RESPAWN_LIMIT = 20
```

So that
```suggestion
    if (!kitted && ticksUntilRespawn() <= 0 && age >= QUICK_RESPAWN_LIMIT) {
```
Or something similar...

KingOfSquares: Reason is I think it would be easier to understand when reading, but its a personal preference

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
Pablete1234: The modification makes no behavior change. I considered doing isAllowed, but that does imply a behavior change (abstaining used to be removed, now they would be accepted) which is why i made it not-denied (which is exactly the same behavior as current)

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
Pablete1234: Probably leaning more towards `MIN_KIT_TCKS` or similar, it's just to avoid giving you the item to leave/switch team instantly if you die and min respawn time is instant, and there isn't auto respawn. It's quite a stretch, but you never know who may make use of such combination in a map.

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
KingOfSquares: Not sure I understand, 

(From Filter.java)
```java
public boolean isDenied() {
      return this == DENY;
}
```
This implies that ABSTAIN will pass right? As `!DENY == ALLOW || ABSTAIN`

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
KingOfSquares: I like your naming better, and I agree it makes sense to limit giving the kit when the respawn is under one second(ish).

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
Pablete1234: Ignore everything i said, i was under the impression isAllowed was `allow || abstain`, and isDenied was `deny || abstain`, which does not seem to be the case at all.

I'll revert to keep the behavior unmodified

Electroid (approved):
🎉

PR #1174, opened 2023-04-22
Gamemode: resolve mode title typos/inconsistencies

calcastor:


TBG1000:
Taking advantage of this PR, @Cazadorian's comments on #dev could possibly be addressed here? When making #1168 I went for "Bedwars" and "Skywars" because that's what I'm used to, but didn't consider other servers/perspectives. If Hypixel is to be taken as the leading example, they use "Bed Wars" and "SkyWars", like Cazadorian said. So should those be resolved here too are they fine as is?

Pablete1234:
destoy the core, my fav gamemode

imo Bedwars and Skywars are fine, but up to mapdevs/the community to decide what they prefer

Pablete1234 (dismissed):

TheRealPear:
Going to be a little biased here, but I prefer Bed Wars and SkyWars. When I wrote the first Bodiam xml, I specifically used the capitalization+space in the gamemode title. From a quick Google search, it looks like Hypixel isn't the only server that follows this convention.

calcastor:
I would prefer to keep it as is personally, at last for Skywars; "SkyWars" gives me the impression of a sort-of "branded" version of the gamemode as it is on Hypixel and I don't think that a PGM implementation of that gamemode should make a point of being 1-to-1. Indifferent on "Bed Wars" over "Bedwars" but it makes sense to have the two words separate.

Maybe we could do "Sky Wars" but I don't know if I've seen the two words separate in many contexts when referring to the gamemode.

This probably needs more discussion and I consider it out of scope of this PR.

calcastor:
> This probably needs more discussion and I consider it out of scope of this PR.

On second thought, I can find frequent mentions of "Bed Wars" compared to "Bedwars" with a quick Google, but can't really say the same for "Sky Wars" and "Skywars". I'll change it to Bed Wars now.

TBG1000:
Agreed on "Bed Wars", would also keep "Skywars", didn't see a major inclination for "SkyWars" or "Sky Wars"

Electroid (approved):

Issue #1042, opened 2022-08-13
Timer tag option for spawners


ZeroFrosty:
A way to show how long until a spawner will spawn an item

Image below is an example of what im thinking of 
![Example](https://cdn.discordapp.com/attachments/785772883305299991/898413495022944256/2021-10-14_18.36.34.png)

KingOfSquares:
An alternative is porting core features from https://github.com/filoghost/HolographicDisplays (permissive license) and the 1.8 NMS stuff... but is it worth it for this thing only? Could we imagine other use cases for holograms in PGM?

Pablete1234:
pgm already does holograms for flag posts, it's not like we don't have the setup in place for such a thing

KingOfSquares:
> pgm already does holograms for flag posts, it's not like we don't have the setup in place for such a thing

I overthought it

arcadeboss:
Repost:

More maps are using bedwars-style store-currency spawners that are slow.
With longer delays it would be nice to inform players on spawner status, similar to bedwars.

Add a countdown option to the spawner feature so that players can see how long until next drop.
It would display numbers in a spatial position relative to the spawner drop position, for example:

3 [fades out],
2 [fades out]
1 [fades out and item spawns]

    countdown - Flag to show countdown or not, default false
    countdown_before - Number of seconds before drop to begin displaying the countdown
    countdown_position - Allow countdown to be positioned anywhere relative to the spawner, example above by 6 blocks.
    countdown_speedup - One or more points in game match time (5m, 10m) where the speed of the spawner can be increaseds

PR #1177, opened 2023-04-25
refactor holograms out of flags

KingOfSquares:
First step towards #1042 

Moves the hologram stuff for flags out into its own API, should create no behaviour changes


core/src/main/java/tc/oc/pgm/util/HologramText.java
```diff
@@ -0,0 +1,121 @@
...
+import org.bukkit.event.entity.EntityDamageByEntityEvent;
+import org.bukkit.event.entity.EntityDamageEvent;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.util.text.TextTranslations;
+
+/** Epic floating text, its like a sci-fi movie! */
+public class HologramText implements Listener {
```
Electroid: Instead of a util class, I think this should follow the `Feature` / `MatchModule` style.

```java
public class HologramMatchModule implements MatchModule {}
public class Hologram implements FeatureDefinition {} // see Portal for example
```

Electroid (changes requested):

core/src/main/java/tc/oc/pgm/util/HologramText.java
Pablete1234: keep in mind this is not something we're defining in xml, this is an utility that other parts of pgm make use of (eg: flags to render their post)

the plan is not to allow user-defined holograms via xml

note, that's not to say that it cannot be moved to a match module etc

KingOfSquares: Aha, so it's a singleton administrering all holograms? Sure.

core/src/main/java/tc/oc/pgm/flag/state/Uncarried.java
```diff
@@ -72,21 +74,11 @@ protected void placeBanner() {
       PGM.get().getGameLogger().severe("Failed to place banner at " + this.location);
     }
 
-    this.labelEntity =
-        this.location.getWorld().spawn(this.location.clone().add(0, 1.7, 0), ArmorStand.class);
```
Pablete1234, line 76: you're loosing this effectively-speaking location-set, that's relevant because the flag post moves, but your hologram does not
note: the behavior change of moving to constructor should be minimal, but may still carry consequences, so double-check that the behavior hasn't broken

KingOfSquares: Im confused, if the flag would move it would just enter a new state anyways? Or can the location change from when the state change is queued to when it actually happens

KingOfSquares: Digging in code, it seems like the state is created at the time the transition is queued. So it would use the location at the time of transition-start anyways. The only thing happening between constructor and `#enterState` being called is leaving the old state

core/src/main/java/tc/oc/pgm/hologram/Hologram.java
```diff
@@ -0,0 +1,100 @@
...
+   * Spawn in the hologram text at the given location, if not already spawned
+   *
+   * @param location the location to spawn the text at
+   */
+  public void show(@NotNull Location location) {
+    assertNotNull(location, "location");
+    if (this.labelEntity != null) return;
+    this.show();
+    this.setLocation(location);
+  }
```
Pablete1234: This method has a couple issues:
 - If there's already an entity, it won't move it
 - It first shows, then sets location?

```suggestion
  public void show(@NotNull Location location) {
    assertNotNull(location, "location");
    this.setLocation(location);
    this.show();
  }
```
After the change:
 - if there's already an entity, it will teleport (and won't re-show as show checks for entity already existing)
 - if there's not an entity already, it will first set the location (with no TP), and then spawn it directly in that new position.

Pablete1234 (changes requested):

Pablete1234 (dismissed):

Pablete1234:
@Electroid could you take a look at it?

PR #1178, opened 2023-05-01
Add map source location to /map output

KingOfSquares:
Adds the location of the map source files in the output when using /map and have the DEBUG permission. (idea by @Pablete1234)

This is a draft because giving opening for access to `MapSource` stuff through the `MapLibrary` seems a little weird, and I'm not sure what's the best approach. I thought of upgrading the object returned from `#getMap(String)` from `MapInfo` to some new object (basically `Pair<MapSource, MapInfo>`) as well as adding a method for just getting the source directly (e.g `#getMapSource()`). Both seemed wrong as it would then imply that you could load maps directly from the `MapSource` and not through `MapLibrary#loadNewMaps()` and `MapLibrary#loadExistingMap()`. 

My final attempt is the current state of this draft PR, which seemed like a better approach, but still feels a little weird since it still gives access to some "low level" map stuff. Feedback appreciated!

Draft also because javadocs and strings are WIP

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/map/MapLibrary.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/MapLibrary.java
@@ -1,9 +1,11 @@
 package tc.oc.pgm.api.map;
 
+import java.net.URI;
 import java.util.Iterator;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Stream;
 import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.map.exception.MapMissingException;
 import tc.oc.pgm.api.map.includes.MapIncludeProcessor;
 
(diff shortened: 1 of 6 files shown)


Pablete1234:
Tbh the approach is fine, exposing an url for the map is not a problem

Main things that i think still need to be addressed:
 - Paths shown should be relative to the server folder, so just like "maps/some-repo/mapname/map.xml" or similar
 - If the map is loaded from a git repo, it'd be nice if the url was the link to the file in the repo (eg: the github link)
 - If the map is loaded from a *public* (non password-protected) git repo, it'd be nice if the link was public, regardless of debug permission
  - If it's a link to a git repo, make it clickable

KingOfSquares:
Putting this on hold waiting on refactor of relevant code by @Pablete1234 

See #1179

Pablete1234:
I'm going to close this due to innactivity

PR #1179, opened 2023-05-08
Refactor map loading & add support for variants

Pablete1234:
Adds support for map variants in the xml:

```xml
<map>
  <name>Map Name</name>
  <variant id="5v5">5v5</variant>
  <variant id="tournament">TE</variant>
  <variant id="halloween" override="true">Spooky Map Name</variant>

  <constant id="team_size">25</constant>
  
  <if variant="5v5">
    <constant id="team_size">5</constant>
  </if>
  
  <if variant="tournament">
    <constant id="team_size">5</constant>
  </if>
  
  <if variant="halloween">
    <!-- NOTE: Terrain module currently doesn't support this, it's for a future PR -->
    <terrain world="halloween"/>
  </if>

  <team id="red" max="${team_size}"/>
  <team id="blue" max="${team_size}"/>
</map>
```

### Why?
This is a tool to simplify map management for mapdevs, currently if they have slightly different versions of the maps they need to be copy-pasting them and keeping different folders up-to-date, even later on when changes are made. This isn't a great experience.

### How do variants work?
Essentially the xml sample above, will load this as if it was 4 different maps: `Map Name`, `Map Name: 5v5`, `Map Name: TE` and `Spooky Map Name`.
If you /maps, they're 4 different maps, same for /setnext, or any other pgm command that takes maps in. It was designed this way to avoid as most breakage as possible, anything relying on map names or map ids/slugs will still be able to use those systems, as the name + variant will be the new slug.

### What features do variants enable?
 - Conditionals: `if`, `unless`, they will enable certain blocks of XML being added or not, depending on the variant
 - Constants: `constant` allows you to define text constants, which get replaced in any xml text along the xml in any text that has `${constant-name}`. This is useful to replace text in attributes conditionally without needing to re-define sometimes large portions of XML just to change one value.

Note: the constants are NOT necessarily tied to variants, you CAN use constants in your xml without variants. This can prove useful for includes. example:

Include:
```xml
<map>
  <constant id="team1-color">red</constant>
  <constant id="team1-name">Red Team</constant>
  <team id="team-1" color="${team1-color}" max="10">${team1-name}</team>
</map>
```
Your map:
```xml
<map>
  <include id="the-other-include-that-defines-teams"/>
  <constants>
    <constant id="team1-color">magenta</constant>
    <constant id="team1-name">Magenta Team</constant>
  </constant>
</map>
```
In this example the include defines the team (it is assumed that the rest of the include would use the team id etc). Now your map can simply override the team color & name for a team, without breaking or needing to have parts of the include in your own XML.

 
 ### Other changes
 This PR also adds changes to how git map sources/factories load stuff, this simplifies the codebase around that and should also make adding map url display in /map easier (see #1178 )
 
 ### Future work (for future PRs)
 - Terrain module should add support for world, like [1.9+ PGM docs](https://docs.oc.tc/modules/world) have
 - Map pools should have a "variants" optional attribute, that lets them say they prefer a certain variant if one exists. This makes xmas/halloween pools much easier to make (no need to rename all maps to their variant version)
  - Have some visibility of variants in /map, maybe display children variants and parent variant with clickable text or smthg.


core/src/main/java/tc/oc/pgm/api/map/MapContext.java
```diff
@@ -3,12 +3,12 @@
...
 
   /**
-   * Get a {@link MapSource} to access the maps's files.
+   * Get an immutable {@link MapInfo} which doesn't hold strong references to the context.
    *
-   * @return A {@link MapSource}.
+   * @return A {@link MapInfo} for this context.
    */
-  MapSource getSource();
+  MapInfo getMapInfo();
```
Electroid: I think `getInfo` is fine

core/src/main/java/tc/oc/pgm/api/map/MapSource.java
```diff
@@ -17,6 +19,21 @@ public interface MapSource {
    */
   String getId();
 
+  /**
+   * The variant of the map this is for
+   *
+   * @return the variant the source, null for the parent source
+   */
+  String getVariant();
```
Electroid, line 29: Should mark as `@Nullable`

core/src/main/java/tc/oc/pgm/api/map/MapSource.java
```diff
@@ -17,6 +19,21 @@ public interface MapSource {
...
+
+  /**
+   * A copy of the map source, tailored to a specific variant
+   *
+   * @param variant variant to use
+   * @return a new instance of map source, using the variant
+   */
+  MapSource withVariant(String variant);
```
Electroid: `asVariant`

Electroid (dismissed):
Looks good

PR #1186, opened 2023-05-28
Clear a players cached decoration on quit

Pugzy:
When a player is assigned or removed from a group with a flair (suffix or prefix) the current cache prevents this from updating live.

This can be confusing for players to know if they've received the rank, the next logical step for a player is to relog to see if that updates their flairs, the cache remains after players quit so this has no effect. Clearing the cache on player quit allows users to relog to invalidate their existing cache record, something that staff members could instruct a player to do.

It would also be nice to have a command that applies the same logic so that this can be run by other plugins (such as Buycraft) so that rank purchases are shown instantly. 

Couldn't come up with a good name for this command as the username cache doesn't just store the flairs but calling it something related to username cache may be confusing, I wouldn't expect players (staff or not) to be using this command so that shouldn't really matter, `/usernamecache reload [username]`, `/flair reload [username]`, `/decoration reload [username]` 🤷 ?

### DIFF
--- a/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistryImpl.java
+++ b/core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistryImpl.java
@@ -17,6 +17,7 @@ import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
 import org.bukkit.event.player.PlayerJoinEvent;
+import org.bukkit.event.player.PlayerQuitEvent;
 import org.bukkit.metadata.FixedMetadataValue;
 import org.bukkit.metadata.MetadataValue;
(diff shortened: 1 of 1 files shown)


Pablete1234 (dismissed):

core/src/main/java/tc/oc/pgm/namedecorations/NameDecorationRegistryImpl.java
```diff
@@ -71,6 +72,12 @@ public void onPartyChange(PlayerPartyChangeEvent event) {
     player.setDisplayName(getDecoratedName(player, party == null ? null : party.getColor()));
   }
 
+  @EventHandler
```
Pablete1234: You may want this on monitor, so that the leave message can make use of cached name

Pablete1234 (approved):

Electroid (approved):

PR #1187, opened 2023-05-28
Refactor NMSHacks to make multi-version support easier

cswhite2000:
## New description

This PR includes changes to make supporting other versions easier, and _should_ not include any change that alters behavior for 1.8.
This refactors the NMSHacks Interface into multiple interfaces and classes.

PGM will still only run on 1.8 with this PR.

## Old description
Felt like doing some work towards https://github.com/PGMDev/PGM/issues/852

This adds NoOp fallbacks for NMS code, which does cause some things to fail silently which is probably not exactly what we want.

This technically does allow PGM to run (poorly) on 1.9 and probably some other versions.
For example: countdowns don't work, but you can play a match with /start 0 and /cycle 0, the 5 maps that come with PGM seemed to work

Also, spigot 1.8.8 should work correctly again with this

Not sure if the style of `NMSHacksProvider.get()` is optimal, but it was easy to do / implement. I'm open to other suggestions.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/actions/FillAction.java
+++ b/core/src/main/java/tc/oc/pgm/action/actions/FillAction.java
@@ -9,6 +9,7 @@ import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.region.Region;
 import tc.oc.pgm.filters.query.BlockQuery;
+import tc.oc.pgm.util.nms.NMSHacks;
 
 public class FillAction extends AbstractAction<Match> {
 
@@ -32,7 +33,7 @@ public class FillAction extends AbstractAction<Match> {
       if (filter != null && filter.query(new BlockQuery(block)).isDenied()) continue;
 
       BlockState newState = block.getState();
-      newState.setMaterialData(materialData);
+      NMSHacks.setBlockStateData(newState, materialData);
 
       if (events) {
         BlockFormEvent event = new BlockFormEvent(block, newState);
--- a/core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
@@ -48,8 +48,10 @@ import tc.oc.pgm.events.PlayerPartyChangeEvent;
 import tc.oc.pgm.flag.event.FlagStateChangeEvent;
 import tc.oc.pgm.util.MapUtils;
 import tc.oc.pgm.util.MethodHandleUtils;
+import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.collection.ContextStore;
 import tc.oc.pgm.util.event.PlayerCoarseMoveEvent;
+import tc.oc.pgm.util.nms.NMSHacks;
 
 /**
  * Handler of dynamic filter magic.
@@ -105,6 +107,9 @@ public class FilterMatchModule implements MatchModule, FilterDispatcher, Tickabl
 
   // Filterables that need a check in the next tick (cleared every tick)
   private final Set<Filterable<?>> dirtySet = new HashSet<>();
+
+  // cleanup players from onPartyChange for non SportPaper servers
+  private Set<MatchPlayer> nonSportCleanUpSet = new HashSet<>();
 
   public ContextStore<? super Filter> getFilterContext() {
     return filterContext;
@@ -387,6 +392,15 @@ public class FilterMatchModule implements MatchModule, FilterDispatcher, Tickabl
   }
 
   public void tick() {
+    // Always empty when the server is running sportpaper
+    if (!nonSportCleanUpSet.isEmpty()) {
+      for (MatchPlayer matchPlayer : nonSportCleanUpSet) {
+        dirtySet.remove(matchPlayer);
+        this.lastResponses.columnKeySet().remove(matchPlayer);
+      }
+      nonSportCleanUpSet.clear();
+    }
+
     final Set<Filterable<?>> checked = new HashSet<>();
     Set<Filterable<?>> checking;
     // Collect Filterables that are dirty, and have not already been checked in this tick
@@ -483,7 +497,7 @@ public class FilterMatchModule implements MatchModule, FilterDispatcher, Tickabl
 
     if (player != null) {
       this.invalidate(player);
-      PGM.get().getServer().postToMainThread(PGM.get(), true, this::tick);
+      NMSHacks.postToMainThread(PGM.get(), true, this::tick);
     }
   }
 
@@ -530,11 +544,16 @@ public class FilterMatchModule implements MatchModule, FilterDispatcher, Tickabl
                 }
               });
 
-      event.yield();
+      if (BukkitUtils.isSportPaper()) {
+        event.yield();
 
-      // Wait until after the event to remove them, in case they get invalidated during the event.
-      dirtySet.remove(event.getPlayer());
-      this.lastResponses.columnKeySet().remove(event.getPlayer());
+        // Wait until after the event to remove them, in case they get invalidated during the event.
+        dirtySet.remove(event.getPlayer());
+        this.lastResponses.columnKeySet().remove(event.getPlayer());
+      } else {
+        // Clean up at the start of the next tick. Most platforms don't include event.yield()
+        nonSportCleanUpSet.add(event.getPlayer());
+      }
     }
   }
 
--- a/core/src/main/java/tc/oc/pgm/snapshot/BudgetWorldEdit.java
+++ b/core/src/main/java/tc/oc/pgm/snapshot/BudgetWorldEdit.java
@@ -8,6 +8,7 @@ import org.bukkit.util.BlockVector;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.region.Region;
 import tc.oc.pgm.util.BlockData;
+import tc.oc.pgm.util.nms.NMSHacks;
 
 /**
  * Utils to save, remove and paste blocks in some {@link Region} in some {@link Match} using the
@@ -33,7 +34,7 @@ class BudgetWorldEdit {
     for (BlockData blockData : snapshot.getMaterials(region)) {
 
       BlockState state = blockData.getBlock(world, offset).getState();
-      state.setMaterialData(blockData.getMaterialData());
+      NMSHacks.setBlockStateData(state, blockData.getMaterialData());
       state.update(true, true);
     }
   }
--- a/core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
+++ b/core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
@@ -52,7 +52,8 @@ public class WorldSnapshot {
     ChunkSnapshot chunkSnapshot = chunkSnapshots.get(chunkVector);
     if (chunkSnapshot != null) {
       BlockVector chunkPos = chunkVector.worldToChunk(x, y, z);
-      state.setMaterialData(
+      NMSHacks.setBlockStateData(
+          state,
           new MaterialData(
               chunkSnapshot.getBlockTypeId(
                   chunkPos.getBlockX(), chunkPos.getBlockY(), chunkPos.getBlockZ()),
--- a/core/src/main/java/tc/oc/pgm/tnt/TNTMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/tnt/TNTMatchModule.java
@@ -69,6 +69,17 @@ public class TNTMatchModule implements MatchModule, Listener {
     }
   }
 
+  private static Sound FUSE_SOUND = chooseFuseSound();
+
+  private static Sound chooseFuseSound() {
+    try {
+      return Sound.FUSE;
+    } catch (NoSuchFieldError error) {
+      // TODO: make or use a 1.8 -> 1.9+ sound conversion api
+      return Sound.valueOf("ENTITY_TNT_PRIMED");
+    }
+  }
+
   @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
   public void handleInstantActivation(BlockPlaceEvent event) {
     if (this.properties.instantIgnite && event.getBlock().getType() == Material.TNT) {
@@ -88,7 +99,7 @@ public class TNTMatchModule implements MatchModule, Listener {
 
       if (callPrimeEvent(tnt, event.getPlayer())) {
         event.setCancelled(true); // Allow the block to be placed if priming is cancelled
-        world.playSound(tnt.getLocation(), Sound.FUSE, 1, 1);
+        world.playSound(tnt.getLocation(), FUSE_SOUND, 1, 1);
 
         ItemStack inHand = event.getPlayer().getItemInHand();
         if (inHand.getAmount() == 1) {
--- a/core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
@@ -68,18 +68,19 @@ public class TNTRenderMatchModule implements MatchModule, Listener {
 
     public PrimedTnt(TNTPrimed entity) {
       this.entity = entity;
-      this.lastLocation = currentLocation = entity.getLocation().toBlockLocation();
+      this.lastLocation = currentLocation = toBlockLocation(entity.getLocation());
     }
 
     public boolean update() {
       if (entity.isDead()) {
-        for (MatchPlayer viewer : viewers)
+        for (MatchPlayer viewer : viewers) {
           NMSHacks.sendBlockChange(currentLocation, viewer.getBukkit(), null);
+        }
         return true;
       }
 
       this.lastLocation = currentLocation;
-      this.currentLocation = entity.getLocation().toBlockLocation();
+      this.currentLocation = toBlockLocation(entity.getLocation());
       this.moved = !currentLocation.equals(lastLocation);
 
       for (MatchPlayer player : match.getPlayers()) {
@@ -100,5 +101,14 @@ public class TNTRenderMatchModule implements MatchModule, Listener {
         NMSHacks.sendBlockChange(lastLocation, player.getBukkit(), null);
       }
     }
+  }
+
+  private static Location toBlockLocation(Location location) {
+    // Spigot 1.8 doesn't have Location.toBlockLocation()
+    Location blockLoc = location.clone();
+    blockLoc.setX(blockLoc.getBlockX());
+    blockLoc.setY(blockLoc.getBlockY());
+    blockLoc.setZ(blockLoc.getBlockZ());
+    return blockLoc;
   }
 }
--- a/util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
+++ b/util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
@@ -1,253 +1,377 @@
 package tc.oc.pgm.util.nms;
 
-import com.google.common.collect.HashMultimap;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.SetMultimap;
-import com.mojang.authlib.GameProfile;
-import com.mojang.authlib.properties.Property;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
-import java.util.*;
+import java.util.Collection;
+import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
-import net.md_5.bungee.api.chat.BaseComponent;
-import net.minecraft.server.v1_8_R3.*;
-import net.minecraft.server.v1_8_R3.Item;
-import net.minecraft.server.v1_8_R3.WorldBorder;
-import org.bukkit.*;
+import java.util.logging.Logger;
+import org.bukkit.Bukkit;
 import org.bukkit.Chunk;
 import org.bukkit.ChunkSnapshot;
+import org.bukkit.GameMode;
+import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.World;
-import org.bukkit.craftbukkit.v1_8_R3.CraftChunk;
-import org.bukkit.craftbukkit.v1_8_R3.CraftWorld;
-import org.bukkit.craftbukkit.v1_8_R3.block.CraftBlock;
-import org.bukkit.craftbukkit.v1_8_R3.entity.*;
-import org.bukkit.craftbukkit.v1_8_R3.inventory.CraftItemStack;
-import org.bukkit.craftbukkit.v1_8_R3.scoreboard.CraftTeam;
-import org.bukkit.craftbukkit.v1_8_R3.util.CraftMagicNumbers;
-import org.bukkit.entity.*;
+import org.bukkit.WorldCreator;
+import org.bukkit.block.Block;
+import org.bukkit.block.BlockState;
 import org.bukkit.entity.Entity;
+import org.bukkit.entity.Fireball;
+import org.bukkit.entity.Firework;
+import org.bukkit.entity.Item;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.entity.Player;
 import org.bukkit.event.player.PlayerTeleportEvent;
-import org.bukkit.inventory.DoubleChestInventory;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.meta.ItemMeta;
 import org.bukkit.inventory.meta.SkullMeta;
 import org.bukkit.material.MaterialData;
+import org.bukkit.plugin.Plugin;
 import org.bukkit.potion.PotionEffectType;
 import org.bukkit.scoreboard.NameTagVisibility;
 import org.bukkit.util.Vector;
 import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.util.ClassLogger;
+import tc.oc.pgm.util.attribute.AttributeMap;
 import tc.oc.pgm.util.attribute.AttributeModifier;
 import tc.oc.pgm.util.block.RayBlockIntersection;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
-import tc.oc.pgm.util.bukkit.ViaUtils;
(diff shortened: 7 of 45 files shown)


util/src/test/java/tc/oc/pgm/util/text/TextExceptionTest.java
```diff
@@ -1,25 +1,20 @@
 package tc.oc.pgm.util.text;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
```
cswhite2000: Should probably not have disabled this test

KingOfSquares:
Maybe the NMSHacks interface can default to some type of error logging when trying to call unsupported methods? To enable the possibility for (server-owner) debug on why some feature might not be working. Then you would also not need to implement empty methods for every NMSHacks type.

I like this framework idea! Should make it easier to add support for higher versions. Maybe there is a point in adding an abstraction layer(Or renaming some class in the current proposed model) to make it more like a "Version dependent implementation" thing? Such that it could be an NMS hack, or it could be native api. 
Imagine some feature that got api in the newest paper API as an example. Code would refer to some version dependent code provider which would for the old version supply NMSHacks but for the new version native api. Or different native api if it ever changes etc...
At that point though it could also be a discussion of being able to abstract PGM out of Bukkit, even. (Imagine this "Version Dependent" provider thing called everywhere...)

util/src/main/java/tc/oc/pgm/util/nms/NMSHacksProvider.java
```diff
@@ -0,0 +1,25 @@
...
+    try {
+      if (BukkitUtils.isSportPaper()) {
+        return new NMSHacksSportPaper();
+      } else {
+        return new NMSHacks1_8();
+      }
+    } catch (Throwable t1) {
+      return new NMSHacksNoOp();
+    }
+  }
```
KingOfSquares: Adding some debug statement here posting which NMSHacks was chosen is probably also a good idea.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacksSportPaper.java
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  @Override
+  public void setAffectsSpawning(Player player, boolean affectsSpawning) {
+    player.spigot().setAffectsSpawning(affectsSpawning);
+  }
+
+  @Override
+  public void entityAttach(Player player, int entityID, int vehicleID, boolean leash) {
+    sendPacket(player, new PacketPlayOutAttachEntity(entityID, vehicleID, leash));
```
KingOfSquares, line 61:
```suggestion
    NMSHacksProvider.get().sendPacket(player, new PacketPlayOutAttachEntity(entityID, vehicleID, leash));
```
I saw you doing something like this at some other place in your code and I think it should be the standard way of calling methods. In case some NMSHacks instance down the line overrides some other this would call the "most correct" implementation of the method.

Pablete1234:
Personally i'd try to keep the `NMSHacks.method()` format across the code (instead of `NMSHacksProvider.get().method()`), and swallow the required changes internally in nmshacks.

What that means is essentally you'll have:
```java
interface NMSHacks {
  private final NMSHacksPlatform INSTANCE = buildNmsHacksPlatform();

  static void sendPacketToAll(Packet packet) {
    for (Player player: Bukkit.getOnlinePlayers()) {
        INSTANCE.sendPacket(packet, player);
    }
  }

  static void sendPacket(Packet packet, Player player) {
    INSTANCE.sendPacket(packet, player);
  }

}

interface NMSHacksPlatform {
  void sendPacket(Packet packet, Player player);
}

class NMSHacks1_8 implements NMSHacksPlatform {
  void sendPacket(Packet packet, Player player) {
    // send it
  }
}
class NMSHacksSportPaper implements NMSHacksPlatform {
  void sendPacket(Packet packet, Player player) {
    // send it
  }
}
```

The methods that must be different for diff implementations can go inside the "nms hacks platform" (maybe got a suggestion for a better name?) then the more generic stuff that just re-uses calls to diff methods can stay in the static nmshacks method in the interface.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacksProvider.java
cswhite2000: Yeah I'll add something here to that effect

cswhite2000:
With regards to @Pablete1234's comment
> Personally i'd try to keep the `NMSHacks.method()` format across the code (instead of `NMSHacksProvider.get().method()`), and swallow the required changes internally in nmshacks.
> ...
> The methods that must be different for diff implementations can go inside the "nms hacks platform" (maybe got a suggestion for a better name?) then the more generic stuff that just re-uses calls to diff methods can stay in the static nmshacks method in the interface.

I can live with that
This would also mean that other plugins that use PGM's NMSHacks, eg Community, wouldn't need to be updated for this change.
`NMSHacksPlatform` seems fine to me, its better than anything I can think of.

to @KingOfSquares 
> I like this framework idea! Should make it easier to add support for higher versions. Maybe there is a point in adding an abstraction layer(Or renaming some class in the current proposed model) to make it more like a "Version dependent implementation" thing? Such that it could be an NMS hack, or it could be native api.
> Imagine some feature that got api in the newest paper API as an example. Code would refer to some version dependent code provider which would for the old version supply NMSHacks but for the new version native api. Or different native api if it ever changes etc...
> At that point though it could also be a discussion of being able to abstract PGM out of Bukkit, even. (Imagine this "Version Dependent" provider thing called everywhere...)

Yeah, I had similar thoughts, but that's work for future PRs. With regards to abstracting PGM out of Bukkit, the two are too tightly coupled at this time to really start working on it.

Electroid:
I like Pab's suggestion. In the future, if we start to separate the PGM and Bukkit APIs from each other, it would likely refactor this code again. But that's a big if whether we get to it. And therefore, I think his suggested approach would be good.

cswhite2000:
Fixed countdowns not working in 1.9+

I believe I've addressed all of the comments so far.

Can I get any further reviews/feedback?

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
```diff
@@ -42,847 +32,277 @@
...
 
-  static void sendPacket(Player bukkitPlayer, Object packet) {
-    if (bukkitPlayer.isOnline()) {
-      EntityPlayer nmsPlayer = ((CraftPlayer) bukkitPlayer).getHandle();
-      nmsPlayer.playerConnection.sendPacket((Packet) packet);
-    }
-  }
+  static NMSHacksPlatform chooseNMSHacks() {
+    NMSHacksPlatform choice;
+    Logger logger = Logger.getLogger("NMSHacks");
```
KingOfSquares: Maybe use `tc.oc.pgm.util.ClassLogger` here?

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
```diff
@@ -42,847 +32,277 @@
...
+        choice = new NMSHacksSportPaper();
+        logger.info("Using NMSHacksSportPaper");
+      } else {
+        choice = new NMSHacks1_8();
+        logger.info("Using NMSHacks1_8");
       }
+    } catch (Throwable t1) {
+      choice = new NMSHacksNoOp();
+      logger.warning("Using Fallback NMSHacksNoOp!");
     }
```
KingOfSquares:
```suggestion
      if (BukkitUtils.isSportPaper()) {
        choice = new NMSHacksSportPaper();
      } else {
        choice = new NMSHacks1_8();
      }
      logger.info("Using " + choice.getClass().getSimpleName());
    } catch (Throwable t1) {
      choice = new NMSHacksNoOp();
      logger.log(Level.WARNING, t1, () -> "Using Fallback NMSHacksNoOp!");
    }
```

Can't remember if the log entry includes the throwable in the terminal by default, but nice to include if it is :)

core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
```diff
@@ -101,4 +102,12 @@ private void updatePlayer(MatchPlayer player) {
...
     }
   }
+
+  private static Location toBlockLocation(Location location) {
+    Location blockLoc = location.clone();
+    blockLoc.setX(blockLoc.getBlockX());
+    blockLoc.setY(blockLoc.getBlockY());
+    blockLoc.setZ(blockLoc.getBlockZ());
+    return blockLoc;
+  }
```
KingOfSquares: If this is necessary because of multi-platform/version compat maybe add a comment about it?

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
cswhite2000: How would I get the parent PGM logger from a static instance in Util?
I guess I could just make a new ClassLogger with no parent, but that would defeat the purpose of the ClassLogger class.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
cswhite2000: Would the error actually be useful?
It's just going to print class not found, which is expected if you aren't using 1.8 and try to use 1.8 NMS code.

core/src/main/java/tc/oc/pgm/tntrender/TNTRenderMatchModule.java
cswhite2000: Yeah, it's needed for version compatibility. I can add a comment

cswhite2000:
Did some further testing with this and pgm seems playable on spigot 1.9 -1.11.2. Although I did not test anything requiring multiple players. Spigot 1.12 crashed.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
KingOfSquares: Does it not handle some prefix stuff? Or is that standard logger behaviour (thinking of `->[PGM]<- This is a log`)

cswhite2000: I was able to have it use Classlogger.get() without a parent logger, which is fine for NMSHacks

cswhite2000:
Should now be playable on Spigot 1.12 as well. 1.13 will require more significant changes, and is out of scope for this PR. I was able to test combat between players, which worked.

core/src/main/java/tc/oc/pgm/action/actions/FillAction.java
```diff
@@ -32,7 +32,8 @@ public void trigger(Match match) {
       if (filter != null && filter.query(new BlockQuery(block)).isDenied()) continue;
 
       BlockState newState = block.getState();
-      newState.setMaterialData(materialData);
+      newState.setType(materialData.getItemType());
+      newState.setData(materialData);
```
Pablete1234: A change like this is bound to break performance for large fills.

The current method is a direct setter just doing two simple number assignment, the replacement is a look-up of int to material, then assigning, then checking validity of data, then setting.
https://github.com/Electroid/SportPaper/blob/5258d6c131b1d0167c6f29bbef8b1c952e0895f0/patches/server/0053-Improve-BlockState-API.patch#L19

While the change may not look like much, making 10k+ calls to getMaterialById *will* inevitably add-up

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
```diff
@@ -530,7 +535,9 @@ public void onPartyChange(PlayerPartyChangeEvent event) throws EventException {
                 }
               });
 
-      event.yield();
+      if (BukkitUtils.isSportPaper()) {
+        event.yield(); // TODO: what does not having this, when using spigot / paper, break?
```
Pablete1234: What can break, is a memory leak. This will occur if anything reacts to this event on a later priority and a dynamic filter is listening, as that may re-add the player to dirtyset/listeners. If you want proper logic, not being able to yield means having to delay the rest too

core/src/main/java/tc/oc/pgm/snapshot/BudgetWorldEdit.java
```diff
@@ -33,7 +34,9 @@ public void placeBlocks(Region region, BlockVector offset) {
     for (BlockData blockData : snapshot.getMaterials(region)) {
 
       BlockState state = blockData.getBlock(world, offset).getState();
-      state.setMaterialData(blockData.getMaterialData());
+      MaterialData materialData = blockData.getMaterialData();
+      state.setType(materialData.getItemType());
+      state.setData(materialData);
```
Pablete1234: As before, for large places this may hurt perf

core/src/main/java/tc/oc/pgm/tnt/TNTMatchModule.java
```diff
@@ -88,7 +88,12 @@ public void handleInstantActivation(BlockPlaceEvent event) {
 
       if (callPrimeEvent(tnt, event.getPlayer())) {
         event.setCancelled(true); // Allow the block to be placed if priming is cancelled
-        world.playSound(tnt.getLocation(), Sound.FUSE, 1, 1);
+        try {
+          world.playSound(tnt.getLocation(), Sound.FUSE, 1, 1);
+        } catch (NoSuchFieldError error) {
```
Pablete1234: A no-such-field happening every time tnt places isnt great, move the sound to a static final field and compute it once

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
```diff
@@ -483,7 +484,11 @@ public void onPlayerMove(PlayerCoarseMoveEvent event) {
 
     if (player != null) {
       this.invalidate(player);
-      PGM.get().getServer().postToMainThread(PGM.get(), true, this::tick);
+      if (BukkitUtils.isSportPaper()) {
+        PGM.get().getServer().postToMainThread(PGM.get(), true, this::tick);
+      } else {
+        PGM.get().getServer().getScheduler().runTask(PGM.get(), this::tick);
```
Pablete1234: These are fundamentally different as one runs at the very beginning of any other main thread operation, while the other schedules to run arbitrarily later.
This is bound to subtly break certain maps which abuse regions as dynamic filters, as 2 move packets received shortly after another will not be processed the same.

Current behavior will process one move (then post ticking at the front), then the other.
Meanwhile, non-sb code processes one move, schedules tick at the end, then takes the 2nd one, then schedules tick at the end, and then it processes both those scheduled ticks, at which point the new position may no longer be the prior one, and filters will react differently.

This, in short, can let players walk thru a 1-thick region undetected if they lag a bit.

core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
```diff
@@ -52,13 +52,15 @@ public BlockState getOriginalBlock(int x, int y, int z) {
...
+      MaterialData materialData =
           new MaterialData(
               chunkSnapshot.getBlockTypeId(
                   chunkPos.getBlockX(), chunkPos.getBlockY(), chunkPos.getBlockZ()),
               (byte)
                   chunkSnapshot.getBlockData(
-                      chunkPos.getBlockX(), chunkPos.getBlockY(), chunkPos.getBlockZ())));
+                      chunkPos.getBlockX(), chunkPos.getBlockY(), chunkPos.getBlockZ()));
+      state.setType(materialData.getItemType());
+      state.setData(materialData);
```
Pablete1234: Same here

util/src/main/java/tc/oc/pgm/util/tablist/TabDisplay.java
```diff
@@ -84,19 +81,12 @@ public TabDisplay(Player viewer, int width) {
...
-          .add(NMSHacks.playerListPacketData(listRemovePacket, uuid, renderedPlayerName));
+      NMSHacks.addPlayerInfoToPacket(
+          listAddPacket, uuid, name, GameMode.SURVIVAL, 9999, null, renderedPlayerName);
+      NMSHacks.addPlayerInfoToPacket(listRemovePacket, uuid, renderedPlayerName);
     }
   }
 
-  public int getWidth() {
-    return width;
-  }
```
Pablete1234, line 98: Any particular reason this is going? 1.7 supports 3 to 10 columns, and this class is modeled after thst tablist display

Pablete1234 (changes requested):
I have the general feeling this pr isnt really solving issues with compatibility but rather hiding them under the rug, while possibly creating new issues.
Im not sure if this is somthing we should be aiming to instead of finding proper non-breaking compatibility changes

core/src/main/java/tc/oc/pgm/action/actions/FillAction.java
cswhite2000: Should I make a method in NMSHacks for this? We'll need to make an api for all materialdata usage later regardless, as more recent versions have no support for MaterialData.

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
cswhite2000: That sounds like it would require some more significant reworking. Could I, for this PR, just simply disable dynamic filters when a server isn't running SportPaper?

core/src/main/java/tc/oc/pgm/tnt/TNTMatchModule.java
cswhite2000: Will do

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
cswhite2000: Do you know if there is an equivalent alternative in spigot? If not I'm leaning toward simply disabling dynamic filters if a server isn't running SportPaper.

util/src/main/java/tc/oc/pgm/util/tablist/TabDisplay.java
cswhite2000: I believe I removed it because it was unused. Although I didn't check to see if any other plugins might be using it.

cswhite2000:
> I have the general feeling this pr isnt really solving issues with compatibility but rather hiding them under the rug, while possibly creating new issues. Im not sure if this is something we should be aiming to instead of finding proper non-breaking compatibility changes

It is not really my intention to fix compatibility issues with this pr. My intention is more to allow for proper compatibility to be added piecemeal. Essentially my goal is for NMSHacksNoOp to be a fallback that will allow most PGM features to work, but not all.

I do think that this PR does lack proper communication of what features get broken.
We could also hide NMSHacksNoOp behind an experimental config option, and then remove it in a future PR that does add that communication.

core/src/main/java/tc/oc/pgm/action/actions/FillAction.java
Pablete1234: I'd go for that, yeah

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
Pablete1234: I'd really advise against that idea, its going to break half the maps. Thats just a non-functioning pgm you're getting

cswhite2000: Ah yeah, I see what you mean from the [docs](https://pgm.dev/docs/modules/mechanics/filters#player-filters) about how many filters are dynamic. Ideally, I'd like to avoid refactoring dynamic filters in this PR. Do you have any suggestions for a strategy for this?

Pablete1234: I think you're gonna need to search for work-arounds even if they're ugly (reflection/nms) for these. I dont think there's a better way around it.

cswhite2000: The idea behind a NoOp NMS is basically as a fallback for when NMS doesn't work. ie bukkit only, no craftbukkit

Pablete1234: Why are we evwn trying to supoort that? Pgm fundamentqlly relies on craftbukkit/nms. What's the target we expect to run on?

cswhite2000: Well really in this case the target is all pgm versions between 1.8 and 1.20. A NoOp implementation in general would make it easier to implement specific version implementations. Specifically, for my purposes, NMSHacksNoOp is targeting only bukkit so that I can increment the Minecraft version number and make PGM able to run with the changes introduced in each version, without having to upgrade an entire NMSHacks implementation. I can then get each change properly reviewed and allow future PRs to be smaller, and therefore less likely to miss any issues. I believe that if we left this on a separate branch and just kept adding changes until we have a version that supports 1.8 and 1.20 that it simply wouldn't happen, and with that approach we will never have any sort of multi-version support.

cswhite2000: Also, given how many features do work with no NMS in this PR, I don't think PGM is as tied to craftbukkit/nms as you think. It is definitely a lesser experience, but not as much as you might think.

Pablete1234: If the no-op impl is only to be used as a base, and it itself cant really function, why not make it abstract and leave methods which cannot be reasonably implemented within bukkit api unimplemented, forcing children (which can have some reliance on CF/nms) to implement those?

Pablete1234: > given how many features do work with no NMS in this PR, I don't think PGM is as tied to craftbukkit/nms as you think

I think the 80/20 rule applies here, where 80% of functionality can be achieved with 20% of the work, and the remaining 20% of the functionality takes 80% of the work.
PGM will "mostly" work on a generic bukkit implementation (non-craftbukkit) but there will be the 20% of missing/broken/subtly-broken features, or bugs that will hide there and no one will fix. I'd rather not claim we support non-craftbukkit bukkit API implementations, and just plainly tell such users on startuo we don't support whatever thing they're trying to run on

cswhite2000: I'm good with keeping it abstract. I agree that we don't want people to try and run the NoOp implementation thinking it is properly supported.

core/src/main/java/tc/oc/pgm/map/MapLibraryImpl.java
```diff
@@ -161,8 +161,8 @@ public CompletableFuture<?> loadNewMaps(boolean reset) {
               }
 
               try (Stream<? extends MapSource> stream =
-                  mapSources.stream().flatMap(Function.identity()).parallel().unordered()) {
```
cswhite2000: This commit https://github.com/PGMDev/PGM/commit/e1ddd1bda55e5884d265b5fd57d96f4e84b6759f broke compilation for me, this fixes that. This does basically the same thing

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
cswhite2000: Made an NMSHacks method for the spigot 1.8 version of this

core/src/main/java/tc/oc/pgm/filters/FilterMatchModule.java
cswhite2000: Added a cleanup at the beginning of the next tick for when the platform don't support yielding

core/src/main/java/tc/oc/pgm/map/MapLibraryImpl.java
Pablete1234: how would this break compilation!?

cswhite2000: I honestly don't know.
It gave the error: `[164,88] incompatible types: java.util.stream.Stream<java.lang.Object> cannot be converted to java.util.stream.Stream<? extends tc.oc.pgm.api.map.MapSource>`
Which made no sense to me, because that should certainly not have failed. Also given that failure, my fix really shouldn't have worked either.

cswhite2000:
I believe I've addressed everything so far. This PR itself only allows PGM to run on 1.8, it simply makes adding support for other versions easier. This _should_ have no behavior changes for 1.8.

I do not believe any further work is needed for this PR, unless prompted by reviewers.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacksNoOp.java
```diff
@@ -0,0 +1,192 @@
...
+        return Enchantment.LURE;
+      default:
+        return Enchantment.getByName(key);
+    }
+  }
+
+  @Override
+  public long getMonotonicTime(World world) {
+    return Instant.now().toEpochMilli() / 50;
+  }
```
Pablete1234: i believe this could break some stuff that expects the same time to be returned during the whole tick, as calling this twice within the same tick would return different values. For example see WorldTickClock#now(), usually it'd only create a `new Tick` if the value changed, now it'd create it every time that's called if it's gone over a 50ms border and leaked into the next millisecond.

util/src/main/java/tc/oc/pgm/util/nms/NMSHacksNoOp.java
```diff
@@ -0,0 +1,192 @@
...
+
+  @Override
+  public int getPing(Player player) {
+    return 0;
```
Pablete1234: i'd use a more reasonable default like 100ms for this

cswhite2000: Will do 👍

util/src/main/java/tc/oc/pgm/util/nms/NMSHacksNoOp.java
cswhite2000: Given that this doesn't seem to change much between versions it could actually be handled fairly well with reflection by my class `util/src/main/java/tc/oc/pgm/util/reflect/MinecraftReflectionUtils.java` from https://github.com/PGMDev/PGM/pull/1199. Given that this code won't actually ever be run in this pull request, could I simply handle it in that PR?

cswhite2000: I've gone ahead and handled this in the other PR. Let me know if you want it in this PR, although doing so would require more rebasing than I'd like to do for my other PR.

Pablete1234: I think it'd be nice to merge in this pr, as it can be merged already once this is adressed, rather than waiting and the other PR merging all at once

Pablete1234 (approved):

Electroid (approved):
Really appreciate all the hard work you put into this, excited to see it land! 🎉

PR #1197, opened 2023-06-26
Fix compile error on newer jdks & cleanup

Pablete1234:
Apparently java complains about some of the generics involved here but only on some jdk versions, because it can't quite figure out the bounds. There was no reason for the generics to be so complicated and simplifying them fixes it.


cswhite2000:
This fixes the compilation issues I was having 👍

Electroid (approved):

PR #1198, opened 2023-07-02
Add Parkour to supported gamemodes

HSQuillin:
Parkour maps have been a genre that have been growing and the most recent submission being two weeks ago. Has gradually populated the map repo and some examples include Parkour: Jungle, Parkour: Debris, spins such as Build: Parkour, and has expanded to Team parkour with Team Parkour: Ice & Team Parkour: Jungle. 33 pages of mentions in Discord of Parkour in OCC. Relevant to include, other forks like Stratus have Parkour in default rotations and are played frequently and has grown in PGM.

Saw other files have translated components commented out, wasn't sure if those were necessary to include in this PR since I don't have the translations for it. Happy to incude them if necessary.


arcadeboss:
I see the existing Parkour games seem to be "scorebox" gamemode, even though they don't designate it explicitly.

Parkour could be a logical alternate name for scorebox mode, if the desired behavior is the same, with the point being to perhaps better convey the game to players via PGM labels.   Is this the current intended benefit of these commits?

It would be interesting, if parkour gamemode had different rules from scorebox to make this a distinct gamemode, but perhaps there are no relevant differences.

HSQuillin:
Currently, the way of making Parkour legitimately work is by making the objective a scorebox. Although this approach is slightly hacky (not the intended use for a scorebox), it works quite well. My line of thinking is that, since this is a game that the community is fleshing out and there are multiple demonstrated uses of Parkour in PGM existing across servers, it deserves to have its own gamemode.

The potential for Parkour is to develop in its own style to include things like checkpoints, spirals (blocks that rotate you automatically when you stand on them a-la Spiral Parkour mode), etc. which can be supported later down the line.

Pablete1234:
There's a difference between tags and gamemodes. Scorebox is a tag, for one of pgm's features, but it doesn't define the map. It is often used in either arcade, or TDM maps. In this case, parkour would be a sub-genere of the arcade gamemode, and i don't see any reason why it couldn't be there as a gamemode to more closely represent what the map is about. I believe there's enough parkour maps in pgm so that we can justify having this as a separate gamemode.

Pablete1234:
@Electroid any opinions on the matter? should we promote parkour to be its own pgm gamemode, or leave it to stay within arcade?

TheRealPear:
I'm not Electroid, but this would be useful for Warzone as well. We have a few parkour maps that we wouldn't necessarily consider arcade. It would be important to us that they have a separate tag for stat tracking purposes.

util/src/main/i18n/translations/gamemode_en_GB.properties
TheRealPear, line 1: This should be done in the [gamemode.properties](https://github.com/PGMDev/PGM/blob/dev/util/src/main/i18n/templates/gamemode.properties) file instead. This file is for English UK.

BennyDoesStuff:
Would a potential alternative here be adding support for custom map tags? Parkour maps could be tagged as parkour accordingly and used to look up and filter where needed.

Pablete1234:
Map tags are, as of now, considered "facts" about a map: "this map has scoreboxes", "this map has monuments", "this map has wools".
The gamemode is a subjective "this map is arcade" or "this map is a CTW map", often derived from the tags that make up a map, but optionally overriden by the map author's will to use either a specific gamemode, or an arbitrary string to define the mode (custom gamemodes).

That's not to say that this has to be the system forever or that it can't be changed, but at least as of now, it makes more sense for this to be a gamemode (or, just a `<game>Parkour</game>` for arbitrary game string, which you can already do), than to be a tag

PR #1199, opened 2023-07-03
Support 1.9 using ProtocolLib

cswhite2000:
Depends on https://github.com/PGMDev/PGM/pull/1187
Leaving this as a draft until the other PR is merged. Although I'm also open to simply going with this PR.

From my work on multi-version support, incremental version support seems easier than skipping versions. Therefore this adds support for 1.9.

Notes:
- This relies on ProtocolLib, ProtocolLib is added as a soft Dependency for PGM
    - It functions as a hard dependency for versions higher than 1.8 
- I've added @Pablete1234's [TablistResizer](https://github.com/Pablete1234/TablistResizer) into PGM itself
- This adds a dependency on [NBTEditor](https://github.com/BananaPuncher714/NBTEditor)
    - This is a single class Reflection based NBT Editing Library, supports 1.8 - 1.20, has an MIT license
    - I've shaded this into the PGM plugin jar
- I've left weapons and such to use vanilla behavior, server owners can use includes with attributes or other plugins to change PVP to the pre-1.9 style if they wish.
- I'm avoiding adding maven dependencies on different bukkit/spigot versions until it is necessary (1.13)
- I've done significant testing on this and PGM on 1.9 has all the same features that PGM 1.8 spigot has

Future Work:
- 1.10 - 1.12 support will not require significant changes. I'm leaving those out of this PR to keep the size of this PR Smaller
- After this PR I'll add version support sequentially, going through each version
- 1.13 will require more significant work (Material changes). This will be easier if working off of a PGM that supports 1.12.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -90,6 +90,7 @@ public final class PGMConfig implements Config {
   // ui.*
   private final boolean showSideBar;
   private final boolean showTabList;
+  private final boolean resizeTabList;
   private final boolean showTabListPing;
   private final boolean showProximity;
   private final boolean showFireworks;
@@ -183,6 +184,7 @@ public final class PGMConfig implements Config {
     this.showProximity = parseBoolean(config.getString("ui.proximity", "false"));
     this.showSideBar = parseBoolean(config.getString("ui.sidebar", "true"));
     this.showTabList = parseBoolean(config.getString("ui.tablist", "true"));
+    this.resizeTabList = parseBoolean(config.getString("ui.tablist-resize", "false"));
     this.showTabListPing = parseBoolean(config.getString("ui.ping", "true"));
     this.participantsSeeObservers =
         parseBoolean(config.getString("ui.participants-see-observers", "true"));
@@ -571,6 +573,11 @@ public final class PGMConfig implements Config {
   @Override
   public boolean showTabList() {
     return showTabList;
+  }
+
+  @Override
+  public boolean resizeTabList() {
+    return resizeTabList;
   }
 
   @Override
--- a/core/src/main/java/tc/oc/pgm/PGMPlugin.java
+++ b/core/src/main/java/tc/oc/pgm/PGMPlugin.java
@@ -73,6 +73,8 @@ import tc.oc.pgm.util.concurrent.BukkitExecutorService;
 import tc.oc.pgm.util.listener.ItemTransferListener;
 import tc.oc.pgm.util.listener.PlayerBlockListener;
 import tc.oc.pgm.util.listener.PlayerMoveListener;
+import tc.oc.pgm.util.nms.NMSHacks;
+import tc.oc.pgm.util.tablist.TablistResizer;
 import tc.oc.pgm.util.text.TextException;
 import tc.oc.pgm.util.text.TextTranslations;
 import tc.oc.pgm.util.xml.InvalidXMLException;
@@ -113,6 +115,9 @@ public class PGMPlugin extends JavaPlugin implements PGM, Listener {
     } catch (IllegalArgumentException e) {
       return; // Indicates the plugin failed to load, so exit early
     }
+
+    // Sanity test PGM is running on a supported version before doing any work
+    NMSHacks.allocateEntityId();
 
     Permissions.registerAll();
 
@@ -214,6 +219,14 @@ public class PGMPlugin extends JavaPlugin implements PGM, Listener {
 
     if (config.showTabList()) {
       matchTabManager = new MatchTabManager(this);
+
+      if (config.resizeTabList()) {
+        if (this.getServer().getPluginManager().isPluginEnabled("ProtocolLib")) {
+          TablistResizer.registerAdapter(this);
+        } else {
+          logger.warning("ProtocolLib is required when 'ui.resize' is enabled");
+        }
+      }
     }
 
     if (!config.getUptimeLimit().isNegative()) {
--- a/util/src/main/java/tc/oc/pgm/util/bukkit/BukkitUtils.java
+++ b/util/src/main/java/tc/oc/pgm/util/bukkit/BukkitUtils.java
@@ -23,7 +23,7 @@ public interface BukkitUtils {
     return PLUGIN.get();
   }
 
-  Boolean isSportPaper = Bukkit.getServer().getVersion().contains("SportPaper");
+  Boolean isSportPaper = Platform.SERVER_PLATFORM == Platform.SPORTPAPER_1_8;
 
   static boolean isSportPaper() {
     return isSportPaper;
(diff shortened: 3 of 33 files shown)


util/src/main/java/tc/oc/pgm/util/nms/NMSHacks.java
```diff
@@ -52,15 +54,27 @@ static NMSHacksPlatform chooseNMSHacks() {
...
 
     try {
-      if (BukkitUtils.isSportPaper()) {
-        choice = new NMSHacksSportPaper();
-        logger.info("Using NMSHacksSportPaper");
+      if (MinecraftReflectionUtils.MINECRAFT_VERSION.lessThanOrEqualTo(
+          NBTEditor.MinecraftVersion.v1_8)) {
+        if (BukkitUtils.isSportPaper()) {
+          choice = new NMSHacksSportPaper();
+          logger.info("Using NMSHacksSportPaper");
```
Electroid: Let's change these to be more user-friendly:

```
[PGM] Detected server: SportPaper (1.8)
[PGM] Detected server: Spigot (1.8)
[PGM] Detected server: Paper (1.9)
...
```

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -183,6 +184,7 @@ public final class PGMConfig implements Config {
     this.showProximity = parseBoolean(config.getString("ui.proximity", "false"));
     this.showSideBar = parseBoolean(config.getString("ui.sidebar", "true"));
     this.showTabList = parseBoolean(config.getString("ui.tablist", "true"));
+    this.resizeTabList = parseBoolean(config.getString("ui.resize", "true"));
```
Electroid: Let's set the default to be `false`, since this will log a warning if `ProtocolLib` is not loaded.

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -214,6 +219,14 @@ public void onEnable() {
 
     if (config.showTabList()) {
       matchTabManager = new MatchTabManager(this);
+
+      if (config.resizeTabList()) {
+        if (this.getServer().getPluginManager().isPluginEnabled("ProtocolLib")) {
+          TablistResizer.registerAdapter(this);
+        } else {
+          logger.warning("You cannot resize the tablist without ProtocolLib!");
```
Electroid:
```suggestion
          logger.warning("ProtocolLib is required when 'ui.resize' is enabled");
```

core/src/main/java/tc/oc/pgm/PGMConfig.java
cswhite2000: Will do

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -183,6 +184,7 @@ public final class PGMConfig implements Config {
     this.showProximity = parseBoolean(config.getString("ui.proximity", "false"));
     this.showSideBar = parseBoolean(config.getString("ui.sidebar", "true"));
     this.showTabList = parseBoolean(config.getString("ui.tablist", "true"));
+    this.resizeTabList = parseBoolean(config.getString("ui.resize", "false"));
```
```suggestion
    this.resizeTabList = parseBoolean(config.getString("ui.tablist-resize", "false"));
```
i think "ui.resize" is just too vague, resize what part of the ui?

util/src/main/java/tc/oc/pgm/util/bukkit/BukkitUtils.java
```diff
@@ -24,11 +24,25 @@ static Plugin getPlugin() {
   }
 
   Boolean isSportPaper = Bukkit.getServer().getVersion().contains("SportPaper");
+  Boolean isSpigot = Bukkit.getServer().getVersion().contains("Spigot");
+  Boolean isPaper = Bukkit.getServer().getVersion().contains("Paper");
```
Pablete1234: given that you may only be one of them at once, maybe just make an enum of supported server softwares, the enum can have the display name, the string to search for, and what nms hacks class to use (probs a method ref to the constructor, like `SportPaperNmsHacks::new`)

cswhite2000: That would certainly be cleaner than what I have now. I'll give it a try

cswhite2000:
Still need to implement the latest comments here from pablete. Upon his suggestion, I've tried using a different annotation-based reflection style.
Edit: Done

util/src/main/java/tc/oc/pgm/util/nms/reflect/ReflectionProxy.java
```diff
@@ -0,0 +1,353 @@
...
+  public static <T> T getProxy(Class<T> classType) {
+    return (T)
+        Proxy.newProxyInstance(
+            ReflectionProxy.class.getClassLoader(), new Class[] {classType}, INSTANCE);
+  }
+
+  @Override
+  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+    Class<?> declaringClass = method.getDeclaringClass();
+    Class<?> parentClass = getAnnotatedClass(declaringClass);
```
Pablete1234: should also go inside compute if absent, we want this to have minimal overhead after the first call

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
```diff
@@ -0,0 +1,82 @@
...
+    if (this.requiresProtocolLib
+        && !Bukkit.getServer().getPluginManager().isPluginEnabled("ProtocolLib")) {
+      Bukkit.getServer().getPluginManager().disablePlugin(BukkitUtils.getPlugin());
+      throw new UnsupportedOperationException(
+          "ProtocolLib is required for PGM to run on " + this.toString());
+    }
+    if (this == SERVER_PLATFORM) {
+      try {
+        logger.info("Detected server: " + this.toString());
+        return nmsHacksClass.getConstructor().newInstance();
```
Pablete1234: since java 8 i'm more of a fan of functions vs calling constructors with reflection.
Instead of asking for a `Class<? extends NMSHacksPlatform>` ask for a `Supplier<? extends NMSHacksPlatform>`, the supplier can be simply `NmsHacks1_8::new`, and you can call it with `supplier.get();`

cswhite2000: I'm fine with doing it that way. Will do

cswhite2000: NMSHacks1_8 caused this not work.

Pablete1234 (approved):

cswhite2000:
I've rebased these changes onto the latest commits from dev, I made no other changes in the most recent force push

Electroid (approved):
Welcome to back 1.9, PGM.

PR #1201, opened 2023-07-07
Restore Objectives as a translatable gamemode

zzufx:
When gamemode titles were hardcoded in https://github.com/PGMDev/PGM/pull/1091, it got suggested that the generic "Objectives" text that appears when some gamemodes are mixed (like DTC and DTM) could still be kept as translatable, as it's more of a special case and not an actual gamemode.

I didn't think it'd cause any confusion, so I tried doing that by simply undoing a couple of changes from that commit. I tested it on a local server and it seemed to work fine.

image


Pablete1234 (approved):
It's missing the sign-off, but i agree with this change. "Objectives" is not a gamemode, it's just a name for when there's mixed objectives and they're too long to fit on the sidebar

Electroid (approved):

KingOfSquares (approved):

PR #1202, opened 2023-07-07
Add support for Spigot 1.10-1.12

cswhite2000:
Depends on https://github.com/PGMDev/PGM/pull/1199
Leaving as a draft until that PR is merged.

Adds support for Spigot versions 1.10 - 1.12.2.
I was originally going to do this in three parts, but the changes to support 1.11 and 1.12 using the 1.10 NMSHacks were ~5 lines of code.

Same basic information as the 1.9 PR, supports everything Spigot 1.8 supports.

Next step:
- Refactor MaterialData usage to use a Wrapper


util/src/main/java/tc/oc/pgm/util/nms/v1_10_12/NMSHacks1_10_12.java
```diff
@@ -0,0 +1,155 @@
...
+    for (Object item : (List<Object>) nbtTagList.getListField(canDestroyList)) {
+      String blockString = nbtTagString.getString(item);
+      Object nmsBlock = reflBlock.getBlockByName(blockString);
+      int id = reflBlock.getId(nmsBlock);
+      Material material = Material.getMaterial(id);
+      materialSet.add(material);
+    }
+
+    return materialSet;
+  }
```
Pablete1234: this is an identical copy-paste to the one in 1.9 nms hacks

The point of reflection is that the same code can operate in multiple underlying types, this should be no different. This piece of code should not be replicated for each version, only the reflection part should use different methods

cswhite2000: Yeah, I was unsure how to structure the inheritance.

I was trying to keep reflection out of the protocol lib NMSHacks, because it was breaking things with the old static initializations, but given the new reflection annotations that should no longer be necessary. I'll consolidate the versioned NMSHacks since the separation should no longer be necessary.

core/src/main/java/tc/oc/pgm/wool/WoolMatchModule.java
```diff
@@ -205,19 +206,22 @@ public void placementCheck(final BlockTransformEvent event) {
...
-
-      if (playerHolder != null && result != null && result.getType() == Material.WOOL) {
-        for (MonumentWool wool : this.wools.values()) {
-          if (wool.getDefinition().isObjectiveWool(result)) {
-            if (!wool.getDefinition().isCraftable()) {
-              playerHolder.sendWarning(
-                  translatable("wool.craftingDisabled", wool.getComponentName()));
-              event.getInventory().setResult(null);
+    Recipe recipe = event.getRecipe();
+    if (recipe != null) {
```
Pablete1234:
```suggestion
    Recipe recipe = event.getRecipe();
    if (recipe == null) return;
```
Avoids the whole method being pushed one indentation level in

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
```diff
@@ -26,7 +27,13 @@ public enum Platform {
...
       false),
   SPIGOT_1_9("Spigot", "1.9", NMSHacks1_9.class, true),
-  PAPER_1_9("Paper", "1.9", NMSHacks1_9.class, true);
+  PAPER_1_9("Paper", "1.9", NMSHacks1_9.class, true),
+  SPIGOT_1_10("Spigot", "1.10", NMSHacks1_10_12.class, true),
+  PAPER_1_10("Paper", "1.10", NMSHacks1_10_12.class, true),
+  SPIGOT_1_11("Spigot", "1.11", NMSHacks1_10_12.class, true),
+  PAPER_1_11("Paper", "1.11", NMSHacks1_10_12.class, true),
+  SPIGOT_1_12("Spigot", "1.12", NMSHacks1_10_12.class, true),
+  PAPER_1_12("Paper", "1.12", NMSHacks1_10_12.class, true);
```
Pablete1234: maybe some representation that allows a range?
```
  SPIGOT_1_10_TO_1_12("Spigot", "1.10-1.12", NMSHacks1_10_12.class, true),
  PAPER_1_10_TO_1_12("Paper", "1.10-1.12", NMSHacks1_10_12.class, true);
```

Also regarding the classes above, being referenced via reflection, would a Supplier fix it? ie: `() -> NMSHacks1_8.class`  and used as a `Supplier<Class<? extends NMSHacksPlatform>>` ?

util/src/main/java/tc/oc/pgm/util/nms/v1_9/NMSHacks1_9.java
```diff
@@ -151,4 +194,593 @@ public void postToMainThread(Plugin plugin, boolean priority, Runnable task) {
...
+    // TODO: use radius defined in spigot.yml / sportpaper.yml
+    List<Player> players = new ArrayList<>();
+    for (Entity nearbyEntity :
+        entity.getWorld().getNearbyEntities(entity.getLocation(), 64, 64, 64)) {
+      if (nearbyEntity instanceof Player && nearbyEntity.getEntityId() != entity.getEntityId()) {
+        players.add((Player) nearbyEntity);
+      }
+    }
+    return players;
+  }
```
Pablete1234: i have the feeling this can go very, very wrong.

Not getting the viewing players from a real place (ie: entity tracker) and instead *assuming* who can see the entity can and will crash older clients. Every time a player dies and they're put to bed, if you receive that packet in 1.7 while not seeing the player (eg: too far away, or the player was vanished or in obs), you'll crash. This is not reliable, a player may have just crossed into view but not have been sent to you yet, or may have moved away but not been removed yet.

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
cswhite2000: Thinking for a range to have it be done via overloading the constructor to have an extra argument ie:
```java
SPIGOT_1_10_TO_1_12("Spigot", "1.10", "1.12", NMSHacks1_10_12.class, true),
```
I'll give using a supplier a try

core/src/main/java/tc/oc/pgm/wool/WoolMatchModule.java
cswhite2000: Will do

util/src/main/java/tc/oc/pgm/util/nms/v1_9/NMSHacks1_9.java
cswhite2000: I should be able to use the reflection annotations for this. I'll rewrite this method to use that.

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
cswhite2000: Supplier fixed it

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
```diff
@@ -2,62 +2,103 @@
...
+    this.majorVersionFirst = majorVersion;
+    this.majorVersionLast = majorVersion;
+    this.nmsHacksClassSupplier = nmsHacksClassSupplier;
+    this.requiresProtocolLib = requiresProtocolLib;
+  }
+
+  Platform(
+      String variant,
+      String majorVersion,
+      Class<? extends NMSHacksPlatform> nmsHacksClass,
```
Pablete1234: if you keep on adding constructors for each possible different combination, this is going to explode quickly, consider just having supplier-based constructors, make all of them pass in the supplier, and maybe even force major & minor version, and just pass in the same version

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
```diff
@@ -2,62 +2,103 @@
...
       }
     }
     return UNKNOWN;
   }
 
   private final String variant;
-  private final String majorVersion;
-  private final Class<? extends NMSHacksPlatform> nmsHacksClass;
+  private final String majorVersionFirst;
+  private final String majorVersionLast;
```
Pablete1234: you can probably turn these into `Version fromVersion, toVersion`, and use PGM's utilities around versioning (constructor can still take in a string, and you can make it like 1.8.0 etc

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
cswhite2000: Yeah I can simplify this

util/src/main/java/tc/oc/pgm/util/bukkit/Platform.java
cswhite2000: Presumably, I'd need to, for example with 1.9, have `"1.9.0", "1.9.4"`. I was hoping to continue ignoring the minor version/patch since the breaking changes so far all seem to be across major versions.

util/src/main/java/tc/oc/pgm/util/nms/reflect/Refl.java
```diff
@@ -157,9 +193,13 @@ interface Block {
     Object getBlockData(Object self);
 
     @Reflect.Method("q")
+    @Reflect.Method(value = "q", parameters = IBlockData.class)
```
Pablete1234: are these two redundant?

Pablete1234 (dismissed):

util/src/main/java/tc/oc/pgm/util/nms/reflect/Refl.java
cswhite2000: So it turns out that the first "q" was incorrect. I didn't catch this because this method was never called due to a bug in blockdrops meaning that `wrong-tool` would never apply. I've fixed that and tested it in sportpaper, 1.9, and 1.12

Pablete1234 (approved):

Electroid (approved):
Exciting!

PR #1204, opened 2023-07-09
Add support for world folder in variants

Pablete1234:
Allows for a world attribute in variants, that specifies the world to run in:

```xml
<map>
  <name>Normal Map</name>
  <variant id="xmas" override="true" world="christmas">Christmas Map</variant>

  ...
</map>
```

Folder structure would look like:
- normal_map
  - map.xml
  - level.dat
  - region
    - 0.0.mca
  - chistmas
    - level.dat
    - region
      - 0.0.mca

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/map/MapSource.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/MapSource.java
@@ -39,10 +39,11 @@ public interface MapSource {
   /**
    * Download the {@link org.bukkit.World} files to a local directory.
    *
+   * @param folder subfolder to download, null for parent
    * @param dir An existent, but empty directory.
    * @throws MapMissingException If an error occurs while creating the files.
    */
-  void downloadTo(File dir) throws MapMissingException;
+  void downloadTo(String folder, File dir) throws MapMissingException;
 
   /**
    * Get an {@link InputStream} of the map's xml document.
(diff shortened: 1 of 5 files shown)


calcastor:
Tested [here](https://github.com/OvercastCommunity/PublicMaps/pull/31) and works as expected!

core/src/main/java/tc/oc/pgm/api/map/MapSource.java
```diff
@@ -39,10 +39,11 @@ public interface MapSource {
   /**
    * Download the {@link org.bukkit.World} files to a local directory.
    *
+   * @param folder subfolder to download, null for parent
    * @param dir An existent, but empty directory.
    * @throws MapMissingException If an error occurs while creating the files.
    */
-  void downloadTo(File dir) throws MapMissingException;
+  void downloadTo(String folder, File dir) throws MapMissingException;
```
Electroid, line 46: I don't think we should change the signature here, can the callee do the resolving?

Pablete1234: The MapSource is created *before* the map has been parsed, so the world folder to use isn't known at the time, and i'd rather keep MapSource immutable

Electroid (approved):

PR #1205, opened 2023-07-10
Allow filters on class selection

cswhite2000:
My intent with this is to allow team-specific classes, similar to kits with portals, but for classes. Adding filters for class selection seemed like the best way to do it, while also allowing other usage.
You should also be able to use this for https://github.com/PGMDev/PGM/issues/800

Usage would be something like:
```xml
<class name="potato" description="lettuce" filter="only-blue" deny-message="Only the blue team can pick this class!">
...
</class>
```

This could also do some other interesting things if we added a class change action as well.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/classes/ClassMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/classes/ClassMatchModule.java
@@ -1,15 +1,21 @@
 package tc.oc.pgm.classes;
 
+import static net.kyori.adventure.text.Component.text;
 import static tc.oc.pgm.util.Assert.assertNotNull;
 import static tc.oc.pgm.util.Assert.assertTrue;
 import static tc.oc.pgm.util.text.TextException.exception;
 
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import net.kyori.adventure.text.format.NamedTextColor;
+import net.kyori.adventure.text.format.TextDecoration;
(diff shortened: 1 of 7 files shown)


cswhite2000:
I realized that for this to work each team would need to be able to have a default class. This now allows multiple default classes and randomly chooses one that the player can use.

Pablete1234:
I feel like this is going to be more troublesome than we think. Players can select classes before the game even starts, if your filters try to enforce anything other than a team i'm not sure that they'll be able to handle the "consequences" of the kit no longer being applicable later, etc.
Are we going to just change their kit randomly each time they respawn just because their previous isn't available atm? and we'll do so based on just "the first default that passes the filter"? what if all defaults don't pass, do they stay classless?

I think we need to take a step back and find a better prespective from where to tackle the problem, that maintains the flexibility of being able to implement arbitrary filters.

cswhite2000:
Well, so the way I currently have it is, that it will pick a random other available default class if the current class no longer is available. And if no default class is available it will pick one of the default classes regardless and log a warning.

Possible ways to improve this:
- Keep the current class if no defaults pass
- Search through all classes if no defaults pass
- Have an additional parameter for whether to check the filter on respawn
- Have a separate `respawn-filter`, which if missing defaults to the main filter
- Have an enum with several strategies for picking fallback classes

Thoughts?

Pablete1234:
I'm moving this to draft for the time being, we've already talked about how this could be improved (by moving the class selection screen to after picking a team) but it doesn't seem to be in focus given cs is working on other stuff.

It'll be nice to see if this gets reworked such that it'll yield good ux but for the time being this would just be too confusing for players.

Pablete1234:
Currently closing as @cswhite2000 has expressed no motivation to continue with this at the moment, can be re-taken at a later time

PR #1208, opened 2023-07-23
Add per team player stats with extra trackers

Pugzy:
Rewrite the stats logic so that statistics can be per team. This allows the stats at the end of the match to show contributions towards both teams.

- Introduces a table to store the team, player, stats values and create new stat entries with a reference to its global stats parent.
- Resets a player's active kills streak when switching teams.
- Store snapshot of player's name in a component inside a teams player stats.
- Rewrite team stats GUI to make use of per-team stats (no longer need to create online/observers/offline players separately.
- Attempts to fix the current issues with the team stats GUI where nicknames and skins are revealed.
- Adds extra stats related to objective completion (rather than just flags).

Still a WIP and requires a bit more testing but general feedback is more than welcome.

Additional changes:

- Tracking activity time for players on a team within the `StatsMatchModule`.
- Ability to check which team was a players primary team (based on how long they played on each team or obs).
- Fixed a PGM issue where any goal change event would trigger not only the invalidation of winners but a check to see if anyone has won. The order of these events meant that a match could be ended before the event of lets say a wool capture would be fired so at the point of match end that wool wouldn't count in a players stats.
- The above should also fix an issue where the announcement of the objective being completed would display after the match over text.
- A slight tweak to the block vector parser so it allows `.5` coordinates that are then translated (this makes structures more backwards compatible as this is the same behaviour as old new PGM).


core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -97,8 +107,12 @@ public StatsMatchModule(Match match) {
...
   }
 
-  public Map<UUID, PlayerStats> getStats() {
-    return Collections.unmodifiableMap(allPlayerStats);
+  public Map<UUID, PlayerStats> getGlobalStats() {
+    return allPlayerStats;
+  }
+
+  public Table<Team, UUID, PlayerStats> getStats() {
+    return Tables.unmodifiableTable(stats);
```
Pablete1234: keep getStats signature intact (returning global stats) and have getParticipationStats as the new one returning a table, for api backwards-compatibility

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -138,13 +152,57 @@ public void onDestroyableBreak(DestroyableHealthChangeEvent event) {
...
+
+  @EventHandler(priority = EventPriority.MONITOR)
+  public void onCoreLeak(CoreLeakEvent event) {
+    event
+        .getCore()
+        .getContributions()
+        .forEach(
+            leaker -> {
+              if (leaker.getPlayerState() != null) {
+                // TODO: wrong method
```
Pablete1234: ?

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -138,13 +152,57 @@ public void onDestroyableBreak(DestroyableHealthChangeEvent event) {
...
+    }
   }
 
   @EventHandler(priority = EventPriority.MONITOR)
-  public void onFlagHold(FlagPickupEvent event) {
-    getPlayerStat(event.getCarrier()).onFlagPickup();
+  public void onWoolCapture(PlayerWoolPlaceEvent event) {
+    if (event.getPlayer() != null && event.getPlayer().getPlayer().isPresent()) {
+      getPlayerStat(event.getPlayer()).onWoolCapture();
+    }
```
Pablete1234: is this about getting the player off of a player state? If so, that's not needed, keep it as:
```suggestion
    if (event.getPlayer() != null) {
      getPlayerStat(event.getPlayer()).onWoolCapture();
    }
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/stats/PlayerStats.java
```diff
@@ -39,17 +55,33 @@ public class PlayerStats {
   // See StatsMatchModule#sendLongHotbarMessage
   private Future<?> hotbarTaskCache;
 
+  public PlayerStats() {
+    this.parent = null;
+    this.component = null;
+    this.timePlayed = Duration.ZERO;
+    this.inTime = null;
```
Pablete1234:
```suggestion
    this(null, null);
```

core/src/main/java/tc/oc/pgm/stats/PlayerStats.java
```diff
@@ -187,4 +274,32 @@ public Future<?> getHotbarTask() {
...
+    if (parent != null) parent.endParticipation();
+  }
+
+  public Duration getTimePlayed() {
+    // If not ended yet add the current session time up
+    return inTime == null ? timePlayed : timePlayed.plus(getActiveSessionDuration());
+  }
+
+  public Duration getActiveSessionDuration() {
+    return (inTime == null) ? Duration.ZERO : Duration.between(inTime, Instant.now());
```
Pablete1234, line 300:
```suggestion
    return inTime == null ? Duration.ZERO : Duration.between(inTime, Instant.now());
```

core/src/main/java/tc/oc/pgm/stats/menu/items/TeamStatsMenuItem.java
```diff
@@ -40,43 +42,28 @@ public class TeamStatsMenuItem implements MenuItem {
...
-      Competitor team,
-      Collection<MatchPlayer> relevantObservers,
-      Collection<UUID> relevantOfflinePlayers) {
-    StatsMatchModule smm = match.needModule(StatsMatchModule.class);
+  public TeamStatsMenuItem(Match match, Competitor team, Map<UUID, PlayerStats> playerStats) {
 
     this.team = team;
     this.members = Lists.newArrayList();
-    this.stats = new TeamStats(team, smm);
+    this.stats = new TeamStats(team, new ArrayList<>(playerStats.values()));
```
Pablete1234: i'd avoid the relocation, just pass accept a Collection<PlayerStats> in the constructor instead
```suggestion
    this.stats = new TeamStats(team, playerStats.values());
```

Pablete1234 (commented):
just a few tweaks

Pablete1234 (approved):

core/src/main/java/tc/oc/pgm/goals/GoalMatchModule.java
```diff
@@ -147,7 +147,9 @@ protected void updateProgress(Goal goal) {
     for (Competitor competitor : competitorsByGoal.get(goal)) {
       progressByCompetitor.put(competitor, new GoalProgress(competitor));
     }
-    match.calculateVictory();
+
+    // Forces team rankings to be invalidated
+    match.getWinners();
```
Electroid, line 152: This is subtle behavior which is not great, it’s better if we could do something explicit. But I’ll still merge for now.

Electroid (approved):

PR #1210, opened 2023-07-26
Tweak flags & gamemodes, add gamemode to votebook

Pablete1234:
Adds the gamemode to the votebook hover, and refactors a few things around pgm.
 - MapTags representing a gamemode now directly store a reference to the gamemode itself
 - MapInfo#getGamemodes now will always be populated, if not by the map author, by tags with a gamemode.
 - Implemented tab-completing for tags in `/maps -t `[tab] so it would now suggest wool, monument, etc.
   -  Note this is affected by https://github.com/Incendo/cloud/pull/460 so tab-completion stops after you type part of the tag.

image
image

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapCommand.java
@@ -15,6 +15,8 @@ import cloud.commandframework.annotations.CommandMethod;
 import cloud.commandframework.annotations.Flag;
 import cloud.commandframework.annotations.specifier.Greedy;
 import cloud.commandframework.annotations.specifier.Range;
+import cloud.commandframework.annotations.suggestions.Suggestions;
+import cloud.commandframework.context.CommandContext;
 import com.google.common.collect.ImmutableList;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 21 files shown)


core/src/main/java/tc/oc/pgm/command/MapCommand.java
```diff
@@ -115,6 +119,20 @@ public Component format(MapInfo map, int index) {
...
 
+  @Suggestions("maptags")
+  public List<String> suggestMapTags(CommandContext<CommandSender> sender, String input) {
+    int commaIdx = input.lastIndexOf(',');
+
+    final String prefix = input.substring(0, commaIdx == -1 ? 0 : commaIdx + 1);
+    final String toComplete =
+        input.substring(commaIdx + 1).toLowerCase(Locale.ROOT).replace("!", "");
+
+    return Stream.of("wool", "monument")
```
cswhite2000: This will only suggest wool and monument

Pablete1234: leftover from testing, nice catch

cswhite2000 (approved):

Electroid (approved):

PR #1212, opened 2023-08-07
Update dependencies and use non-deprecated methods

calcastor:
I swapped over a bunch of `Title.Times.of` usage in https://github.com/PGMDev/PGM/pull/1008 but missed these it seems.

Aside from that, this PR updates dependencies, and switches to using ViaVersion's distributed API rather than ViaVersion directly, which lets us remove special casing for a past build error. If there's a reason we weren't using the API in the first place I can drop this change.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/state/Carried.java
+++ b/core/src/main/java/tc/oc/pgm/flag/state/Carried.java
@@ -199,7 +199,7 @@ public class Carried extends Spawned implements Missing {
     if (!message.equals(this.lastMessage)) {
       this.lastMessage = message;
       this.carrier.showTitle(
-          title(empty(), message, Title.Times.of(Duration.ZERO, fromTicks(5), fromTicks(35))));
+          title(empty(), message, Title.Times.times(Duration.ZERO, fromTicks(5), fromTicks(35))));
     }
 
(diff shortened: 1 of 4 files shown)


cswhite2000:
You could also update FastBoard from `1.2.1` -> `2.0.0`. They recently added support for adventure, although that may be constrained to more recent paper versions.

calcastor:
> You could also update FastBoard from `1.2.1` -> `2.0.0`. They recently added support for adventure, although that may be constrained to more recent paper versions.

It does seem to be constrained to modern Paper versions. Regardless, also updated it.

cswhite2000:
https://github.com/MrMicky-FR/FastBoard/blob/master/src/main/java/fr/mrmicky/fastboard/adventure/FastBoard.java#L91
It may actually handle it reasonably well regardless, but we can revisit that in a future pr.

Pablete1234:
> https://github.com/MrMicky-FR/FastBoard/blob/master/src/main/java/fr/mrmicky/fastboard/adventure/FastBoard.java#L91 It may actually handle it reasonably well regardless, but we can revisit that in a future pr.

won't work, we need per-player rendering, this renders without context

Pablete1234 (approved):

Electroid (approved):

PR #1214, opened 2023-08-11
Add support for variables acting on Blitz Lives

cswhite2000:
New Description:
---

So decided to go with a different approach for this. This now adds support for different types of variables. Specifically, this adds a blitz lives variable type that directly sets or gets blitz lives for a player.

Example Usage:
```xml
<variable id="playerlives" scope="player" type="lives"/>
...
<set var="playerlives" value="playerlives+1"/>
```

With this, the type can be either `dummy` or `lives`. `dummy` being the original variable type, and `lives` getting or writing blitz lives. `type` is optional and defaults to `dummy`.

Original description
---
 :warning: Syntax beyond this point is no longer valid. :warning: 

Adds a `modify-lives` action. At @zzufx's request

You can add, subtract, or set lives.
Usage would be as follows:

```xml
<modify-lives operation="add" amount="2" show-title="true"/>
```
`show-title` is optional and can be left out
```xml
<modify-lives operation="set" amount="1"/>
or
<modify-lives operation="subtract" amount="1"/>
```

The action will act on all players in a given scope, so for example you could have a team scoped set lives to 1 that would set the lives remaining for all players to 1 after an objective is completed.

Edit:
amount now supports variables and expressions

### DIFF
--- a/core/src/main/java/tc/oc/pgm/variables/BlitzVariable.java
+++ b/core/src/main/java/tc/oc/pgm/variables/BlitzVariable.java
@@ -1,0 +1,33 @@
+package tc.oc.pgm.variables;
+
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.blitz.BlitzMatchModule;
+import tc.oc.pgm.filters.Filterable;
+
+public class BlitzVariable implements Variable<MatchPlayer> {
+
+  private final VariableDefinition<MatchPlayer> definition;
+
+  public BlitzVariable(VariableDefinition<? extends Filterable<?>> definition) {
+    this.definition = (VariableDefinition<MatchPlayer>) definition;
+  }
+
+  @Override
+  public VariableDefinition<MatchPlayer> getDefinition() {
+    return definition;
+  }
+
+  @Override
+  public double getValue(Filterable<?> context) {
+    MatchPlayer matchPlayer = context.getFilterableAncestor(MatchPlayer.class);
+    BlitzMatchModule blitzMatchModule = matchPlayer.moduleRequire(BlitzMatchModule.class);
+    return blitzMatchModule.getNumOfLives(matchPlayer.getId());
+  }
+
+  @Override
+  public void setValue(Filterable<?> context, double value) {
+    MatchPlayer matchPlayer = context.getFilterableAncestor(MatchPlayer.class);
+    BlitzMatchModule blitzMatchModule = matchPlayer.moduleRequire(BlitzMatchModule.class);
+    blitzMatchModule.setLives(matchPlayer, Math.max((int) value, 0));
+  }
+}
--- a/core/src/main/java/tc/oc/pgm/variables/Variable.java
+++ b/core/src/main/java/tc/oc/pgm/variables/Variable.java
@@ -1,52 +1,16 @@
 package tc.oc.pgm.variables;
 
-import java.util.HashMap;
-import java.util.Map;
 import tc.oc.pgm.api.feature.Feature;
-import tc.oc.pgm.filters.FilterMatchModule;
 import tc.oc.pgm.filters.Filterable;
 
-public class Variable<T extends Filterable<?>> implements Feature<VariableDefinition<T>> {
-
-  private final VariableDefinition<T> definition;
-  private final Map<T, Double> values;
-
-  public Variable(VariableDefinition<T> definition) {
-    this.definition = definition;
-    this.values = new HashMap<>();
-  }
+public interface Variable<T extends Filterable<?>> extends Feature<VariableDefinition<T>> {
 
   @Override
-  public String getId() {
-    return definition.getId();
+  default String getId() {
+    return getDefinition().getId();
   }
 
-  @Override
-  public VariableDefinition<T> getDefinition() {
-    return definition;
-  }
+  double getValue(Filterable<?> context);
 
-  public double getValue(Filterable<?> context) {
-    return values.computeIfAbsent(getAncestor(context), k -> definition.getDefault());
-  }
-
-  public void setValue(Filterable<?> context, double value) {
-    T ctx = getAncestor(context);
-    values.put(ctx, value);
-    // For performance reasons, let's avoid launching an event for every variable change
-    context.getMatch().needModule(FilterMatchModule.class).invalidate(ctx);
-  }
-
-  private T getAncestor(Filterable<?> context) {
-    T filterable = context.getFilterableAncestor(definition.getScope());
-    if (filterable != null) return filterable;
-
-    throw new IllegalStateException(
-        "Wrong variable scope for '"
-            + getId()
-            + "', expected "
-            + definition.getScope().getSimpleName()
-            + " which cannot be found in "
-            + context.getClass().getSimpleName());
-  }
+  void setValue(Filterable<?> context, double value);
 }
--- a/core/src/main/java/tc/oc/pgm/variables/VariableType.java
+++ b/core/src/main/java/tc/oc/pgm/variables/VariableType.java
@@ -1,0 +1,33 @@
+package tc.oc.pgm.variables;
+
+import java.util.function.Function;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.filters.Filterable;
+
+public enum VariableType {
+  DUMMY(DummyVariable::new, Filterable.class),
+  LIVES(BlitzVariable::new, MatchPlayer.class);
+
+  private final Function<VariableDefinition<?>, Variable<?>> supplierFunction;
+  private final Class<?>[] supportedScopes;
+
+  VariableType(
+      Function<VariableDefinition<? extends Filterable>, Variable<?>> supplierFunction,
+      Class<?>... supportedScopes) {
+    this.supplierFunction = supplierFunction;
+    this.supportedScopes = supportedScopes;
+  }
+
+  public boolean supports(Class<?> cls) {
+    for (Class<?> supportedScope : supportedScopes) {
+      if (supportedScope.isAssignableFrom(cls)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  public Variable<?> buildInstance(VariableDefinition<?> definition) {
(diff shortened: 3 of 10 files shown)


Pablete1234:
Make number an expression supporting variables, adds alot more play

core/src/main/java/tc/oc/pgm/action/Operation.java
```diff
@@ -0,0 +1,30 @@
+package tc.oc.pgm.action;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public enum Operation {
+  ADD("add", "addition", "plus"),
```
Pablete1234: Why do we have 3 diff names for the same thing? Id honestly avoid giving different syntax for the same thing, decide on one

core/src/main/java/tc/oc/pgm/action/actions/BlitzLivesAction.java
```diff
@@ -0,0 +1,87 @@
...
+      blitzMatchModule.showLivesTitle(player);
+    }
+  }
+
+  @Override
+  public void trigger(T t) {
+    modifyLivesConsumer.accept(t);
+  }
+
+  private Consumer<T> chooseScopeConsumer(Class<T> cls) {
```
Pablete1234: Yeah... no.
Make this a player-scoped action. If the mapdev wants it to affect all players on a team they can use switch-scope action.
If in the future we add team-scoped lives in blitz (ie: whole team shares a pool of lives, maybe mixed in with player lives too) then this would conflict with that, as now you have to break maps to decide if lives would go to the team (pool) or the player.

Pablete1234 (changes requested):
Besides what i pointed out, i have doubts about the whole add/remove/set thing.
In the original draft for variables (#881) i had a similar system, but that eventually went out the window in favor of formulas with variables.
Want to add 5? Use `<set var="thing" value="thing + 5"/>`
Want to halve it? Use `<set var"thing" value="thing / 2">`
Want to multiply it by the cos of another variable? Use `<set var"thing" value="thing * cos(other)">`

Its much more flexible than hard-coded allowed operations.

The biggest roadblock is lack of a way to reference lives left within that. I mentioned somewhere that it'd be nice to have more variable types, so you could define a: `<variable id="lives" type="blitz-lives">` or similar. Some other candidates would be player curr and max health, or team score. What that allows is to directly operate on blitz lives as if it was just another variable. You can arbitratily `<set var="lives" value="lives + 5">` which automatically gets you this behavior, plus you can now operate on so much more (eg: set player max heath based on amount of lives, so last live you have less max hp). Also there's a consideration for potentially read-only variable types, maybe health should be read-only and can be set via a kit if you wish, etc.

I'm fully aware that much of what i mentioned here is well out of scope for this pr, but i'm just trying to give my long-term vision on the matter, in hopes that we can get some middle ground that both aligns this towards that but doesn't necessarily mean a full rewrite

core/src/main/java/tc/oc/pgm/action/Operation.java
cswhite2000: Yeah I can change it to just a single name

core/src/main/java/tc/oc/pgm/action/actions/BlitzLivesAction.java
cswhite2000: Will do

cswhite2000:
I also would prefer to use variables directly for things like this, and I'm willing to go with that approach if we can come up with a suitable design.

My main sticking point for why I didn't originally go with that approach is that it seemed like I'd have to create default variables for those things and update them every time the underlying value was changed, while also adding an update for every time the variable changed to change the underlying value.

core/src/main/java/tc/oc/pgm/variables/BlitzVariable.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public double getValue(Filterable<?> context) {
+    MatchPlayer matchPlayer = (MatchPlayer) context;
+    BlitzMatchModule blitzMatchModule = matchPlayer.getMatch().needModule(BlitzMatchModule.class);
+    return blitzMatchModule.getNumOfLives(matchPlayer.getId());
+  }
+
+  @Override
+  public void setValue(Filterable<?> context, double value) {
+    MatchPlayer matchPlayer = (MatchPlayer) context;
```
Pablete1234:
```suggestion
    MatchPlayer matchPlayer = context.getFilterableAncestor(MatchPlayer.class);
```
although we don't have a children that could have match player as parent and this will always return just itself, it's a safer method (won't get an illegal cast).
(note: same for getValue)

core/src/main/java/tc/oc/pgm/variables/BlitzVariable.java
```diff
@@ -0,0 +1,38 @@
...
+    BlitzMatchModule blitzMatchModule = matchPlayer.getMatch().needModule(BlitzMatchModule.class);
```
Pablete1234:
```suggestion
    BlitzMatchModule blitzMatchModule = matchPlayer.getMatch().requireModule(BlitzMatchModule.class);
```

core/src/main/java/tc/oc/pgm/variables/Variable.java
```diff
@@ -1,52 +1,18 @@
...
-
-    throw new IllegalStateException(
-        "Wrong variable scope for '"
-            + getId()
-            + "', expected "
-            + definition.getScope().getSimpleName()
-            + " which cannot be found in "
-            + context.getClass().getSimpleName());
-  }
+  VariableType getType();
```
Pablete1234: is this used anywhere? i have the feeling it's just unneeded extra code in each variable implementation

core/src/main/java/tc/oc/pgm/variables/VariableType.java
```diff
@@ -0,0 +1,47 @@
...
+
+  static final Map<String, VariableType> stringTypeMap = new ConcurrentHashMap<>();
+
+  static {
+    for (VariableType value : VariableType.values()) {
+      stringTypeMap.put(value.name().toLowerCase(Locale.ROOT), value);
+    }
+  }
+
+  static VariableType parse(Node variableTypeNode, VariableType defaultType)
```
Pablete1234: pretty sure there's XMLUtil to parse enum, you don't need all of the setup, or the stringTypeMap stuff

core/src/main/java/tc/oc/pgm/variables/VariableType.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.filters.Filterable;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.Node;
+
+public enum VariableType {
+  DUMMY(Filterable.class),
+  LIVES(MatchPlayer.class);
```
Pablete1234:
```suggestion
  DUMMY(DummyVariable::new, Filterable.class),
  LIVES(BlitzVariable::new, MatchPlayer.class);
```

core/src/main/java/tc/oc/pgm/variables/VariablesModule.java
```diff
@@ -74,7 +75,16 @@ public static <T extends Filterable<?>> VariableCache<T> of(
   @Override
   public VariablesMatchModule createMatchModule(Match match) throws ModuleLoadException {
     for (VariableDefinition<?> varDef : this.variables) {
-      match.getFeatureContext().add(new Variable<>(varDef));
+      switch (varDef.getVariableType()) {
```
Pablete1234:
```suggestion
      match.getFeatureContext().add(varDef.buildInstance());
```
buildInstance method can call the builder in type with itself as parameter

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/variables/BlitzVariable.java
cswhite2000: There doesn't seem to be a method with the signature requireModule

core/src/main/java/tc/oc/pgm/variables/Variable.java
cswhite2000: Ah no its not, I'll remove it

core/src/main/java/tc/oc/pgm/variables/VariableType.java
cswhite2000: Ah you're correct, I'll use that

core/src/main/java/tc/oc/pgm/variables/BlitzVariable.java
Pablete1234: apparently it's `moduleRequire`, which you may even be able to run on the matchPlayer itself (as it's filterable, which extends match query)

`matchPlayer.moduleRequire(BlitzMatchModule.class)`

Pablete1234 (changes requested):
The one last thing i see missing is preventing variable filters for non-dummy variable types being created. If you made a blitz variable and filtered on it, it wouldn't work dynamically

core/src/main/java/tc/oc/pgm/variables/VariableType.java
```diff
@@ -0,0 +1,44 @@
...
+    return supplierFunction.apply(definition);
+  }
+
+  static final Map<String, VariableType> stringTypeMap = new ConcurrentHashMap<>();
+
+  static {
+    for (VariableType value : VariableType.values()) {
+      stringTypeMap.put(value.name().toLowerCase(Locale.ROOT), value);
+    }
+  }
```
Pablete1234:
```suggestion
```

Pablete1234 (changes requested):

Pablete1234 (approved):
Great work

Pablete1234:
Please note, given recent changes, this should be merged alongside #1220. Syntax has changed.

Electroid (approved):

PR #1218, opened 2023-08-17
Add a play-sound action

cswhite2000:
Adds a `sound` action. Essentially implements my proposal from https://github.com/PGMDev/PGM/issues/1216
Usage would be as follows:
```xml
<sound preset="alert"/>
...
<sound key="random.click" volume="0.5" pitch="0.7"/>
...
<sound/>
```
I've made an enum of some of the sounds pgm uses, besides the ones like match end or match start.
The `key` parameter is the same as defined here https://docs.advntr.dev/sound.html
`preset` refers to all of the presets in the `SoundType` enum.
All parameters are optional

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -1,10 +1,13 @@
 package tc.oc.pgm.action;
 
+import static net.kyori.adventure.key.Key.key;
+import static net.kyori.adventure.sound.Sound.sound;
 import static net.kyori.adventure.text.Component.empty;
 
 import com.google.common.collect.ImmutableList;
 import java.lang.reflect.Method;
 import java.util.Map;
+import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.title.Title;
 import org.bukkit.inventory.ItemStack;
@@ -19,6 +22,7 @@ import tc.oc.pgm.action.actions.MessageAction;
 import tc.oc.pgm.action.actions.ReplaceItemAction;
 import tc.oc.pgm.action.actions.ScopeSwitchAction;
 import tc.oc.pgm.action.actions.SetVariableAction;
+import tc.oc.pgm.action.actions.SoundAction;
 import tc.oc.pgm.api.feature.FeatureValidation;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.Filterables;
@@ -237,6 +241,26 @@ public class ActionParser {
           "Expected at least one of text, title, subtitle or actionbar", el);
 
     return new MessageAction(text, actionbar, title);
+  }
+
+  @MethodParser("sound")
+  public SoundAction parseSoundAction(Element el, Class<?> scope) throws InvalidXMLException {
+    SoundType soundType =
+        XMLUtils.parseEnum(
+            Node.fromAttr(el, "preset"), SoundType.class, "preset", SoundType.CUSTOM);
+    Node resourceNode = Node.fromAttr(el, "key");
+    String resource = resourceNode == null ? soundType.getResource() : resourceNode.getValue();
+
+    float volume =
+        Math.min(
+            1f,
+            XMLUtils.parseNumber(Node.fromAttr(el, "volume"), Float.class, soundType.getVolume()));
+    float pitch =
+        XMLUtils.parseNumber(Node.fromAttr(el, "pitch"), Float.class, soundType.getPitch());
+
+    Sound sound = sound(key(resource, ':'), Sound.Source.MASTER, volume, pitch);
+
+    return new SoundAction(sound);
   }
 
   @MethodParser("set")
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -239,6 +243,23 @@ public MessageAction parseChatMessage(Element el, Class<?> scope) throws Invalid
...
 
+  @MethodParser("play-sound")
+  public SoundAction parseSoundAction(Element el, Class<?> scope) throws InvalidXMLException {
+    SoundType soundType =
+        XMLUtils.parseEnum(Node.fromAttr(el, "sound"), SoundType.class, "sound", SoundType.CUSTOM);
+    Node resourceNode = Node.fromAttr(el, "resources");
+    String resource = resourceNode == null ? soundType.getResource() : resourceNode.getValue();
+
+    float volume =
+        XMLUtils.parseNumber(Node.fromAttr(el, "volume"), Float.class, soundType.getVolume());
```
Pablete1234: i'd restrict volume to at most x1

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -239,6 +243,23 @@ public MessageAction parseChatMessage(Element el, Class<?> scope) throws Invalid
     return new MessageAction(text, actionbar, title);
   }
 
+  @MethodParser("play-sound")
+  public SoundAction parseSoundAction(Element el, Class<?> scope) throws InvalidXMLException {
+    SoundType soundType =
+        XMLUtils.parseEnum(Node.fromAttr(el, "sound"), SoundType.class, "sound", SoundType.CUSTOM);
+    Node resourceNode = Node.fromAttr(el, "resources");
+    String resource = resourceNode == null ? soundType.getResource() : resourceNode.getValue();
```
Pablete1234: instead of "sound" and "resources", maybe "preset" and "key"?

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -239,6 +243,23 @@ public MessageAction parseChatMessage(Element el, Class<?> scope) throws Invalid
     return new MessageAction(text, actionbar, title);
   }
 
+  @MethodParser("play-sound")
```
```suggestion
  @MethodParser("sound")
```
i'd avoid two-words whenever possible, it's not "place-blocks" it's just "fill", it's not "send-message", it's just "message", i'd follow the trend and call this one simply a "sound" action

core/src/main/java/tc/oc/pgm/action/SoundType.java
```diff
@@ -0,0 +1,56 @@
...
+  OBJECTIVE_COMPLETION_ENEMY("mob.blaze.death", 0.8f, 0.8f),
+  OBJECTIVE_MODE("mob.zombie.remedy", 0.15f, 1.2f),
+  CONTROL_POINT_CAPTURE("portal.travel", 0.35f, 2f),
+  CONTROL_POINT_LOSE("mob.blaze.death", 0.4f, 0.8f),
+  FLAG_PICKUP_OWN("mob.wither.idle", 0.7f, 1.2f),
+  FLAG_DROP_OWN("mob.wither.hurt", 0.7f, 1f),
+  FLAG_RETURN_OWN("mob.zombie.infect", 1.1f, 1.2f),
+  FLAG_PICKUP("entity.firework_rocket.blast_far", 1f, 0.7f),
+  FLAG_DROP("entity.firework_rocket.twinkle_far", 1f, 1f),
+  FLAG_RETURN("entity.firework_rocket.twinkle_far", 1f, 1f),
```
Pablete1234: i really wonder if we *really* need *all* of these
i think it's good to give a few (5-10) preset sounds and if they want smthg else they can use the resource key, but giving every single pgm sound (~30) as a preset seems overkill, because no one will know them all and listing the 30 sounds in documentation without proper explanation as to which each of them is, will be pain.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/action/ActionParser.java
cswhite2000: Will do

core/src/main/java/tc/oc/pgm/action/ActionParser.java
cswhite2000: Yeah those are slightly more clear. I'll change it

core/src/main/java/tc/oc/pgm/action/SoundType.java
cswhite2000: Yeah I was mostly going for completion. It would be a pain to document all of these. I'll try and reduce this list.

cswhite2000: I've reduced it to 12

Pablete1234 (approved):

Pablete1234:
Exercise caution when merging, needs to be merged after #1220 and with prior modification as in e66ee8d9386e5fc512352fe9d5db6553e77e3126

Electroid (approved):

PR #1220, opened 2023-08-19
Add Score & team-rescoping variables

Pablete1234:
Depends on #1214 being merged first.

Implements score variable, which is a way to directly access (read/write) a competitor's score. The variable is scoped to team.
`<score id="team_score" />`

Also implements a team-rescoping variable, which allows using a specific teams' value from a team-scoped variable, as a match-scoped variable.
`<with-team id="red_team_score" var="team_score" team="red"/>`

A more fully-fledged example:
```xml
<variables>
  <score id="team_score" />
  <with-team id="red_team_score" var="team_score" team="red" />

  <variable id="other" scope="team" /> <!-- A dummy team variable -->
  <!-- A match-scoped variable for the value of other for red team -->
  <with-team id="other_red" var="other" team="red" />
  <!-- A match-scoped variable for the value of other for blue team -->
  <with-team id="other_blue" var="other" team="blue" />
</variables>

<actions>
  <trigger scope="player" filter="some-filter">
    <action>
      <set var="red_team_score" value="10"/>
      <set var="team_score" value="team_score+5" />
    <action>
  </trigger>
</actions>
```
Whenever some-filter matches for a player (eg: entering a region), it will trigger an action that sets red team's score to exactly 10, then adds 5 to the score of the team the player is in.
In other words, if a player in red team triggers some-filter, red is reset to 10 + 5 = 15, no other scores are affected.
if a player in blue team triggers some-filter, red is reset to 10, and blue wins 5 points.


Pablete1234:
Either merge after #1218, or simply close #1218 after merging this, as it superceeds it.

cswhite2000 (dismissed):

Electroid (dismissed):

Electroid:
Conflicts need to be resolved, but lgtm.

Pablete1234:
@Electroid conflicts fixed, please re-review to merge. This is now urgent as dev branch has a broken syntax for variables that we don't want map authors relying upon

cswhite2000 (approved):

PR #1221, opened 2023-08-20
Add unused xml detection

Pablete1234:
Lists all elements and attributes that are present in the map XML, but pgm never gets to read. Very useful to find stuff you're doing wrong in xml, where pgm won't even read it, or typos:

A (real) example:
```xml
...
<kit id="experts-spade-kit" fdeduct-tools="false" ilter="experts-spade-equipped">
...
```
```
community-maps/ctw/standard/nyxis - 'fdeduct-tools' attribute of 'kit' element @ line 182 to 191: Unused node, maybe a typo?
community-maps/ctw/standard/nyxis - 'ilter' attribute of 'kit' element @ line 182 to 191: Unused node, maybe a typo?
```

Additionally fixed include errors being shown as part of the original file (but in the includes' line numbers), now maps will show (if they have a variant and it happened inside an include) as:
`path/to/map[variant]#include - description @ line: message`

A setting to be able to control if messages are issued has been added, defaults to true. Recommended to set to false for production servers, but keep it enabled for mapdev enviroments.

`<tutorial>` tag is explicitly whitelisted to avoid maps that define them being flagged. Also namespaced attributes or elements are ignored to allow 3rd party tags not messing with pgm, eg: `<ns:something/>`

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -63,6 +63,7 @@ public final class PGMConfig implements Config {
   private final List<MapSourceFactory> mapSourceFactories;
   private final Path mapPoolFile;
   private final Path includesDirectory;
+  private final boolean showUnusedXml;
 
   // countdown.*
   private final Duration startTime;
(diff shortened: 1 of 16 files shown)


cswhite2000 (approved):

Electroid (approved):
Very cool, we thought this wasn’t possible for the longest time, glad you found a way.

PR #1223, opened 2023-08-22
Revert "Update dependencies and use non-deprecated methods"

calcastor:
This is breaking Community on OCC; as such let's revert these dependency upgrades for now.


Pablete1234:
Closing in favor of #1224

PR #1224, opened 2023-08-22
Revert "Update dependencies and use non-deprecated methods"

Pablete1234:
Reverts PGMDev/PGM#1212

Updated adventure makes use of JsonObject.size(), which doesn't exist unstil gson 2.7 while sportpaper uses 2.2. This breaks certain types of text serialization, and, in particular, breaks community's cross-server adminchat.


cswhite2000 (approved):

calcastor:
https://github.com/Electroid/SportPaper/pull/136 should resolve this on SportPaper's end. There are implications for plugins which depend on PGM and Adventure, and want to support Spigot unfortunately. However, Community itself does require SportPaper to my knowledge.

It seems difficult to reconcile this if we want to use up to date libraries.

Electroid (approved):

PR #1226, opened 2023-08-23
Implement TNT map tag if any kit has enough tnt

Pablete1234:
Adds a `#tnt` map tag, which will be shown in maps where any kit gives 16 tnt or more. The 16 limit is so that maps giving just a few tnt blocks on a kit (maybe as killreward) are not really considered "tnt maps"

### DIFF
--- a/core/src/main/java/tc/oc/pgm/kits/KitModule.java
+++ b/core/src/main/java/tc/oc/pgm/kits/KitModule.java
@@ -2,10 +2,13 @@ package tc.oc.pgm.kits;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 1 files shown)


cswhite2000:
Will this also tag maps with tnt in a shop? Such as [Extraterrestrial](https://github.com/OvercastCommunity/CommunityMaps/blob/master/dtcm/standard/extraterrestrial/map.xml#L447)
Or in kill rewards? [Streamline II](https://github.com/OvercastCommunity/PublicMaps/blob/7458163729f987a732fc006e4ab7120845fe3986/dtcm/standard/streamline_ii/map.xml#L168)

Pablete1234:
This would not tag either of those, they're not kits

cswhite2000:
Well, I guess my concern then is that this might give the false impression during voting that a map isn't a tnt map due to it not having the tnt tag.

Electroid (approved):

PR #1228, opened 2023-08-30
Additional SoundType Enums

arcadeboss:
A few extra sounds for the sound module.


Pablete1234 (changes requested):
You can use these via the "key" option, unless they have widespread use i don't think they should be included.

arcadeboss:
> You can use these via the "key" option, unless they have widespread use i don't think they should be included.

What is a key option?  How is it used?

Ah, nevermind, I think I found it.

> sound(key("mob.irongolem.hit"), Sound.Source.MASTER, 1, 1)

Wow, if this works, I had no idea we weren't limited to just the enum'd sounds.  

... No, wasn't able to find a way to use it.  The call requires the Enum.

arcadeboss:
Now that I know how to use key for sound, this is not needed.

Thanks.

PR #1237, opened 2023-09-10
Add power attribute for projectiles

### LINKED ISSUE #1232: Custom Projectile - Explosiveness Control
I would like more control over the explosion factor of a Fireball, as I can only control damage to players, not damage to the environment.  (Everyone seems to feel it is under-powered in BW context.)

A powerful new custom projectile attribute called "power" would allow control of a the boom factor of either all projectiles (if practical) or for a set of applicable projectiles including Fireball. It would work just like the same parameter for TNT.  Most projectiles would default to 0, although a Fireball might default to a positive number.

cswhite2000:
Resolves https://github.com/PGMDev/PGM/issues/1232

Adds a power attribute for projectiles. If the projectile implements `Explosive` and a `power` is supplied it will set the explosive power to the specified amount.

Example usage:
```xml
<projectile id="throwable-fireball" name="Fireball" projectile="Fireball" power="5" velocity=".5" click="right" damage="8" cooldown="2s"/>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
@@ -8,6 +8,7 @@ import java.util.Set;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Entity;
+import org.bukkit.entity.Explosive;
 import org.bukkit.entity.Fireball;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
@@ -104,6 +105,9 @@ public class ProjectileMatchModule implements MatchModule, Listener {
           projectile =
               player.getWorld().spawn(player.getEyeLocation(), projectileDefinition.projectile);
           projectile.setVelocity(velocity);
+        }
+        if (projectileDefinition.power != null && projectile instanceof Explosive) {
+          ((Explosive) projectile).setYield(projectileDefinition.power);
         }
         projectile.setMetadata(
             "projectileDefinition", new FixedMetadataValue(PGM.get(), projectileDefinition));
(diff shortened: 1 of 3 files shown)


Pablete1234:
I believe yield is the probability of items dropping, 1=100% of items, 0.5=50% items etc, not the power of the explosion

arcadeboss:
> I believe yield is the probability of items dropping, 1=100% of items, 0.5=50% items etc, not the power of the explosion

We did a test on CS's test server last night where he had hard-coded the value to 10.   It was super explosive in both destruction of materials and allowing high Fireball jumping.

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -105,6 +106,9 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
               player.getWorld().spawn(player.getEyeLocation(), projectileDefinition.projectile);
           projectile.setVelocity(velocity);
         }
+        if (projectileDefinition.yield != null && projectile instanceof Explosive) {
+          ((Explosive) projectile).setYield(projectileDefinition.yield);
```
arcadeboss: Indention off?

cswhite2000: This was run through the formatter so the indentation is correct.

cswhite2000:
> I believe yield is the probability of items dropping, 1=100% of items, 0.5=50% items etc, not the power of the explosion

So the `setYield` on `Explosive` sets the explosion's power / explosion size.
While `setYield` on `EntityExplodeEvent` sets the block drop rate.

Pablete1234:
> > I believe yield is the probability of items dropping, 1=100% of items, 0.5=50% items etc, not the power of the explosion
> 
> So the `setYield` on `Explosive` sets the explosion's power / explosion size. While `setYield` on `EntityExplodeEvent` sets the block drop rate.

Well, that sounds like a stupid decision on the bukkit api or other upstream sources, can we avoid passing it down to pgm?
What i mean by that is, can we not call it `yield="x"`?
Currently pgm has
```xml
<tnt>
  <power>5</power>
  <yield>0.5</yield>
</tnt>
```
and i wouldn't want that yield and this one to be confused.

Maybe just call it power (like tnt has)?

cswhite2000:
Sure will do

Pablete1234 (approved):

Electroid (approved):

PR #1242, opened 2023-09-17
Use more exact username matching when available

cswhite2000:
Current username matching is overly broad for commands and results.

For example:
Running `/maps -a cs_` will not show any maps by `cs_` on the first page of results.
In this case, I've changed it such that if an exact username match is found it will only show those matches, otherwise it uses existing behavior.

Suggestions for player-targeting commands are overly broad
Running `/team force ma` and then tab completing currently prioritizes usernames like `CincoDeMayo` over something like `math1234`.
I've changed it such that if a prefix match is found it will only show those usernames that match the prefix, otherwise it will use the current behavior.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapCommand.java
@@ -47,7 +47,6 @@ import tc.oc.pgm.rotation.pools.MapPool;
 import tc.oc.pgm.util.Audience;
 import tc.oc.pgm.util.LiquidMetal;
 import tc.oc.pgm.util.PrettyPaginatedComponentResults;
-import tc.oc.pgm.util.StringUtils;
 import tc.oc.pgm.util.named.MapNameStyle;
 import tc.oc.pgm.util.named.NameStyle;
 import tc.oc.pgm.util.text.TextFormatter;
@@ -83,14 +82,20 @@ public final class MapCommand {
       search = search.filter(map -> matchesTags(map, tagSet.get(false), tagSet.get(true)));
     }
 
-    if (author != null) {
-      String query = StringUtils.normalize(author);
-      search = search.filter(map -> matchesAuthor(map, query));
-    }
-
     // FIXME: change when cloud gets support for default flag values
     final Phase finalPhase = phase == null ? Phase.PRODUCTION : phase;
     search = search.filter(map -> map.getPhase() == finalPhase);
+
+    if (author != null) {
+      String query = author;
+
+      List<MapInfo> collect = search.collect(Collectors.toList());
+      boolean exactMatch =
+          collect.stream().anyMatch((suggestion) -> matchesAuthor(suggestion, query, true));
+      search =
+          collect.stream()
+              .filter((mapSuggestion -> matchesAuthor(mapSuggestion, query, exactMatch)));
+    }
 
     Set<MapInfo> maps = search.collect(Collectors.toCollection(TreeSet::new));
     int resultsPerPage = 8;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/command/MapCommand.java
```diff
@@ -83,15 +83,25 @@ public void maps(
...
-      String query = StringUtils.normalize(author);
-      search = search.filter(map -> matchesAuthor(map, query));
-    }
-
     // FIXME: change when cloud gets support for default flag values
     final Phase finalPhase = phase == null ? Phase.PRODUCTION : phase;
     search = search.filter(map -> map.getPhase() == finalPhase);
 
+    if (author != null) {
+      String query = StringUtils.normalizeUsername(author);
```
Pablete1234: Why normalize usernames, its not like they can have random chars, they're a-z0-9_

core/src/main/java/tc/oc/pgm/util/Players.java
```diff
@@ -34,10 +34,16 @@ public static String getVisibleName(CommandSender viewer, Player other) {
...
-        .map(p -> Players.getVisibleName(sender, p))
-        .filter(n -> LiquidMetal.match(n, query))
+    List<String> playerSuggestions =
+        Bukkit.getOnlinePlayers().stream()
+            .filter(p -> Players.isVisible(sender, p))
+            .map(p -> Players.getVisibleName(sender, p))
+            .filter(n -> LiquidMetal.match(n, query))
+            .collect(Collectors.toList());
+    boolean prefixMatchFound =
+        playerSuggestions.stream().anyMatch((suggestion) -> suggestion.startsWith(query));
```
Pablete1234: Chech starts with, but ignoring case
Also, instead of streaming for a boolean, just stream to a list, if its empty return the prev one, if non empty return the new one

util/src/main/java/tc/oc/pgm/util/StringUtils.java
```diff
@@ -140,6 +140,14 @@ public static String normalize(String text) {
             .toLowerCase(Locale.ROOT);
   }
 
+  public static String normalizeUsername(String text) {
+    return text == null
+        ? ""
+        : Normalizer.normalize(text, Normalizer.Form.NFD)
```
Pablete1234: This is literally a no-op for a playee name

util/src/main/java/tc/oc/pgm/util/StringUtils.java
```diff
@@ -140,6 +140,14 @@ public static String normalize(String text) {
...
+            .replaceAll("[^A-Za-z0-9_ ]", "")
```
Pablete1234: Same for this

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/command/MapCommand.java
Electroid: It's rare, but there are invalid usernames. I think it's fine

core/src/main/java/tc/oc/pgm/util/Players.java
cswhite2000: Right that would be more efficient

util/src/main/java/tc/oc/pgm/util/StringUtils.java
cswhite2000: In the past there have been legacy usernames with spaces ` ` or `.`. However I'm unable to find any at this point https://namemc.com/search?q=tempusername3 as it seems like every one I search for was changed to `tempusername#` on January 19th 2023. Not sure if there are still invalid usernames that I've missed. Not sure whether to  continue normalizing usernames given this.

core/src/main/java/tc/oc/pgm/command/MapCommand.java
```diff
@@ -154,9 +163,20 @@ private static boolean matchesTags(
...
+
   private static boolean matchesAuthor(MapInfo map, String query) {
+    String lowerCaseName = query.toLowerCase(Locale.ROOT);
     for (Contributor contributor : map.getAuthors()) {
-      if (StringUtils.normalize(contributor.getNameLegacy()).contains(query)) {
+      String name = contributor.getNameLegacy();
+      if (name != null && name.toLowerCase(Locale.ROOT).contains(lowerCaseName)) {
         return true;
       }
     }
```
Pablete1234: Can probably merge these into a :
`private static boolean matchesAuthor(MapInfo map, String query, boolean exact) {`

Pablete1234 (approved):

PR #1249, opened 2023-09-27
Add max and min to expressions

cswhite2000:
Adds `max` and `min` as functions that can be used in `<set/>` actions

Example:
```xml
<set var="player_score" value="min(player_score,3)"/>
or
<set var="player_score" value="max(player_score,10)"/>
```


Pablete1234:
Just to leave a record of it, we discussed trying to support things like `max(a, b, c, d)` but Exp4j expects a specific number of parameters for a function, no support for variable amount of args.

Would be great to long-term have that tho, but for the time being 2-arg is the best we can do

Pablete1234 (approved):

Electroid (approved):

PR #1257, opened 2023-10-13
Implement parser for inline filters

Pablete1234:
Adds support for formula-like definitions of inline filters.

```xml
<filters>
  <holding id="holding-gap"><item>golden apple</item></holding>
  <pulse id="can-give-gap" duration="0.05s" period="0.1s" filter="all(not(holding-gap), missing_gaps=1..)"/>
</filters>
```

Note how the pulse filter depends on an in-line defined equivalent of:
```xml
<all>
  <not><filter id="holding-gap"></not>
  <variable var="missing_gaps">1..</variable>
</all>
```

Also added some syntax highlighting to the errors:
image
image

<details>
<summary> Grand dueling TM II broke, but was already fixed </summary>

Tested with all current maps loaded in occ repos and one map broke, grand dueling tournament II, due to the use of ( and ) in the filter ids:
image
</detauls>


cswhite2000 (approved):
I've made it so that the Grand Dueling Tournament II no longer conflicts with these changes

Electroid (approved):

PR #1260, opened 2023-10-22
Add a timelimit variable

cswhite2000:
Adds a match scoped `timelimit` action

Behaves like the command

Example:
```xml
<timelimit duration="10s" result="objectives"/>
```

All attributes are optional.

This has all of the same optional arguments as the command.

For @zzufx

### DIFF
--- a/core/src/main/java/tc/oc/pgm/variables/types/TimeLimitVariable.java
+++ b/core/src/main/java/tc/oc/pgm/variables/types/TimeLimitVariable.java
@@ -1,0 +1,51 @@
+package tc.oc.pgm.variables.types;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.timelimit.TimeLimit;
+import tc.oc.pgm.timelimit.TimeLimitMatchModule;
+import tc.oc.pgm.variables.VariableDefinition;
+
+public class TimeLimitVariable extends AbstractVariable<Match> {
+
+  private TimeLimit oldTimeLimit;
+  private TimeLimitMatchModule tlmm;
+
+  public TimeLimitVariable(VariableDefinition<Match> definition) {
+    super(definition);
+    oldTimeLimit =
+        new TimeLimit(null, Duration.of(0, ChronoUnit.SECONDS), null, null, null, null, true);
+  }
+
+  @Override
+  public void postLoad(Match match) {
+    tlmm = match.moduleRequire(TimeLimitMatchModule.class);
+  }
+
+  @Override
+  protected double getValueImpl(Match obj) {
+    Duration remaining = tlmm.getFinalRemaining();
+    return remaining == null ? -1 : remaining.getSeconds();
+  }
+
+  @Override
+  protected void setValueImpl(Match obj, double value) {
+    TimeLimit existingTimeLimit = tlmm.getTimeLimit();
+    if (value < 0) {
+      if (existingTimeLimit != null) {
+        oldTimeLimit = existingTimeLimit;
+      }
+
+      tlmm.cancel();
+      return;
+    }
+
+    TimeLimit newTimeLimit =
+        new TimeLimit(
+            existingTimeLimit != null ? existingTimeLimit : oldTimeLimit,
+            Duration.of((long) value, ChronoUnit.SECONDS));
+    tlmm.setTimeLimit(newTimeLimit);
+    tlmm.start();
+  }
+}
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -309,6 +313,18 @@ public ReplaceItemAction parseReplaceItem(Element el, Class<?> scope) throws Inv
     return new ReplaceItemAction(matcher, item, keepAmount, keepEnchants);
   }
 
+  @MethodParser("timelimit")
+  public TimeLimitAction parseTimeLimit(Element el, Class<?> scope) throws InvalidXMLException {
+    Duration duration = XMLUtils.parseDuration(el.getAttribute("duration"), Duration.ZERO);
```
Pablete1234: Reuse the parsing code in time limits, and just pass in a TimeLimit to the action

core/src/main/java/tc/oc/pgm/timelimit/TimeLimitModule.java
```diff
@@ -83,14 +85,25 @@ public TimeLimitModule parse(MapFactory factory, Logger logger, Document doc)
       return oldTimeLimit;
     }
 
-    private static @Nullable TimeLimit parseTimeLimit(MapFactory factory, Element el)
+    private static @Nullable TimeLimit parseTimeLimit(MapFactory factory, Element el, String child)
         throws InvalidXMLException {
-      el = el.getChild("time");
+      el = el.getChild(child);
```
Pablete1234: Doesn't seem like something that would need changing, given "time" is the only param ever used?
Maybe just rename this function to `parseChildTimeLimit` and rename `el` to `parent`, to better reflect what it does

core/src/main/java/tc/oc/pgm/timelimit/TimeLimitModule.java
```diff
@@ -83,14 +85,25 @@ public TimeLimitModule parse(MapFactory factory, Logger logger, Document doc)
...
 
+      return parseTimeLimit(factory, el);
+    }
+
+    public static TimeLimit parseTimeLimit(MapFactory factory, Element el)
+        throws InvalidXMLException {
+      String durationText = el.getTextNormalize();
+      Duration duration =
+          durationText == null || durationText.isEmpty()
+              ? XMLUtils.parseDuration(el.getAttribute("duration"))
```
Pablete1234: I'd say let's not modify the syntax and keep it as 
`<timelimit>5m</timelimit>` instead of `duration="5m"`

at the end of the day it's the timelimit syntax

Pablete1234 (commented):
Besides the two minor things i mentioned, looks good to me

However, i thought we were going to have timelimit actually implemented as a variable instead of an action, which gives the flexibility to change it more arbitrarily (eg: +120 for adding 2 mins to tl)

cswhite2000:
> However, i thought we were going to have timelimit actually implemented as a variable instead of an action

Right, I forgot we had this discussion
Probably would be better to implement it that way

cswhite2000:
Switched to a timelimit variable with example usage:

```xml
<variables>
	<timelimit id="tl_2"/>
</variables>

...
in actions:
<set var="tl_2" value="60"/>
```

Other behaviour:

Negative time limits cancel the timer.
A time-limit of 0 will instantly end the match

core/src/main/java/tc/oc/pgm/variables/types/TimeLimitVariable.java
```diff
@@ -0,0 +1,77 @@
...
+import tc.oc.pgm.timelimit.TimeLimitMatchModule;
+import tc.oc.pgm.variables.VariableDefinition;
+
+public class TimeLimitVariable extends AbstractVariable<Match> {
+
+  VictoryCondition result = null;
+  Duration overtime = null;
+  Duration maxOvertime = null;
+  Duration endOvertime = null;
+  boolean show = true;
```
Pablete1234: could probs just be a TimeLimit object, also private?
it should be fine as long as TL is just a plain immutable object with no extra logic, which i believe it is

cswhite2000: Oh yeah that would be better

core/src/main/java/tc/oc/pgm/variables/types/TimeLimitVariable.java
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+
+    TimeLimit newTimeLimit;
+    if (existingTimeLimit == null) {
+      newTimeLimit = new TimeLimit(oldTimeLimit, Duration.of((long) value, ChronoUnit.SECONDS));
+    } else {
+      newTimeLimit =
+          new TimeLimit(existingTimeLimit, Duration.of((long) value, ChronoUnit.SECONDS));
+    }
```
Pablete1234:
```suggestion
    TimeLimit newTimeLimit = new TimeLimit(
        existingTimeLimit != null ? existingTimeLimit : oldTimeLimit, 
        Duration.of((long) value, ChronoUnit.SECONDS));
```

Pablete1234 (approved):

PR #1261, opened 2023-10-26
Cleanup material filter & matcher legacy usages

Pablete1234:
The `BlockFilter` class does exactly the same as `MaterialFilter` does with the appropriate material matcher. It's a legacy class that can be simplified away by just migrating to material filter. Additionally does a few other bits of cleanup, no behavior changes should occur.

Also gets rid of `BlockQueryCustomLocation` in favor of just `BlockQuery` with a set location, which again, should behave identically.


cswhite2000 (approved):
This should make the diff for the final 1.13 material changes slightly smaller 👍

PR #1271, opened 2023-11-05
Fix shading of exp4j

applenick:
Fixes a `NoClassDef` caused by the recent merge of #1268


Pablete1234:
Fixed in #1270

applenick:
@Pablete1234 Turns out it should have been `com.github.Pablete1234` not `me.github.Pablete1234`

PR #1277, opened 2023-11-22
Fix map load exception with MaxBuildVariable

applenick:
Here's a quick fix PR! This has been causing issues where PGM will not startup properly, particularly when a max build variable is defined in a global include.

Should work as expected 👍 Let me know if you'd like any additional changes made.


Pablete1234 (approved):

Pablete1234:
Turns out this change is actually not correct. The issue here isn't that the region match module is optional, but an issue in load order. I'll open a separate PR to address it.

Issue #1114, opened 2022-12-28
Add player tracking compass item ?


lefckntea:
Simply a compass which points to the direction of the nearest enemy player. Right clicking could show you their name and distance in chat but there's various ways the final thing could look like.
Never seen this discussed anywhere in the pgm community before so i guess no one really needs it but it's pretty common in a lot of other gamemodes so maybe it's simply being overlooked a little. I could see this being useful on some maps like some blitz stuff or hide and seek after some time.

Cazadorian:
This actually used to exist in PGM as part of Ghost Squadron but was limited to Ghost Squadron, and is no longer present. I've wanted it for a few blitz maps I've made to avoid people hiding and stalling down the timer instead. 

When I've discussed it with people in the past the general idea of how to implement it is to have an attribute or default behaviour for compasses that enables tracking, and being able to tie it to a filter with it pointing to the nearest player who matches that filter. In a blitz map for example, it could be set to the blitz filter and so would point to the nearest enemy with blitz enabled. In a CTW map, it could be set to track the nearest enemy player who is carrying a piece of wool. Would be good to see it implemented again properly.

TBG1000:
I want to chip in and say that Warzone used to have a similar functionality but exclusive to Infection maps. The compass would be given in the last few minutes to all of the infected and would point towards the nearest human (not infected) player. It would be great to use it in the same way Cazadorian is describing for such maps.

Pablete1234:
Cazadorian explained the design/implementation i proposed pretty well, but yes, it would be given via kits (similar to how grenades are given) with a special item with that functionality, potential syntax:
```xml
<kit id="...">
  <item material="compass" tracking-filter="..."/>
</kit>
```

arcadeboss:
This would be a nice feature.   My suggestion for how it could work:

Right clicking the compass would provide the distance between the compass holder and the nearest player not on their team. 

A compass attribute could configure the granularity (1 block, 10 blocks, etc.)

PR #1286, opened 2023-12-18
Add flag-tracking compass functionality

OhPointFive:
Firstly, sorry for just making this without discussing the design first. Turns out I needed to get this done much sooner than I expected. Luckily it's a pretty easy implementation so if it gets thrown away & replaced by something better it's not much work lost.

This PR adds the ability to have compasses that track flags. Specifically, it uses this XML syntax:

```xml
<flag compass-filter="always">
```
or
```xml
<flag compass-filter="red-team">
<flag compass-filter="blue-team">
```

**Can you have multiple different compasses?**
I'm pretty sure a player cannot have multiple compasses with different targets in a vanilla client.

**Why put this on flags instead of compass items?**
Putting the filter on the flags, instead of on compass items, will make it clear to map makers they can't treat different compasses differently. It also makes the filter simpler if other kinds of targets are added — hills or cores could also have compass filters, and the filters will always filter players, and not some weird mishmash of objectives and locations.

**What happens if multiple filters match the player?**
Right now it would probably visually just track one, while sending updates for both. Practically I'm not really sure what could be done here. This is one of the major drawbacks of this design.

**What about tracking players?**
I'm not sure how this design could be extended for player tracking, which is the other major drawback of this design. Perhaps there's a clean way that I'm missing at 3 am, or perhaps we'll need to rethink this entirely. See #1114 for discussion on this.

**What about tracking multiple things & clicking the compass?**
This could easily be added — increment a variable when the player clicks the compass and filter on variables. I don't think right click actions are currently available, but they are planned so that's an easy future extension of this.

---

My other concern with this PR is how it filters all the players every tick. Some quick testing seems to indicate that this isn't an issue, and in my mind I'd be surprised if it was an issue expect with pathological filters, but I'm not familiar enough with PGM to be confident.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/FlagDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/flag/FlagDefinition.java
@@ -39,6 +39,8 @@ public class FlagDefinition extends ProximityGoalDefinition {
   private final Filter pickupFilter; // Filter players who can pickup this flag
   private final Filter captureFilter; // Filter players who can capture this flag
   private final Filter dropFilter; // Filter players who can drop the flag
+  private final @Nullable Filter
+      compassFilter; // Filter for players who have their compass set to this flag
   private final @Nullable Kit pickupKit; // Kit to give on flag pickup
   private final @Nullable Kit dropKit; // Kit to give carrier when they drop the flag
   private final @Nullable Kit carryKit; // Kit to give to/take from the flag carrier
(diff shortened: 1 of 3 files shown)


Pablete1234:
Having this param filter be part of the flags is pretty weird, and even if a single compass is allowed at once i think it makes more sense for this to be defined in the compass item given in the kit itself, or in a separate module that defines the compass pointing rules

Pablete1234:
Closing in favor of a more generic approach in #1308

PR #1287, opened 2023-12-22
Allow for multiple game sub-elements

zzufx:
Since the reintroduction of includes, there's been a rise in (or comeback of) "unofficial" gamemodes. These gamemode include files tend to have a `<game>GAMEMODE NAME</game>` sub-element each. 

This means that currently, gamemodes like King's Conquest, Ghost Squadron, Blasternauts, Monster Mash, OITQ or _Lifesteal_ can't be mixed with one another, as PGM will complain about `Multiple values for 'game'`

I believe allowing for multiple of these sub-elements and just using the last defined one makes the most sense here, as the other option would be to have no `<game>` element in includes and always having to define it per map.

This change would allow you to combine, for instance, Ghost Squadron and Lifesteal, a compatible combination which wasn't previously possible due to each include file having `<game>Ghost Squadron</game>` and `<game>Lifesteal</game>` respectively.

With this, you'd be able to combine them _and_ add a custom `<game>GS and Lifesteal</game>` line to that specific map's XML after its include calls (or just use the gamemode from the last defined include, if you wanted to)

This concept could go even further to also allow for multiple objectives (for includes that have those), but that's not part of this PR at the moment.

I have tested this in a local server.


Pablete1234:
I'm not convinced that this is the way we should handle multiple game elements, instead of combining them like we do for gamemodes (where a map being blitz and rage, will automatically understand it to be both blitz and rage, and display as such in the scoreboard and elsewhere).
Maybe we just need a way for game elements to be handled less like "A random free-text string to put on scoreboard" and more of a "this is a custom non-standard gamemode that this map uses". The main difference i think would be to add a secondary="true" or a type="primary/secondary" to the tag, so that we can differentiate primary (or main) and secondary (or add-on) gamemodes

Pablete1234:
Although as explained above i think we could do much better, this is probably an ok and pretty trivial solution for the time

We can explore better options down the line but this is an annoying blocker for certain include stuff so it's better off if changed

Pablete1234 (approved):

Pablete1234 (approved):

cswhite2000 (approved):

PR #1288, opened 2024-01-01
Avoid reflection when using SportPaper

calcastor:
~~Depends on https://github.com/Electroid/SportPaper/pull/139~~

The reflection used in NMSHacks1_8 breaks in modern Java versions, but we can modify SportPaper and avoid that here.

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/nms/v1_8/NMSHacksSportPaper.java
+++ b/util/src/main/java/tc/oc/pgm/util/nms/v1_8/NMSHacksSportPaper.java
@@ -16,7 +16,9 @@ import org.bukkit.Bukkit;
 import org.bukkit.ChunkSnapshot;
 import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.World;
 import org.bukkit.block.BlockState;
+import org.bukkit.craftbukkit.v1_8_R3.CraftWorld;
 import org.bukkit.craftbukkit.v1_8_R3.inventory.CraftItemStack;
 import org.bukkit.craftbukkit.v1_8_R3.scoreboard.CraftTeam;
 import org.bukkit.entity.EntityType;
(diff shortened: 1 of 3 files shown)


calcastor:
This has been updated to add a fallback for older versions of SportPaper. However, it does now depend on newer versions of SportPaper to build thus the action failure.

util/src/main/java/tc/oc/pgm/util/nms/v1_8/NMSHacksSportPaper.java
```diff
@@ -185,6 +191,25 @@ public Skin getPlayerSkinForViewer(Player player, Player viewer) {
...
+      Field worldServerField = ReflectionUtils.getField(CraftWorld.class, "world");
+      Field dimensionField = ReflectionUtils.getField(WorldServer.class, "dimension");
+      Field modifiersField = ReflectionUtils.getField(Field.class, "modifiers");
+
+      try {
+        modifiersField.setInt(dimensionField, dimensionField.getModifiers() & ~Modifier.FINAL);
+        dimensionField.set(worldServerField.get(world), 11);
+      } catch (IllegalAccessException ex) {
+        // No-op, newer version of Java have disabled modifying final fields
+      }
```
Pablete1234: just call `super.resetDimension(world)`

calcastor:
@Electroid The relevant PR for this has been merged in SportPaper however it does not seem to be deployed to the Maven repo for SportPaper. Any ideas?

calcastor:
GitHub actions doesn't seem to like the pgm.fyi maven repo, but with snapshots enabled for the SportPaper maven repo the build succeeds here now. That seems to be only permitted for authenticated GitHub users, however the pgm.fyi maven repo does seem to work fine for local builds.

cswhite2000 (approved):

Electroid (approved):

PR #1289, opened 2024-01-03
Fix nicked player sorting in player list

zzufx:
Currently, when a player with a rank is disguised, they keep their "real" position on the player list (appearing higher on the list than unranked players), making it extremely easy to differentiate  nicked players from regular players as they're not following alphabetical order.

This PR adds some extra checks to the "group permission check" stage and after in _PlayerOrder_.

I have tested this in a local server with 4 clients.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/tablist/PlayerOrder.java
+++ b/core/src/main/java/tc/oc/pgm/tablist/PlayerOrder.java
@@ -64,15 +64,17 @@ public class PlayerOrder implements Comparator<MatchPlayer> {
     if (aStaff && !bStaff) return -1;
     else if (bStaff && !aStaff) return 1;
 
-    // Compare the nicknames of 'a' and 'b' if both are nicked, skip group permission check
+    // Compare the nicknames of 'a' and 'b'. If both are nicked, skip group permission check
     if (aNick != null && bNick != null) return aNick.compareToIgnoreCase(bNick);
 
     // If players have different permissions, the player with the highest ranked perm
-    // that the other one does't have is first. Disguised players effectively have no perms.
+    // that the other one doesn't have is first. Disguised players effectively have no perms.
     for (Config.Group group : PGM.get().getConfiguration().getGroups()) {
+      if (group.getId().equalsIgnoreCase("default")) continue;
+
       Permission permission = group.getPermission();
-      boolean aPerm = a.hasPermission(permission);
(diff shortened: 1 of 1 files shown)


Pablete1234:
Be aware, this will put tbem always at the end, since everyone has the default group

core/src/main/java/tc/oc/pgm/tablist/PlayerOrder.java
```diff
@@ -64,23 +64,26 @@ public int compare(MatchPlayer ma, MatchPlayer mb) {
...
+      } else if (bPerm && (!aPerm && bNick == null)) {
         return 1;
       }
     }
 
+    if (aNick != null && bNick == null) return aNick.compareToIgnoreCase(b.getName());
+    else if (aNick == null && bNick != null) return a.getName().compareToIgnoreCase(bNick);
+
     // All else equal, order the players alphabetically
     return a.getName().compareToIgnoreCase(b.getName());
```
Pablete1234:
```suggestion
    // All else equal, order the players alphabetically
    return (aNick == null ? a.getName() : aNick).compareToIgnoreCase(bNick == null ? b.getName() : bNick)
```

core/src/main/java/tc/oc/pgm/tablist/PlayerOrder.java
```diff
@@ -64,24 +64,25 @@ public int compare(MatchPlayer ma, MatchPlayer mb) {
...
+    // Compare the nicknames of 'a' and 'b'. If both are nicked, skip group permission check
     if (aNick != null && bNick != null) return aNick.compareToIgnoreCase(bNick);
 
     // If players have different permissions, the player with the highest ranked perm
-    // that the other one does't have is first. Disguised players effectively have no perms.
+    // that the other one doesn't have is first. Disguised players effectively have no perms.
     for (Config.Group group : PGM.get().getConfiguration().getGroups()) {
       Permission permission = group.getPermission();
       boolean aPerm = a.hasPermission(permission);
       boolean bPerm = b.hasPermission(permission);
```
Pablete1234:
```suggestion
      boolean aPerm = a.hasPermission(permission) && aNick == null;
      boolean bPerm = b.hasPermission(permission) && bNick == null;
```
If just included here, makes the follow-up if simpler

Pablete1234 (approved):

Electroid (approved):

PR #1298, opened 2024-02-08
Different style for nicked + vanished players

zzufx:
This adds an easy way to tell if a player is both nicked and vanished at the same time, visible when they send a message in chat and in commands like `/list` and `/nicks` (Community)

nick_style
(notice how both the real name and the nick are strikethrough in the third example)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/util/player/PlayerRenderer.java
+++ b/core/src/main/java/tc/oc/pgm/util/player/PlayerRenderer.java
@@ -26,8 +26,6 @@ import tc.oc.pgm.util.named.NameStyle;
 @SuppressWarnings("UnstableApiUsage")
 public class PlayerRenderer {
   private static final TextColor DEAD_COLOR = NamedTextColor.DARK_GRAY;
-  private static final Style NICK_STYLE =
-      Style.style(TextDecoration.ITALIC).decoration(TextDecoration.STRIKETHROUGH, false);
 
   private final LoadingCache<PlayerCacheKey, Component> nameCache;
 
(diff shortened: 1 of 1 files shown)


Pablete1234:
Nicked and vanished isnt a valid state, if you're vanished your nick should be inactive

zzufx:
We've been able to be both vanished and nicked for a long time now (wasn't the case before), which is extremely useful as it lets you hide join/leave messages while nicked, and spectate players who are breaking the rules while vanished without having to relog if you joined with a nick.

If that's an unintended feature (which I wasn't aware of) I personally think it's way better to keep it or find a way to properly support it instead.

Pablete1234:
> and spectate players who are breaking the rules while vanished without having to relog if you joined with a nick.

So, you should be able to vanish while nicked, but it essentially makes you no longer nicked, just vanished. To other players online the "nick" disconnected. If you later un-vanish, your nick should become active again, and if your nick is different than the one you used to have it should appear updated (ie: "new nick" joined), but i'm aware that this last part has never worked, and you can't enable, disable, or change your nick by vanishing, you need to rejoin for it to apply

zzufx:
Ah, I see. I personally like the current setup - I haven't seen people change their nick often enough for that to be a problem. The only thing I didn't like I changed in this PR, and I like that this shows "more info" than just hiding the nick (as, for instance, if a nicked staff member unvanishes, I already know I should refer to them as their nick if I'm talking in public chat)

If the idea is to fix that behaviour you mentioned though, I guess this PR wouldn't really relevant (otherwise I'd love to have this).

PR #1303, opened 2024-02-29
fix(ActionScopeValidation): use action scope in class

TTtie:
Corrects cases when specifying an invalid action scope results in incorrect errors of "Wrong action scope, got SCOPE but expected SCOPE" where the expected scope is the scope from the action definition. This PR corrects the oversight and uses the scope from the class instead.


Pablete1234 (dismissed):
Would probs be better to just rename the local variable to something else so it doesn't shadow the field, but this is fine too

Pablete1234 (approved):

PR #1304, opened 2024-03-03
Implement basic command flag suggestion filter

Pugzy:
Our current no-filter approach suggests command flags first (and incorrectly) instead of partly formed inputs.

Currently running `/ban den[tab]` changes the input to `/ban --silent` and then `/ban --time`.

This change means that command flags will only be suggested if:
- Input string currently empty
- Input string starts with a flag `-`/`--`
- Matches the start of the command flag

The above `ban` command example is from Community which extends the PGM command graph, to see an example baked in to PGM the `/setnext` command can be used which has `--force` and `--reset` flags.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/util/CommandGraph.java
+++ b/core/src/main/java/tc/oc/pgm/command/util/CommandGraph.java
@@ -22,6 +22,7 @@ import cloud.commandframework.exceptions.InvalidSyntaxException;
 import cloud.commandframework.exceptions.NoPermissionException;
 import cloud.commandframework.exceptions.parsing.ParserException;
 import cloud.commandframework.execution.CommandExecutionCoordinator;
+import cloud.commandframework.execution.FilteringCommandSuggestionProcessor;
 import cloud.commandframework.extra.confirmation.CommandConfirmationManager;
 import cloud.commandframework.meta.CommandMeta;
 import cloud.commandframework.minecraft.extras.MinecraftHelp;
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/command/util/CommandGraph.java
```diff
@@ -85,9 +86,11 @@ protected PaperCommandManager<CommandSender> createCommandManager() throws Excep
...
             context.getCommandContext().store(CommandKeys.INPUT_QUEUE, context.getInputQueue()));
 
-    // By default, suggestions run by a filtered processor.
-    // By default, it prevents suggestions like "s" -> "Something" or "someh" -> "Something"
-    manager.commandSuggestionProcessor((cpc, strings) -> strings);
+    // Basic suggestion filtering processor which avoids suggesting flags when not applicable
+    manager.commandSuggestionProcessor(
+        new FilteringCommandSuggestionProcessor<>(
+            FilteringCommandSuggestionProcessor.Filter.Simple.contextFree(
+                (s, i) -> i.isEmpty() || s.startsWith("-") == s.startsWith(i))));
```
Pablete1234: Starts with should be case-insenitive so that `--FOR` can tab-in `--force` the same way `/JO` can tab-in to `/join`

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #1305, opened 2024-03-04
Add take-payment action

Pablete1234:
As per mentioned https://github.com/PGMDev/PGM/pull/1254#issuecomment-1759728344, this adds a take-payment action, that is able to programatically (via action) trigger the user "paying" with items in their inventory:

```xml
<take-payment>
  <!-- This is syntax taken from the shops, you can define multiple items required as payment, but can be simpler for single-item payments -->
  <payment material="diamond" price="5"/>
  <payment material="emerald" price="5"/>
  <payment price="1">
      </item>
  </payment>
  <success-action>
    <message><text>Success!</text></message>
    <kit>
        <chestplate material="diamond chestplate"/>
    </kit>
  </success-action>
  <fail-action>
    <message><text>You can't afford this!</text></message>
  </fail-action>
</take-payment>
```

When using attributes and not needing complicated prices (single item cost) it can be simplified:
```xml
<take-payment id="buy-chestplate" currency="diamond" price="5" success-action="success" fail-action="too-poor"/>

<action id="success">
  <message><text>Success!</text></message>
  <kit>
      <chestplate material="diamond chestplate"/>
  </kit>
</action>
<message id="too-poor"><text>You can't afford this!</text></message>
```

Note: has not been tested yet.
@CoWinkKeyDinkInc if you want to give this a test and confirm if it solves your use-case that'd be greatly appreciated

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -28,10 +28,12 @@ import tc.oc.pgm.action.actions.ReplaceItemAction;
 import tc.oc.pgm.action.actions.ScopeSwitchAction;
 import tc.oc.pgm.action.actions.SetVariableAction;
 import tc.oc.pgm.action.actions.SoundAction;
+import tc.oc.pgm.action.actions.TakePaymentAction;
 import tc.oc.pgm.api.feature.FeatureValidation;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.Filterables;
 import tc.oc.pgm.api.map.factory.MapFactory;
+import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.features.FeatureDefinitionContext;
 import tc.oc.pgm.features.XMLFeatureReference;
 import tc.oc.pgm.filters.Filterable;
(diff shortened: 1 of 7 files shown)


CoWinkKeyDinkInc:
I've tested this and it perfectly solves my use case.

```xml
<actions>
    <take-payment id="buy-eagle" currency="emerald" price="1" success-action="switch-eagle" fail-action="too-poor"/>
    <action id="switch-eagle" scope="player">
        <kit id="eagle"/>
    </action>
    <message id="too-poor"><text>You can't afford this!</text></message>
    <trigger filter="eagle-portal" action="buy-eagle" scope="player"/>
</actions>
<filters>
    <all id="redeem-eagle">
        <region id="eagle-portal"/>
        <!-- Use items unique to each class -->
        <carrying>
            <item>cooked fish</item>
        </carrying>
    </all>
</filters>
<portals>
    <!-- teleports players after they buy the eagle class (when they have item unique to eagle) -->
    <portal forward="redeem-eagle" destination="invaders-game-spawn" yaw="@-90"/>
</portals>
```

Pablete1234:
glad to hear this works for you, just as a minor thing:
> ```
>     <action id="switch-eagle" scope="player">
>         <kit id="eagle"/>
>     </action>
> ```
this isn't needed

you can just use `eagle` as a player-scoped action (because kits are a type of action), so:
`<take-payment id="buy-eagle" currency="emerald" price="1" success-action="eagle" fail-action="too-poor"/>`
would do the same in this case

cswhite2000 (approved):

PR #1306, opened 2024-03-04
Refactor Channels

dentmaged:
This PR refactors channels and adds some new features:
- A new `Channel` class that allows other plugins (such as [PGMSquads](https://github.com/PabloPintor/PGMSquads)) to implement custom channels
- A proper broadcasts API
- `/msg <target>` is now a channel, and therefore can be set to your default - messages will be sent to your selected target without running `/msg` or `/r`
- Suggestions when starting a message with a channel shortcut now work
- Messages sent in admin or global chat during a cycle will be visible to all viewers (at the moment, messages can be lost as some players are in old `Match`, and others in the new `Match`)


core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+import org.bukkit.command.CommandSender;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.event.ChannelMessageEvent;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.setting.SettingValue;
+import tc.oc.pgm.util.Players;
+import tc.oc.pgm.util.named.NameStyle;
+
+public interface Channel<T> {
```
TheMolkaPL, line 22: This could extend Adventure's `Audience`.

Pablete1234: given the extra overhead that a channel implies, it probably shouldn't.
You don't want someone treating this like it was just another casual audience, when it may not act like one at all (eg: requirement for a *target* to send within the channel). Effectively speaking the only channel that could properly implement audience would be global chat

Pablete1234, line 22: This API needs to be fully documented, explaining each method's purpose and params

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+
+  T getTarget(MatchPlayer sender, Map<String, ?> arguments);
+
+  Collection<MatchPlayer> getViewers(T target);
+
+  default Collection<MatchPlayer> getBroadcastViewers(T target) {
+    return getViewers(target);
+  }
+
+  default void sendMessage(ChannelMessageEvent<T> event) {}
```
Pablete1234, line 55:
```suggestion
  void sendMessage(ChannelMessageEvent<T> event);
```
either force the downstream to implement this, or have an *actual* default impl, but don't just leave an empty no-op method for something this critical.
I'd personally advocate for having a default impl here, just like broadcastMessage has

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+
+  default Component formatMessage(T target, @Nullable MatchPlayer sender, Component message) {
```
Pablete1234, line 57: i believe this method should not default to global-chat behavior, force downstream to implement it too

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+    return text()
+        .append(text("<", NamedTextColor.WHITE))
+        .append(sender.getName(NameStyle.VERBOSE))
+        .append(text(">: ", NamedTextColor.WHITE))
+        .append(message)
+        .build();
+  }
+
+  default void registerCommand(PaperCommandManager<CommandSender> manager) {
+    String name = getAliases()[0];
```
Pablete1234, line 68: if aliases is an empty array, this throws an out of bounds exception. Throw your own illegal argument exception instead

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+    String[] aliases = new String[getAliases().length - 1];
+    System.arraycopy(getAliases(), 1, aliases, 0, aliases.length);
```
Pablete1234, line 70: This needs to go into some util, you could also use `Arrays.copyOfRange`:
```suggestion
    String[] aliases = Arrays.copyOfRange(getAliases(), 1, getAliases().length);
```

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+                    .build())
+            .handler(
+                context -> {
+                  MatchPlayer sender =
+                      context.inject(MatchPlayer.class).orElseThrow(IllegalStateException::new);
+                  PGM.get().getChannelManager().process(this, sender, context.asMap());
+                }));
+  }
+
+  default Map<String, Object> processChatShortcut(MatchPlayer sender, String message) {
```
Pablete1234, line 98:
```suggestion
  default Map<String, Object> processShortcutMessage(MatchPlayer sender, String message) {
```
you're not processing the shortcut (with a chat), you're processing the message (with a shortcut).

Make sure the docs explain the rationale of why this is even needed cause it's nonobvious

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+    Map<String, Object> arguments = new HashMap<String, Object>();
+    if (message.length() == 1) {
+      PGM.get().getChannelManager().setChannel(sender, this);
+      return arguments;
```
Pablete1234, line 102:
```suggestion
      return Collections.emptyMap();
```

core/src/main/java/tc/oc/pgm/api/channels/Channel.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    arguments.put("message", message.substring(1).trim());
+    return arguments;
```
Pablete1234, line 106:
```suggestion
    return Collections.singletonMap("message", message.substring(1).trim());
```

core/src/main/java/tc/oc/pgm/api/event/ChannelMessageEvent.java
```diff
@@ -0,0 +1,72 @@
...
+  public Channel<T> getChannel() {
+    return channel;
+  }
+
+  @Nullable
+  public MatchPlayer getSender() {
+    return sender;
+  }
+
+  public void setSender(@Nullable MatchPlayer sender) {
```
Pablete1234, line 42: probably no reason to override the sender

core/src/main/java/tc/oc/pgm/api/integration/Integration.java
```diff
@@ -79,6 +91,14 @@ public static boolean setVanished(MatchPlayer player, boolean vanish, boolean qu
     return VANISH.get().setVanished(player, vanish, quiet);
   }
 
+  public static Set<Channel<?>> getRegisteredChannels() {
+    return Collections.unmodifiableSet(CHANNELS);
+  }
+
+  public static void finishRegisteringChannels() {
+    CHANNELS = null;
+  }
```
Pablete1234, line 100: I'd suggest going for `pollRegisteredChannels`, signifying that you're getting the list and setting it to null, both within the same method. Without that, you cannot guarantee that the state of what is used and what throws exception remains consistent.
Someone could call getRegisteredChannels, then do some stuff (causing registerChannel to be called), then call finishRegisteringChannels, and it would silently be hidden away that those channels didn't register because they were set after registering started.

core/src/main/java/tc/oc/pgm/api/event/ChannelMessageEvent.java
```diff
@@ -0,0 +1,72 @@
...
+import tc.oc.pgm.api.channels.Channel;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.event.PreemptiveEvent;
+
+public class ChannelMessageEvent<T> extends PreemptiveEvent {
+
+  private final Channel<T> channel;
+  private @Nullable MatchPlayer sender;
+  private final T target;
+  private final List<MatchPlayer> viewers;
```
Pablete1234, line 17: you probably should be able to set this, to be able to change who sees the message

core/src/main/java/tc/oc/pgm/api/event/ChannelMessageEvent.java
```diff
@@ -0,0 +1,72 @@
...
+  public ChannelMessageEvent(
+      Channel<T> channel,
+      @Nullable MatchPlayer sender,
+      T target,
+      Collection<MatchPlayer> viewers,
+      String message) {
+    this.channel = channel;
+    this.sender = sender;
+    this.target = target;
+    this.viewers = new ArrayList<MatchPlayer>(viewers);
```
Pablete1234, line 29: i'd avoid the array list, just keep the collection passed in

core/src/main/java/tc/oc/pgm/channels/AdminChannel.java
```diff
@@ -0,0 +1,104 @@
...
+  private static final Sound SOUND = sound(key("random.orb"), Sound.Source.MASTER, 1f, 0.7f);
+
+  @Override
+  public String getDisplayName() {
+    return "admin";
+  }
+
+  @Override
+  public String[] getAliases() {
+    return new String[] {"a"};
```
Pablete1234, line 44: make this a constant, we don't want java to allocate a new one every time this is called

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+
+  public void processChat(MatchPlayer sender, String message) {
+    if (message.isEmpty()) return;
+
+    Channel<?> channel = getSelectedChannel(sender);
+    Map<String, Object> arguments = new HashMap<String, Object>();
+    arguments.put("message", message);
+
+    Channel<?> shortcut = shortcuts.get(message.charAt(0));
+    if (shortcut != null && shortcut.canSendMessage(sender)) {
```
Pablete1234, line 77: feels like shortcut should be the first thing to process, and fallback to the getSelectedChannel stuff if none is found

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+      arguments = shortcut.processChatShortcut(sender, message);
+      channel = shortcut;
+    }
+
+    if (arguments.containsKey("message")) process(channel, sender, arguments);
```
Pablete1234, line 82: if there's no message, should it change your default chat? eg: typing `!` change default to global chat?
of course can only do that within the shortcut if, the no-shortcut version can't do that

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  public void process(Channel<?> channel, MatchPlayer sender, Map<String, ?> arguments) {
+    process0(calculateChannelRedirect(channel, sender), sender, arguments);
+  }
+
+  private <T> void process0(Channel<T> channel, MatchPlayer sender, Map<String, ?> arguments) {
```
Pablete1234, line 89: method names like process0 are terrible, try to find better names or simply merge this method with process and do calculateChannelRedirect as the first line of the method

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+
+    T target = channel.getTarget(sender, arguments);
+    Collection<MatchPlayer> viewers = channel.getViewers(target);
+
+    final AsyncPlayerChatEvent asyncEvent =
+        new AsyncPlayerChatEvent(
+            false,
+            sender.getBukkit(),
+            (String) arguments.get("message"),
+            viewers.stream().map(MatchPlayer::getBukkit).collect(Collectors.toSet()));
```
Pablete1234, line 101: how many times are we re-packaging those viewers? consider a transformed view of a list to cut on these

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+    CHAT_EVENT_CACHE.put(asyncEvent, true);
+    sender.getMatch().callEvent(asyncEvent);
+    if (asyncEvent.isCancelled()) return;
+
+    // The actual message is sent in sendMessage(ChannelMessageEvent)
+    final ChannelMessageEvent<T> channelEvent =
+        new ChannelMessageEvent<T>(channel, sender, target, viewers, asyncEvent.getMessage());
+    channel.sendMessage(channelEvent);
+    sender.getMatch().callEvent(channelEvent);
```
Pablete1234, line 110: what does this comment mean? the message shouldn't be sent until the event has finished (and allowed to be cancelled by whoever wanted to cancel it)

should properly send the message if it wasn't cancelled, *after* the call event

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  private Channel<?> calculateChannelRedirect(Channel<?> channel, MatchPlayer sender) {
+    if (Integration.isVanished(sender.getBukkit()) && !(channel instanceof AdminChannel)) {
+      if (!channel.supportsRedirect()) throw exception("vanish.chat.deny");
+      return adminChannel;
+    }
+
+    if (sender.getMatch().isFinished() || sender.getParty() instanceof Tribute) {
+      if (channel instanceof TeamChannel) return globalChannel;
```
Pablete1234, line 120: should probably also call to supportsRedirect before doing this

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+      return;
+    }
+
+    Component finalMessage =
+        event
+            .getChannel()
+            .formatMessage(event.getTarget(), event.getSender(), text(event.getMessage()));
+    Identity senderId = identity(event.getSender().getId());
+    event.getViewers().forEach(player -> player.sendMessage(senderId, finalMessage));
+  }
```
Pablete1234, line 150: cancellation should be handled by the manager, sending should be managed by the channel, but neither should be in response to the event

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+
+  @EventHandler
+  public void onPlayerTabComplete(PlayerChatTabCompleteEvent event) {
+    if (event.getChatMessage().trim().equals(event.getLastToken())) {
```
Pablete1234, line 154: `event.getChatMessage().contains(' ')`

You just need to check that the message has no spaces for  you to try to suggest adding something

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+      char first = event.getLastToken().charAt(0);
```
Pablete1234, line 155: pretty sure higher versions can get a tab completion with no chars, meaning, index out of bounds

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+      if (shortcuts.get(first) != null) {
```
Pablete1234, line 156:
```suggestion
      if (shortcuts.containsKey(first)) {
```

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+    try {
+      processChat(player, message);
+    } catch (TextException e) {
+      // Allow sub-handlers to throw command exceptions just fine
+      player.sendWarning(e);
+    }
+  }
+
+  private Channel<?> findChannelBySetting(SettingValue setting) {
+    for (Channel<?> channel : channels) if (setting == channel.getSetting()) return channel;
```
Pablete1234, line 196: did fmt agree to this one-liner? if it did, shame on it

make this occupy multiple lines

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+
+    return globalChannel;
```
Pablete1234, line 198: wasn't default team chat?

core/src/main/java/tc/oc/pgm/channels/ChannelManager.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  public void setChannel(MatchPlayer player, SettingValue value) {
+    selectedChannel.put(player.getBukkit(), findChannelBySetting(value));
+  }
+
+  public void setChannel(MatchPlayer player, Channel<?> channel) {
+    Channel<?> previous = selectedChannel.get(player.getBukkit());
+    selectedChannel.put(player.getBukkit(), channel);
```
Pablete1234, line 207:
```suggestion
    Channel<?> previous = selectedChannel.put(player.getBukkit(), channel);
```
Also, consider using UUIDs for maps instead of players or match players. Avoids mem leaks from holding onto entities you shouldn't

core/src/main/java/tc/oc/pgm/channels/GlobalChannel.java
```diff
@@ -0,0 +1,53 @@
...
+public class GlobalChannel implements Channel<Void> {
+
+  @Override
+  public String getDisplayName() {
+    return "global";
+  }
+
+  @Override
+  public String[] getAliases() {
+    return new String[] {"g", "all", "shout"};
```
Pablete1234, line 21: same as before, should keep this as a private static final array pre-defined

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+    return '@';
+  }
+
+  @Override
+  public MatchPlayer getTarget(MatchPlayer sender, Map<String, ?> arguments) {
+    MatchPlayer target = (MatchPlayer) arguments.get("target");
+    if (target == null) {
+      return getSelectedTarget(sender);
+    } else {
+      checkSettings(target, sender);
```
Pablete1234, line 72: checkAllowed? checkSettings is too generic, check for *what*??

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+          && !(manager.getSelectedChannel(sender) instanceof MessageChannel)) {
+        this.lastMessagedBy.put(
+            sender.getBukkit(), new MessageSenderIdentity(sender.getBukkit(), target.getBukkit()));
+      }
+
+      if (!sender.equals(getSelectedTarget(target))
+          && !(manager.getSelectedChannel(target) instanceof MessageChannel)) {
+        this.lastMessagedBy.put(
+            target.getBukkit(), new MessageSenderIdentity(target.getBukkit(), sender.getBukkit()));
+      }
```
Pablete1234, line 84: this should probably update as part of post-sending hooks (ie: same place ping sounds are placed at)

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+        || value.equals(SettingValue.SOUNDS_DM)) target.playSound(SOUND);
+
+    Bukkit.getPluginManager()
+        .callEvent(
+            new ChannelMessageEvent<MatchPlayer>(
+                event.getChannel(),
+                sender,
+                event.getTarget(),
+                Collections.singletonList(sender),
+                event.getMessage()));
```
Pablete1234, line 113: no, sending one message should not trigger a separate message event

it simply should be that sending itself happens within this method, and it just so happens that for DMs one message sends two chat messages (one with "From" and one with "To"). This is fine: in global chat one message sends N messages (to N recipients) with just one event.

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+                        TypeToken.get(MatchPlayer.class), "target")
+                    .withParser(
+                        manager
+                            .getParserRegistry()
+                            .createParser(
+                                TypeToken.get(MatchPlayer.class), ParserParameters.empty())
+                            .orElseThrow(IllegalStateException::new))
+                    .build())
+            .argument(
+                StringArgument.<CommandSender>builder("message")
```
Pablete1234, line 175: you can chain commands like these.

after the `.handler(context -> ...)` you can append the next .argument and another .handler. The first handler is used if the command is up to the first argument, the second is used when the command also has the 2nd param

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+                  if (target == null) throw exception("command.message.noReply", text("/msg"));
+                  context.set("target", target);
+
+                  PGM.get().getChannelManager().process(this, sender, context.asMap());
+                }));
+  }
+
+  @Override
+  public Map<String, Object> processChatShortcut(MatchPlayer sender, String message) {
+    if (message.length() == 1) throw usage("/msg <player> [message]");
```
Pablete1234, line 222: if it's a shortcut, it probably isn't a /command syntax

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+    Map<String, Object> arguments = new HashMap<String, Object>();
+
+    int spaceIndex = message.indexOf(" ");
```
Pablete1234, line 225:
```suggestion
    int spaceIndex = message.indexOf(' ');
```

core/src/main/java/tc/oc/pgm/channels/MessageChannel.java
```diff
@@ -0,0 +1,292 @@
...
+
+    return arguments;
+  }
+
+  private void checkSettings(MatchPlayer target, MatchPlayer sender) {
+    if (sender.equals(target)) throw exception("command.message.self");
+
+    SettingValue option = sender.getSettings().getValue(SettingKey.MESSAGE);
+    if (option.equals(SettingValue.MESSAGE_OFF))
+      throw exception("command.message.disabled", text("/toggle dm", NamedTextColor.RED));
```
Pablete1234, line 248: could be made into a clickable text to toggle

core/src/main/java/tc/oc/pgm/goals/TouchableGoal.java
```diff
@@ -192,15 +195,13 @@ protected void sendTouchMessage(@Nullable ParticipantState toucher, boolean incl
     Component message = getTouchMessage(toucher, false);
     Audience.console().sendMessage(message);
 
-    if (!showEnemyTouches()) {
-      message = text().append(toucher.getParty().getChatPrefix()).append(message).build();
-    }
-
-    for (MatchPlayer viewer : getMatch().getPlayers()) {
-      if (shouldShowTouched(toucher.getParty(), viewer.getParty())
```
Pablete1234, line 200: Pretty sure this is a behavior change with matches with +2 teams. Before it would show this to all players in a team that had touched, while now it will only show to the team that touched.
Eg, a 4-team DTC map:
Blue touches red core <- Blue is notified
Green touches red core <- Before, both green and blue are notified, now only green is.

Pablete1234 (changes requested):

PR #1308, opened 2024-03-11
Add Tracking Compass Module

cswhite2000:
Adds a module for tracking players with compasses.
Tracking other things is to be added in later PRs.
This acts on all compasses

Example XML for an FFA:
```xml
<compass show-distance="true">
        <player filter="always" show-player="true"/>
</compass>
```

thingy

Other things this supports:
- Custom Names:
```xml
<compass>
    <player filter="always" name="Wool Carrier"/>
</compass>
```
- Multiple Options:
```xml
<compass order="FIRST_DEFINED">
    <player filter="wool-carrier" name="Wool Carrier"/>
    <player filter="enemy" show-player="true" name="Nearest Enemy"/>
</compass>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/compass/CompassMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/compass/CompassMatchModule.java
@@ -1,0 +1,190 @@
+package tc.oc.pgm.compass;
+
+import static net.kyori.adventure.text.Component.text;
+import static net.kyori.adventure.text.Component.translatable;
+import static net.kyori.adventure.text.format.Style.style;
+import static tc.oc.pgm.util.text.TextTranslations.translateLegacy;
+
+import com.google.common.collect.ImmutableList;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Stream;
+import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.TextComponent;
+import net.kyori.adventure.text.format.NamedTextColor;
+import org.bukkit.Material;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.Listener;
+import org.bukkit.event.entity.ItemSpawnEvent;
+import org.bukkit.event.inventory.InventoryMoveItemEvent;
+import org.bukkit.event.player.PlayerItemHeldEvent;
+import org.bukkit.event.player.PlayerQuitEvent;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.PlayerInventory;
+import org.bukkit.inventory.meta.ItemMeta;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchModule;
+import tc.oc.pgm.api.match.Tickable;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.time.Tick;
+import tc.oc.pgm.events.PlayerResetEvent;
+import tc.oc.pgm.spawns.events.ParticipantKitApplyEvent;
+
+public class CompassMatchModule implements MatchModule, Tickable, Listener {
+
+  private static final long REFRESH_TICKS = 20;
+  private static final long REFRESH_TICKS_UNFOCUSED = 20 * 10;
+  private final Match match;
+  private final Map<UUID, Long> lastRefresh;
+  private final ImmutableList<CompassTarget> compassTargets;
+  private final OrderStrategy orderStrategy;
+  private final boolean showDistance;
+
+  public CompassMatchModule(
+      Match match,
+      ImmutableList<CompassTarget> compassTargets,
+      OrderStrategy orderStrategy,
+      boolean showDistance) {
+    this.match = match;
+    this.lastRefresh = new ConcurrentHashMap<>();
+    this.compassTargets = compassTargets;
+    this.orderStrategy = orderStrategy;
+    this.showDistance = showDistance;
+  }
+
+  @Override
+  public synchronized void tick(Match match, Tick tick) {
+    for (Map.Entry<UUID, Long> lastRefreshEntry : lastRefresh.entrySet()) {
+      long ticksSince = tick.tick - lastRefreshEntry.getValue();
+      if (ticksSince >= REFRESH_TICKS) {
+        UUID uuid = lastRefreshEntry.getKey();
+        MatchPlayer player = this.match.getPlayer(uuid);
+        if (player == null || player.getInventory() == null || !player.isAlive()) {
+          lastRefresh.remove(player.getId());
+          continue;
+        }
+        if (!Material.COMPASS.equals(player.getInventory().getItemInHand().getType())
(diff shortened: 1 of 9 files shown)


Pugzy:
I'd remove the bold from the "Tracking: ......" item name, there's no reason for it to be bold.

core/src/main/java/tc/oc/pgm/compass/CompassMatchModule.java
```diff
@@ -0,0 +1,177 @@
...
+                        "compass.tracking.distance",
+                        style(NamedTextColor.AQUA, TextDecoration.BOLD),
+                        text((int) compassResult.get().getDistance())));
+          }
+
+          itemNameComponent = builder.build();
+        } else {
+          itemNameComponent =
+              translatable(
+                  "compass.tracking.unknown", style(NamedTextColor.WHITE, TextDecoration.BOLD));
```
Pablete1234: As pugzy mentioned, i think this is too much bold text. The point of bold is to emphasize or bring attention to some specific part, if all of it is bold then nothing is emphasized

core/src/main/java/tc/oc/pgm/compass/CompassMatchModule.java
```diff
@@ -0,0 +1,177 @@
...
+    PlayerInventory inventory = player.getInventory();
+    if (inventory == null) {
+      return;
+    }
+    ItemStack[] contents = inventory.getContents();
+    if (contents == null) {
+      return;
+    }
+
+    for (ItemStack content : contents) {
```
Pablete1234: Realistically updating the text isn't required unless it's held; you can't see the name anyways
It's a trivial optimization that will save on *a lot* of text re-renders since most of the times players won't be holding the compass. You probably do need to do an explicit update if they decide to hold it tho (ie: don't wait on the full refresh ticks for it to update once they select it, do one immediate re-render, then every 2s)

core/src/main/java/tc/oc/pgm/compass/OrderStrategy.java
```diff
@@ -0,0 +1,6 @@
+package tc.oc.pgm.compass;
+
+public enum OrderStrategy {
+  FIRST_DEFINED,
```
Pablete1234: i don't really like this "first defined" name, makes it sound not like an order strategy at all, because you're not going to pick the first one, you'll just happen to pick them in the order they're defined, whichever one happens to match first. maybe "as defined", or maybe we can come up with something better

core/src/main/java/tc/oc/pgm/compass/CompassTargetResult.java
```diff
@@ -0,0 +1,32 @@
...
+  public double getDistance() {
+    return distance;
+  }
+
+  public Component getComponent() {
+    return component;
+  }
+
+  public int compareTo(CompassTargetResult other) {
+    return (int) (distance - other.distance);
```
Pablete1234:
```suggestion
    return Double.compare(this.distance, other.distance);
```

core/src/main/java/tc/oc/pgm/compass/CompassModule.java
```diff
@@ -0,0 +1,71 @@
...
+      for (Element compassRoot : doc.getRootElement().getChildren("compass")) {
+        orderStrategy =
+            XMLUtils.parseEnum(
+                Node.fromAttr(compassRoot, "order"), OrderStrategy.class, orderStrategy);
+        showDistance =
+            XMLUtils.parseBoolean(Node.fromAttr(compassRoot, "show-distance"), showDistance);
+        for (Element compassTarget : compassRoot.getChildren()) {
+          if (parser.isTarget(compassTarget)) {
+            compassTargets.add(parser.parseCompassTarget(compassTarget));
+          }
```
Pablete1234:
```suggestion
        compassTargets.add(parser.parseCompassTarget(compassTarget));
```
Let the parser complain that it's not a valid target

core/src/main/java/tc/oc/pgm/compass/CompassModule.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+      }
+
+      return new CompassModule(compassTargets.build(), orderStrategy, showDistance);
```
Pablete1234, line 71: Avoid creating a CompassModule as a whole if there is no `<compass>` in the xml.
Throw an invalid XML exception if there's a compass tag but `compassTargets` is empty

core/src/main/java/tc/oc/pgm/compass/targets/PlayerCompassTarget.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Override
+  public Filter getFilter() {
+    return filter;
+  }
+
+  @Override
+  public Component getName(Match match, MatchPlayer player) {
+    return showPlayerName ? player(player, NameStyle.FANCY) : name;
```
Pablete1234: pretty sure this shouldn't be the behavior, it's not an either-or kinda deal, should be possible to show both

name="FlagCarrier" & show name: `Flag carrier: Pablete1234 (34m)`
name="FlagCarrier": `Flag carrier (34m)`
just show name: `Pablete1234 (34m)`
in all of those the (34m) at the end is optional

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/compass/CompassMatchModule.java
cswhite2000: I'm fine with changing this

core/src/main/java/tc/oc/pgm/compass/CompassMatchModule.java
cswhite2000: That is a very good point. Will do

core/src/main/java/tc/oc/pgm/compass/OrderStrategy.java
cswhite2000: maybe `DEFINITION_ORDER`?

core/src/main/java/tc/oc/pgm/compass/CompassTargetResult.java
cswhite2000: will do

core/src/main/java/tc/oc/pgm/compass/CompassModule.java
cswhite2000: makes sense. Will do

core/src/main/java/tc/oc/pgm/compass/CompassModule.java
cswhite2000: will do

core/src/main/java/tc/oc/pgm/compass/targets/PlayerCompassTarget.java
cswhite2000: that's a good point

core/src/main/java/tc/oc/pgm/compass/OrderStrategy.java
Pablete1234: sounds good

Pablete1234 (approved):

arcadeboss:
Thank you!

PR #1311, opened 2024-04-27
Username resolving rework

Pablete1234:
Rewrites how username resolving works in pgm to make it more reliable. At the moment `Unknown` prevails and is shown way too often, this rewrite fixes a big amount of hidden bugs and quirks with the original system like, to name a few:
- SQL was only used to resolve usernames after you called getNameLegacy
- API based resolution was queued up as soon as you asked for a username
  -  We shouldn't need to hit the API is local SQL resolves correctly!!
- API based resolution queue only processed after /newmaps, or on match end due to stats side-effect
- SQL was not batched, meaning 100s of queries to retrieve all author names
- SQL would query, then save itself back into the database with updated expiry

The new resolution system isn't flawless, but it's much better than the predecessor.

There's a series of `UsernameResolver`s, they can work in batch mode (in which case they no-op until the end of the batch) or in normal mode. They return a `CompletableFuture` with the result of the resolution. You can setup multiple of them to work in a chain, where the first resolver is tried, if it gives a good answer that's returned via the future, otherwise the next resolver is tried, until one of them gives a good answer, or we run out of resolvers. They can also chain in batching mode where the first will resolve everything (sending downstream the not-well-resolved ones), then the 2nd tries to resolve the remaining batch, and so on.

The main caveat that still seems not to work too well is bukkit's username cache, which seems to either not keep users or say their last played time was 0. Using vanilla's user cache via nmshacks seems to work around the issue and allow for "instant resolving" to work just fine.


util/src/main/java/tc/oc/pgm/util/usernames/BukkitUsernameResolver.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return UsernameResponse.empty();
+  }
+
+  private static class SyncExecutor implements Executor {
+
+    @Override
+    public void execute(@NotNull Runnable command) {
+      if (Bukkit.isPrimaryThread()) command.run();
+      else Bukkit.runOnMainThread(PGM, false, command);
```
cswhite2000: This won't work on non sportpaper
 https://github.com/Electroid/SportPaper/blob/main/patches/api/0080-Mojang-task-queue-access.patch#L27

cswhite2000 (changes requested):

util/src/main/java/tc/oc/pgm/util/usernames/BukkitUsernameResolver.java
Pablete1234: modified to use the NMSHacks method instead

cswhite2000 (approved):

PR #1318, opened 2024-05-12
Cleanup minor map & default repo issues

Pablete1234:
Fixes a few minor issues with current map loading, especially when default configs are used for first-time PGM users:
 - Hardcoded default repo used `default-maps` while the config had a commented out `maps-default`, both of those now align
 - Improved logging around timing out while loading maps, which can happen if you configure a big git repo on a slow connection, or have loads of usernames to resolve first time
 - Fixed links to map xml in `/map` having incorrect paths if server runs on windows

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMPlugin.java
+++ b/core/src/main/java/tc/oc/pgm/PGMPlugin.java
@@ -173,23 +173,13 @@ public class PGMPlugin extends JavaPlugin implements PGM, Listener {
 
     datastore = new CacheDatastore(datastore);
 
-    try {
-      mapLibrary.loadNewMaps(false).get(30, TimeUnit.SECONDS);
-    } catch (ExecutionException | InterruptedException | TimeoutException e) {
-      e.printStackTrace();
-    }
-
-    if (!mapLibrary.getMaps().hasNext()) {
+    if (!loadInitialMaps()) {
+      logger.warning("No maps found, adding default repository as a fallback.");
       PGMConfig.registerRemoteMapSource(mapSourceFactories, PGMConfig.DEFAULT_REMOTE_REPO);
-      try {
-        mapLibrary.loadNewMaps(false).get(30, TimeUnit.SECONDS);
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -394,6 +384,17 @@ private void registerListeners() {
     registerEvents(new JoinLeaveAnnouncer(matchManager));
   }
 
+  private boolean loadInitialMaps() {
+    try {
+      mapLibrary.loadNewMaps(false).get(300, TimeUnit.SECONDS);
+    } catch (TimeoutException e) {
+      getLogger().warning("Loading all maps took >30s, other maps will keep loading async.");
```
TTtie, line 391: Possible oversight: is the new map load timeout supposed to be 5 minutes, or 30 seconds?

```suggestion
      getLogger().warning("Loading all maps took >5m, other maps will keep loading async.");
```

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -394,6 +384,17 @@ private void registerListeners() {
     registerEvents(new JoinLeaveAnnouncer(matchManager));
   }
 
+  private boolean loadInitialMaps() {
+    try {
+      mapLibrary.loadNewMaps(false).get(300, TimeUnit.SECONDS);
```
TTtie: Possible oversight: 5 minutes or 30 seconds?

```suggestion
      mapLibrary.loadNewMaps(false).get(30, TimeUnit.SECONDS);
```

TTtie (changes requested):

core/src/main/java/tc/oc/pgm/PGMPlugin.java
Pablete1234: the oversight is actually the 300, it was 30 and supposed to stay at 30, added a 0 for debugging purposes

cswhite2000 (approved):

PR #1321, opened 2024-05-19
Consistently use JetBrains' annotations

calcastor:
Swap remaining Checker Framework annotations to JetBrains' for consistency


Pablete1234:
All of these are related to cloud, because cloud itself uses or used to use checkerframework annotations. It's probably best to do this inside the cloud 2 PR

PR #1323, opened 2024-05-25
Switch to Palantir via Spotless for formatting

calcastor:
This PR switches away from our previous formatter to Palantir via Spotless, using Palantir's Google formatter.

This format avoids butchering lambdas and also also makes improvements for code review as noted in the project's [README](https://github.com/palantir/palantir-java-format#motivation--examples).

This requires building with a more modern version of Java than 8, however PGM will still run on 8 at the moment.

It's configured to ratchet against `origin/dev` to avoid reformatting the whole project at once.


Pablete1234 (dismissed):

Pablete1234:
Let's first merge other PRs, as otherwise it would force those already open PRs to reformat

.github/workflows/build.yml
```diff
@@ -11,11 +11,13 @@ jobs:
     - id: checkout
       name: Checkout
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
calcastor, line 15: I'm not sure if we have concerns over storage space or otherwise with this.

cswhite2000 (approved):

PR #1328, opened 2024-06-10
Make PGM depend on intersection jar, migrate deps to platforms

Pablete1234:
This modifies PGM's main bukkit API dependency to a special "intersected" jar that only contains classes/methods/fields that exist all the way from 1.8.8 to 1.20.6 bukkit/spigot APIs.

It means that now PGM relies only on classes & methods from the bukkit/spigot api that have existed and continue to exist from 1.8.8 to 1.20.6 (or later versions if we migrate later), and so prevents a developer mistake of using something that breaks future version compatibility, which there were a bunch of instances of. This PR also takes care of abstracting all of those instances into platform-specific interfaces, and provides implementations for sportpaper. As far as i could tell, sportpaper is working as it was before. Spigot 1.8-1.12 compat could've been compromised but i don't think that was great to begin with either. Future PRs will take care of implementing future platforms.

This jar was built with a tool that can be found at https://github.com/Pablete1234/JarIntersector, and the jar result has been put in repo.pgm.fyi for the convenience of anyone trying to build PGM.


platform/platform-shared/pom.xml
```diff
@@ -21,6 +21,18 @@
       <version>1.8.8-R0.1-SNAPSHOT</version>
       <scope>provided</scope>
     </dependency>
+<!--    <dependency>-->
```
cswhite2000: Is this supposed to be a reminder for something?

util/src/main/java/tc/oc/pgm/util/material/Materials.java
```diff
@@ -1,19 +1,121 @@
...
+  Material GOLD_PICKAXE = parse("GOLD_PICKAXE", "GOLDEN_PICKAXE");
+  Material WOOD_SHOVEL = parse("WOOD_SPADE", "WOODEN_SHOVEL");
+  Material STONE_SHOVEL = parse("STONE_SPADE", "STONE_SHOVEL");
+  Material GOLD_SHOVEL = parse("GOLD_SPADE", "GOLDEN_SHOVEL");
+  Material IRON_SHOVEL = parse("IRON_SPADE", "IRON_SHOVEL");
+  Material DIAMOND_SHOVEL = parse("DIAMOND_SPADE", "DIAMOND_SHOVEL");
+  Material WOOD_HOE = parse("WOOD_HOE", "WOODEN_HOE");
+  Material GOLD_HOE = parse("GOLD_HOE", "GOLDEN_HOE");
+
+  Material SIGN_POST = parse("SIGN_POST", "LEGACY_SIGN_POST");
```
cswhite2000, line 32: This, or the logic surrounding this, will need modification when we go for version support. 
Example: `ACACIA_SIGN != LEGACY_SIGN_POST`

This is however fine for now.

util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
```diff
@@ -1,18 +1,53 @@
...
+    return Collections.emptyList();
+  }
+
+  @Override
+  public ChunkData generateChunkData(
+      World world, Random random, int chunkX, int chunkZ, BiomeGrid biome) {
+    ChunkData chunkData = super.createChunkData(world);
+
+    // For everyblock in the chunk set the biome to plains
+    for (int x = 0; x < 16; x++) {
```
cswhite2000, line 42: Nitpick: we can skip this for 1.8 / sportpaper. I believe this only becomes necessary as of 1.13.

Also this method not returning null will bypass the call to `generate(`. The underlying code looks similar performance wise however.

cswhite2000 (dismissed):
A few minor comments. Nothing blocking.

platform/platform-shared/pom.xml
Pablete1234: This was something i used during development to have the IDE find me 1.20.6 classes to check what was diff
Can remove

util/src/main/java/tc/oc/pgm/util/material/Materials.java
Pablete1234: Yeah, all of the LEGACY_* material names should be considered placeholders, and will need to be revisited for future version support

cswhite2000 (approved):
Took a while to review. I have no concerns with this. Approved 👍

PR #1331, opened 2024-06-15
Booting 1.20.6 support!

Pablete1234:
After these changes, pgm is able to boot up and is playable in 1.20.6 paper!

## Feature set

Most if not all PGM features have been reimplemented in the newer features, with some exceptions:
 - CanPlaceOn/CanBreak
   -  Removed in 1.20.6 spigot/paper due to changes in vanilla.
   -  Hoping to add them back when they re-add them in 1.21 or something.
 - Knockback reduction kit
   - This relied on a sportpaper only patch to set a multiplier to the kb taken
   - Can maybe be replaced with a knockback resistance attribute, since i belive that's what it does in newer versions (in older versions it was a % chance of receiving 0 kb, instead of a multiplier on kb taken)
   - If not, paper now launches an event when KB is taken and we can modify it there
 - Extra data in ping packet
   - This was also a sportpaper exclusive, we can't freely modify the ping packet to add extra data in paper
 - Observers seeing invisible players
   - sportpaper exclusive. Obs will not be able to see invisible players. 
   - Maybe can get this back by modifying packets in protocol lib.
 - Hide all potion particles for players
   - sportpaper feature that PGM enabled by default on all kits who didn't explicitly disable it, only ghost squadron really disabled it.
   - Modified it so instead it will make kit potion particles ambient, no need to fully hide them.
   - Can probably get this fully back by modifying packets in protocol lib, but is a good candidate for an actual behavior change in legacy too.
 - Can't track entities spawned by dispensers (BlockDispenseEntityEvent)
   - If the spawned entity (arrow, spawn egg, tnt, etc) deals damage to someone, it doesn't count as yours for having placed the dispenser
   - This is likely very hard to work around, if at all possible
 - Can't track who caused an explosion (eg: tnt or creeper) (ExplosionPrimeByEntityEvent)
   - When you shoot a tnt with flame, or light a creeper with flint & steep, it can't count that explosion as yours
   - When a TNT you lit, makes another tnt go off, it can't count it as yours either
   - This is likely very hard to work around, if at all possible
 - EntityExtinguishEvent is missing, this should only be an internal technical issue as certain state doesn't get cleaned up, but shouldn't see feature differences.

Some stuff currently known broken:
 - No inventory editing in observers (paper bug)
 - Max health kits seem not to work

## Performance
While we can't test in-game performance until we run this in a production environment with real players, so far map loading seems not to be significantly slower than original sportpaper, in fact it was faster. I've backported this speedup to sportpaper impl too in this pr and the load times for the 1900 maps look like this:
Sportpaper 1.8 with added optz
image
Paper 1.20.6:
image

Overall modern has a couple second overhead due to having to load CraftLegacy, but this is a one-time cost on startup and after that seems not to affect pgm negatively.

## Stability

This PR has gotten up to the point i'd consider viable playing in it, all major features have been implemented and i've done a fair share of testing. That's not to say there won't still be a bunch more issues to find in the future


core/src/main/java/tc/oc/pgm/command/util/CommandGraph.java
```diff
@@ -72,20 +71,18 @@ protected CommandManager<CommandSender> createCommandManager() {
     manager.settings().set(ManagerSetting.LIBERAL_FLAG_PARSING, true);
 
     // Register Brigadier mappings
-    if (manager.hasCapability(CloudBukkitCapabilities.BRIGADIER)) manager.registerBrigadier();
+    // if (manager.hasCapability(CloudBukkitCapabilities.NATIVE_BRIGADIER))
+    // manager.registerBrigadier();
```
cswhite2000: commented out code

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
```diff
@@ -197,16 +198,16 @@ public void show(MatchPlayer player) {
     }
 
     private void spawn(Player player, FakeEntity entity) {
-      location().ifPresent(l -> entity.spawn(l).send(player));
+      location().ifPresent(l -> entity.spawn(l.add(0, 3, 0)).send(player));
```
cswhite2000: This seems hazardous given that `location.add(` modifies the underlying value. Please check if this is a concern here

core/src/main/java/tc/oc/pgm/regions/RegionMatchModule.java
```diff
@@ -404,19 +405,13 @@ private ParticipantState getActor(BlockTransformEvent event) {
...
-  }
-
-  private static Material getHangingType(Hanging hanging) {
+  private static MaterialData getHangingType(Hanging hanging) {
     if (hanging instanceof Painting) {
-      return Material.PAINTING;
+      return MaterialData.of(Material.PAINTING);
     } else if (hanging instanceof ItemFrame) {
-      return Material.ITEM_FRAME;
+      return MaterialData.of(Material.ITEM_FRAME);
```
cswhite2000: Probably need to handle [glow item frames](https://minecraft.wiki/w/Glow_Item_Frame)

platform/platform-1_20_6/src/main/java/tc/oc/pgm/platform/v1_20_6/ModernPlayerUtils.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Override
+  public void setAffectsSpawning(Player player, boolean affectsSpawning) {
+    player.setAffectsSpawning(affectsSpawning);
+  }
+
+  @Override
+  public void setCollidesWithEntities(Player player, boolean collides) {
+    player.setCollidable(collides);
+    // player.setNoPhysics(true);
```
cswhite2000: remove

platform/platform-1_20_6/src/main/java/tc/oc/pgm/platform/v1_20_6/material/ModernColorUtils.java
```diff
@@ -0,0 +1,142 @@
...
+    registerColorable("DYE", Material.LEGACY_INK_SACK);
+    registerColorable("CARPET", Material.LEGACY_CARPET);
+    registerColorable("TERRACOTTA", Material.LEGACY_STAINED_CLAY);
+    registerColorable("GLAZED_TERRACOTTA");
+    registerColorable("STAINED_GLASS", Material.LEGACY_STAINED_GLASS);
+    registerColorable("STAINED_GLASS_PANE", Material.LEGACY_STAINED_GLASS_PANE);
+
+    registerColorable("BANNER", Material.LEGACY_BANNER, Material.LEGACY_STANDING_BANNER);
+    registerColorable("WALL_BANNER", Material.LEGACY_WALL_BANNER);
+    // TODO: Any others?
```
cswhite2000: We could consider some sort of automatic detection based off of colors in the material names

cswhite2000 (commented):
I've gone through about half the files. Will check the other half later.

core/src/main/java/tc/oc/pgm/flag/LegacyFlagBeamMatchModule.java
Pablete1234: It isn't because it is always guaranteed to be a clone, regardless i'm cleaning up a bit of this to make it more explicit

platform/platform-1_20_6/src/main/java/tc/oc/pgm/platform/v1_20_6/packets/ModernEntityPackets.java
```diff
@@ -0,0 +1,143 @@
...
+public class ModernEntityPackets implements EntityPackets {
+
+  private static final EntityDataAccessor<Byte> ENTITY_FLAGS =
+      new EntityDataAccessor<>(0, EntityDataSerializers.BYTE);
+  private static final EntityDataAccessor<Integer> ENTITY_AIR =
+      new EntityDataAccessor<>(1, EntityDataSerializers.INT);
+
+  @Override
+  public Packet spawnArmorStand(Location loc, int entityId, Vector velocity) {
+    /*
```
cswhite2000: commented out code

platform/platform-1_20_6/src/main/java/tc/oc/pgm/platform/v1_20_6/packets/PlPacket.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Override
+  public void send(Player viewer) {
+    PL.sendServerPacket(viewer, packet);
+  }
+
+  @Override
+  public void sendToViewers(Entity entity, boolean excludeSpectators) {
+    if (excludeSpectators)
+      throw new UnsupportedOperationException("Protocol lib packets can't send to all except N");
```
cswhite2000: incomplete message

pom.xml
```diff
@@ -2,8 +2,8 @@
     <modelVersion>4.0.0</modelVersion>
 
     <properties>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>21</maven.compiler.source>
```
cswhite2000, line 5: Please update the contributing readme as well https://github.com/PGMDev/PGM/blob/dev/docs/CONTRIBUTING.md?plain=1#L30

util/src/main/java/tc/oc/pgm/util/material/Materials.java
```diff
@@ -68,48 +68,66 @@ static Material parse(String... names) {
...
-          WOOD_HOE,
-          STONE_HOE,
-          GOLD_HOE,
-          IRON_HOE,
-          DIAMOND_HOE,
-          BOW,
-          FLINT_AND_STEEL,
-          SHEARS,
-          STICK);
+  Set<Material> WEAPONS = ImmutableSet.of(
```
cswhite2000: This will probably need to go platform specific. For example tridents, or netherite tools. Fine with a TODO here given this should only effect Death messages

cswhite2000 (commented):
And that's the rest of the files

util/src/main/java/tc/oc/pgm/util/material/Materials.java
Pablete1234: I think quite a few of the methods in materials could go platform-specific, but i'm keeping them as is for the time being, future improvements i think can go in a future PR for the sake of merging this colossal one already

Pablete1234: Made it not platform specific but still should add the newer version items in

platform/platform-1_20_6/src/main/java/tc/oc/pgm/platform/v1_20_6/material/ModernColorUtils.java
Pablete1234: For the time being i've added all i could find, can elaborate on it further later

cswhite2000 (dismissed):

cswhite2000 (approved):

PR #1334, opened 2024-06-22
Update modern PGM to 1.21!

Pablete1234:
Moves classes from the package/module `1_20_6` to just `modern`, as it is expected that this will probably keep up with modern in the future when breaking changes occur.

Additionally has stopped using an intersection jar of spigot 1.8 to paper 1.20.6, instead relies on intersection of sportpaper 1.8 to paper 1.21. This adds a few more opportunities to merge code in the future, mainly thinking about attributes here.


cswhite2000 (approved):
I have no concerns with this 👍

PR #1335, opened 2024-06-22
Drop platform-specific attribute impl

Pablete1234:
By dropping pgm's Attribute and using bukkit's, we directly gain support for all new attributes in modern server versions. This means maps made to play in new versions can use attributes to change player scale or reach, added in newer versions


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/ModernInventoryUtil.java
```diff
@@ -60,12 +59,10 @@ public void copyAttributeModifiers(ItemMeta destination, ItemMeta source) {
 
   @Override
   public void applyAttributeModifiers(
-      SetMultimap<tc.oc.pgm.util.attribute.Attribute, AttributeModifier> attributeModifiers,
+      SetMultimap<org.bukkit.attribute.Attribute, AttributeModifier> attributeModifiers,
```
cswhite2000: nitpick: import this

core/src/main/java/tc/oc/pgm/inventory/ViewInventoryMatchModule.java
```diff
@@ -177,7 +174,10 @@ public void updateMonitoredClick(final InventoryClickEvent event) {
           continue invLoop;
 
         for (int i = 0; i < inventory.getViewers().size(); i++) {
-          if (!inventory.getViewers().get(i).equals(tracker.getWatched().getViewers().get(i))) {
+          if (!inventory
```
cswhite2000, line 177: Not related to your changes but we should see about simplifying this in the future

cswhite2000 (dismissed):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/ModernInventoryUtil.java
Pablete1234: nice catch

cswhite2000 (approved):

PR #1346, opened 2024-06-30
Add velocity action, teleport action, and related variable

### LINKED ISSUE #1133: Velocity Actions
Currently, velocity can only be applied with `<apply>`s. I think it makes sense as a player-scoped action.

It would also be interesting if velocities could be applied relative to the player's yaw and optionally pitch, although I am unsure how this should be represented or how the math would work here if it were relative.

cswhite2000:
Resolves https://github.com/PGMDev/PGM/issues/1133

Adds a velocity & teleport action
also adds a player location variable

example:
```xml
<actions>
    <trigger scope="player">
        <filter>
            <carrying ignore-metadata="true">
                <item material="lava bucket"/>
            </carrying>
        </filter>
        <action scope="player">
		<velocity x="0" y="1" z="0"/>
		<teleport x="x" y="y+10" z="z"/>
        </action>
    </trigger>
</actions>
<variables>
	<player-location id="x" component="x"/>
	<player-location id="y" component="y"/>
	<player-location id="z" component="z"/>
</variables>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -12,6 +12,7 @@ import java.text.DecimalFormat;
 import java.text.NumberFormat;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import net.kyori.adventure.sound.Sound;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.title.Title;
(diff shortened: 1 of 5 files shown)


Pablete1234 (approved):
There's one thing that isn't too well handled and it's teleporting looking at a certain block, you could probs do it by calculating pitch/yaw with trig, but maybe we can offer an "angle" x,y,z in the teleport that will cause the player to look at that direction. But we can do that at a later time.

cswhite2000:
That's not a bad idea. But yeah we can add it at a later time.

PR #1349, opened 2024-07-03
Add a paste-structure action

cswhite2000:
Adds a paste structure action. Accepts variable x y z coordinates to allow relative pasting.

Example usecase would be a consumable item that builds a little castle around you.

Example xml:

```xml
<actions>
    <trigger scope="player">
        <filter>
            <carrying ignore-metadata="true">
                <item material="lava bucket"/>
            </carrying>
        </filter>
        <action scope="player">
                <message text="`f Carrying `6lava bucket`f."/>
                <paste-structure x="x+3" y="y" z="z+3" structure="structy"/>
        </action>
    </trigger>
</actions>
<structures>
    <structure clear="false" id="structy"><region><cuboid min="72,15,75" size="3,3,5"/></region></structure>
</structures>
<variables>
        <timelimit id="tl_2"/>
        <player-location id="x" component="x"/>
        <player-location id="y" component="y"/>
        <player-location id="z" component="z"/>
</variables>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -25,6 +25,7 @@ import tc.oc.pgm.action.actions.ExposedAction;
 import tc.oc.pgm.action.actions.FillAction;
 import tc.oc.pgm.action.actions.KillEntitiesAction;
 import tc.oc.pgm.action.actions.MessageAction;
+import tc.oc.pgm.action.actions.PasteStructureAction;
 import tc.oc.pgm.action.actions.ReplaceItemAction;
 import tc.oc.pgm.action.actions.ScopeSwitchAction;
 import tc.oc.pgm.action.actions.SetVariableAction;
@@ -48,6 +49,7 @@ import tc.oc.pgm.regions.BlockBoundedValidation;
 import tc.oc.pgm.regions.RegionParser;
 import tc.oc.pgm.shops.ShopModule;
 import tc.oc.pgm.shops.menu.Payable;
+import tc.oc.pgm.structure.StructureDefinition;
 import tc.oc.pgm.util.Audience;
 import tc.oc.pgm.util.MethodParser;
 import tc.oc.pgm.util.MethodParsers;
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/action/ActionModule.java
```diff
@@ -48,7 +49,7 @@ public static class Factory implements MapModuleFactory<ActionModule> {
 
     @Override
     public Collection<Class<? extends MapModule<?>>> getWeakDependencies() {
-      return ImmutableList.of(VariablesModule.class);
+      return ImmutableList.of(VariablesModule.class, StructureModule.class);
```
Pablete1234: You should not need this dependency. if you just use references you don't need to force load order

core/src/main/java/tc/oc/pgm/structure/StructureDefinition.java
```diff
@@ -48,10 +50,22 @@ public Vector getOrigin() {
...
+  @Override
+  public void place(BlockVector offset) {
+    throw new UnsupportedOperationException("Cannot call place from StructureDefinition!");
+  }
+
+  @Override
+  public void placeAbsolute(Match match, Vector vector) {
+    WorldStructure worldStructure = match.getFeatureContext().get(getId(), WorldStructure.class);
+    worldStructure.placeAbsolute(match, vector);
+  }
```
Pablete1234: i'm not sure structure definition should implement structure, often times there's a clear cut difference between the match-scoped object and the map-scoped object (definition). 

Sometimes the lines are a bit blurrier for objects that don't have separate match and map scopes, (eg: filters, actions, or regions, who are stateless), but do note that stateful objects (eg: VariableDefinition & Variable, or TeamFactory & Team, or CoreFactory & Core) do have clearly defined separate definitions and impls, and often times a way to map (VariableDefinition#getVariable, or GoalDefinition#getGoal)

You can bridge the gap with a `Structure getStructure(Match match)` method

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -405,4 +407,21 @@ public Action<? super MatchPlayer> parseTeleport(Element el, Class<?> scope)
...
+    Formula<T> xFormula =
+        Formula.of(Node.fromRequiredAttr(el, "x").getValue(), variables.getContext(scope));
+    Formula<T> yFormula =
+        Formula.of(Node.fromRequiredAttr(el, "y").getValue(), variables.getContext(scope));
+    Formula<T> zFormula =
+        Formula.of(Node.fromRequiredAttr(el, "z").getValue(), variables.getContext(scope));
+
+    XMLFeatureReference<StructureDefinition> structure =
+        features.addReference(new XMLFeatureReference<>(
+            features, Node.fromRequiredAttr(el, "structure"), StructureDefinition.class));
```
```suggestion
    XMLFeatureReference<StructureDefinition> structure = 
          features.createReference(Node.fromRequiredAttr(el, "structure"), StructureDefinition.class);
```

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -48,4 +48,9 @@ public Region getRegion() {
   public void place(BlockVector offset) {
     snapshot.placeBlocks(region, offset);
   }
+
+  public void placeAbsolute(Vector vector) {
```
Pablete1234:
```suggestion
  public void placeAbsolute(BlockVector vector) {
```

Pablete1234 (commented):
Just a few minor nits

Pablete1234 (approved):

PR #1350, opened 2024-07-03
Add targeted block components to PlayerLocationVariable

cswhite2000:
Adds a few new components to `player-location` variables.

```java
    HAS_TARGET,
    TARGET_X,
    TARGET_Y,
    TARGET_Z,
    PLACE_X,
    PLACE_Y,
    PLACE_Z,
```

not sure whether caching here is worth it.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
+++ b/core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
@@ -3,33 +3,29 @@ package tc.oc.pgm.variables.types;
 import static java.lang.Math.toRadians;
 import static tc.oc.pgm.util.nms.PlayerUtils.PLAYER_UTILS;
 
-import java.util.HashMap;
+import org.bukkit.Location;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.util.block.RayBlockIntersection;
-import tc.oc.pgm.util.nms.NMSHacks;
 import tc.oc.pgm.variables.VariableDefinition;
 
 public class PlayerLocationVariable extends AbstractVariable<MatchPlayer> {
 
   private final Component component;
-  private final HashMap<MatchPlayer, RayBlockIntersection> locationTargetMap;
-  private long tickCached;
+  private Location lastLocation;
+  private RayBlockIntersection lastRayCast;
 
   public PlayerLocationVariable(VariableDefinition<MatchPlayer> definition, Component component) {
     super(definition);
     this.component = component;
-    this.locationTargetMap = new HashMap<>();
-    this.tickCached = -1;
   }
 
   private RayBlockIntersection intersection(MatchPlayer player) {
-    long tick = NMSHacks.NMS_HACKS.getMonotonicTime(player.getWorld());
-    if (tickCached != tick) {
-      tickCached = tick;
-      locationTargetMap.clear();
+    if (player.getLocation().equals(lastLocation)) {
+      return lastRayCast;
     }
-    return locationTargetMap.computeIfAbsent(
-        player, matchPlayer -> PLAYER_UTILS.getTargetedBlock(matchPlayer.getBukkit()));
+    lastLocation = player.getLocation().clone();
+    lastRayCast = PLAYER_UTILS.getTargetedBlock(player.getBukkit());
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
```diff
@@ -1,17 +1,35 @@
...
+import java.util.HashMap;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.block.RayBlockIntersection;
+import tc.oc.pgm.util.nms.NMSHacks;
 import tc.oc.pgm.variables.VariableDefinition;
 
 public class PlayerLocationVariable extends AbstractVariable<MatchPlayer> {
 
   private final Component component;
+  private final HashMap<MatchPlayer, RayBlockIntersection> locationTargetMap;
```
Pablete1234: i would not bother caching a map, just cache literally the latest intersection:
```
Location inputCache;
RayBlockIntersection resultCache;
```

then just:
```
private RayBlockIntersection(Player player) {
  if (Objects.equals(player, inputCache) return resultCache;
  inputCache = player.getLocation().clone();
  return resultCache = PLAYER_UTILS.getTargetedBlock(player);
}
```

Pablete1234 (commented):
consider also adding `HAS_TARGET` with 1 or 0 based on if the raycast found anything or not. I'm unsure as to what getBlock() returns if there was nothing found, but if there's potential for a NPE you should instead return 0 or -1 or something.

core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
cswhite2000: that would be easier

core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
```diff
@@ -30,6 +44,13 @@ protected double getValueImpl(MatchPlayer player) {
       case VEL_X -> player.getBukkit().getVelocity().getX();
       case VEL_Y -> player.getBukkit().getVelocity().getY();
       case VEL_Z -> player.getBukkit().getVelocity().getZ();
+      case TARGET_X -> intersection(player).getBlock().getX();
```
Pablete1234: These all still have potential for NPE if a map author forgets to check for HAS_TARGET = 1

cswhite2000: right

core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
```diff
@@ -1,19 +1,36 @@
...
 
+import java.util.function.ToDoubleFunction;
+import org.bukkit.Location;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.block.RayBlockIntersection;
 import tc.oc.pgm.variables.VariableDefinition;
 
 public class PlayerLocationVariable extends AbstractVariable<MatchPlayer> {
 
+  private static final double NULL_VALUE = -1000000; // negative 1 million
```
Pablete1234: Honestly this sounds problematic, if someone tries to start placing blocks 1 million blocks out it's probs gonna force those chunks to load.

just make -1 the default or smthg

core/src/main/java/tc/oc/pgm/variables/types/PlayerLocationVariable.java
```diff
@@ -1,19 +1,36 @@
...
 
+import java.util.function.ToDoubleFunction;
+import org.bukkit.Location;
 import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.block.RayBlockIntersection;
 import tc.oc.pgm.variables.VariableDefinition;
 
 public class PlayerLocationVariable extends AbstractVariable<MatchPlayer> {
 
+  private static final double NULL_VALUE = -1; // negative 1 million
```
Pablete1234:
```suggestion
  private static final double NULL_VALUE = -1;
```

Pablete1234 (approved):

Issue #1352, opened 2024-07-04
Allow setting death messages to squad members


Brottweiler:
An option of `/toggle dms` should be `squad` members, so that only the death messages of players in your squad shows up.

PR #1353, opened 2024-07-04
Add squad death message setting

applenick:
This PR adds squad support to death messages, resolving issue #1352. Additionally, it includes a cleanup of the death message logic for better readability and maintainability. As always, I've tested all changes and everything is in working order 👍 Please let me know if there's any additional feedback/changes you'd like to see made.

Thanks!

### DIFF
--- a/core/src/main/java/tc/oc/pgm/death/DeathMessageMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/death/DeathMessageMatchModule.java
@@ -1,5 +1,6 @@
 package tc.oc.pgm.death;
 
+import java.util.function.BiPredicate;
 import java.util.logging.Logger;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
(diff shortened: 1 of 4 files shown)


Pablete1234:
Should this just be `own`, `friends`, `friends & squad`, `all` ?
i think it'd make more sense

applenick:
That makes sense! I kept the name as `squad` since no prior setting has used more than one word for the name. If you'd like me to change that, please let me know. Otherwise, I have adjusted the logic and settings menu description to reflect friends as well.

core/src/main/java/tc/oc/pgm/death/DeathMessageMatchModule.java
```diff
@@ -41,25 +42,35 @@ public void handleDeathBroadcast(MatchPlayerDeathEvent event) {
...
+              || isStaff) {
+            viewer.sendMessage(message.decoration(TextDecoration.ITALIC, isStaff));
           }
           break;
         case DEATH_ALL:
-          if (event.isInvolved(viewer) || event.isInvolved(viewer.getSpectatorTarget())) {
+          if (involved) {
             viewer.sendMessage(message.decoration(TextDecoration.BOLD, true));
           } else {
             viewer.sendMessage(message);
```
Pablete1234: This switch was already a monstruosity before, and it just gets worse.

Maybe changing this to a switch expression that returns true/false for if the msg should show, then just a simple check after to decide if bold (own) or italic (staff & teamgrief) is applied ontop.


```java
boolean show = involved || isStaff || switch (viewer.getSettings().getValue(SettingKey.DEATH)) {
        case DEATH_OWN -> false;
        case DEATH_FRIENDS -> isFriendInvolved(...);
        case DEATH_SQUAD -> isFriendInvolved(...) ||  isSquadInvolved(...);
        case DEALTH_ALL -> true;
}
if (show) {
  if (involved) message = message.decoration(TextDecoration.BOLD, true);
  else if (isStaff) message = message.decoration(TextDecoration.ITALIC, true);
  viewer.sendMessage(message);
}
```

Pablete1234 (commented):
Looks good but it'd be nice if you could adopt what i mentioned.

If you don't feel like making the change we can merge it as is and make the change afterwards

zzufx:
I can see a lot of value in just having a squad-only option, especially if you've got a lot of people on your friends list - would having both new options feel too "bloaty"?

Pablete1234:
I think it'd be too bloaty

Technically our data model doesn't prohibit two enum flags being toggled at once, so potentially this could be made so you can toggle the friends and squad bits independently, but it does require a bit of a refactor on pgm side to take that multiple options could be present

core/src/main/java/tc/oc/pgm/death/DeathMessageMatchModule.java
applenick: Great suggestion! I wasn't sure how much I could simplify before, but I'm glad we were able to improve it even more by getting rid of that awful-looking switch statement. Let me know if there's anything else you'd like to see changed. Thanks 👍

Pablete1234 (approved):

cswhite2000 (approved):

PR #1356, opened 2024-07-06
Add variant support for slugs & server version limitation

Pablete1234:
```xml
<map min-server-version="1.21">
    <name>New map</name>
    <variant id="old" min-server-version="1.8" max-server-version="1.17" world="legacy">Old map</variant>
</map>
```

The map itself can have `min-server-version` and `max-server-version` and individual variants can also override this

# Slugs
Currently slugs are broken in variants, as one variant can't know the slug of other variants, since it's not part of the XML it sees (part of conditionals)
This PR fixes that by adding slug as an attribute in variants:

```xml
<map>
  <name>New map</name>
  <variant id="old" slug="map_old">Old map</variant>
</map>
```
If a slug is defined in the variant, that slug is used, otherwise:
For variants `override=true`: the slug defaults to the variant name slugified
For variants `override=false` & no default `<slug>`: the slug will be the full variant name slugified
For variants `override=false` & and default `<slug>`: the slug will be `defaultslug + "_" + variantId`

This aligns closely with what the behavior would usually be if it went thru the old rules, but won't always be the same

### DIFF
--- a/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
+++ b/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
@@ -398,6 +398,7 @@ public class MapInfoImpl implements MapInfo {
 
         String variantSlug = variantEl.getAttributeValue("slug");
         if (variantSlug != null) slug = variantSlug;
+        else if (override) slug = null;
         else if (slug != null) slug += "_" + variantId;
 
         minVer = fallback(Node.fromAttr(variantEl, "min-server-version"), minVer);
--- a/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
+++ b/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
@@ -39,6 +39,7 @@ import tc.oc.pgm.teams.TeamModule;
 import tc.oc.pgm.util.StreamUtils;
 import tc.oc.pgm.util.StringUtils;
 import tc.oc.pgm.util.Version;
+import tc.oc.pgm.util.bukkit.MiscUtils;
 import tc.oc.pgm.util.named.MapNameStyle;
 import tc.oc.pgm.util.named.NameStyle;
 import tc.oc.pgm.util.text.TextFormatter;
@@ -119,12 +120,11 @@ public class MapInfoImpl implements MapInfo {
   }
 
   @NotNull
-  private static Map<String, VariantInfo> createVariantMap(Element root)
-      throws InvalidXMLException {
+  private Map<String, VariantInfo> createVariantMap(Element root) throws InvalidXMLException {
     ImmutableMap.Builder<String, VariantInfo> variants = ImmutableMap.builder();
-    variants.put(DEFAULT_VARIANT, new VariantData(root, null));
+    variants.put(DEFAULT_VARIANT, new VariantData(root, this, null));
     for (Element el : root.getChildren("variant")) {
-      VariantData vd = new VariantData(root, el);
+      VariantData vd = new VariantData(root, this, el);
       variants.put(vd.variantId, vd);
     }
     return variants.build();
@@ -366,13 +366,17 @@ public class MapInfoImpl implements MapInfo {
   }
 
   private static class VariantData implements VariantInfo {
+    // taken from https://minecraft.wiki/w/Data_version#Java_Edition
+    private static final int MAP_DATA_VERSION_1_13 = 1519;
+    private static final Version VERSION_1_13 = new Version(1, 13, 0);
     private final String variantId;
     private final String mapName;
     private final String mapId;
     private final String world;
     private final Range<Version> serverVersions;
 
-    public VariantData(Element root, @Nullable Element variantEl) throws InvalidXMLException {
+    public VariantData(Element root, MapInfo mapInfo, @Nullable Element variantEl)
+        throws InvalidXMLException {
       String name = assertNotNull(Node.fromRequiredChildOrAttr(root, "name").getValueNormalize());
       String slug = assertNotNull(root).getChildTextNormalize("slug");
       Node minVer = Node.fromAttr(root, "min-server-version");
@@ -404,7 +408,9 @@ public class MapInfoImpl implements MapInfo {
       this.mapId = assertNotNull(slug != null ? slug : StringUtils.slugify(mapName));
 
       this.serverVersions = XMLUtils.parseClosedRange(
-          minVer, XMLUtils.parseSemanticVersion(minVer), XMLUtils.parseSemanticVersion(maxVer));
+          minVer,
+          parseOrInferMinimumVersion(mapInfo, minVer),
+          XMLUtils.parseSemanticVersion(maxVer));
     }
 
     @Override
@@ -430,6 +436,30 @@ public class MapInfoImpl implements MapInfo {
     @Override
     public Range<Version> getServerVersions() {
       return serverVersions;
+    }
+
+    @Nullable
+    private Version parseOrInferMinimumVersion(MapInfo mapInfo, @Nullable Node minVer)
+        throws InvalidXMLException {
+      if (minVer != null) {
+        return XMLUtils.parseSemanticVersion(minVer);
+      } else {
+        /*
+         * Infer the map version from the DataVersion field in level.dat. We only need to know
+         * if the world is 1.13+ to avoid server crashes in legacy versions due to changes
+         * in chunk formatting.
+         */
+        var sourceDir = mapInfo.getSource().getAbsoluteDir();
+        var levelDatPath =
+            (world != null ? sourceDir.resolve(world) : sourceDir).resolve("level.dat");
+
+        var mapDataVersion = MiscUtils.MISC_UTILS.getWorldDataVersion(levelDatPath);
+        if (mapDataVersion >= MAP_DATA_VERSION_1_13) {
+          return VERSION_1_13;
+        } else {
+          return null;
+        }
+      }
     }
   }
 }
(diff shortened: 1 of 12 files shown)


core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -367,10 +370,13 @@ private static class VariantData implements VariantInfo {
     private final String mapName;
     private final String mapId;
     private final String world;
+    private final Range<Version> serverVersions;
 
     public VariantData(Element root, @Nullable Element variantEl) throws InvalidXMLException {
       String name = assertNotNull(Node.fromRequiredChildOrAttr(root, "name").getValueNormalize());
       String slug = assertNotNull(root).getChildTextNormalize("slug");
+      Node minVer = Node.fromAttr(root, "min-server-version");
```
TTtie, line 383: Modern (1.9+) worlds have got version metadata in `level.dat`, maybe the minimum server version could be inferred from there if not specified?

Pablete1234: Having this in the XML makes it more visible to mapdevs, and the behavior more predictable. Additionally, there's times where a map may be made in 1.9 or 1.12 and yet be loaded and play in 1.8 just fine. Forcibly making those not load would be a behavior change.

I'm fine with having a default min-server-version of 1.13 if the world is in 1.13+ format, as that is never going to be able to load in 1.8, but otherwise keep it xml defined

calcastor: > I'm fine with having a default min-server-version of 1.13 if the world is in 1.13+ format, as that is never going to be able to load in 1.8, but otherwise keep it xml defined

This seems ideal and aligns similarly with how PGM tends to treat `level.dat` information at present.

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/SpMiscUtil.java
```diff
@@ -83,4 +89,19 @@ public ThrownPotion spawnPotion(Location loc, ItemStack item) {
...
+  @Override
+  public int getWorldDataVersion(Path levelDatPath) {
+    try {
+      var compoundTag =
+          NBTCompressedStreamTools.a(Files.newInputStream(levelDatPath, StandardOpenOption.READ));
+      var dataTag = compoundTag.getCompound("Data");
+      return dataTag.hasKeyOfType("DataVersion", NBT_TAG_ANY_NUMERIC)
+          ? dataTag.getInt("DataVersion")
+          : -1;
+    } catch (Exception ignored) {
```
TTtie: style: might be worthwhile to catch all throwables to keep it consistent with modern
```suggestion
    } catch (Throwable ignored) {
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/SpMiscUtil.java
```diff
@@ -83,4 +89,19 @@ public ThrownPotion spawnPotion(Location loc, ItemStack item) {
   public double getArrowDamage(Arrow arrow) {
     return arrow.spigot().getDamage();
   }
+
+  @Override
+  public int getWorldDataVersion(Path levelDatPath) {
+    try {
+      var compoundTag =
+          NBTCompressedStreamTools.a(Files.newInputStream(levelDatPath, StandardOpenOption.READ));
```
TTtie: style: keep the root tag name consistent with modern
```suggestion
      var root =
          NBTCompressedStreamTools.a(Files.newInputStream(levelDatPath, StandardOpenOption.READ));
```

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -384,9 +393,23 @@ public VariantData(Element root, @Nullable Element variantEl) throws InvalidXMLE
...
+
+        String variantSlug = variantEl.getAttributeValue("slug");
+        if (variantSlug != null) slug = variantSlug;
+        else if (slug != null) slug += "_" + variantId;
+
+        Node minVerVariant = Node.fromAttr(variantEl, "min-server-version");
+        if (minVerVariant != null) minVer = minVerVariant;
+
+        Node maxVerVariant = Node.fromAttr(variantEl, "max-server-version");
+        if (maxVerVariant != null) minVer = minVerVariant;
```
TTtie: 🚨💥 variant-level max-server-version setting is being straight up ignored
```suggestion
        if (maxVerVariant != null) maxVer = maxVerVariant;
```

testing diff over OCC PublicMaps:

```diff
diff --git a/ctf/directrix/map.xml b/ctf/directrix/map.xml
index 9811fc9e..525771c0 100644
--- a/ctf/directrix/map.xml
+++ b/ctf/directrix/map.xml
@@ -3,8 +3,8 @@
 <variant id="halloween" world="halloween" override="true">DirecTricks</variant>
 <variant id="christmas" world="christmas" override="true">DirectriXmas</variant>
 <variant id="classic" world="classic">Classic</variant>
-<variant id="king_ranch" world="king_ranch" override="true">King Ranch</variant>
-<variant id="king_ranch_og" world="king_ranch_og" override="true">King Ranch: OG</variant>
+<variant id="king_ranch" world="king_ranch" override="true" min-server-version="1.0" max-server-version="1.8.8">King Ranch</variant>
+<variant id="king_ranch_og" world="king_ranch_og" override="true" min-server-version="1.21">King Ranch: OG</variant>
 <unless variant="king_ranch,king_ranch_og">
     <version>1.1.5</version>
 </unless>
```

TTtie (changes requested):
uh-oh! it blew up :sob:

add'l issues/notes:
- `/map <map>` lists variants not loaded due to version restrictions 
- it might be worthwhile to note somewhere in the docs that `max-server-version="1.8"` means "limit to 1.8.0", not "limit to 1.8.8" (depending on how common the belief of "`max-server-version="1.8"` includes `1.8.8` servers" is going to get)

Pablete1234:
image
image

Variants that can't load will now appear in red and have an explicit hover message

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -349,44 +356,64 @@ protected void setContext(MapContextImpl context) {
...
         this.world = variantEl.getAttributeValue("world");
-        if (slug != null) slug += "_" + variantId;
+
+        String variantSlug = variantEl.getAttributeValue("slug");
+        if (variantSlug != null) slug = variantSlug;
+        else if (override) slug = null;
+        else if (slug != null) slug += "_" + variantId;
+
+        minVer = fallback(Node.fromAttr(variantEl, "min-server-version"), minVer);
+        maxVer = fallback(Node.fromAttr(variantEl, "max-server-version"), maxVer);
```
TTtie, line 405: ~~supposed to be other way around? (otherwise root-level map settings will take priority over variant-level settings)~~ not the issue
```suggestion
        minVer = fallback(minVer, Node.fromAttr(variantEl, "min-server-version"));
        maxVer = fallback(maxVer, Node.fromAttr(variantEl, "max-server-version"));
```

TTtie (changes requested):

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
Pablete1234: Nope, it takes the first non null param

This is just "take the variant's node, otherwise fallback to the map's node"

TTtie: ~~Currently, if both the default variant and a variant define a version bound (such as the example in the cover letter), the first non-null parameter in that case the root-level `{min,max}-server-version`, i. e. the `{min,max}-server-version` in `<map>` would take precedence over the attributes in `<variant>`.~~

~~In other words, root-level/default variant version bound is used first, falling back to the variant-level bound if not present 🤷‍♀️~~ 

Wait, nevermind.

TTtie: This was observed and read incorrectly on my side, sorry! (╯°□°)╯︵ ┻━┻


There's actually a different issue with this – the root-level version bounds get unused node warnings when a variant overrides them… although that's correct, that should get silenced as it is intended to not be used

Pablete1234:
There's one pending thing and it's unused node detection can mark the attributes on the root as unused for the variants. How exactly we tackle unused node detection in variants is something i want to give more thought and a more final solution later down the road, so it'll be ignored for now.

core/src/main/java/tc/oc/pgm/map/MapLibraryImpl.java
```diff
@@ -226,16 +223,22 @@ private MapContext loadMap(MapSource source, @Nullable String mapId) throws MapE
     }
 
     MapInfo info = context.getInfo();
-    maps.merge(
-        info.getId(), info, (m1, m2) -> m2.getVersion().isOlderThan(m1.getVersion()) ? m1 : m2);
+    // Only if from a supported version, add it to our library
+    if (info.isServerSupported()) {
+      maps.merge(
```
cswhite2000, line 228: Will this register as a failure and should it? I feel like there would be some value in logging it specifically as a version incompatibility. I feel like people might be confused if we don't log anything.

cswhite2000 (commented):
No major concerns

core/src/main/java/tc/oc/pgm/map/MapLibraryImpl.java
Pablete1234: It will simply skip maps that are not for the current server version. It is intended as you may maintain a repo that contains both 1.8 and 1.21 maps, and simply when running in 1.8 the 1.21 maps are skipped and all is good

I'm not sure about logging as it'd likely lead to log-spam that is not really serving any use, as in, whoever set those maps knows what they're doing

cswhite2000 (approved):

PR #1357, opened 2024-07-06
Add support for projectiles shooting falling blocks

cswhite2000:
Adds support for projectiles shooting falling blocks

Example xml:
```xml
<kits>
    <kit id="spawn-kit">
        <item slot="3" material="stick" projectile="web"/>
    </kit>
</kits>
<projectiles>
        <projectile id="web" name="web" throwable="false" projectile="FallingBlock" material="web"/>
</projectiles>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
@@ -10,6 +10,7 @@ import java.util.Set;
 import java.util.logging.Logger;
 import org.bukkit.entity.Arrow;
 import org.bukkit.entity.Entity;
+import org.bukkit.entity.FallingBlock;
 import org.bukkit.potion.PotionEffect;
 import org.jdom2.Document;
 import org.jdom2.Element;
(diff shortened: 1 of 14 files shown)


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/material/BlockDataIterator.java
```diff
@@ -73,4 +75,11 @@ public org.bukkit.block.data.BlockData getBlock() {
   public BlockVector getBlockVector() {
     return blockVector;
   }
+
+  @Override
+  public FallingBlock spawnFallingBlock(Location location) {
+    return location
+        .getWorld()
+        .spawn(location, FallingBlock.class, fallingBlock -> fallingBlock.setBlockData(data));
+  }
```
Pablete1234: impl can be moved over to ModernBlockMaterialData to avoid it being duplicated

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -54,20 +55,18 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+            Node.fromChildOrAttr(projectileElement, "velocity"), Double.class, 1.0);
+        ClickAction clickAction = XMLUtils.parseEnum(
+            Node.fromAttr(projectileElement, "click"), ClickAction.class, ClickAction.BOTH);
         Class<? extends Entity> entity =
             XMLUtils.parseEntityTypeAttribute(projectileElement, "projectile", Arrow.class);
-        Float power =
-            XMLUtils.parseNumber(
-                Node.fromChildOrAttr(projectileElement, "power"), Float.class, (Float) null);
+        BlockMaterialData blockMaterial =
+            XMLUtils.parseBlockMaterialData(Node.fromAttr(projectileElement, "material"));
```
Pablete1234: only parse material if the entity type is a falling block, that way it will launch an unused node if you try to use it for smthg else

Pablete1234 (approved):

PR #1358, opened 2024-07-06
fix: avoid loading 1.13+ maps in legacy versions

TTtie:
This patch (based on #1356) adds the ability to prevent 1.13+ maps without a specified minimum version in `map.xml` from accidentally being loaded into legacy servers, causing undefined behavior including server crashes when cycled to.

This is achieved by reading the data version (an integer increased with any Minecraft version, including snapshots [^1]) from `level.dat` bundled with the map and inferring whether the map is a 1.13+ map.

The patch is in a preliminary state, some things might need to be moved around as the version-gating PR develops.

[^1]: See https://minecraft.wiki/w/Data_version for more information.


Pablete1234:
Has already been cherry-picked into the branch

PR #1359, opened 2024-07-07
chore: fold code owners into a single line

TTtie:
Per GitHub's documentation, two or more code owners must be on a single line to be considered "code owners", otherwise only the last entry is taken into account. [^1] This results in review requests currently only being requested only from a single maintainer. As such, this patch folds everyone into a single line so all maintainers get a review request as (probably) intended.

[^1]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,5 +2,4 @@
 # https://help.github.com/articles/about-codeowners/
 
 * @Electroid
-* @Pablete1234
-* @cswhite2000
+* @Pablete1234 @cswhite2000


Pablete1234:
No, we don't want electroid to be required for reviews, if anything just collapse me and cs and leave elec as separate line

Pablete1234 (approved):

cswhite2000 (approved):

PR #1360, opened 2024-07-07
Add some tracking workarounds for modern versions

cswhite2000:
Some workarounds for non implemented tracking in modern versions.

Needs testing

### DIFF
--- a/core/src/main/java/tc/oc/pgm/listeners/BlockTransformListener.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/BlockTransformListener.java
@@ -42,7 +42,6 @@ import org.bukkit.event.block.BlockPlaceEvent;
 import org.bukkit.event.block.BlockSpreadEvent;
 import org.bukkit.event.entity.EntityChangeBlockEvent;
 import org.bukkit.event.entity.EntityExplodeEvent;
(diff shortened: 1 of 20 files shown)


Pablete1234:
Did a bit of testing on this with mame,

What works:
- Shooting tnt with a flame bow gives credit
- Dispensing an arrow gives kill credit
- Creeper egg on the monument gives credit
- Placing TNT minecart gives credit

What doesn't work:
- Dispensing tnt does not give credit
- Dispensing a mob egg (zombie) still attacks you, so it's not tracking
- Dispensing a creeper egg does not give credit
- Dispensing a tnt minecart does not give credit

Finally, this causes a NPE due to the dispense event using null as an item for projectile dispensing:
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.bukkit.inventory.ItemStack.clone()" because "this.item" is null
	at org.bukkit.event.block.BlockDispenseEvent.getItem(BlockDispenseEvent.java:37) ~[paper-mojangapi-1.21.1-R0.1-SNAPSHOT.jar:?]
	at PGM.jar/tc.oc.pgm.listeners.BlockTransformListener.onDispenserDispense(BlockTransformListener.java:531) ~[PGM.jar:?]
```

PR #1362, opened 2024-07-07
Add "5CP" to the list of gamemodes

OhPointFive:
This PR adds 5CP (5 Control Points) to the list of gamemodes. From searching through some discords, it seems that `Points` is always plural, which makes sense but for some reason isn't what I expected.

This is following [discussion in the discord](https://discord.com/channels/730855489767997511/730855489767997515/1259601301994733568).

I alphabetically sorted `5` as coming before `a`, since I've never seen `five` spelled out in 5cp.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
+++ b/core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
@@ -1,31 +1,31 @@
 package tc.oc.pgm.api.map;
 
 public enum Gamemode {
-  FIVE_CONTROL_POINT("5cp", "5 Control Points", "5CP"),
-  ATTACK_DEFEND("ad", "Attack/Defend", "A/D"),
   ARCADE("arcade", "Arcade", "Arcade"),
+  ATTACK_DEFEND("ad", "Attack/Defend", "A/D"),
   BEDWARS("bedwars", "Bed Wars", "Bed Wars"),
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/api/map/Gamemode.java
```diff
@@ -1,6 +1,7 @@
 package tc.oc.pgm.api.map;
 
 public enum Gamemode {
+  FIVE_CONTROL_POINT("5cp", "5 Control Points", "5CP"),
```
Pablete1234: gamemodes are sorted alphabetically, `FIVE` should go between FREE_FOR_ALL and FLAG_FOOTBALL

OhPointFive: They seem to be alphabetized by ID, for some reason — `DEATHMATCH` is after `SURVIVAL_GAMES`, which I assumed was because of `tdm` going after `sg`.

I can rearrange them if you want, though, since that would probably make more sense

Pablete1234: yeah... i think sorting by enum constant makes more sense

OhPointFive: Done — apparently there were many others that got rearranged slightly, that I hadn't noticed before.

Pablete1234 (approved):

PR #1364, opened 2024-07-10
fix(wool): use platform-specific ColorUtils for wool matching

TTtie:
On 1.21 servers, CraftItemStack#getData() returns unsubclassed MaterialData for wool item stacks, hence we cannot check for `instanceof Wool` there. This patch changes the appropriate code to a manual comparison, enabling pickup and proximity tracking to work regardless of whether the material data is correctly subclassed.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
+++ b/core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
@@ -8,7 +8,6 @@ import org.bukkit.DyeColor;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
 import org.bukkit.inventory.ItemStack;
-import org.bukkit.material.Wool;
 import org.bukkit.util.Vector;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.feature.FeatureInfo;
@@ -18,6 +17,7 @@ import tc.oc.pgm.goals.ProximityMetric;
 import tc.oc.pgm.goals.ShowOptions;
 import tc.oc.pgm.teams.TeamFactory;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
+import tc.oc.pgm.util.material.Materials;
 import tc.oc.pgm.util.text.TextFormatter;
 
 @FeatureInfo(name = "wool")
@@ -123,7 +123,10 @@ public class MonumentWoolFactory extends ProximityGoalDefinition {
   }
 
   public boolean isObjectiveWool(org.bukkit.material.MaterialData material) {
-    return material instanceof Wool && ((Wool) material).getColor() == this.color;
+    // On 1.21, CraftItemStack#getData() on wool item stacks returns unsubclassed MaterialData,
+    // hence we're comparing the item type and color manually
+    return material.getItemType() == Materials.WOOL
+        && DyeColor.getByWoolData(material.getData()) == color;
   }
 
   public boolean isHolding(InventoryHolder holder) {
--- a/core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
+++ b/core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
@@ -6,7 +6,6 @@ import static tc.oc.pgm.wool.WoolModule.WOOL;
 import net.kyori.adventure.text.Component;
 import org.apache.commons.lang.StringUtils;
 import org.bukkit.DyeColor;
-import org.bukkit.block.BlockState;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
 import org.bukkit.inventory.ItemStack;
@@ -20,6 +19,7 @@ import tc.oc.pgm.goals.ShowOptions;
 import tc.oc.pgm.teams.TeamFactory;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.material.ColorUtils;
+import tc.oc.pgm.util.material.MaterialData;
 import tc.oc.pgm.util.text.TextFormatter;
 
 @FeatureInfo(name = "wool")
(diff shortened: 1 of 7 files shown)


core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
```diff
@@ -123,7 +123,10 @@ public boolean isObjectiveWool(ItemStack stack) {
   }
 
   public boolean isObjectiveWool(org.bukkit.material.MaterialData material) {
-    return material instanceof Wool && ((Wool) material).getColor() == this.color;
+    // On 1.21, CraftItemStack#getData() on wool item stacks returns unsubclassed MaterialData,
+    // hence we're comparing the item type and color manually
+    return material.getItemType() == Materials.WOOL
```
cswhite2000: For 1.13+ the Wool material is always in the format `COLOR_WOOL`. This won't work for modern versions.

cswhite2000 (changes requested):

core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java
Pablete1234: it will work because we are manually defining the wools with legacy wool, but it's not something we should do anyways

core/src/main/java/tc/oc/pgm/wool/WoolMatchModule.java
```diff
@@ -83,7 +80,7 @@ public Multimap<Team, MonumentWool> getWools() {
   }
 
   private boolean isObjectiveWool(ItemStack stack) {
-    if (WOOL.matches(stack.getType())) {
+    if (WoolModule.WOOL.matches(stack.getType())) {
```
cswhite2000: nitpick: static importing this would remove the diff

core/src/main/java/tc/oc/pgm/wool/WoolModule.java
```diff
@@ -29,11 +29,13 @@
 import tc.oc.pgm.teams.TeamMatchModule;
 import tc.oc.pgm.teams.TeamModule;
 import tc.oc.pgm.teams.Teams;
+import tc.oc.pgm.util.material.MaterialMatcher;
 import tc.oc.pgm.util.xml.InvalidXMLException;
 import tc.oc.pgm.util.xml.Node;
 import tc.oc.pgm.util.xml.XMLUtils;
 
 public class WoolModule implements MapModule<WoolMatchModule> {
+  static final MaterialMatcher WOOL = MaterialMatcher.of(m -> m.name().endsWith("WOOL"));
```
cswhite2000: nitpick: Style: prefer to have an explicit access modifier over an implicit one

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpColorUtils.java
```diff
@@ -71,6 +71,11 @@ public boolean isColor(BlockState block, DyeColor color) {
     return color.getWoolData() == block.getRawData();
   }
 
+  @Override
+  public boolean isColor(ItemStack stack, DyeColor color) {
+    return color.getWoolData() == stack.getDurability();
```
cswhite2000: If this is called on other things with data / durability this may return true even if color is not relevant.
Not sure if this is currently an issue based off of current usage of the method, but it seems a candidate for easily introducing bugs if kept as is.
If you double check current usage and see no issues, then I'm fine with keeping it as is for this pr.

cswhite2000 (commented):
nothing major, few minor things

core/src/main/java/tc/oc/pgm/wool/WoolModule.java
TTtie: This is intentionally package-private, how do I document this? Or should I avoid package-private access level and bump this to `public`?

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpColorUtils.java
TTtie: Currently, this method is used only here: (the mentioned method is called by other 4 callers, but the preceding wool query filters that away)

https://github.com/PGMDev/PGM/blob/0065f1527c2e85683f66d6d2d2d6f8c0b69b6aa1/core/src/main/java/tc/oc/pgm/wool/MonumentWoolFactory.java#L123-L125

As for this method being called for un-colorable items, I agree this might open a possibility for other bugs, but none of the other methods in legacy ColorUtils check for whether the block can be colored either – fixing this should go in another PR as it would be out of scope of this PR and #1363 for which this PR is set to be merged into.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/material/ModernColorUtils.java
```diff
@@ -103,9 +103,18 @@ public void setColor(Block block, DyeColor color) {
     block.setType(setColor(block.getType(), color));
   }
 
+  private boolean setAndCompare(Material mat, DyeColor color) {
+    return setColor(mat, color) == mat;
+  }
+
```
Pablete1234:
```suggestion
```

i'd inline it, it's such a trivial thing

util/src/main/java/tc/oc/pgm/util/material/ColorUtils.java
```diff
@@ -29,6 +29,8 @@ public interface ColorUtils {
 
   boolean isColor(BlockState block, DyeColor color);
 
+  boolean isColor(ItemStack stack, DyeColor color);
```
Pablete1234: This should probably just use pgm's MaterialData as param instead, so it can work for both blockstte and itemstacks: `MaterialData.of(blockstate)` or `MaterialData.of(stack)`

core/src/main/java/tc/oc/pgm/wool/WoolModule.java
cswhite2000: If you have a reason for it to be packaged private over protected than that is fine. In any case this isn't a sufficient issue to block merging

cswhite2000 (approved):

PR #1366, opened 2024-07-19
Fix objective event performance issues

Pablete1234:
image

In an FFA map with a ton of hills being captured by a ton of players, re-sorting the list of competitors every time a hill is captured is a real performance issue.


cswhite2000 (dismissed):

cswhite2000:
seems you forgot to run the formatter pablo

cswhite2000 (approved):

PR #1367, opened 2024-07-21
chore!: remove Docker support, fold CI into single workflow

TTtie:
The Docker image is using the SP+PGM dev server build, which is prone to breakages in production due to the differences between a proper PGM deployment and the development server. With recent updates, in fact, the Docker image for PGM has stopped working entirely. Following a discussion on Discord[^1] and to avoid bad practices, let's stop offering a Docker image altogether as most real-world deployments build their own Docker images anyway.

This patch additionally unifies all CI into a single workflow as there's no need to maintain two separate workflows that do the same thing (mvn deploy implies mvn install and works without edits only on the dev branch at the mainline PGM repo).

[^1]: https://discord.com/channels/730855489767997511/730857379729506335/1259887964062879774


.github/workflows/build.yml
```diff
@@ -2,10 +2,14 @@ name: Build
 on:
   pull_request:
   workflow_dispatch:
+  push:
+    branches:
+      - dev
```
Pablete1234, line 7: won't this break builds for prs?

TTtie: It won't - per the [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events) the whole `on` object means "trigger on one of (PR, manual dispatch, push to `dev`)"

In fact - this PR's CI [already runs off of the mentioned file](https://github.com/PGMDev/PGM/actions/runs/10024239030/workflow?pr=1367)!

TTtie:
I see no complaints against removal, undrafting – feel free to allow a bit more time in case a stray complaint arrives

cswhite2000 (approved):

PR #1378, opened 2024-08-05
Make map search prefer name over slug

### LINKED ISSUE #1377: Slugs (map IDs) are prioritised over map names when it doesn't make sense to do so
Slugs are used to maintain a map's ID across different names in a database. With the command `/map`, exact string matches to a map's slug will resolve to that map: for instance `/map coastline` on OCC will resolve to `Coastline: Classic`.

Following 8ceb97834f506bfbce84d29dd810b9ccce6d1991, this has extended to map pools (not exactly correct, see https://github.com/PGMDev/PGM/issues/1377#issuecomment-2267975193). This is problematic given that map pools have, until this commit, resolved against name only. To use `Coastline` as an example again, we list this map in our default map pool on OCC; prior to this commit, the map that was in the pool was `Coastline` which has a slug of `coastline_redux`, and now it is `Coastline: Classic` which has a slug of `coastline`.

There are very little situations in which it is appropriate for slug to take precedence over a map's name. Slugs should function as an internal mechanism only to avoid conflicts between maps in a database. Having slugs take precedence in situations that I have outlined above makes map pools more difficult to maintain and confuses end users and players.

Resolving this issue will require consideration for computational expense. Matching against ID is quick as, typically, a slugified map name is directly connected to that map because most maps do not have an explicitly defined slug. See the following screenshot for some further context:
image

Pablete1234:
Fixes #1377 

Modifies map search to prefer matching a name, and only if multiple (or no) matches by name, it falls back to slug. Note that for maps where slug is just the normalized name, it will still stick to the result of the search by slug since it's a faster match.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
@@ -95,7 +95,7 @@ public abstract class MapPool implements MapOrder, Comparable<MapPool> {
     for (String varId : variantIds) {
       VariantInfo variant = variants.get(varId);
       if (variant == null) continue;
-      MapInfo variantMap = maps.getMap(variant.getId());
+      MapInfo variantMap = maps.getMapById(variant.getId());
       if (variantMap != null) {
         return variantMap;
       } else {
@@ -154,7 +154,6 @@ public abstract class MapPool implements MapOrder, Comparable<MapPool> {
    * @param match The match that is currently ending
    */
   public void unloadPool(Match match) {}
-  ;
 
   @Override
   public int compareTo(MapPool o) {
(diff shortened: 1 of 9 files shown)


core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
```diff
@@ -152,7 +153,8 @@ public void setNextMap(MapInfo map) {}
    *
    * @param match The match that is currently ending
    */
-  public void unloadPool(Match match) {};
+  public void unloadPool(Match match) {}
+  ;
```
cswhite2000: nitpick: random semicolon

cswhite2000 (dismissed):

core/src/main/java/tc/oc/pgm/rotation/pools/MapPool.java
Pablete1234: fixed it and also did a minor fix

cswhite2000 (approved):

PR #1374, opened 2024-08-02
Fix dynamic pickup-filter NPE

Pablete1234:
Should fix the following NPE:
```
[20:54:27 WARN]: java.lang.NullPointerException: Cannot invoke "tc.oc.pgm.variables.Variable.getValue(tc.oc.pgm.filters.Filterable)" because "variable" is null
[20:54:27 WARN]:        at tc.oc.pgm.filters.matcher.match.VariableFilter.getValue(VariableFilter.java:55)
[20:54:27 WARN]:        at tc.oc.pgm.filters.matcher.match.VariableFilter.queryTyped(VariableFilter.java:51)
[20:54:27 WARN]:        at tc.oc.pgm.filters.matcher.match.VariableFilter$Generic.queryTyped(VariableFilter.java:75)
[20:54:27 WARN]:        at tc.oc.pgm.filters.matcher.WeakTypedFilter.query(WeakTypedFilter.java:21)
[20:54:27 WARN]:        at tc.oc.pgm.filters.operator.AllFilter.query(AllFilter.java:18)
[20:54:27 WARN]:        at tc.oc.pgm.filters.operator.AllFilter.query(AllFilter.java:18)
[20:54:27 WARN]:        at tc.oc.pgm.filters.operator.AllFilter.query(AllFilter.java:18)
[20:54:27 WARN]:        at tc.oc.pgm.flag.FlagDefinition.canPickup(FlagDefinition.java:192)
[20:54:27 WARN]:        at tc.oc.pgm.flag.Flag.<init>(Flag.java:123)
[20:54:27 WARN]:        at tc.oc.pgm.flag.FlagMatchModule.<init>(FlagMatchModule.java:42)
[20:54:27 WARN]:        at tc.oc.pgm.flag.FlagModule.createMatchModule(FlagModule.java:52)
[20:54:27 WARN]:        at tc.oc.pgm.flag.FlagModule.createMatchModule(FlagModule.java:25)
[20:54:27 WARN]:        at tc.oc.pgm.match.MatchImpl$ModuleLoader.createModule(MatchImpl.java:807)
```

Issue #1267, opened 2023-10-31
PGM proto 1.5


Pablete1234:
An idea came up to implement more default ids into pgm, the same way `never` or `everywhere` are default ids for `<never/>` and `<everywhere/>`. This can't be done without potentially breaking old maps that could be using whatever we decide on id names, as an already existing filter id for something else. That's when the idea to make a proto 1.5 came.

With a new proto, maps on older (1.4.2 and below) protos simply wouldn't get those defaults registered, and would continue to operate just fine. New maps or maps that get migrated to 1.5 will start getting these new default ids.

Here's a list of proposals/potential candiates:
 - `observing` & `participating` filters
 - `alive` & `dead` filters
 - `match-idle`, `match-starting`, `match-running`, `match-finished` & `match-started` match phase filters
   -  kill generic `<match-phase>` filter
   - add specific idle & starting filters (atm you're forced to use `<match-phase>idle</match-phase>` etc)
 - `crouching`, `walking`, `sprinting`, `flying` and `grounded` player pose filters
 - `void` filter
 - `maxbuildheight` variable
 - `timelimit.duration` variable
   - this doesn't exist yet, see #1260 
   - maybe also `timelimit.overtime` if we want that to be variable-modifiable?
   - Unsure about the use of `.` for hierarchy-like keys, but we can't use `-` in variables so...
 - Rename `<walking/>` filter to `<standing/>` to better reflect what it means?
   - The filter simply passes if the player is neither sprinting, nor crouching, so either walking or standing still pass.

PR #1379, opened 2024-08-11
Start working towards proto 1.5

Pablete1234:
The road towards proto 1.5.0 has begun. Originally started in #1267

## New default IDs
A bunch of different features you'd have to define even if they are dummy just to be able to use them as an id, for example:
```xml
<filters>
  <alive id="alive"/>
</filters>
```
This adds them by default, so you don't need to worry about defining them. But it also means that these IDs are now reserved by PGM and you can't use them freely to define other things or it will be invalid xml.
### Filters
 - `observing` & `participating` filters
 - `alive` & `dead` filters
 - `match-idle`, `match-starting`, `match-running`, `match-finished` & `match-started` match phase filters
 - `crouching`, `walking`, `sprinting`, `grounded`, `flying` and `can-fly` player pose filters
 - `void` filter
### Variables
- `lives`
- `score`
- `timelimit`
- `maxbuildheight`
- `player.x`, `player.y`, `player.z`, `player.pitch`, `player.yaw` current player position
- `player.dir_x`, `player.dir_y`, `player.dir_z` for the direction player is looking in
- `player.vel_x`, `player.vel_y`, `player.vel_z` for the player velocity
- `player.has_target` for if the player is targeting (looking at) a block
- `player.target_x`, `player.target_y`, `player.target_z` for the block the player is targeting
- `player.place_x`, `player.place_y`, `player.place_z` for the block the player would place on (ie: the block adjacent to the targeted block, in the blockface the player is looking in)

## Triggers & switch-scope actions exclude observers
A common mistake when dealing with actions is that they will apply to observers, even when it often is not the intent of the map author.
Triggers & switch-scope actions for scopes other than match (ie: team & player) will default to not caring about obs, you may override this behavior by explicitly setting it.
For maps 1.4.2 and below, observers defaults to true to keep backwards-compat identical, for 1.5.0 and up, it will default to false.
`<trigger scope="player" observers="true" filter="your-filter">` 
`<switch-scope inner="player" observers="true">` 

## Actions untrigger by default
Actions can trigger, but they can also untrigger. This is not very commonplace and the only action that does anything on untrigger are removable kits, however this may change in the future if we add more stuff.
For backwards-compat, we were forced to make `untrigger-filter` default to never (although this seems not to even be documented), that changes with proto 1.5 to `always`.
`<action untrigger-filter="always">` is the new default

## Other internal changes
If you're not a developer, you can skip this explanation as it does not affect map development at all.

I've refactored most of how variables were originally coded. Originally we had a clear distinction between VariableDefinition and Variable, because the former was map-scoped and the later match-scoped. With increasing amounts of variables that are stateless and very few requiring state, this made less and less sense.

Variables now work more closely to how filters do, where `Variable` is both the variable and the definition, and as an implementation-detail, whichever variable needs a state, can create one per-match.
The concept of `reactors` that filters already enjoyed has been generalized to `StateHolder<T>` that anything can implement to create state in the match that it can later retrieve. Reactors now extend this state mechanism.

## Fixes
Interactions on flag load that could cause filters that acted on variables that weren't yet loaded to fail, has been fixed. This was thru a small refactor in how certain features are loaded (this is, after initialization has already happened), now the order is variables, filters, flags, goals, which means flags can use filters that require variables. (Fixes #1374)


cswhite2000 (approved):
Its a rather large diff, so I could have missed something, but overall everything looks good to me.

Pablete1234:
Even tho i don't fully consider 1.5 finished, i think it's better to merge and keep working on it over time
If a change is very minor (eg: a new id for something very unlikely any map is already using) append it into 1.5 and if it's substantial then make a 1.5.1

PR #1392, opened 2024-08-25
Reimplement Blitz title for older protos

calcastor:
`MapProtos` lists `REMOVE_BLITZ_TITLE` as a change in proto 1.4.0, but modern PGM has never seemed to use `<title>` within `<blitz>`. This PR reimplements this, attempting to follow the implementation in [ProjectAres](https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/modules/InfoModule.java#L131-L141).

### DIFF
--- a/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
+++ b/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
@@ -28,6 +28,7 @@ import tc.oc.pgm.api.map.Gamemode;
 import tc.oc.pgm.api.map.MapContext;
 import tc.oc.pgm.api.map.MapInfo;
 import tc.oc.pgm.api.map.MapModule;
+import tc.oc.pgm.api.map.MapProtos;
 import tc.oc.pgm.api.map.MapSource;
 import tc.oc.pgm.api.map.MapTag;
 import tc.oc.pgm.api.map.Phase;
@@ -102,7 +103,7 @@ public class MapInfoImpl implements MapInfo {
             Node.fromLastChildOrAttr(root, "difficulty"), Difficulty.class, Difficulty.NORMAL)
         .ordinal();
     this.world = parseWorld(root);
-    this.gamemode = XMLUtils.parseFormattedText(Node.fromLastChildOrAttr(root, "game"));
+    this.gamemode = parseGamemode(root);
     this.gamemodes = parseGamemodes(root);
     this.phase =
         XMLUtils.parseEnum(Node.fromLastChildOrAttr(root, "phase"), Phase.class, Phase.PRODUCTION);
@@ -263,6 +264,23 @@ public class MapInfoImpl implements MapInfo {
     return XMLUtils.flattenElements(root, "rules", "rule").stream()
         .map(Element::getTextNormalize)
         .collect(StreamUtils.toImmutableList());
+  }
+
+  private Component parseGamemode(Element root) throws InvalidXMLException {
+    Component gamemode = XMLUtils.parseFormattedText(root, "game");
+
+    if (gamemode == null) {
+      for (Element title : XMLUtils.flattenElements(root, "blitz", "title")) {
+        if (this.getProto().isNoOlderThan(MapProtos.REMOVE_BLITZ_TITLE)) {
+          throw new InvalidXMLException(
+              "<title> inside <blitz> is no longer supported, use <map game=\"...\"> or <game>",
+              title);
+        }
+        return XMLUtils.parseFormattedText(new Node(title));
+      }
+    }
+
+    return gamemode;
   }
 
   private static @NotNull List<Gamemode> parseGamemodes(Element root) throws InvalidXMLException {


core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -265,6 +266,26 @@ public Component getStyledName(MapNameStyle style) {
         .collect(StreamUtils.toImmutableList());
   }
 
+  private Component parseGamemode(Element root) throws InvalidXMLException {
+    Component gamemode = XMLUtils.parseFormattedText(Node.fromLastChildOrAttr(root, "game"));
+
+    if (gamemode == null) {
+      final Element blitz = root.getChild("blitz");
+      if (blitz != null) {
+        final Element title = blitz.getChild("title");
```
Pablete1234: I think with xml utils we can flatten these more easily

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -265,6 +266,25 @@ public Component getStyledName(MapNameStyle style) {
...
+    for (Element blitzTitle : XMLUtils.flattenElements(root, "blitz", "title")) {
+      if (this.getProto().isNoOlderThan(MapProtos.REMOVE_BLITZ_TITLE)) {
+        throw new InvalidXMLException(
+            "<title> inside <blitz> is no longer supported, use <map game=\"...\"> or <game>",
+            blitzTitle);
+      } else if (gamemode != null && blitzTitle != null) {
+        throw new InvalidXMLException(
+            "<title> inside <blitz> is overriden by <map game=\"...\"> or <game>", blitzTitle);
+      } else if (gamemode == null) {
+        gamemode = XMLUtils.parseFormattedText(Node.fromNullable(blitzTitle));
```
Pablete1234:
```suggestion
        gamemode = XMLUtils.parseFormattedText(new Node(blitzTitle));
```
this isn't nullable in this case, you know for a fact it's there

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -265,6 +266,25 @@ public Component getStyledName(MapNameStyle style) {
...
 
+  private Component parseGamemode(Element root) throws InvalidXMLException {
+    Component gamemode = XMLUtils.parseFormattedText(Node.fromLastChildOrAttr(root, "game"));
+
+    for (Element blitzTitle : XMLUtils.flattenElements(root, "blitz", "title")) {
+      if (this.getProto().isNoOlderThan(MapProtos.REMOVE_BLITZ_TITLE)) {
+        throw new InvalidXMLException(
+            "<title> inside <blitz> is no longer supported, use <map game=\"...\"> or <game>",
+            blitzTitle);
+      } else if (gamemode != null && blitzTitle != null) {
```
Pablete1234: we know blitzTitle is always non-null, so this condition is not useful

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -265,6 +266,25 @@ public Component getStyledName(MapNameStyle style) {
...
+        throw new InvalidXMLException(
+            "<title> inside <blitz> is overriden by <map game=\"...\"> or <game>", blitzTitle);
```
Pablete1234: this exception did not exist in original pgm, so it's a breaking change. It was instead ignored if an actual game element was already defined earlier.

just add an if != null and return the value

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -265,6 +266,25 @@ public Component getStyledName(MapNameStyle style) {
...
+      } else if (gamemode == null) {
+        gamemode = XMLUtils.parseFormattedText(Node.fromNullable(blitzTitle));
+      }
+    }
+
+    return gamemode;
+  }
```
Pablete1234: I believe this should have behavior identical to original for maps in the prior protos:

```suggestion
    Component gamemode = XMLUtils.parseFormattedText(root, "game");
    if (gamemode != null) return gamemode;

    for (Element title : XMLUtils.flattenElements(root, "blitz", "title")) {
      if (this.getProto().isNoOlderThan(MapProtos.REMOVE_BLITZ_TITLE)) {
        throw new InvalidXMLException(
            "<title> inside <blitz> is no longer supported, use <map game=\"...\"> or <game>",
            title);
      }
      return XMLUtils.parseFormattedText(new Node(title));
    }
  }
```
for reference (<https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/modules/InfoModule.java#L131>)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
calcastor: I was trying to be explicit about the override to avoid confusion about which title is used because whatever I did before didn't result in `<title>` being regarded as an unused node. Should be fine in latest push.

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
calcastor: Needed a return at the end, but yes.

Pablete1234 (approved):

PR #1399, opened 2024-09-15
Implement a cross-version sounds interface

### LINKED ISSUE #1395: Sound keys need to be version-dependant
Most if not all of the sound keys around pgm look like this:

`Sound WARNING_SOUND = sound(key("note.bass"), Sound.Source.MASTER, 1f, 0.75f);`

In modern versions, the keys have changed, eg `note.bass` in 1.8 is  `block.note_block.bass`, these will all need to be updated (and it probably makes sense to put them all into a `Sounds` class similar to how we got [Enchantments.java](https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/bukkit/Enchantments.java) and similar ones for other bukkit enums that have changed thru time.

Should also probably make them function the same (use string references to enum constants in `org.bukkit.Sound` so that when they break we know on startup.

calcastor:
Closes https://github.com/PGMDev/PGM/issues/1395

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
+++ b/util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
@@ -1,0 +1,19 @@
+package tc.oc.pgm.util.bukkit;
+
+import static net.kyori.adventure.sound.Sound.sound;
+import static tc.oc.pgm.util.bukkit.MiscUtils.MISC_UTILS;
+
+import net.kyori.adventure.sound.Sound;
+
+public interface Sounds {
+  Sound WARNING = sound("NOTE_BASS", "BLOCK_NOTE_BLOCK_BASS", 1f, 0.75f);
+
+  static Sound sound(String legacyConstant, String modernConstant, float volume, float pitch) {
+    return sound(
+        MISC_UTILS.getSound(
+            BukkitUtils.parse(org.bukkit.Sound::valueOf, legacyConstant, modernConstant)),
+        Sound.Source.MASTER,
+        volume,
+        pitch);
+  }
+}
--- a/util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
+++ b/util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
@@ -1,15 +1,65 @@
 package tc.oc.pgm.util.bukkit;
 
-import static net.kyori.adventure.sound.Sound.sound;
 import static tc.oc.pgm.util.bukkit.MiscUtils.MISC_UTILS;
 
 import net.kyori.adventure.sound.Sound;
 
 public interface Sounds {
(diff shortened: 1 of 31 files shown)


util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
```diff
@@ -0,0 +1,69 @@
...
+import static tc.oc.pgm.util.bukkit.MiscUtils.MISC_UTILS;
+
+import net.kyori.adventure.sound.Sound;
+
+public interface Sounds {
+  Sound ADMIN_CHAT = sound("ORB_PICKUP", "ENTITY_EXPERIENCE_ORB_PICKUP", 1f, 0.7f);
+  Sound ALERT = sound("NOTE_PLING", "BLOCK_NOTE_BLOCK_PLING", 1f, 2f);
+  Sound CONTROL_POINT_GOOD = sound("PORTAL_TRAVEL", "BLOCK_PORTAL_TRAVEL", 0.35f, 2f);
+  Sound CONTROL_POINT_BAD = sound("BLAZE_DEATH", "ENTITY_BLAZE_DEATH", 0.4f, 0.8f);
+  Sound CUSTOM = sound("NOTE_PLING", "BLOCK_NOTE_BLOCK_PLING");
```
calcastor: I don't like the name for this; according to the docs it is the fallback sound for sound actions so `FALLBACK` seems more appropriate.

util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
```diff
@@ -0,0 +1,69 @@
...
+  Sound ALERT = sound("NOTE_PLING", "BLOCK_NOTE_BLOCK_PLING", 1f, 2f);
+  Sound CONTROL_POINT_GOOD = sound("PORTAL_TRAVEL", "BLOCK_PORTAL_TRAVEL", 0.35f, 2f);
+  Sound CONTROL_POINT_BAD = sound("BLAZE_DEATH", "ENTITY_BLAZE_DEATH", 0.4f, 0.8f);
+  Sound DEATH_OWN =
+      sound("IRONGOLEM_DEATH", "ENTITY_IRON_GOLEM_DEATH"); // Own death is normal pitch, full volume
+  Sound DEATH_KILLED = sound(
+      "IRONGOLEM_DEATH",
+      "ENTITY_IRON_GOLEM_DEATH",
+      0.75f,
+      4f / 3f); // Kill is higher pitch, quieter
```
Pablete1234: Move the comment(s) above the sound itself to try and avoid them extending several lines:
```suggestion
  // Kill is higher pitch, quieter
  Sound DEATH_KILLED = sound("IRONGOLEM_DEATH", "ENTITY_IRON_GOLEM_DEATH", 0.75f, 4f / 3f);
```

Pablete1234 (dismissed):
Just a small nit, otherwise looks good

Pablete1234 (approved):

PR #1404, opened 2024-09-30
Destroyable - Add sparks-match parameter

arcadeboss:
- Adds a new parameter "sparks-match" to Destroyable

- "sparks" must be active for "sparks-match" to have any effect

- With the default value of "true" there is no change to current behavior for Destroyable sparks.

- With value of "false" PGM will skip the far (+64 blocks away) logic

- Purpose is to reduce the "noise" for players not actually near the action, especially for large Destroyables

- Tested and verified that sound behavior changed for players and observers


arcadeboss:
Not signed correctly

calcastor:
> Not signed correctly

You can just amend the commit to resolve that once able.

calcastor:
(and then re-open the PR after force pushing the amended commit)

PR #1405, opened 2024-09-30
Destroyable - Add new sparks parameter value of "near"

arcadeboss:
Adds a new parameter "sparks-match" to Destroyable

"sparks" must be active for "sparks-match" to have any effect

With the default value of "true" there is no change to current behavior for Destroyable sparks.

 With value of "false" PGM will skip the far (+64 blocks away) logic

 Purpose is to reduce the "noise" for players not actually near the action, especially for large Destroyables

 Tested and verified that sound behavior changed for players and observers

### DIFF
--- a/core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
+++ b/core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
@@ -330,7 +330,7 @@ public class Destroyable extends TouchableGoal<DestroyableFactory>
     if (deltaHealth < 0) {
       touch(player);
 
-      if (this.definition.hasSparks()) {
+      if (this.definition.isSparksActive()) {
         Location blockLocation = BlockVectors.center(oldState);
         Instant now = Instant.now();
 
@@ -356,14 +356,12 @@ public class Destroyable extends TouchableGoal<DestroyableFactory>
             NMS_HACKS.skipFireworksLaunch(firework);
           }
 
-          if (this.definition.hasSparksMatch()) {
+          if (this.definition.isSparksAll()) {
             // Players more than 64m away will not see or hear the fireworks, so play sound for them
             for (MatchPlayer listener : this.getOwner().getMatch().getPlayers()) {
               if (listener.getBukkit().getLocation().distance(blockLocation) > 64) {
-                listener.playSound(
-                    sound(key("fireworks.blast_far"), Sound.Source.MASTER, 0.75f, 1f));
-                listener.playSound(
-                    sound(key("fireworks.twinkle_far"), Sound.Source.MASTER, 0.75f, 1f));
+                listener.playSound(Sounds.OBJECTIVE_FIREWORKS_FAR);
+                listener.playSound(Sounds.OBJECTIVE_FIREWORKS_TWINKLE);
               }
             }
           }
--- a/core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
+++ b/core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
@@ -356,12 +356,15 @@ public class Destroyable extends TouchableGoal<DestroyableFactory>
             NMS_HACKS.skipFireworksLaunch(firework);
           }
 
-          // Players more than 64m away will not see or hear the fireworks, so just play the sound
-          // for them
-          for (MatchPlayer listener : this.getOwner().getMatch().getPlayers()) {
-            if (listener.getBukkit().getLocation().distance(blockLocation) > 64) {
-              listener.playSound(Sounds.OBJECTIVE_FIREWORKS_FAR);
-              listener.playSound(Sounds.OBJECTIVE_FIREWORKS_TWINKLE);
+          if (this.definition.hasSparksMatch()) {
+            // Players more than 64m away will not see or hear the fireworks, so play sound for them
+            for (MatchPlayer listener : this.getOwner().getMatch().getPlayers()) {
+              if (listener.getBukkit().getLocation().distance(blockLocation) > 64) {
+                listener.playSound(
+                    sound(key("fireworks.blast_far"), Sound.Source.MASTER, 0.75f, 1f));
+                listener.playSound(
+                    sound(key("fireworks.twinkle_far"), Sound.Source.MASTER, 0.75f, 1f));
+              }
             }
           }
         }
--- a/core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
+++ b/core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
@@ -1,6 +1,7 @@
 package tc.oc.pgm.destroyable;
 
 import com.google.common.collect.ImmutableSet;
+import org.jdom2.Element;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.feature.FeatureInfo;
 import tc.oc.pgm.api.region.Region;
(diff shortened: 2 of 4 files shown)


Pablete1234:
I think it needs a better name, maybe consider allowing sparks to have true, false, and 'near' that disables the global sound instead of a diff attr

core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
```diff
@@ -356,12 +356,15 @@ public DestroyableHealthChange handleBlockChange(
...
-          for (MatchPlayer listener : this.getOwner().getMatch().getPlayers()) {
-            if (listener.getBukkit().getLocation().distance(blockLocation) > 64) {
-              listener.playSound(Sounds.OBJECTIVE_FIREWORKS_FAR);
-              listener.playSound(Sounds.OBJECTIVE_FIREWORKS_TWINKLE);
+          if (this.definition.hasSparksMatch()) {
+            // Players more than 64m away will not see or hear the fireworks, so play sound for them
+            for (MatchPlayer listener : this.getOwner().getMatch().getPlayers()) {
+              if (listener.getBukkit().getLocation().distance(blockLocation) > 64) {
+                listener.playSound(
+                    sound(key("fireworks.blast_far"), Sound.Source.MASTER, 0.75f, 1f));
```
Pablete1234: There's a reason all sounds are defined in the Sounds class, please follow the standard so multi-version works

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
arcadeboss: Hmm, I didn't change anything with respect to sound, so don't really understand the build error.

How did this go wrong?

arcadeboss:
> I think it needs a better name, maybe consider allowing sparks to have true, false, and 'near' that disables the global sound instead of a diff attr

I considered that, but felt wrong to change the type, but if you think it's okay, then alright.

core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
cswhite2000: You switched it from `Sounds.OBJECTIVE_FIREWORKS_FAR` to `sound(key("fireworks.blast_far"), Sound.Source.MASTER, 0.75f, 1f)`. We put all sounds constants in [Sounds](https://github.com/PGMDev/PGM/blob/dfa78531f9364c1194632ae6f4b453884b709548/util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java) since the sound keys are not version agnostic.

Pablete1234:
> but felt wrong to change the type

as long as it's backwards-compatible in xml, i don't think it's an issue

arcadeboss:
> as long as it's backwards-compatible in xml, i don't think it's an issue

I have updated the code to use an Enum to enforce the 3 spark type values of "true", "false" and "near"

The code is backwards compatible, so current uses of "true" and "false" remain the same, and "near" provides the new logic of having sparks, but no far match sounds.

There is a check for valid sparks values. and a corresponding exception thrown if non-valid value provided.

core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
```diff
@@ -33,15 +56,17 @@ public DestroyableFactory(
...
+      Element el)
+      throws InvalidXMLException {
     super(id, name, required, showOptions, owner, proximityMetric);
     this.region = region;
     this.materials = materials;
     this.destructionRequired = destructionRequired;
     this.modeList = modeList;
     this.showProgress = showProgress;
-    this.sparks = sparks;
+    this.sparks = SparksType.fromString(sparks, el);
```
Pablete1234: parsing should not occur on the constructor, the constructor should take in the already parsed attributes.

Additionally we have XMLUtils to parse enums, you likely didn't need to add your own fromString

core/src/main/java/tc/oc/pgm/destroyable/DestroyableModule.java
```diff
@@ -127,7 +127,9 @@ public DestroyableModule parse(MapFactory context, Logger logger, Document doc)
 
         boolean showProgress =
             XMLUtils.parseBoolean(destroyableEl.getAttribute("show-progress"), false);
-        boolean sparks = XMLUtils.parseBoolean(destroyableEl.getAttribute("sparks"), false);
+        String sparks = destroyableEl.getAttributeValue("sparks") != null
+            ? destroyableEl.getAttributeValue("sparks")
+            : "false";
```
Pablete1234: use XMLUtils.parseEnum instead, eg:

`SparkType sparks = XMLUtils.parseEnum(Node.fromAttr(destroyableEl, "sparks"), SparkType.class, SparkType.NONE);`

core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
```diff
@@ -10,15 +11,37 @@
 import tc.oc.pgm.modes.Mode;
 import tc.oc.pgm.teams.TeamFactory;
 import tc.oc.pgm.util.material.MaterialMatcher;
+import tc.oc.pgm.util.xml.InvalidXMLException;
 
 @FeatureInfo(name = "destroyable")
 public class DestroyableFactory extends ProximityGoalDefinition {
+  public static enum SparksType {
```
Pablete1234: To keep the behavior of `true` and `false` you can implement the `Aliased` interface in your enum to give them aliases:
```
  @NotNull
  @Override
  public Iterator<String> iterator() {
    return Iterators.forArray(name(), name);
  }
```

core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
```diff
@@ -10,15 +11,37 @@
...
+    SPARKS_ALL("true"),
+    SPARKS_NONE("false"),
+    SPARKS_NEAR("near");
```
Pablete1234:
```suggestion
    NONE("false"),
    NEAR("near"),
    ALL("true");
```
I think it makes more sense in order, from least to most (or if you prefer from most to least), but not with jumps. Also since this is a SparkType there's no need to prefix it with SPARKS all the time

core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
```diff
@@ -33,15 +56,17 @@ public DestroyableFactory(
       double destructionRequired,
       @Nullable ImmutableSet<Mode> modeList,
       boolean showProgress,
-      boolean sparks,
-      boolean repairable) {
+      String sparks,
+      boolean repairable,
+      Element el)
```
Pablete1234: As mentioned in another comment, don't pass the element here, do the parsing outside

```suggestion
      SparksType sparks,
      boolean repairable)
```

core/src/main/java/tc/oc/pgm/destroyable/DestroyableFactory.java
```diff
@@ -65,8 +90,12 @@ public boolean getShowProgress() {
...
   }
 
-  public boolean hasSparks() {
-    return this.sparks;
+  public boolean isSparksActive() {
+    return this.sparks != SparksType.SPARKS_NONE;
+  }
+
+  public boolean isSparksAll() {
+    return this.sparks != SparksType.SPARKS_NEAR;
```
Pablete1234: It would return true if you have set this to sparks none, it kinda makes no sense

```suggestion
    return this.sparks == SparksType.ALL;
```

arcadeboss:
Updated from review comments.   Abandoned aliasing, because couldn't make it work with parseEnum, but got everything else.

Pablete1234:
> Abandoned aliasing, because couldn't make it work with parseEnum, but got everything else.

Should be working out of the box, it's built-in already as long as you implement the aliasing interface in pgm like settings do

arcadeboss:
> > Abandoned aliasing, because couldn't make it work with parseEnum, but got everything else.
> 
> Should be working out of the box, it's built-in already as long as you implement the aliasing interface in pgm like settings do

This is what I implemented.  It complies, but at run time gives following error:

>  'sparks' attribute of 'destroyables' > 'destroyable' element @ line 2002 to 2004: Invalid format: 'false' is not a type.sparkstype (did you mean 'none'?)


See anything wrong?

> public static enum SparksType implements Aliased {
>    NONE("false"),
>    NEAR("near"),
>    ALL("true");
>
>    private final String name;
>
>    private SparksType(String name) {
>      this.name = name;
>    }
>
>    public String getName() {
>      return name;
>    }
>
>    @NotNull
>    @Override
>    public Iterator<String> iterator() {
>      return Iterators.forArray(name(), name);
>    }
>  }

arcadeboss:
The first attempt at iterator() above was based on the MapPoolType class enum which has the same iterator method that you shared.

But you, also mentioned following a settings enum, so I tried implementing like SettingKey class does it, but it gets the same results at run time that first one did.  It looks like this:

> public static enum SparksType implements Aliased {
>   NONE("false"),
>    NEAR("near"),
>    ALL("true");
>
>    private final List<String> aliases;
>
>    SparksType(String name) {
>      this(Collections.singletonList(name));
>    }
>
>    SparksType(List<String> aliases) {
>      this.aliases = ImmutableList.copyOf(aliases);
>    }
>
>    @NotNull
>    @Override
>    public Iterator<String> iterator() {
>      return aliases.iterator();
>    }
>  }

Pablete1234:
Looked into why aliased enums were being rejected myself and submitted a fix into the PR myself, should be good to go now

Pablete1234:
Also note, for backwards-compat we need to support "true", "yes", and "on" variants as those are all valid boolean parsing in pgm xml (ie: you could do `sparks="off"` or `sparks="on"` and they'd  work the same as true/false)

arcadeboss:
Wow, that's awesome, thank you for those updates!

I was poking around in the text parser, but never would have figured out those changes.  Nice on the extended enums.

Pablete1234 (approved):

PR #1408, opened 2024-10-09
Implement fluent parser & start migration

Pablete1234:
For a long while, PGM has had an explosion of parsing methods to support the different ways in which we can parse. Overloads upon overloads for parsing a node, an element, an attribute, required or with a default, maybe with a validation in there too.

That doesn't create the cleanest programming model, and it's why i want to move away from it. Welcome fluent parsing api:

```java
// Before (yes, all diff 10  overloads exist for filters):
var a1 = parseProperty(el, "filter"); // Defaults to null, non-obvious
var b1 = parseProperty(el, "filter", StaticFilter.ALLOW); // Defaults to ALLOW
var c1 = parseProperty(el, "filter", DynamicFilterValidation.ANY); // Defaults to null, non-obvious
var d1 = parseProperty(el, "filter", StaticFilter.ALLOW, DynamicFilterValidation.ANY);

var a2 = parseProperty(Node.fromChildOrAttr(el, "filter", "trigger"));
var b2 = parseProperty(Node.fromChildOrAttr(el, "filter", "trigger"), StaticFilter.ALLOW);
var c2 = parseProperty(Node.fromChildOrAttr(el, "filter", "trigger"), DynamicFilterValidation.ANY);
var d2 = parseProperty(Node.fromChildOrAttr(el, "filter", "trigger"), StaticFilter.ALLOW, DynamicFilterValidation.ANY);

var a3 = parseRequiredProperty(el, "filter");
var b3 = parseRequiredProperty(el, "filter", DynamicFilterValidation.ANY);

// After
var a1 = parser.filter(el, "filter").orNull();
var b1 = parser.filter(el, "filter").orAllow();
var c1 = parser.filter(el, "filter").dynamic().orNull();
var d1 = parser.filter(el, "filter").dynamic().orAllow();

var a2 = parser.filter(el, "filter", "trigger").orNull();
var b2 = parser.filter(el, "filter", "trigger").orAllow();
var c2 = parser.filter(el, "filter", "trigger").dynamic().orNull();
var d2 = parser.filter(el, "filter", "trigger").dynamic().orAllow();

var a3 = parser.filter(el, "filter").required();
var b3 = parser.filter(el, "filter").dynamic().required();
```
This is just a specific example for filters (listing all 10 former methods) but the new api also comes with more methods  like `optional(defaultValue)` of which `orAllow` is just an overload specifically for filters. The beauty is we can compose these calls without having an explosion in method overloads.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/util/xml/parsers/Builder.java
+++ b/core/src/main/java/tc/oc/pgm/util/xml/parsers/Builder.java
@@ -1,0 +1,114 @@
+package tc.oc.pgm.util.xml.parsers;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.jdom2.Element;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.util.function.ThrowingSupplier;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.Node;
+
+public abstract class Builder<T, B extends Builder<T, B>> {
+  protected final @Nullable Element el;
+  protected final String[] prop;
+
+  protected List<Validator<T>> validators;
+  protected boolean attr;
+  protected boolean child;
+  protected boolean self;
+
+  public Builder(@Nullable Element el, String... prop) {
+    this.el = el;
+    this.prop = prop;
+    this.attr = prop.length > 0;
+    this.child = prop.length > 0;
+    this.self = prop.length == 0;
+  }
+
+  /** Make it so this parser will only run on attributes, ignoring child elements */
+  public B attr() {
+    assert prop.length != 0;
+    attr = true;
+    child = false;
+    return getThis();
+  }
+
+  /** Make it so this parser will only run on child elements, ignoring attributes */
+  public B child() {
+    assert prop.length != 0;
+    attr = false;
+    child = true;
+    return getThis();
+  }
+
+  public B validate(Validator<T> validator) {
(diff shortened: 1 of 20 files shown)


core/src/main/java/tc/oc/pgm/util/xml/parsers/Builder.java
```diff
@@ -0,0 +1,114 @@
...
+    this.child = prop.length > 0;
+    this.self = prop.length == 0;
+  }
+
+  /** Make it so this parser will only run on attributes, ignoring child elements */
+  public B attr() {
+    assert prop.length != 0;
+    attr = true;
+    child = false;
+    return getThis();
```
cswhite2000, line 34: I don't understand the purpose of having `getThis()` instead of returning `this`.

Anyways this is not a blocking conern

cswhite2000 (approved):

core/src/main/java/tc/oc/pgm/util/xml/parsers/Builder.java
Pablete1234: Extended builders

If you return `this` in the parent `Builder`, you can't know what you're returning, because a parent can't know what it's subclass is.

In other words, doing:

`parser.filter(el, "filter").attr().orAllow()` wouldn't be  possible, because `attr()` (parent) would not return a `FilterBuilder` (which contains the `orAllow()` method

cswhite2000: Ah, that makes sense

PR #1412, opened 2024-10-24
Implement time penalties for switching

Pablete1234:
Adds options for PGM to penalize certain team-switches or rejoins. Aims to nerf certain kinds of unwanted behaviors on players:
1) team switching to stack (ie: switching team right after match start)
2) team swapping for your own benefit (ie: get wool from one team to another, or switching to the winning team)
3) rejoin spam to gather resources
4) rejoining to get your team over max (ie: join obs to make room for a non-premium user)

Some safeguards exist to avoid punishing legitimate switches, like switching to un-stack teams will avoid penalizing or a single rejoin to the same team wont penalize either.

Additionally what is considered a "stacked" team has been lowered from 20% to 10% for cased where all teams are equally sized (and increased to 25% when differing sizes)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -23,6 +23,7 @@ import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.EnumMap;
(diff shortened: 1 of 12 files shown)


cswhite2000 (commented):
Edit: nevermind saw the single rejoin part

cswhite2000 (approved):

PR #1416, opened 2024-10-29
Fix pitch and yaw in teleport action

### LINKED ISSUE #1415: Teleport action yaw and pitch are reversed
Given the following teleport action, the expected result is that the player will be teleported facing yaw 90, however instead they are teleported with their pitch set to 90.

```xml
<teleport x="16.5" y="18.5" z="-9.5" yaw="90"/>
```
image

Changing `yaw` to `pitch` then teleports the player facing yaw 90.

```xml
<teleport x="16.5" y="18.5" z="-9.5" pitch="90"/>
```

image

image

Pablete1234:
Fixes #1415


mitchts:
Can confirm now working as expected

cswhite2000 (approved):

PR #1420, opened 2024-11-01
Implement enchant item action

Pablete1234:
Works similar to replace-item action, but instead of replacing the item enchants it with a level which can be variables/formulas

```xml
<enchant-item enchantment="efficiency" level="some_var+1" ignore-metadata="true">
  <find material="diamond_pickaxe"/>
</enchant-item>
```

The inner element `find` is an item definition of what to search for in the inventory. Works the same as replace-item, where you can define extra item matching attributes like ignore-metadata and such.

Ontop of that you have the attributes:
`enchantment` the enchantment to apply
`level` a formula that says what level. Like all formulas it can be just a constant too like `level="5"`

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/actions/EnchantItemAction.java
+++ b/core/src/main/java/tc/oc/pgm/action/actions/EnchantItemAction.java
@@ -1,0 +1,45 @@
+package tc.oc.pgm.action.actions;
+
+import java.util.Objects;
+import org.bukkit.enchantments.Enchantment;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.PlayerInventory;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.inventory.ItemMatcher;
+import tc.oc.pgm.util.math.Formula;
+
+public class EnchantItemAction extends AbstractAction<MatchPlayer> {
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/action/actions/EnchantItemAction.java
```diff
@@ -0,0 +1,45 @@
...
+  public EnchantItemAction(ItemMatcher matcher, Enchantment enchant, Formula<MatchPlayer> level) {
+    super(MatchPlayer.class);
+    this.matcher = matcher;
+    this.enchant = enchant;
+    this.level = level;
+  }
+
+  @Override
+  public void trigger(MatchPlayer player) {
+    PlayerInventory inv = Objects.requireNonNull(player.getInventory());
```
cswhite2000, line 26: nit: The `Objects.requireNonNull` seems kind of pointless given it will throw a null pointer anyways on line 30 if its null.

cswhite2000 (approved):

core/src/main/java/tc/oc/pgm/action/actions/EnchantItemAction.java
Pablete1234: it is pointless, its just to make the IDE not complain about it being potentially nullable: it's not

PR #1421, opened 2024-11-03
Implement resize transformed region

Pablete1234:
# Resized regions

Resized regions take in another region and resize it by a certain amount. I think the best way to visualize this is with an example:

```xml
    <action id="fill-regions" expose="true" scope="match">
      <fill material="glass">
        <resize min="5,6,5" max="5,4,5" region="monument-region"/>
      </fill>
      <fill material="stained_glass:1">
        <resize min="1,1,1" max="1,1,1"  relative="true" region="monument-region"/>
      </fill>
      <fill material="obsidian">
        <region id="monument-region"/>
      </fill>
    </action>
```
This will fill an area 5 blocks sideways, 6 down and 4 up with glass, then will fill a region that expands by 100% of the original size to each direction with orange glass, and finally places back the obsidian.
Result is as follows:
image

### Attributes
- `min` How much to expand in the -x, -y and -z directions
- `max` How much to expand in the +x, +y and +z directions
- `relative` If the min&max should be relative to the original size of the region (proportion) or absolute (number of blocks)
  - `false` (default):  min & max are a number of blocks to expand in each direction
  - `true`: min & max represent a multiplier of the original size, eg: 2 = means 200% in each direction
  - Example: to expand a region by 50% you would `relative="true" min="0.5,0.5,0.5" max="0.5,0.5,0.5"`
- `region` Can be an attribute, or be a direct child.

## Relative resize

Expand the region by a factor of 100% in each direction (notice: relative="true"), so because the original is 2 tall, and 1 wide, it expands 2 up/down, and 1 to each side.
```xml
<resize min="1,1,1" max="1,1,1"  relative="true" region="monument-region"/> <!-- Filled with orange glass -->
```
image

## Absolute resize

Expand the region by an absolute 5 blocks sideways, 6 down, 4 up:
```xml
<resize min="5,6,5" max="5,4,5" region="monument-region"/> <!-- Filled with regular glass -->
```
image

## Caution with unions

**BE AWARE: Multiple children will act as a union. You probably do not want to use this region type with unions, as it will likely not create the effect you desire!!**

What happens with a union:
```xml
    <action id="fill-regions" expose="true" scope="match">
      <fill material="glass">
        <resize min="5,6,5" max="5,4,5" region="monument-region"/>
      </fill>
      <fill material="stained_glass:1">
        <resize min="1,1,1" max="1,1,1"  relative="true" region="monument-region"/>
      </fill>
      <fill material="obsidian">
        <region id="monument-region"/>
      </fill>
    </action>
    <action id="fill-regions-union" expose="true" scope="match">
      <fill material="glass">
        <resize min="5,6,5" max="5,4,5">
          <region id="monument-region1"/>
          <region id="monument-region2"/>
          <region id="monument-region3"/>
        </resize>
      </fill>
      <fill material="stained_glass:1">
        <resize min="1,1,1" max="1,1,1"  relative="true">
          <region id="monument-region1"/>
          <region id="monument-region2"/>
          <region id="monument-region3"/>
        </resize>
      </fill>
    </action>
```
**Relative result:**
The union is like 34 blocks across, so the relative resize by an extra factor of 100% in each direction, it triples the size and the result of the resize is a region 102 blocks wide (orange glass).
image

**Absolute result:**
The absolute region ends up looking pretty broken, because it's "only" enlarging by 5 blocks a region that is over 30 blocks across, meaning its a very small % increase, and barely creates extra area.
So the region did in fact grow by 4 blocks up and 5 sideways, and the area of the monument (the two blocks) enlarge to this 2x4 glass column:
image

# Other changes/fixes
The PR includes a couple minor fixes too:
- You can now use item matchers that ignore all meta *except* name/enchants, before pgm wouldn't allow it.
- Enchant item action calls setItem so that the update is instantly sent to the client instead of waiting a few ticks for the next inventory sync

### DIFF
--- a/core/src/main/java/tc/oc/pgm/regions/ResizedRegion.java
+++ b/core/src/main/java/tc/oc/pgm/regions/ResizedRegion.java
@@ -1,0 +1,63 @@
+package tc.oc.pgm.regions;
+
+import org.bukkit.util.Vector;
+import tc.oc.pgm.api.region.Region;
+import tc.oc.pgm.util.math.TransformMatrix;
+
+public class ResizedRegion extends TransformedRegion {
+  private final Vector min, max;
+  private final boolean relative;
+  private TransformMatrix matrix;
+  private TransformMatrix inverse;
+
+  public ResizedRegion(Region region, Vector min, Vector max, boolean relative) {
+    super(region);
+    this.min = min;
+    this.max = max;
+    this.relative = relative;
+  }
+
+  @Override
+  protected Vector transform(Vector point) {
+    if (matrix == null) ensureInitialized();
+    return matrix.transform(point);
+  }
+
+  @Override
+  protected Vector untransform(Vector point) {
+    if (inverse == null) ensureInitialized();
+    return inverse.transform(point);
+  }
+
+  @Override
+  public Bounds getBounds() {
+    ensureInitialized();
+    return super.getBounds();
+  }
+
+  private void ensureInitialized() {
+    if (matrix != null) return;
+
+    var oldBounds = region.getBounds();
+    var oldSize = oldBounds.getSize();
+
+    if (relative) {
+      min.multiply(oldSize);
+      max.multiply(oldSize);
+    }
+
+    this.bounds =
+        new Bounds(oldBounds.getMin().subtract(min), oldBounds.getMax().add(max));
+    var newSize = bounds.getSize();
+
+    this.matrix = TransformMatrix.concat(
+        TransformMatrix.untranslate(oldBounds.getMin()),
+        TransformMatrix.scale(newSize.clone().divide(oldSize)),
+        TransformMatrix.translate(bounds.getMin()));
+
+    this.inverse = TransformMatrix.concat(
+        TransformMatrix.untranslate(bounds.getMin()),
+        TransformMatrix.scale(oldSize.clone().divide(newSize)),
+        TransformMatrix.translate(oldBounds.getMin()));
+  }
+}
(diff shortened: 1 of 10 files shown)


util/src/main/java/tc/oc/pgm/util/math/TransformMatrix.java
```diff
@@ -0,0 +1,81 @@
+package tc.oc.pgm.util.math;
+
+import org.bukkit.util.Vector;
+
+public class TransformMatrix {
```
cswhite2000, line 5: I'll trust you on this one. I don't want to revisit linear algebra to this extent

core/src/main/java/tc/oc/pgm/regions/ResizedRegion.java
```diff
@@ -0,0 +1,63 @@
...
+      max.multiply(oldSize);
+    }
+
+    this.bounds =
+        new Bounds(oldBounds.getMin().subtract(min), oldBounds.getMax().add(max));
+    var newSize = bounds.getSize();
+
+    this.matrix = TransformMatrix.concat(
+        TransformMatrix.untranslate(oldBounds.getMin()),
+        TransformMatrix.scale(newSize.clone().divide(oldSize)),
```
cswhite2000, line 61: Should we be doing a check here for zero? 
 suppose the possible presence of infinity or -infinity could be intended Although NaN could pose issues

Also applies to the lines of code just below

Pablete1234: Yeah for empty regions it probs makes sense to just leave these matrices as identities as its irrelevant

cswhite2000 (approved):

PR #1426, opened 2024-11-06
Make item matching ignore modifier uuids

### LINKED ISSUE #1167: Shop items that cost an item with an <attribute> aren't purchasable
Take the following XML as an example.

 You first purchase the `Red Bowling Shoes` for a single emerald. These can then be traded up with 2 emeralds for the `Blue Bowling Shoes`. However, while the `<attribute ../>` modifier is present on the item you receive and the payment, PGM never registeres that you have the required payment needed to purchase the item.

Removing the `<attribute ../>` tags from the purchasable item and the payment requirement allows you to purchase the blue shoes.

```xml
<shops>
    <shop id="ball-shop" name="Bowling Ball Upgrades">
        <category id="boots" name="`rSpeed" material="leather boots">
            <item material="leather boots" color="993333" name="`rRed Bowling Shoes">
                <attribute operation="add" amount="0.1">generic.movementSpeed</attribute>
                <payment currency="emerald" price="1"/>
            </item>
            <item material="leather boots" color="334CB2" name="`rBlue Bowling Shoes">
                <attribute operation="add" amount="0.2">generic.movementSpeed</attribute>
                <payment currency="emerald" price="2"/>
                <payment price="1">
                    <item material="leather boots" color="993333" name="`rRed Bowling Shoes">
                        <attribute operation="add" amount="0.1">generic.movementSpeed</attribute>
                    </item>
                </payment>
            </item>
        </category>
    </shop>
</shops>
<shopkeepers>
    <shopkeeper name="`bAdam" shop="ball-shop">
        <point yaw="-90">-48,7,-0</point>
    </shopkeeper>
</shopkeepers>
```

The same behaviour is seen with non-armour items as well.

Using PGM v0.16-SNAPSHOT (git-7da1edd)

Pablete1234:
Fixes #1167 

Essentially makes it so item-matching in PGM checks attribute modifiers ignoring their UUID, and  if they're all equal then strips the modifiers from the meta that is passed down to the "equals" check

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/inventory/InventoryUtils.java
+++ b/util/src/main/java/tc/oc/pgm/util/inventory/InventoryUtils.java
@@ -25,6 +25,7 @@ import org.bukkit.inventory.meta.ItemMeta;
 import org.bukkit.inventory.meta.PotionMeta;
 import org.bukkit.potion.PotionEffect;
 import org.bukkit.potion.PotionEffectType;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.platform.Platform;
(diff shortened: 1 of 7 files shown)


util/src/main/java/tc/oc/pgm/util/inventory/InventoryUtils.java
```diff
@@ -188,6 +189,40 @@ default void setUnbreakable(ItemStack item, boolean unbreakable) {
     void applyAttributeModifiers(
         SetMultimap<Attribute, AttributeModifier> modifiers, ItemMeta meta);
 
+    boolean attributesEqual(ItemMeta meta1, ItemMeta meta2);
+
+    default boolean modifiersDiffer(
+        Collection<AttributeModifier> a, Collection<AttributeModifier> b) {
```
cswhite2000, line 195: nit: style: imo this method is going too far with shortening variables names and makes it a bit harder to follow

Admittedly its less of a problem here given that what the method does is pretty simple

cswhite2000 (approved):

util/src/main/java/tc/oc/pgm/util/inventory/InventoryUtils.java
Pablete1234: the method is rather trivial, so it's either A and B or modifiersA modifiersB, or modifiers1, modifiers2

given how trivial it is i don't think it matters much

PR #1442, opened 2024-11-26
Implemement `player-kit`

calcastor:
This PR implements/backports `player-kit` from ProjectAres. `player-kit` is a kit applied alive players in the match on spawn. On Overcast Network, the use case seemed to be applying a global `generic.attackSpeed` attribute multiplier to all players. This does not differ greatly from the intended use case now.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
@@ -3,6 +3,7 @@ package tc.oc.pgm.spawns;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
```diff
@@ -66,17 +74,22 @@ public SpawnModule parse(MapFactory factory, Logger logger, Document doc)
       FilterParser filterParser = factory.getFilters();
       SpawnParser parser = new SpawnParser(factory, new PointParser(factory));
       List<Spawn> spawns = Lists.newArrayList();
+      List<Kit> playerKits = new ArrayList<>();
```
TheMolkaPL, line 77:
```suggestion
      List<Kit> playerKits = Lists.newArrayList();
```

TheMolkaPL (changes requested):

core/src/main/java/tc/oc/pgm/spawns/SpawnModule.java
Pablete1234: This became obsolete in java 7, the main reason we do the bottom one is to avoid having to type the type arguments twice, since nowadays they're inferred i don't believe there's a benefit to using guava's method over just java's `new ArrayList<>();`

Pablete1234 (approved):

PR #1447, opened 2024-11-28
Add support for slotted attributes

### LINKED ISSUE #1446: Implement slot-specific attribute modifiers
In ProjectAres, it was possible to specify a slot in which an attribute is applied to an item, as exemplified below (an excerpt from the global include for Overcast Network):

```xml
<item-mods>
    <rule>
        <match>
            <material>diamond leggings</material>
        </match>
        <modify show-attributes="false">
            <attribute amount="6" slot="slot.armor.legs">generic.armor</attribute>
            <attribute amount="400" slot="slot.armor.legs">generic.armorToughness</attribute>
        </modify>
    </rule>
</item-mods>
```

In this example, this option is helpful because it avoids applying these attributes when the item is held, only applying it if worn.

Pablete1234:
Fixes #1446 

Allows `attributes` to specify a `slot` that the item has to be in to work:
```xml
<attribute amount="3" slot="mainhand">generic.attackDamage</attribute>
```
Note: This will only work when running in modern paper, legacy (1.8) minecraft didn't have support for this.

Slots can be specified using `EquipmentSlotGroup` which is the modern way, allowing for things like `armor` (for any armor piece) or `hand` (for either hand), but it also, for backwards-compat reasons, supports slots too like `slot.armor.head`.

## Modern values (recommended)
```
any -> any equipment slot, the default if you don't set one
mainhand
offhand
hand  -> either main or off-hand
feet
legs
chest
head
armor -> any armor piece
body -> Used for entities with no feet/legs/chest/head distinction (eg: horses when they wear armor)
``` 

## Legacy (deprecated) slots
They only work for backwards-compat reasons, and realistically do not need to be documented, focus on documenting the above ones instead.

All of these may have the `slot.` prefix, or not, like `weapon.mainhand` or `slot.weapon.mainhand`:
```
weapon.mainhand
weapon.offhand
armor.feet
armor.legs
armor.chest
armor.head
```


util/src/main/java/tc/oc/pgm/util/inventory/Slot.java
```diff
@@ -0,0 +1,500 @@
...
+    public static @Nullable Container forIndex(int index) {
+      return forIndex(Container.class, index);
+    }
+
+    Container(String key, int index) {
+      super(Container.class, key, index);
+    }
+  }
+
+  public abstract static class Player extends Slot {
```
cswhite2000, line 239: Interesting use of composition

cswhite2000 (approved):

PR #1457, opened 2024-12-11
Permit specifying world for default map variant

calcastor:
I am proposing this change however there are some differences to consider when comparing to ProjectAres.

In ProjectAres, the terrain module had more options that it currently does in present-day PGM, in that it was possible to specify both a world folder and to disable pre-match block physics. Further, in PGM we can define dimension in the terrain module, while ProjectAres moved this to the main map element as `dimension`, separate from the terrain module.

The idea with this change is that we are able to define a `world` conditionally for certain versions, while avoiding an entirely new variant for a map. An example where this may be used is with the map Azure: Azure has path blocks in its original 1.12 map world, however we have simply NBT converted the map to be used on OCC 1.8. Preferably, we do not need to create an entirely new variant just to use the original world on relevant server versions.

Alternatively to this PR, we could implement the world folder feature in the terrain module but given the pre-existing framework in place for variants this does seem more trivial.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
+++ b/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
@@ -375,7 +375,7 @@ public class MapInfoImpl implements MapInfo {
       if (variantEl == null) {
         this.variantId = DEFAULT_VARIANT;
         this.mapName = name;
-        this.world = null;
+        this.world = assertNotNull(root).getChildTextNormalize("world");
       } else {
         this.variantId = XMLUtils.parseRequiredId(variantEl);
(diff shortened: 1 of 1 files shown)


Pablete1234 (changes requested):
I think this needs to be thought thru a bit more.

It creates an inconsistency between the default variant, where it can have a `<world>folder</world>` while the other variants aren't allowed to have that.

Ontop of that this creates inconsistencies with conditionals, as other map variants would not know about the world of the default variant, this was already an issue before with allowing slugs to be defined in the main section of the xml instead of within the variant

Pablete1234:
Closing in favor of #1512

PR #1459, opened 2024-12-15
feat: support switch-case replacements and definitions at root level

TTtie:
Adds support for switch-case style replacement in message actions, inspired by pattern matching found in modern programming languages, for example (preliminary syntax, some details may be subject to change):

```xml
<message text="The outcome is {outcome}">
    <replacements>
        <!-- 
             value supports formulas and may be omitted 
             (in this case, the match attribute in <case>
             is disallowed)
        -->
        <switch id="outcome" value="variable">
            <!-- matches if the formula result is 0 -->
            <case match="0" result="Outcome 1"/>
            
            <!-- matches if the formula result is 1 AND only-team-1 allows -->
            <case match="1" filter="only-team-1" result="`aOutcome 2"/>
            
            <!-- matches if only-team-2 allows regardless of formula result -->
            <case filter="only-team-2" result="`bOutcome 3"/>
            
            <!--
                full syntax (matches if the formula result is between 2 and 5,
                and the inner filter allows)
            -->
            <case>
                <match>2..5</match>
                <filter>
                    <all id="not-moving">
                        <variable var="player.vel_x">0</variable>
                        <variable var="player.vel_y">..0</variable>
                        <variable var="player.vel_z">0</variable>
                    </all>
                </filter>
                <result>`cOutcome 4</result>
            </case>
            
            <!-- a fallback value if no cases match (defaults to an empty component) -->
            <fallback>`dA fallback outcome</fallback>
        </switch>
    </replacements>
</message>
```

Additionally, replacements can now be specified at the root level:
```xml
<map>
    <replacements>
        <!--
            scope is required for all replacements except <player />
            and it will be enforced if specified
        -->
        <decimal id="decimal" value="money" scope="player" format="0.00"/>
        <switch id="team" scope="team">
            <case filter="only-team-1" result="`aLime"/>
        </switch>
        
        <!-- 
            This will error as only-team-1 is team-scoped
            Not true as of commit 54666274 anymore. `only-team-1` will match as it abstains.
        -->
        <switch id="team2" scope="match">
            <case filter="only-team-1" result="`aLime"/>
        </switch>
    </replacements>
    <actions>
        <action id="action" scope="match">
            <switch-scope inner="player">
                <message text="The variable is {variable} and your team is {team}!">
                    <replacements>
                        <!--
                            the syntax for replacement references is
                            <replacement id="local-id">global-id</replacement>
                        -->
                        <replacement id="variable">decimal</replacement>
                        <replacement id="team">team</replacement>
                    </replacements>
                </message>
            </switch-scope>
        </action>
    </actions>
</map>
```

---

Launch checklist:
- [x] Cannot use higher-scope filters in `<case>` than defined in `<switch>`
- [x] Cannot specify higher-scope replacements than defined in `<action>`
- [x] Cannot use lower-scope variables in `<switch>` value
- [x] Validation???
  - [x] Using a reference of lower scope replacement should not work in a higher-scope action
  - [x] Declaring a lower scope replacement in a higher scope action should not work?

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
+++ b/core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
@@ -1,0 +1,26 @@
+package tc.oc.pgm.action.replacements;
+
+import net.kyori.adventure.text.ComponentLike;
+import tc.oc.pgm.api.feature.FeatureDefinition;
+import tc.oc.pgm.filters.Filterable;
+
+/** A replacement object provides replacement text in message actions. */
+public interface Replacement extends FeatureDefinition {
+  /**
+   * Tests if the replacement can be used with the passed filterable.
+   *
+   * @param filterable The filterable to test
+   * @return Whether the replacement can be used with the passed filterable
+   */
+  default boolean canUse(Class<? extends Filterable<?>> filterable) {
+    return true;
+  }
+
+  /**
+   * Creates a replacement component tailored to the given filterable.
+   *
+   * @param filterable The filterable to use when creating the replacement component
+   * @return The replacement component
+   */
+  ComponentLike get(Filterable<?> filterable);
+}
--- a/core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
+++ b/core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
@@ -3,18 +3,20 @@ package tc.oc.pgm.action.replacements;
 import net.kyori.adventure.text.ComponentLike;
 import tc.oc.pgm.api.feature.FeatureDefinition;
 import tc.oc.pgm.filters.Filterable;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.Node;
 
 /** A replacement object provides replacement text in message actions. */
 public interface Replacement extends FeatureDefinition {
   /**
    * Tests if the replacement can be used with the passed filterable.
    *
-   * @param filterable The filterable to test
-   * @return Whether the replacement can be used with the passed filterable
+   * @param scope The filterable to test
+   * @param node The node reference used for parsing
+   * @throws InvalidXMLException if the replacement cannot be used with the passed scope
    */
-  default boolean canUse(Class<? extends Filterable<?>> filterable) {
-    return true;
-  }
+  default void validate(Class<? extends Filterable<?>> scope, Node node)
+      throws InvalidXMLException {}
 
   /**
    * Creates a replacement component tailored to the given filterable.
--- a/core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
@@ -1,0 +1,164 @@
+package tc.oc.pgm.action.replacements;
+
+import static net.kyori.adventure.text.Component.empty;
+import static net.kyori.adventure.text.Component.text;
+
+import com.google.common.collect.Range;
+import java.lang.reflect.Method;
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.Set;
+import net.kyori.adventure.text.Component;
+import org.jdom2.Element;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.api.filter.Filter;
+import tc.oc.pgm.api.filter.Filterables;
+import tc.oc.pgm.api.map.factory.MapFactory;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.features.FeatureDefinitionContext;
+import tc.oc.pgm.filters.Filterable;
+import tc.oc.pgm.filters.matcher.StaticFilter;
+import tc.oc.pgm.util.MethodParser;
+import tc.oc.pgm.util.MethodParsers;
+import tc.oc.pgm.util.named.NameStyle;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.Node;
+import tc.oc.pgm.util.xml.XMLFluentParser;
+
+public class ReplacementParser {
+  private static final NumberFormat DEFAULT_FORMAT = NumberFormat.getIntegerInstance();
+  private final Map<String, Method> methodParsers =
+      MethodParsers.getMethodParsersForClass(getClass());
+  private final FeatureDefinitionContext features;
+  private final XMLFluentParser parser;
+  private final boolean isTopLevel;
+
+  public ReplacementParser(MapFactory factory, boolean topLevel) {
+    features = factory.getFeatures();
+    parser = factory.getParser();
+    isTopLevel = topLevel;
+  }
+
+  public ReplacementParser(MapFactory factory) {
+    this(factory, false);
+  }
+
+  public <B extends Filterable<?>> Replacement parse(Element el, @Nullable Class<B> scope)
+      throws InvalidXMLException {
+    Method parser = getParserFor(el);
+    if (parser != null) {
+      try {
+        return (Replacement) parser.invoke(this, el, scope);
+      } catch (Exception e) {
+        throw InvalidXMLException.coerce(e, new Node(el));
+      }
+    } else {
+      throw new InvalidXMLException("Unknown replacement type: " + el.getName(), el);
+    }
+  }
+
+  public Set<String> replacementTypes() {
+    return methodParsers.keySet();
+  }
+
+  protected Method getParserFor(Element el) {
+    return methodParsers.get(el.getName().toLowerCase());
+  }
+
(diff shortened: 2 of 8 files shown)


core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
```diff
@@ -0,0 +1,27 @@
...
+   */
+  Class<S> getScope();
+
+  /**
+   * Creates a replacement component tailored to the given audience.
+   *
+   * @param audience The audience to use when replacing
+   * @return The replacement component
+   */
+  ComponentLike replace(S audience);
```
Pablete1234: You're not *replacing* you're creating a component, so i feel calling this method replace is a bit misleading
Maybe `create` or `get` would fit better

core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @param <S> The scope of the replacement
+ */
+public interface Replacement<S extends Audience> extends FeatureDefinition {
+  /**
+   * Gets the scope of the replacement.
+   *
+   * @return The scope of the replacement
+   */
+  Class<S> getScope();
```
Pablete1234: You can likely get away with not having a scope, but rather a *checking* function, since some replacements dont need any specific scope.

Think of it in the same spirit as `respondsTo` in filters.

Reason behind this is that if a replacement works for `Match` it should also work for party or player, there's no reason it shouldn't, at least not with the current replacements. If in the future a new replacement requires something else you can still implement that replacement as more strict, but for the ones we have it should be more lenient

core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
```diff
@@ -0,0 +1,193 @@
...
+          "Wrong scope defined for replacement, scope must be " + scope.getSimpleName(), el);
+    return scope;
+  }
+
+  @MethodParser("decimal")
+  public <T extends Filterable<?>> Replacement<T> parseDecimal(Element el, Class<T> scope)
+      throws InvalidXMLException {
+    scope = parseScope(el, scope);
+    Formula<T> formula = parser.formula(scope, el, "value").required();
+    var formatString = parser.string(el, "format").attr().orNull();
```
Pablete1234: `.optional().map(DecimalFormat::new).orElse(DEFAULT_FORMAT);`

core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
```diff
@@ -0,0 +1,193 @@
...
+
+  @MethodParser("player")
+  public <T extends Filterable<?>> Replacement<T> parsePlayer(Element el, Class<T> scope)
+      throws InvalidXMLException {
+    var variable = parser.variable(el, "var").scope(MatchPlayer.class).singleExclusive();
+    var fallback = parser.formattedText(el, "fallback").optional(empty());
+    var nameStyle = parser.parseEnum(NameStyle.class, el, "style").optional(NameStyle.VERBOSE);
+    // Double cast is required, otherwise the compiler complains about type incompatibility
+    @SuppressWarnings("unchecked,RedundantCast")
+    Class<T> _scope = scope == null ? (Class<T>) (Object) Filterable.class : scope;
```
Pablete1234: You're doing the most scope-tricks for the replacement that needs them the least. The player replacement can literally work with any scope, it shouldnt even need to do any figuring-out

core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      if (valueRange != null && formula == null) {
+        throw new InvalidXMLException(
+            "A match attribute is specified but there's no switch value to bind to", innerEl);
+      }
+
+      var result = parser.formattedText(innerEl, "result").required();
+
+      if (filter != null) {
```
Pablete1234: This isn't the place to write this code, make this part of the filter parser so you can `parser.filter(el, "filter").respondsTo(scope).orNull()`

core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
```diff
@@ -0,0 +1,193 @@
...
+                    n);
+              }
+            },
+            Node.fromChildOrAttr(innerEl, "filter"));
+      }
+
+      branches.add(new SwitchReplacement.Branch(
+          result,
+          valueRange == null ? Range.all() : valueRange,
+          filter == null ? StaticFilter.ALLOW : filter));
```
Pablete1234: You can use `orAllow()` at the end of the filter parser to avoid the nullcheck

core/src/main/java/tc/oc/pgm/action/replacements/Replacement.java
```diff
@@ -0,0 +1,27 @@
...
+import net.kyori.adventure.text.ComponentLike;
+import tc.oc.pgm.api.feature.FeatureDefinition;
+import tc.oc.pgm.util.Audience;
+
+/**
+ * A replacement object provides replacement text in message actions.
+ *
+ * @param <S> The scope of the replacement
+ */
+public interface Replacement<S extends Audience> extends FeatureDefinition {
```
Pablete1234: scopes should be `extends Filterable<?>`

core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
```diff
@@ -0,0 +1,193 @@
...
+      throws InvalidXMLException {
+    Class<?> definitionScope = definition.getScope();
+    if (!definitionScope.isAssignableFrom(scope))
+      throw new InvalidXMLException(
+          "Wrong replacement scope, got "
+              + definitionScope.getSimpleName()
+              + " but expected "
+              + scope.getSimpleName(),
+          el);
+  }
```
Pablete1234: As mentioned before, relegate this to a method within the replacement; let it decide what the logic for rejecting a scope should be

Pablete1234 (changes requested):

Pablete1234 (dismissed):

calcastor:
Testing with changes from [this repo](https://github.com/TTtie/occ-community-maps/commits/refactor/use-switch-case-replacements/), with the exception of Grand Dueling and Chromatic Conquest which differ too much upstream for me to rebase the changes right now, everything seems to work as intended.

core/src/main/java/tc/oc/pgm/util/xml/parsers/FilterBuilder.java
```diff
@@ -27,6 +29,26 @@ public FilterBuilder dynamic(Class<? extends Filterable<?>> type) {
...
+        filter,
+        (definition, n) -> {
+          if (!definition.respondsTo(type)) {
+            throw new InvalidXMLException(
+                "Expected a filter that can respond to "
+                    + type.getSimpleName()
+                    + ". The filter "
+                    + definition.getDefinitionType().getSimpleName()
+                    + " only responds to "
+                    + Filterables.scope(definition).getSimpleName()
```
Pablete1234: This can and will throw for filters that do not belong to any scope, which is possible in non-dynamic filters.

core/src/main/java/tc/oc/pgm/action/replacements/ReplacementParser.java
```diff
@@ -0,0 +1,153 @@
...
+    var branches = new ArrayList<SwitchBranch>(children.size());
+
+    for (var innerEl : children) {
+      var valueRange = formula != null ? parser.doubleRange(innerEl, "match").orNull() : null;
+      var filter = parser.filter(innerEl, "filter").optional(() -> {
+        if (valueRange == null)
+          throw new InvalidXMLException(
+              "At least a filter or a match attribute must be specified", innerEl);
+        return StaticFilter.ALLOW;
+      });
```
Pablete1234: If you don't specify a value, and do specify a match, you could end up with a weird error message:

"Unused node 'match'" alongside an "At least a filter or a match..." which may leave you puzzled if you were a mapdev

maybe an option would be:
```suggestion
      var filterParse = parser.filter(innerEl, "filter");
      var filter = valueRange == null ? filterParse.required() : filterParse.orAllow();
```
which will just mark filter as required (without mentioning match at all) if you've not specified match.
not sure if it's really better or worse, maybe the err message should be different depending on if value is specified or not? i really don't know what's better

TTtie: > If you don't specify a value, and do specify a match, you could end up with a weird error message:
> 
> "Unused node 'match'" alongside an "At least a filter or a match..." which may leave you puzzled if you were a mapdev

That will never happen. Parsing errors interrupt the whole map loading process, and unused node warnings are the very last step in that process. The error message probably needs a rewrite given that `match="..."` is ignored when there's no value though.

Pablete1234 (approved):

PR #1467, opened 2024-12-29
Track assist count within player stats

Pugzy:

PR #1469, opened 2024-12-31
Make assists part of stats

Pablete1234:
Supercedes #1467 , makes it so assists are fully integrated into pgm, showing on match end stats, tablist, etc

Adding assists made certain things not work out as well as they used to (eg: "Your stats: " line) so they were modified and enhanded, here's how it looks:

image

image

- Multiple players can be shown for a stat now
- If over 3 players have a stat, it instead goes to "several Players" with a hover
- If over 10 players have a stat, it will just show "many, many players" and no hover (no point in showing the 20 ppl who all got 1 death in blitz)
- Instead of showing "Your stats", they show as "(You: x)"
- "You" is omitted if you're in the list of top players for that stat
- If a stat max is 0 , it is no longer displayed at match end

### DIFF
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -1,30 +1,31 @@
 package tc.oc.pgm.stats;
 
+import static net.kyori.adventure.text.Component.empty;
 import static net.kyori.adventure.text.Component.text;
 import static net.kyori.adventure.text.Component.translatable;
-import static net.kyori.adventure.text.event.HoverEvent.showText;
-import static tc.oc.pgm.util.nms.PlayerUtils.PLAYER_UTILS;
 import static tc.oc.pgm.util.player.PlayerComponent.player;
 import static tc.oc.pgm.util.text.NumberComponent.number;
+import static tc.oc.pgm.util.text.TextFormatter.list;
 
+import com.google.common.collect.Collections2;
 import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Lists;
 import com.google.common.collect.Table;
 import com.google.common.collect.Tables;
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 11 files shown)


core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -238,33 +236,21 @@ public void onFlagDrop(FlagStateChangeEvent event) {
...
-            .getLocation()
-            .distance(((ProjectileInfo) event.getDamageInfo()).getOrigin()));
-      }
-
-      murdererStats.onMurder();
+    getPlayerStat(victim).onDeath(victim);
+
+    if (event.isChallengeKill()) {
+      MatchPlayerState killer = event.getKiller();
+      assert killer != null;
```
cswhite2000, line 246: Why do you have an assert here?

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -238,33 +236,21 @@ public void onFlagDrop(FlagStateChangeEvent event) {
...
+      PlayerStats murdererStats = getPlayerStat(killer);
+      if (event.getDamageInfo() instanceof ProjectileInfo projectile)
+        murdererStats.setLongestBowKill(victim.getLocation().distance(projectile.getOrigin()));
+      murdererStats.onMurder(killer.getPlayer().orElse(null));
+    }
 
-      sendPlayerStats(murderer, murdererStats);
+    if (event.isChallengeAssist()) {
+      MatchPlayerState assister = event.getAssister();
+      assert assister != null;
```
cswhite2000, line 255: and here?

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
Pablete1234: mostly to keep the IDE quiet, if it's a challenge kill we know for certain it's non-null. null killer would be a NONE relationship which doesn't qualify as ENEMY, which is needed for it to be a challenge kill

cswhite2000 (approved):

PR #1477, opened 2025-01-08
Refresh map data when reloading

Pablete1234:
Doing `/pgm reload` will now not only reload the config, but also map data, which in a sense is almost like a config too, storing what maps shouldn't play right away etc


Pugzy:
Should we move this functionality to its own command? Or require an additional flag to clear scores.

Running PGM reload can be done for other minor config change reasons than needing to to wipe everything.

cswhite2000 (approved):

Pablete1234:
> Running PGM reload can be done for other minor config change reasons than needing to to wipe everything.

PGM reload was ALREADY wiping all scores before.

Resyncing does NOT wipe scores, it reloads the values from the database

PR #1482, opened 2025-01-19
Fix shop issues with category filtering

Pablete1234:
Fixes bugs regarding filtering categories in the shop:
- Categories with filters would cause index out of bounds exceptions
- When the categories change due to filters, they get updated instead of keeping the same list until re-opening
- Paginated Categories now properly highlight when going back and forth between pages

### DIFF
--- a/core/src/main/java/tc/oc/pgm/shops/menu/ShopMenu.java
+++ b/core/src/main/java/tc/oc/pgm/shops/menu/ShopMenu.java
@@ -10,6 +10,7 @@ import fr.minuskube.inv.ClickableItem;
 import fr.minuskube.inv.content.InventoryContents;
 import fr.minuskube.inv.content.Pagination;
 import fr.minuskube.inv.content.SlotIterator;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/shops/menu/ShopMenu.java
```diff
@@ -142,35 +144,53 @@ private void setCategory(Category category, int slot) {
...
-  private ClickableItem[] getCategoryItems() {
-    ClickableItem[] items = new ClickableItem[shop.getVisibleCategories(getViewer()).size()];
-    for (int i = 0; i < shop.getCategories().size(); i++) {
-      items[i] = getCategoryItem(shop.getCategories().get(i));
+  private boolean updateCategories() {
+    boolean anyChanged = false;
+    for (CategoryItem category : allCategories) {
+      if (category.changed()) anyChanged = true;
+    }
+    if (categories != null && !anyChanged) return false;
```
cswhite2000, line 152: nit: Assuming we don't need to run `category.changed()` for every category every time:
```suggestion
    if (categories != null && allCategories.stream().noneMatch((CategoryItem::changed)))
      return false;
```

Pablete1234: We do need to check all of them, since that will update the `active` flag on the category.

If we do a `noneMatch` as soon as one is changed, all following ones will not get their filter updated.

For most cases, you'll check `changed` on all categories and none will have changed, in which case that's where it ends. In the odd case that one or more have changed, you will need to re-build the lists etc

cswhite2000 (approved):

PR #1487, opened 2025-02-02
Add ScoreCause to CompetitorScoreChangeEvent and MatchPlayerScoreEvent

OhPointFive:
## Score Causes

This PR adds a `ScoreCause` to the events `CompetitorScoreChangeEvent` and  `MatchPlayerScoreEvent`.

Possible `ScoreCause`s are:
-  `CONTROL_POINT`
 - `DEATH`
 - `FLAG_CAPTURE`
 - `FLAG_TICK`
 - `KILL`
 - `SCOREBOX`
 - `VARIABLE`
 
 ## Motivation
 
In our original use case, I need a way to know when a player has gone into a scorebox, and for how many points they scored. A reasonable alternative to this would be to add an event for when a player enters a scorebox.

This approach seems more generally useful, though, and could be used for making an interactive graph of the match score, for example.


core/src/main/java/tc/oc/pgm/score/ScoreCause.java
```diff
@@ -0,0 +1,11 @@
+package tc.oc.pgm.score;
+
+public enum ScoreCause {
+  CONTROL_POINT,
+  DEATH,
+  FLAG_CAPTURE,
+  FLAG_TICK,
```
Pablete1234: Why flag capturing and ticking is separate, but control point capturing (owning) and ticking isn't?

OhPointFive: I didn't realize that owning behaviour even existed so I didn't pay enough attention to the relevant code — thank you for catching this!

I've split it into `CONTROL_POINT_TICK`, `CONTROL_POINT_OWNED`, and `CONTROL_POINT_LOST`. I'm not sure if distinguishing between the last two is necessary since you can tell by the sign of the score change.

I left `CONTROL_POINT_CAPTURED`, the parallel of the flag `ScoreCause`, undefined for now. I think it might be useful if control points could grant a fixed score on capture that's _not_ revoked when ownership is lost (which CTP maps currently accomplish in a sort of hacky way), and then such a cause could be used then.

I don't have strong opinions on the divides or names if you think something else would be better.

Pablete1234: I guess i just noticed in flags there's also a bit of inconsistency, where ticking could be at a post or while being carried.

I guess we could split them all:
```
CONTROL_POINT_TICK
CONTROL_POINT_OWNED
CONTROL_POINT_LOST
FLAG_CAPTURED
FLAG_CARRIED_TICK
FLAG_RETURNED_TICK
```

the owned & lost for controlpoints i don't really have a strong opinion for merging or splitting tbh

OhPointFive: I've split those up too — may as well be as specific as we can.

Pablete1234 (approved):
LGTM, just missing the build which seems to have failed due to a dependency

PR #1488, opened 2025-02-03
Prevent players from getting doors as observers

OhPointFive:
Players can place doors in observer mode, and the top half remains client-side due to a desync bug which lets them more easily block glitch around.

PGM has the following code, I believe, to prevent players from acquiring blocks that let them create desyncs to block glitch:
```java
    if (BAD_TYPES.contains(item.getType()) || item.getData() instanceof Door) {
      event.setCancelled(true);
    }
```

Unfortunately, at least in 1.8, `instanceof Door` only works on the block materials, not the item materials. I've left it as perhaps it works for higher versions.

I have expanded `BAD_TYPES` to include all the doors available in 1.8.

For now I've not included modern version doors (e.g. copper), since I'm not sure if this desync bug exists in modern versions, I'm not sure if the existing check gets the doors in modern versions, and I'm not set up to test modern versions.

The existing `IRON_DOOR` in `Materials` is an iron door _block_ in 1.8. This is used & needed for some iron door map conversion fix — I tried changing it to the item and it filled maps with barrier blocks upon loading. Thus I have made a new `IRON_DOOR_ITEM` in `Materials` and have named the other doors to match, renaming the old `WOOD_DOOR`.

Some inconsistencies in the door naming in sportpaper:\
`Material.WOOD_DOOR` is the item and `Material.WOODEN_DOOR` is the block.\
`Material.IRON_DOOR` is the item and `Material.IRON_DOOR_BLOCK` is the block.\
`Material.SPRUCE_DOOR_ITEM` is the item and `Material.SPRUCE_DOOR` is the block.\
The rest match the spruce door's pattern.

I have **not** tested this in modern versions yet. I'm going to have it tested soon — please don't merge it before then.

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/material/Materials.java
+++ b/util/src/main/java/tc/oc/pgm/util/material/Materials.java
@@ -53,7 +53,7 @@ public interface Materials {
       .build();
 
   MaterialMatcher DOOR_ITEMS = MaterialMatcher.builder()
-      .addAll(m -> m.name().contains("DOOR") && !m.name().contains("TRAPDOOR") && !m.isBlock())
+      .addAll(m -> m.name().contains("_DOOR") && !m.isBlock())
       .build();
 
   // A set of item types which, when used to interact with the match environment by non-playing
--- a/util/src/main/java/tc/oc/pgm/util/material/Materials.java
+++ b/util/src/main/java/tc/oc/pgm/util/material/Materials.java
@@ -28,7 +28,13 @@ public interface Materials {
   Material SIGN = parse("SIGN", "OAK_SIGN");
   Material SKULL = parse("SKULL_ITEM", "SKELETON_SKULL");
   Material PLAYER_HEAD = parse("SKULL_ITEM", "PLAYER_HEAD");
-  Material WOOD_DOOR = parse("WOOD_DOOR", "OAK_DOOR");
+  Material WOOD_DOOR_ITEM = parse("WOOD_DOOR", "OAK_DOOR");
+  Material ACACIA_DOOR_ITEM = parse("ACACIA_DOOR_ITEM", "ACACIA_DOOR");
+  Material JUNGLE_DOOR_ITEM = parse("JUNGLE_DOOR_ITEM", "JUNGLE_DOOR");
+  Material DARK_OAK_DOOR_ITEM = parse("DARK_OAK_DOOR_ITEM", "DARK_OAK_DOOR");
+  Material SPRUCE_DOOR_ITEM = parse("SPRUCE_DOOR_ITEM", "SPRUCE_DOOR");
+  Material BIRCH_DOOR_ITEM = parse("BIRCH_DOOR_ITEM", "BIRCH_DOOR");
+  Material IRON_DOOR_ITEM = parse("IRON_DOOR");
   Material BOOK_AND_QUILL = parse("BOOK_AND_QUILL", "WRITABLE_BOOK");
   Material EYE_OF_ENDER = parse("EYE_OF_ENDER", "ENDER_EYE");
   Material FIREWORK = parse("FIREWORK", "FIREWORK_ROCKET");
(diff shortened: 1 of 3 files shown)


util/src/main/java/tc/oc/pgm/util/material/Materials.java
```diff
@@ -28,7 +28,13 @@ public interface Materials {
...
   Material SKULL = parse("SKULL_ITEM", "SKELETON_SKULL");
   Material PLAYER_HEAD = parse("SKULL_ITEM", "PLAYER_HEAD");
-  Material WOOD_DOOR = parse("WOOD_DOOR", "OAK_DOOR");
+  Material WOOD_DOOR_ITEM = parse("WOOD_DOOR", "OAK_DOOR");
+  Material ACACIA_DOOR_ITEM = parse("ACACIA_DOOR_ITEM", "ACACIA_DOOR");
+  Material JUNGLE_DOOR_ITEM = parse("JUNGLE_DOOR_ITEM", "JUNGLE_DOOR");
+  Material DARK_OAK_DOOR_ITEM = parse("DARK_OAK_DOOR_ITEM", "DARK_OAK_DOOR");
+  Material SPRUCE_DOOR_ITEM = parse("SPRUCE_DOOR_ITEM", "SPRUCE_DOOR");
+  Material BIRCH_DOOR_ITEM = parse("BIRCH_DOOR_ITEM", "BIRCH_DOOR");
+  Material IRON_DOOR_ITEM = parse("IRON_DOOR");
```
Pablete1234: Given you don't actually want all the different doors individually, just "all doors", make this work for newer versions too. We have utils that can get all the materials based off a predicate (see the weapons example a few lines below)

Your predicate can be just contains "DOOR" and !isBlock

Pablete1234:
>  I tried changing it to the item and it filled maps with barrier blocks upon loading

Prior to 1.8, the trick to get a "barrier" block was to use half iron doors, because the client did not render them. When migrating from 1.7 to 1.8, pgm was made to replace all half iron doors with barrier blocks, that's the history behind that

> I have not tested this in modern versions yet. I'm going to have it tested soon — please don't merge it before then.

Turned the PR into draft, when it's ready to be merged un-draft it

xRob3:
Don't worry guys this "issue" has been proven to be false rumors spread by the occ lizard cabal in attempt to sow divisiveness in the greater community. There is absolutely no block-glitching occuring with doors, trust me.

util/src/main/java/tc/oc/pgm/util/material/Materials.java
```diff
@@ -52,6 +52,17 @@ public interface Materials {
       .addNullable(Material.getMaterial("MACE"))
       .build();
 
+  MaterialMatcher DOOR_ITEMS = MaterialMatcher.builder()
+      .addAll(m -> m.name().contains("DOOR") && !m.name().contains("TRAPDOOR") && !m.isBlock())
```
Pablete1234:
```suggestion
      .addAll(m -> m.name().contains("_DOOR") && !m.isBlock())
```
wouldn't this do the same and avoid the two contains checks?

cswhite2000:
Couldn't we also just replicate the logic that prevents doors placed in deny regions from glitching out while participating?

OhPointFive:
> Couldn't we also just replicate the logic that prevents doors placed in deny regions from glitching out while participating?

This seems like a better solution to me, since that way if people still get a door it won't let them blockglitch. I'll try to figure out how this works for regions once I am able.

Pablete1234:
> > Couldn't we also just replicate the logic that prevents doors placed in deny regions from glitching out while participating?
> 
> This seems like a better solution to me, since that way if people still get a door it won't let them blockglitch. I'll try to figure out how this works for regions once I am able.

pretty sure all it does is force-send the block update for the block above where you placed, but i think this is good enough of a fix as it is, it's not like obs *need* to have doors or that it's some massive downgrade not having them

Pablete1234 (approved):

PR #1489, opened 2025-02-03
Expose kill and death scores

OhPointFive:
This makes the kill and death scores of a `ScoreMatchModule` publicly visible.

### Motivation

We use this to check if a map cares about kills or not.

Specifically, if someone enters a scorebox on a map where there is a kill score, we scale their droplet reward with the number of points the scorebox is worth, since the 10 point scorebox on BlockBlock is more significant than a 3 point scorebox on Facility. On maps with no kill score, the point value of the scorebox likely is not related to how important or difficult it is to score, so we use a fixed value instead.

Since we're using the kill score when looking at a scorebox, we can't easily get the kill score off a `MatchPlayerScoreEvent`.

The death score is also made public since it seems weird to only provide the one.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/score/ScoreDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/score/ScoreDefinition.java
@@ -10,7 +10,7 @@ import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.party.Competitor;
 import tc.oc.pgm.util.named.NameStyle;
 
-public record ScoreConfig(
+public record ScoreDefinition(
     int scoreLimit,
     int deathScore,
     int killScore,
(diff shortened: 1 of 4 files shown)


Pablete1234:
Maybe expose score config instead? It's essentially the "definition" equivalent that many other modes have (eg: CoreDefinition) that you can get simply by `.getDefinition()`

Would probably make sense to make the ScoreConfig be renamed to `ScoreDefinition`, ensure it has public getters for all fields, and expose it as part of the module in a `getDefinition`. Maybe turn it into a record too.

OhPointFive:
Looks like it's already a record — it has now been renamed and exposed

Pablete1234:
I forgot i had already turned it into a record when i last reworked scoring, so yeah that's good already

Pablete1234 (approved):

PR #1490, opened 2025-02-05
Log death messages

OhPointFive:
I find death messages very helpful when they are logged. It helps see what's going on on a server at a glance from a terminal, and it also helps find short clips in the logs since you can search by death messages.

I've just entirely enabled it in this PR since other match messages are logged by default. I'm happy to put it behind a config flag (defaulting whichever way is preferred) if some people wouldn't want this enabled.


Pablete1234 (approved):
lgtm

PR #1494, opened 2025-02-18
Adopt virtual components

Pablete1234:
PGM had a `RenderableComponent` that extended from component directly, and implemented custom logic to be rendered per-player. Adventure has finally added virtual components which allow us to do the same in a less hacky way, since we no longer need to extend the api, we can just use the virtual component api instead.


cswhite2000 (dismissed):

cswhite2000:
Seems there's a build failure

Pablete1234:
build fails due to using a snapshot adventure build, we can wait until they release to merge this

calcastor:
With this rebased on https://github.com/PGMDev/PGM/pull/1508, text for offline map authors appears grey rather than cyan.

Without this PR:
image

With this PR:
image

cswhite2000 (dismissed):

calcastor:
image

All usernames seem to be shown as "Unknown", though authors in `/map` appear as expected.

Pablete1234:
Added a workaround for the "Unknown" bug, which is caused by https://github.com/KyoriPowered/adventure/issues/1299

As far as i could tell, rendering of usernames after that seems to be fine in all the contexts i could see.

util/src/main/java/tc/oc/pgm/util/text/ComponentRenderer.java
```diff
@@ -1,38 +1,87 @@
...
-              .get(Identity.UUID)
-              .<CommandSender>map(Bukkit::getPlayer)
-              .orElse(Bukkit.getConsoleSender());
-      component = ((RenderableComponent) component).render(sender);
-    }
+  protected @NotNull Component renderVirtual(
+      @NotNull VirtualComponent vc, @NotNull Pointered pointer) {
+    var factory = contextFactory.get(vc.contextType());
+    if (factory == null)
+      throw new UnsupportedOperationException("Contest type not supported: " + vc.contextType());
```
cswhite2000:
```suggestion
      throw new UnsupportedOperationException("Context type not supported: " + vc.contextType());
```

cswhite2000 (dismissed):

cswhite2000 (approved):

PR #1495, opened 2025-03-02
fix: CycleMatchModule should not cycle erroneously when players drop below the `join.min-players` amount

jasoryeh:
This fixes an issue that most servers will not encounter, unless they changed the `join.min-players` setting.

`join.min-players` defines the minimum players to start a match, this option is default to `1` meaning most servers will run correctly if their configurations are not modified, however, previously if you change it to a value higher than 1 and you have `<=` that amount of players playing/participating, the matches will cycle/end erroneously when players leave/join mid-match due to an incorrect comparison that triggers a cycle if empty.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -197,7 +197,7 @@ public final class PGMConfig implements Config {
     this.maxExtraVotes = parseInteger(config.getString("votes.max-extra-votes", "5"));
 
     this.minPlayers = parseInteger(config.getString("join.min-players", "1"));
-    this.endEmptyMatches = config.getBoolean("join.end-empty-matches", true);
+    this.endEmptyMatches = config.getBoolean("join.end-empty-matches", minPlayers > 0);
     this.limitJoin = parseBoolean(config.getString("join.limit", "true"));
     this.priorityKick = parseBoolean(config.getString("join.priority-kick", "true"));
     this.balanceJoin = parseBoolean(config.getString("join.balance", "true"));
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -90,7 +90,7 @@ public final class PGMConfig implements Config {
 
   // join.*
   private final long minPlayers;
-  private final long minPlayersMaintain;
+  private final boolean endEmptyMatches;
   private final boolean limitJoin;
   private final boolean priorityKick;
   private final boolean balanceJoin;
@@ -197,7 +197,7 @@ public final class PGMConfig implements Config {
     this.maxExtraVotes = parseInteger(config.getString("votes.max-extra-votes", "5"));
 
     this.minPlayers = parseInteger(config.getString("join.min-players", "1"));
-    this.minPlayersMaintain = parseInteger(config.getString("gameplay.min-players-maintain", "1"));
+    this.endEmptyMatches = config.getBoolean("join.end-empty-matches", true);
     this.limitJoin = parseBoolean(config.getString("join.limit", "true"));
     this.priorityKick = parseBoolean(config.getString("join.priority-kick", "true"));
     this.balanceJoin = parseBoolean(config.getString("join.balance", "true"));
@@ -586,8 +586,8 @@ public final class PGMConfig implements Config {
   }
 
   @Override
-  public long getMinimumPlayersMaintain() {
-    return this.minPlayersMaintain;
+  public boolean allowEndingEmptyMatches() {
+    return this.endEmptyMatches;
   }
 
   @Override
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -90,6 +90,7 @@ public final class PGMConfig implements Config {
 
   // join.*
   private final long minPlayers;
+  private final long minPlayersMaintain;
   private final boolean limitJoin;
   private final boolean priorityKick;
   private final boolean balanceJoin;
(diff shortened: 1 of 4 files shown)


Pablete1234:
I believe this works as intended
setting min-players = 2 should mean that no match should be playing without at least 2 players, that applies to both being able to start the match, as well as continuing the match.

If you wish certain maps to enforce a min player count before starting (eg: tournament-edition versions which require N players to start), you can use the `min` attribute of teams, which works similarly but won't end matches when they drop below

If you wish to change this behavior i think its most appropriate for it to be a separate option.

core/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java
```diff
@@ -47,7 +47,7 @@ public void startCountdown(@Nullable Duration duration) {
   public void onPartyChange(PlayerPartyChangeEvent event) {
     if (event.wasParticipating()
         && match.isRunning()
-        && match.getParticipants().size() < PGM.get().getConfiguration().getMinimumPlayers()) {
+        && match.getParticipants().isEmpty()) {
```
Pablete1234: This would end matches even if `min-players` is set to 0, which is often done in debug/test server scenarios so that your matches don't end on you if you rejoin

Pablete1234 (changes requested):
As mentioned in the comment, consider simply making this an extra config setting

core/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java
jasoryeh: Yes, this scenario makes sense- perhaps a separate configuration option is necessary to more clearly enforce a minimum participant count for continuing a match.

jasoryeh:
> If you wish certain maps to enforce a min player count before starting (eg: tournament-edition versions which require N players to start), you can use the `min` attribute of teams, which works similarly but won't end matches when they drop below

I understand the idea that this configuration might not make sense for enforcing minimum players as each map is different, however, this doesn't make sense from the configuration documentation comments, as the the `config.yml` is clear that the `join.min-players` is designated to enforce the minimum number of players to start a match:
https://github.com/PGMDev/PGM/blob/84bce746dd347722ffada0b8406817cc727cd579/core/src/main/resources/config.yml#L66-L68

Furthermore, using this value as double duty for default minimum player count and as a minimum-continuing-match-player-size creates unnecessary ambiguity.

> I believe this works as intended setting min-players = 2 should mean that no match should be playing without at least 2 players, that applies to both being able to start the match, as well as continuing the match.

I disagree, I believe a minimum player count limit for a match to continue should be described in a separate configuration not in the `join` section and rather in something like `gameplay` or `match`. The `join` section, in my opinion, should only be used to regulate only the pre-match -> start-of-match phases.

> If you wish to change this behavior i think its most appropriate for it to be a separate option.

I think a best case course of action would be to introduce new configuration to more clearly define an option for enforcing a minimum playing player count.

Pablete1234:
> I understand the idea that this configuration might not make sense for enforcing minimum players as each map is different, however, this doesn't make sense from the configuration documentation comments, as the the `config.yml` is clear that the `join.min-players` is designated to enforce the minimum number of players to start a match:
> 
> https://github.com/PGMDev/PGM/blob/84bce746dd347722ffada0b8406817cc727cd579/core/src/main/resources/config.yml#L66-L68
> 

Yeah the problem may be the config's naming, in modern pgm (1.9+) the config wasn't even in the `join` config
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/resources/config.yml#L139

and had a separate config for cycling when match is empty:
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/resources/config.yml#L122
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/cycle/CycleMatchModule.java#L210


In fact, digging into the history it seems that this 1.8-pgm version "simplified" the settings, renaming minimum-players to join.min-players, and removed the match-empty in favor of join.min-players = 0:
https://github.com/PGMDev/PGM/commit/12e9e602e3b7b3064d8a302719f5f5d5d4885ab7#diff-fd22f8387085c525f472af2f02ff3dd726cbecf8d416c1cf9639050d480efe5bR176

https://github.com/PGMDev/PGM/commit/12e9e602e3b7b3064d8a302719f5f5d5d4885ab7#diff-c44e8a03c6e51fede2607ed31511ee21a7022f2e615b7ee442acf97a3e0dd29fL48


So yeah, tldr: this was an intentional change to simplify the settings by merging the cycle.empty.enabled: true/false config into the minimum-players encoded as a 0 or a 1, with little regard to what it means for values >1.

Feel free to revert these so that we have the cycle enabled as a separate config

jasoryeh:
I made it a separate config value and just put it in `join.min-players-maintain` next to the original config that was being used. Feel free to suggest a better name because I wasn't too sure what to call it

Pablete1234 (commented):
So sorry for the late review, missed it

I think it would probs work best as a boolean property: `end-empty-matches`  or similar, and default to true when min-players is >=1, and false when min-players = 0
That's basically what the behavior was prior to the changes that broke this behavior, and i think it probably solves all issues. It is a bit weird to consider the case where non-empty match would end, and this realistically should be just for testing etc, which a boolean suffices

core/src/main/java/tc/oc/pgm/PGMConfig.java
```diff
@@ -196,6 +197,7 @@ public final class PGMConfig implements Config {
     this.maxExtraVotes = parseInteger(config.getString("votes.max-extra-votes", "5"));
 
     this.minPlayers = parseInteger(config.getString("join.min-players", "1"));
+    this.endEmptyMatches = config.getBoolean("join.end-empty-matches", true);
```
```suggestion
    this.endEmptyMatches = config.getBoolean("join.end-empty-matches", minPlayers > 0);
```
This would keep the default in-line with current behavior (where min players=0 is used to avoid it ending when empty)

Pablete1234 (changes requested):
Just one last small nit, otherwise looks good to me

Pablete1234 (approved):

PR #1496, opened 2025-03-05
Add built-in initial match loading

Pablete1234:
Currently pgm will only create the first match once a player attempts to log-in, this can be troublesome (eg: no MOTD because no match) and is the reason for other plugins adding this functionality (ie: https://github.com/applenick/AlwaysLoad), this bakes it right into pgm as there's no good reason for this not to be default behavior


cswhite2000 (dismissed):
No concerns with code, build is failing however

Pablete1234:
It fails due to a dependency not resolving from gh actions, see https://github.com/dmulloy2/ProtocolLib/issues/3353

cswhite2000 (approved):

PR #1497, opened 2025-03-06
Improve tracking of afk players

Pablete1234:
Currently afk status is determined mostly by if you move your camera, this can mean that a tp (ie: a cycle) can end the afk status even if you continue to be afk.

This PR makes it so certain actions do reset afk instantly (eg: clicking an item in an inventory, or sending a chat message or command) but other actions just add-up towards a reset, but need to happen within a certain time, so sparse moves once every many minutes won't add up to un-afk you.

Walking around (without moving the camera) can also get you out of afk status, as long as you're moving fast enough (ie: a typical water-circle anti-afk machine won't do to consider you not afk)

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/listener/AfkTracker.java
+++ b/util/src/main/java/tc/oc/pgm/util/listener/AfkTracker.java
@@ -2,12 +2,15 @@ package tc.oc.pgm.util.listener;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import org.bukkit.Bukkit;
+import org.bukkit.Location;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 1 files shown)


cswhite2000 (approved):

Pugzy:
> Walking around (without moving the camera) can also get you out of afk status, as long as you're moving fast enough (ie: a typical water-circle anti-afk machine won't do to consider you not afk)

I think this allowance may unintentionally mark players as non-afk.

Spawns that drop the player or players getting knocked around would after spawning would cause this to trigger.

Pablete1234:
> I think this allowance may unintentionally mark players as non-afk.
> Spawns that drop the player or players getting knocked around would after spawning would cause this to trigger.

Falling never counts, and you need to move quite a bit for it to count, you'd need to get comboed quite a bit for it to count

PR #1499, opened 2025-04-01
Allow registering player-based formulas to stats match module

chatasma:
This allows mapmakers to define player-based formulas as match statistics that can be shown in the match stats summary at the end of a match.

i.e. adding this to a Bridge map
```xml
<stats>
    <stat value="player_score" name="`6Goals" />
</stats>
```

(where `value` is a player-based formula)

Adds bridge goals as a stat that is appended at the end of the match stats summary

bridge_stat

Creating as a draft pull request to receive feedback and/or thoughts!


core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,87 @@
...
+    public final String key = "match.stats.type." + name().toLowerCase(Locale.ROOT);
 
     @Override
     public Component makeNumber(Number number) {
-      return translatable(blocks, number(number, YELLOW));
+      return number(number, this == DEATHS ? RED : GREEN);
+    }
+
+    @Override
+    public Component conciseComponent(StatHolder stats) {
```
Pablete1234: I don't see the need for this to change from `component` to `conciseComponent`, it's the same as it was as far as i can tell

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,87 @@
...
+    public Component buildMessage(Component valueComponent) {
+      return translatable(key, valueComponent);
     }
-  };
+  }
 
-  public final String key = "match.stats.type." + name().toLowerCase(Locale.ROOT);
-  public static final StatType[] VALUES = values();
+  final class VariableStatType implements StatType {
+    private static final String SEPERATOR = ":";
```
Pablete1234: Odd choice to keep the separator here, then merge it with a whitespace below

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,87 @@
...
+
+    @Override
+    public Component makeNumber(Number number) {
+      return number(number, GREEN);
+    }
+
+    @Override
+    public Component conciseComponent(StatHolder stats) {
+      return friendlyName.append(
+          text(SEPERATOR + " ").append(makeNumber(stats.getStat(this))).color(WHITE));
```
Pablete1234: Should probably have a translation key for generic stat instead, so the same way we have `match.stats.type.kills` introduce a `match.stats.type.generic` that defines the format, so like `{0}: {1}`, where one is the stat name and the other the value

core/src/main/java/tc/oc/pgm/stats/PlayerStats.java
```diff
@@ -61,17 +60,22 @@ public class PlayerStats implements StatHolder {
   private Duration longestFlagHold = Duration.ZERO;
   private Instant longestFlagHoldCache;
 
+  // Variable-based statistics
+  private final Map<Variable<?>, Supplier<Double>> variableStats;
```
Pablete1234: i don't really like PlayerStats being polluted with this. The point of PlayerStats is to hold state that exists nowhere else: how many kills, deaths, etc the player has. These variables already have somewhere else to be, they shouldn't be here.

It should be a higher level of abstraction that implements `StatHolder` that can direct the calls either to this, or to the variable-based one for non "Base" stat types

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -12,16 +10,9 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
Pablete1234: avoid wildcard imports, configure your IDE not to use them

core/src/main/java/tc/oc/pgm/variables/Variable.java
```diff
@@ -52,6 +54,14 @@ default boolean isIndexed() {
     int size();
   }
 
+  interface TrackedHolders<T extends Filterable<?>> extends Variable<T> {
+    Map<T, Double> getTrackedValues(MatchFeatureContext matchFeatureContext);
+  }
+
+  interface DefinesDefault<T extends Filterable<?>> extends Variable<T> {
+    double getDefaultValue();
+  }
```
Pablete1234: The last thing we need is more complexity on variables, please don't. Also you should probably not be needing MatchFeatureContext here, regular features should not be needing to interact with the context manually

core/src/main/java/tc/oc/pgm/stats/TeamStats.java
```diff
@@ -37,11 +37,14 @@ public TeamStats(Collection<PlayerStats> playerStats) {
...
-      case DAMAGE -> damageDone;
-      default -> Double.NaN;
+      case StatType.BaseStatType baseStatType -> switch (baseStatType) {
+        case KILLS -> teamKills;
+        case DEATHS -> teamDeaths;
+        case KILL_DEATH_RATIO -> teamKD;
+        case DAMAGE -> damageDone;
+        default -> Double.NaN;
+      };
+      case StatType.VariableStatType _variableStatType -> Double.NaN;
```
Pablete1234: do not use `_` in variable names

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,90 @@
...
+  private StatsModule(List<MapVariableStat> variables) {
+    this.variables = variables;
+  }
+
+  @Override
+  public @Nullable StatsMatchModule createMatchModule(Match match) throws ModuleLoadException {
+    final List<StatsMatchModule.VariableStat> variableStats = new ArrayList<>();
+    for (MapVariableStat mapVar : variables) {
+      if (mapVar.variable instanceof Variable.TrackedHolders<MatchPlayer>
+          && mapVar.variable instanceof Variable.DefinesDefault<MatchPlayer>) {
```
Pablete1234: if you need both of these, you're just asking for a DummyVariable, which i don't think is a good restriction

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,90 @@
...
+        return new StatsModule(Collections.emptyList());
+      } else {
+        Element additionalStatsElement = XMLUtils.getUniqueChild(statsElement, "additions");
+        List<MapVariableStat> statVars = new ArrayList<>();
+        if (additionalStatsElement != null) {
+          for (Element el : XMLUtils.getChildren(additionalStatsElement, "stat")) {
+            if (el.getAttribute("var") == null) {
+              throw new InvalidXMLException(
+                  "Expected 'var' attribute on custom stat definition", el);
+            }
```
Pablete1234: you do not need this, calling `.required()` on `parser.variable(el, "var").scope(MatchPlayer.class).required();` will already do this logic for you

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,90 @@
...
+            if (el.getAttribute("display-name") == null) {
+              throw new InvalidXMLException(
+                  "Expected 'displayName' attribute on custom stat definition", el);
+            }
```
Pablete1234: same here

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,90 @@
...
+
+            Component displayName = XMLUtils.parseFormattedText(el, "display-name");
+            Variable<MatchPlayer> varDef =
+                parser.variable(el, "var").scope(MatchPlayer.class).required();
+            if (varDef instanceof Variable.TrackedHolders<MatchPlayer>
+                && varDef instanceof Variable.DefinesDefault<MatchPlayer>) {
+              statVars.add(new MapVariableStat(varDef, displayName));
+            } else {
+              throw new InvalidXMLException(
+                  "Expected variable to be a general purpose player-scoped variable", el);
```
Pablete1234: What does "general purpose" even mean here? you're restricting to literally a single type of variable, dummy variables.
Imagine a mapmaker wants to use xp (gather-the-most-xp kinda map), or health (uhc-like) as the end of the match as the stat, this won't let them for no good reason.

Pablete1234 (commented):
Generally speaking having this feature added would be good, however, it needs some changes to how its implemented.

I think using variables for this is the wrong approach, and you should instead be using a formula. Formulas allow any set of variables to be mixed-and-matched as the map author desires, for example they could make it `value="player.health + player.food"` and their final score would be however many hearts and food they were able to keep.
This change alone would remove some of the overhead of dealing with variables, as it simplifies it to an expression that you can call to get the value of at the end of the match.

Ontop of that i'm not sure of the stats -> additions -> stat structure in xml, maybe simplify to:
```xml
<stats>
  <stat name="Score" value="player.health"/>
</stats>
```
i'm guessing that the only reason to have `additions` is a potential future where removing builtin stats is an option, but if that's the case we could probably have them in an attribute like `<stats remove="deaths"/>` or something along the lines

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: More or less arbitrary, I'll just do it all inline w/o this static constant

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: It is the same functionally, I more or less changed it for semantics since I have a separate interface method for building the stats display component as well. Can change it back to `component` though

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Cool can remove

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Cool can remove

core/src/main/java/tc/oc/pgm/stats/TeamStats.java
chatasma: Sure will remove the underscore

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Will address this in the similar comment below

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Yes it is restricting it to the dummy variable implementation.

> this won't let them for no good reason

Well, without this change they cannot use any variable 🙂
But in all seriousness, I figured we could always expand on this in a follow-up. Though, if you prefer allowing other types of variables i.e. player variables in the initial version that is a totally valid sentiment as well! I'll explore supporting all variable types and see what I can do

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: Sure

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Seems sensible can do

core/src/main/java/tc/oc/pgm/variables/Variable.java
chatasma: > The last thing we need is more complexity on variables, please don't.

This does not seem constructive, what is the actionable advice here? I added the "complexity" as a mechanism for the stats module to be able to source a default value for the aggregate statistic and getting the value map of the variable. Looking back I do not think I need this tracked values interface and will remove it. Let me know if you have a preferred way of sourcing a default value for the aggregate stat, i.e. I could change it so the XML element specifies a default value or if you have any other ideas I'm all ears!

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
Pablete1234: Its not about supporting different variable types, its about not using variables at all!! Variables are only needed in features that need write access, for read-only access use formulas instead, which is instant support of any kind of variable for the same price

core/src/main/java/tc/oc/pgm/stats/PlayerStats.java
chatasma: I think I understand where you are coming from on this. Is the recommendation that we create a new StatHolder that contains a references the formula, and have a reference to that stat holder here? Could be misunderstanding but I think I understand the general sentiment

chatasma:
> Generally speaking having this feature added would be good, however, it needs some changes to how its implemented.
> 
> I think using variables for this is the wrong approach, and you should instead be using a formula. Formulas allow any set of variables to be mixed-and-matched as the map author desires, for example they could make it `value="player.health + player.food"` and their final score would be however many hearts and food they were able to keep. This change alone would remove some of the overhead of dealing with variables, as it simplifies it to an expression that you can call to get the value of at the end of the match.
> 
> Ontop of that i'm not sure of the stats -> additions -> stat structure in xml, maybe simplify to:
> 
> ```
> <stats>
>   <stat name="Score" value="player.health"/>
> </stats>
> ```
> 
> i'm guessing that the only reason to have `additions` is a potential future where removing builtin stats is an option, but if that's the case we could probably have them in an attribute like `<stats remove="deaths"/>` or something along the lines

I agree with everything stated here generally, so:
* Change the stats to accept formulas instead of dummy variables
* Remove the `additions` element, we can designate adding/removing as a property anyway

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Yeah makes sense I get you on using the formulas

core/src/main/java/tc/oc/pgm/variables/Variable.java
chatasma: Removed these interfaces in favor of using formulas

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,87 @@
...
-  LONGEST_BOW_SHOT {
-    private final String blocks = key + ".blocks";
+public sealed interface StatType {
+  Component makeNumber(Number number);
+
+  Component buildMessage(Component valueComponent);
+
+  Component component(StatHolder stats);
+
+  enum BaseStatType implements StatType {
```
Pablete1234: `StatType.BaseStatType` is both very long, and redundant in its name, consider something shorter like:
```suggestion
  enum Builtin implements StatType {
```
so it simplifies to `StatType.Builtin`

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,87 @@
...
-      return damageComponent(number.doubleValue(), GREEN);
+    public Component buildMessage(Component valueComponent) {
+      return translatable(key, valueComponent);
     }
-  };
+  }
 
-  public final String key = "match.stats.type." + name().toLowerCase(Locale.ROOT);
-  public static final StatType[] VALUES = values();
+  final class FormulaStatType implements StatType {
```
Pablete1234:
```suggestion
  final class OfFormula implements StatType {
```
Similar as above, would simplify to `StatType.OfFormula`

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -290,14 +294,28 @@ public void onStatsDisplay(MatchStatsEvent event) {
...
-    if (verboseStats) stats.add(new AggStat<>(StatType.DAMAGE, 0d, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.BaseStatType.KILLS, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.BaseStatType.DEATHS, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.BaseStatType.ASSISTS, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.BaseStatType.BEST_KILL_STREAK, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.BaseStatType.LONGEST_BOW_SHOT, 0, new HashSet<>()));
+    if (verboseStats) stats.add(new AggStat<>(StatType.BaseStatType.DAMAGE, 0d, new HashSet<>()));
+    for (FormulaStat formulaStat : formulaStats) {
+      stats.add(new AggStat<>(
+          new StatType.FormulaStatType(formulaStat.formula, formulaStat.statDisplayName),
```
Pablete1234: instead of re-wrapping `FormulaStat` into a StatType, why not just make FormulaStat implement StatType ?

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -290,14 +294,28 @@ public void onStatsDisplay(MatchStatsEvent event) {
...
-    allPlayerStats.forEach((uuid, s) -> stats.replaceAll(stat -> stat.track(uuid, s)));
+    for (Map.Entry<UUID, PlayerStats> playerStats : allPlayerStats.entrySet()) {
+      FormulaStatHolder formulaStatHolder =
+          new FormulaStatHolder(getPlayerFormulaStatSuppliers(playerStats.getKey()));
+      StatHolder compositeStatHolder = (statType) -> switch (statType) {
+        case StatType.BaseStatType baseStatType -> playerStats.getValue().getStat(statType);
+        case StatType.FormulaStatType formulaStatType -> formulaStatHolder.getStat(statType);
+      };
+      stats.replaceAll(stat -> stat.track(playerStats.getKey(), compositeStatHolder));
+    }
```
Pablete1234: i don't like the need for so much complexity here, and i think it points to a bigger code smell

`StatHolder` was made as an abstraction over team vs player stats, and has a `getStat` method that exists for generalization over getting the built-in stats, but it doesn't really make sense for *all* types of stats that we're introducing (stats that come from other sources).

Maybe StatType should specify what type of holder it comes from, eg: `Builtin implements StatType<StatHolder>` while formula could `OfFormula implements StatType<MatchPlayer>`, and then we can simply change the signature of StatHolder to take in StatType.Builtin, as it's the only type it actually holds.

That way, in theory, you should be able to make a single class for the formula-typed stat, that implements StatType, and contains all the logic of how to get the stat in a self-contained way, instead of needing the whole getPlayerFormulaStatSuppliers intermediate set of objects and indirections, and leaves StatHolder untouched except for the rename from StatType to StatType.Builtin

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: It kind of makes sense but a little confused on some parts. Makes sense on keeping StatHolder to built-in stats

> single class for the formula-typed stat, that implements StatType, and contains all the logic of how to get the stat in a self-contained way

But the StatType just defines formatting for the stat type right? How can we retrieve the value of the formula stat type in a unified way? If not in a unified way, should we typecheck the stat type to determine how to retrieve the stat? And then how to supply it to the AggStat to track it? AggStat uses the input stat holder for retrieving the statistic, should it directly take in the statistic to track and have the caller retrieve? But if it is changed to be this way, we cannot pass in the Number we get from the built-in StatHolder.

Sorry for all of the clarification questions, just trying to understand how to make it work!

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: Sure can just do that

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Sure

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Sure

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: I have pushed a new commit that is closer aligned with what you have recommended, though not sure if it necessarily reduces complexity rather than shifts it. In any case, I do not have a preference so happy to change it with respect to your recommendation

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -465,13 +476,27 @@ public Team getPrimaryTeam(UUID uuid, boolean includeObservers) {
...
       int cmp = value.compareTo(newVal);
       if (cmp > 0) return this;
       if (cmp < 0) players.clear();
       players.add(uuid);
       return cmp == 0 ? this : new AggStat<>(type, newVal, players);
     }
+
+    private T getStat(MatchPlayer player, StatHolder stat) {
+      return (T) getStatValue(type, player, stat);
+    }
```
Pablete1234:
```suggestion
    public AggStat<T> track(UUID uuid, MatchPlayer player, StatHolder stat) {
      T newVal = (T) getStatValue(type, player, stat);
      if (newVal == null) return this;
      int cmp = value.compareTo(newVal);
      if (cmp > 0) return this;
      if (cmp < 0) players.clear();
      players.add(uuid);
      return cmp == 0 ? this : new AggStat<>(type, newVal, players);
    }
```
no need for the extra method used exactly once

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,81 @@
...
-    return number(number, this == DEATHS ? RED : GREEN);
+    @Override
+    public Component buildMessage(Component valueComponent) {
+      return translatable(key, valueComponent);
+    }
   }
 
-  public Component component(StatHolder stats) {
-    return translatable(key, makeNumber(stats.getStat(this)));
+  record OfFormula(Formula<MatchPlayer> formula, Component statDisplayName, double defaultValue)
```
Pablete1234:
```suggestion
  record OfFormula(Component name, double defaultValue, Formula<MatchPlayer> formula)
```
I'd simplify naming a bit

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,81 @@
...
+    private static final String KEY = "match.stats.type.generic";
+
+    @Override
+    public Component makeNumber(Number number) {
+      return number(number, GREEN);
+    }
+
+    @Override
+    public Component component(MatchPlayer player) {
+      return translatable(KEY, statDisplayName, makeNumber(formula.apply(player)));
```
Pablete1234:
```suggestion
      return buildMessage(makeNumber(formula.apply(player)));
```
seems like both methods are doing the same, just one takes a component while the other takes the stat holder from which to create the component, maybe both should just `component`?

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -93,6 +93,7 @@ public class StatsMatchModule implements MatchModule, Listener {
   private final Match match;
   private final Map<UUID, PlayerStats> allPlayerStats = new HashMap<>();
   private final Table<Team, UUID, PlayerStats> stats = HashBasedTable.create();
+  private final List<StatType.OfFormula> ofFormulas;
```
Pablete1234: "of formula" is not really the type of this variable, the variable is "formula stat type", just so happens that Formula is inner class to StatType hence it's written as StatType.Formula (and, since formula is already a type elsewhere and the collision would've been annoying to deal with, instead we opted for OfFormula) 

consider overall renaming variable names to formulaStats or similar
```suggestion
  private final List<StatType.OfFormula> formulaStats;
```

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: Sure

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Sure

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Seems reasonable can do

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: Yeah definitely should rename

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -290,14 +292,20 @@ public void onStatsDisplay(MatchStatsEvent event) {
...
-    stats.add(new AggStat<>(StatType.BEST_KILL_STREAK, 0, new HashSet<>()));
-    stats.add(new AggStat<>(StatType.LONGEST_BOW_SHOT, 0, new HashSet<>()));
-    if (verboseStats) stats.add(new AggStat<>(StatType.DAMAGE, 0d, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.Builtin.KILLS, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.Builtin.DEATHS, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.Builtin.ASSISTS, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.Builtin.BEST_KILL_STREAK, 0, new HashSet<>()));
+    stats.add(new AggStat<>(StatType.Builtin.LONGEST_BOW_SHOT, 0, new HashSet<>()));
+    if (verboseStats) stats.add(new AggStat<>(StatType.Builtin.DAMAGE, 0d, new HashSet<>()));
+    for (StatType.OfFormula formulaStats : formulaStats) {
```
Pablete1234:
```suggestion
    for (StatType.OfFormula formulaStat : formulaStats) {
```

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,59 @@
...
+    public StatsModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      XMLFluentParser parser = factory.getParser();
+      Element statsElement = XMLUtils.getUniqueChild(doc.getRootElement(), "stats");
+      if (statsElement == null) {
+        return new StatsModule(Collections.emptyList());
+      } else {
+        List<StatType.OfFormula> formulaStats = new ArrayList<>();
+        for (Element el : XMLUtils.getChildren(statsElement, "stat")) {
+          Component displayName = XMLUtils.parseFormattedText(el, "display-name");
```
Pablete1234:
```suggestion
          Component displayName = XMLUtils.parseFormattedText(el, "name");
```

I'm pretty sure most things in ogm just use name for the display name (eg: objectives, modes)

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,59 @@
...
+          double initialStatValue =
+              XMLUtils.parseNumber(el.getAttribute("initial"), Double.class, 0D);
```
Pablete1234:
```suggestion
              XMLUtils.parseNumber(el.getAttribute("default"), Double.class, 0D);
```
Variables have use default, i'd consider using it here too

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,59 @@
...
+          if (displayName == null) {
+            throw new InvalidXMLException("Expected a display name to be present", el);
```
Pablete1234: You can (and should) the fluent parser for everything, it makes parsing less verbose.
Ie:
`parser.component(el, "name").required()` 
And just add support for component in there (that makes the call go parseFormattedText

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,59 @@
...
+  public @Nullable StatsMatchModule createMatchModule(Match match) throws ModuleLoadException {
+    return new StatsMatchModule(match, formulaStats);
+  }
+
+  public static class Factory implements MapModuleFactory<StatsModule> {
+    @Override
+    public StatsModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      XMLFluentParser parser = factory.getParser();
+      Element statsElement = XMLUtils.getUniqueChild(doc.getRootElement(), "stats");
```
Pablete1234: No need for it to be unique, just use flatten children from XMLUtils

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,81 @@
...
 
-  public Component component(StatHolder stats) {
-    return translatable(key, makeNumber(stats.getStat(this)));
+  record OfFormula(Component name, double defaultValue, Formula<MatchPlayer> formula)
+      implements StatType<MatchPlayer> {
+    private static final String KEY = "match.stats.type.generic";
+
+    @Override
+    public Component makeNumber(Number number) {
+      return number(number, GREEN);
```
Pablete1234: This color should probably be configurable in xml

Pablete1234 (changes requested):
Overall looking pretty good, just a few minor touch-ups

One thing that still seems odd is:
AggStat will always pick the *higher* value for any player, having the initial value be anything other than 0 will probably not work well (eg: imagine you're tracking distance to the goal, winner is 0 distance, the max distance away is 300 blocks, you use default of 300; no player will be able to get better than that, and wont be shown).
You may need to either add the concept of positive/negative stats, or simply get rid of default, make it 0 like all the other stats (they only define it for the sake of specifying the type, ie: int or double) and just assume all stats are positive and count starting from 0

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
chatasma: Oops will fix that

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Cool can change

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Will remove this default in favor of using 0 as referenced by other comment

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Cool can change

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Sounds good

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Sounds good

chatasma:
Makes sense on the default value, I am going to remove the XML default and have it initialize to 0 since that seems sensible for most if not all use-cases

core/src/main/java/tc/oc/pgm/stats/StatType.java
```diff
@@ -1,43 +1,83 @@
...
-    return number(number, this == DEATHS ? RED : GREEN);
+    @Override
+    public Component component(Component valueComponent) {
+      return translatable(key, valueComponent);
+    }
   }
 
-  public Component component(StatHolder stats) {
-    return translatable(key, makeNumber(stats.getStat(this)));
+  record OfFormula(Component name, Formula<MatchPlayer> formula, Color color)
```
Pablete1234:
```suggestion
  record OfFormula(Component name, Formula<MatchPlayer> formula, TextColor color)
```
The datatype here is most definitely not `Color`, those are arbitrary colors and are only used for leather armor.
RGB chat colors are modern and pgm hasn't adapted to them, the codebase relies in bukkit's `ChatColor` in many places still, and more recently, in `NamedTextColor`. Migrating those types eventually is a task we'll probs have to look at some day, but regardless, Color is not the right datatype even for generic color that supports either named or not, the type should be adventure's `TextColor`.

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,56 @@
...
+        throws InvalidXMLException {
+      XMLFluentParser parser = factory.getParser();
+      List<Element> statElements = XMLUtils.flattenElements(doc.getRootElement(), "stats");
+      if (statElements.isEmpty()) {
+        return new StatsModule(Collections.emptyList());
+      } else {
+        List<StatType.OfFormula> formulaStats = new ArrayList<>();
+        for (Element el : statElements) {
+          Component name = parser.component(el, "name").required();
+          Color color = parser.hexColor(el, "color").optional(Color.GREEN);
```
Pablete1234: As for parsing, everything else in pgm just uses `XMLUtils#parseChatColor` and then feeds it thru `TextFormatter#convert` at some point after parsing, that makes sure support for things like `aqua` etc still work, which parsing as hex color won't do.

Unless you're wanting to write a new `TextColor` parser that can take in both (which would be appreciated for future arbitrary rgb color support down the line!), you should stick to those mentioned above methods (but, feel free to wrap them in the xml fluent parser for future expansion for rgb support)

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public StatsModule parse(MapFactory factory, Logger logger, Document doc)
+        throws InvalidXMLException {
+      XMLFluentParser parser = factory.getParser();
+      List<Element> statElements = XMLUtils.flattenElements(doc.getRootElement(), "stats");
+      if (statElements.isEmpty()) {
+        return new StatsModule(Collections.emptyList());
+      } else {
+        List<StatType.OfFormula> formulaStats = new ArrayList<>();
+        for (Element el : statElements) {
```
Pablete1234: Generally the way to use flatten elements is more something like this:
```suggestion
      ImmutableList.Builder<StatType.OfFormula> formulaStats = ImmutableList.builder();
      for (Element el : XMLUtils.flattenElements(doc.getRootElement(), "stats", "stat")) {
```
Which avoids the extra nesting, if it's empty then just formulaStats will be empty at the end.

Also it is good practice to use immutable collections, so you can rest assured that some match won't later modify/break map objects (it would mean if the same map plays again, it may do so with different, and broken, behavior)

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
```diff
@@ -0,0 +1,56 @@
...
+  private StatsModule(List<StatType.OfFormula> formulaStats) {
+    this.formulaStats = formulaStats;
+  }
+
+  @Override
+  public @Nullable StatsMatchModule createMatchModule(Match match) throws ModuleLoadException {
+    return new StatsMatchModule(match, formulaStats);
+  }
+
+  public static class Factory implements MapModuleFactory<StatsModule> {
```
Pablete1234, line 36: Also one thing i forgot: you need to declare a dependency on the variables module, as stats now need to load *after* variables, so that formulas can work well.
By pure coincidence this may work without the explicit dependency, but it is fragile and may break at any point in the future.

overwrite the `getWeakDependencies` method and return a `List.of(VariablesModule.class);` (you can look at ActionModule for an example

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/stats/StatType.java
chatasma: Cool was going to make it a ChatColor but didn't see an existing way to convert it to a native TextColor, I see there's `TextFormatter#convert` as you've pointed out so can use that

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: I'll probably stick to the first method you've mentioned for this PR and wrap it in to the fluent parser to keep this PR relatively small

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Cool seems sensible

core/src/main/java/tc/oc/pgm/stats/StatsModule.java
chatasma: Will add

Pablete1234 (approved):
LGTM, just needs some testing now :)

chatasma:
I did some (very rudimentary) testing with a bridge map (i.e. the example mentioned in the PR) to see if goals were added which looked like they were. Any other test cases I should be evaluating here?

zzufx:
I've retrofitted and tested this module on some of my maps and includes (teams and FFA) and it seemed to work nicely. Maybe a `filter` argument could be useful in the future, but I've managed to do the stuff I wanted just fine without it so far.

~Could also be a good opportunity to stardardize the spelling on the original multi-word stats (I think shot could be lowercase here)~ I can't seem to find the match link stuff here so it might not even be part of PGM, it's been a while (regardless of that, the suggestion would still be to change the OCC one, "Longest Shot" https://github.com/PGMDev/PGM/blob/d20769cc1b06706f73cb4a0b955be9b9f02c3f3b/util/src/main/i18n/templates/match.properties#L57)
image

Pablete1234:
Match link isn't pgm, in the case of occ it's handled by https://github.com/OvercastCommunity/Dewdrop/blob/main/src/main/java/tc/oc/occ/dewdrop/managers/MatchManager.java#L43

Seems a bit out of scope for this pr to change

PR #1501, opened 2025-04-21
Restore channel change when player has no perms

zzufx:
Restores [previous functionality](https://github.com/PGMDev/PGM/blob/fccf7ba2028c8707c5c5d6b5442bfd72f35fd702/core/src/main/java/tc/oc/pgm/listeners/ChatDispatcher.java#L167C7-L167C7) where, if a player sets their chat channel to ADMIN and they don't have the adequate permission their channel setting will be reset the next time they try to send a message. 

There have been cases of newer players doing this (likely by incorrectly trying to use `/a` for "all") and then being left confused and unable to send messages at all, as they'd just get the "_You don't have permission_" error over and over.


Pablete1234 (changes requested):
If done this way, it would mean someone who just types `/a whatever` without perms would *also* get their chat channel reset, even if their default selected chat wasn't a problem at all

Pablete1234 (approved):

PR #1502, opened 2025-05-10
Display entity projectiles

KaevonD:
This feature introduces display entities (specifically block display entities) as projectiles. This feature is only supported for modern PGM.

This can be done in XML, specifying "projectile" to be "BlockDisplay". Speed can be controlled with the "velocity" attribute, damage with the "damage" attribute, and cooldown with the "cooldown" attribute. In addition, the size of the projectile can be specified with the "scale" attribute, where 1 is the default size. The travel time can be specified with the "maxTravelTime" in seconds, where 1 second is the default travel time. Currently, these projectiles travel in a linear path with no falloff.

This is my first PR to PGM, so I understand if I missed some things. I hope I can fix these and make it usable! Thanks for the help.

Example projectile:
``` xml
<projectile
        id="laser"
        name="laser"
        projectile="Block"
        material="RED_STAINED_GLASS"
        velocity="0.5"
        damage="1"
        scale="0.2"
        throwable="false"
        max-travel-time="2s"
        cooldown="1s"/>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
@@ -3,6 +3,7 @@ package tc.oc.pgm.projectile;
 import java.time.Duration;
 import java.util.List;
 import org.bukkit.entity.Entity;
+import org.bukkit.entity.FallingBlock;
 import org.bukkit.potion.PotionEffect;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.filter.Filter;
@@ -15,7 +16,7 @@ public class ProjectileDefinition extends SelfIdentifyingFeatureDefinition {
   protected @Nullable Float power;
   protected double velocity;
   protected ClickAction clickAction;
-  protected Class<? extends Entity> projectile;
+  protected ProjectileEntity projectile;
   protected List<PotionEffect> potion;
   protected Filter destroyFilter;
   protected Duration coolDown;
@@ -30,7 +31,7 @@ public class ProjectileDefinition extends SelfIdentifyingFeatureDefinition {
       @Nullable Float power,
       double velocity,
       ClickAction clickAction,
-      Class<? extends Entity> entity,
+      ProjectileEntity entity,
       List<PotionEffect> potion,
       Filter destroyFilter,
       Duration coolDown,
@@ -50,6 +51,25 @@ public class ProjectileDefinition extends SelfIdentifyingFeatureDefinition {
     this.throwable = throwable;
     this.precise = precise;
     this.blockMaterial = blockMaterial;
+  }
+
+  public sealed interface ProjectileEntity permits RealEntity, BlockEntityType {
+    boolean requiresBlockMaterial();
+  }
+
+  record RealEntity(Class<? extends Entity> entityType) implements ProjectileEntity {
+    @Override
+    public boolean requiresBlockMaterial() {
+      return FallingBlock.class.isAssignableFrom(entityType);
+    }
+  }
+
+  record BlockEntityType(float size, boolean solidBlockCollision, Duration maxTravelTime)
+      implements ProjectileEntity {
+    @Override
+    public boolean requiresBlockMaterial() {
+      return true;
+    }
   }
 
   public @Nullable String getName() {
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
@@ -336,7 +336,7 @@ public class ProjectileMatchModule implements MatchModule, Listener {
       this.currentLocation.setYaw(0);
 
       this.increment = normalizedDirection.clone().multiply(definition.velocity);
-      this.substeps = Math.max(1, (int) (definition.velocity / ce.size()));
+      this.substeps = Math.min(10, Math.max(1, (int) (definition.velocity / Math.max(0.1, ce.size()))));
       this.substep = increment.clone().divide(new Vector(substeps, substeps, substeps));
       if (this.substep.length() < 0.1) this.substep = normalizedDirection.clone().multiply(0.1);
 
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
@@ -4,8 +4,14 @@ import static tc.oc.pgm.util.Assert.assertTrue;
 
 import com.google.common.collect.ImmutableSet;
 import java.util.HashMap;
+import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Explosive;
@@ -37,6 +43,7 @@ import tc.oc.pgm.api.filter.query.Query;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
 import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.party.Party;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.events.ListenerScope;
@@ -44,7 +51,10 @@ import tc.oc.pgm.events.PlayerParticipationStopEvent;
 import tc.oc.pgm.filters.query.BlockQuery;
 import tc.oc.pgm.filters.query.PlayerBlockQuery;
 import tc.oc.pgm.kits.tag.ItemTags;
+import tc.oc.pgm.util.MatchPlayers;
+import tc.oc.pgm.util.TimeUtils;
 import tc.oc.pgm.util.bukkit.MetadataUtils;
+import tc.oc.pgm.util.bukkit.entities.BlockEntity;
 import tc.oc.pgm.util.inventory.InventoryUtils;
 import tc.oc.pgm.util.nms.NMSHacks;
 
@@ -78,67 +88,77 @@ public class ProjectileMatchModule implements MatchModule, Listener {
     ParticipantState playerState = match.getParticipantState(player);
     if (playerState == null) return;
 
-    ProjectileDefinition projectileDefinition =
-        this.getProjectileDefinition(player.getItemInHand());
+    var definition = this.getProjectileDefinition(player.getItemInHand());
 
-    if (projectileDefinition != null
-        && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction)) {
-      // Prevent the original projectile from being fired
-      event.setCancelled(true);
+    if (definition == null || !isValidProjectileAction(event.getAction(), definition.clickAction))
+      return;
 
-      if (this.isCooldownActive(player, projectileDefinition)) return;
+    // Prevent the original projectile from being fired
+    event.setCancelled(true);
 
-      boolean realProjectile = Projectile.class.isAssignableFrom(projectileDefinition.projectile);
-      Vector velocity =
-          player.getEyeLocation().getDirection().multiply(projectileDefinition.velocity);
-      Entity projectile;
-      try {
-        assertTrue(launchingDefinition.get() == null, "nested projectile launch");
-        launchingDefinition.set(projectileDefinition);
-        if (realProjectile) {
-          projectile = player.launchProjectile(
-              projectileDefinition.projectile.asSubclass(Projectile.class), velocity);
-          if (projectile instanceof Fireball fireball && projectileDefinition.precise) {
-            NMSHacks.NMS_HACKS.setFireballDirection(fireball, velocity);
-          }
-        } else {
-          if (FallingBlock.class.isAssignableFrom(projectileDefinition.projectile)) {
-            projectile =
-                projectileDefinition.blockMaterial.spawnFallingBlock(player.getEyeLocation());
+    if (this.isCooldownActive(player, definition)) return;
+
+    var projType = definition.projectile;
+    boolean needsEvent = true;
+    Vector velocity = player.getEyeLocation().getDirection().multiply(definition.velocity);
+    Entity projectile = null;
+    try {
+      assertTrue(launchingDefinition.get() == null, "nested projectile launch");
+      launchingDefinition.set(definition);
+      switch (projType) {
+        case ProjectileDefinition.RealEntity(Class<? extends Entity> entityType) -> {
+          if (Projectile.class.isAssignableFrom(entityType)) {
+            needsEvent = false;
+
+            projectile = player.launchProjectile(entityType.asSubclass(Projectile.class), velocity);
+            if (projectile instanceof Fireball fireball && definition.precise) {
+              NMSHacks.NMS_HACKS.setFireballDirection(fireball, velocity);
+            }
           } else {
-            projectile =
-                player.getWorld().spawn(player.getEyeLocation(), projectileDefinition.projectile);
+            if (FallingBlock.class.isAssignableFrom(entityType)) {
+              projectile = definition.blockMaterial.spawnFallingBlock(player.getEyeLocation());
+            } else {
+              projectile = player.getWorld().spawn(player.getEyeLocation(), entityType);
+            }
+            projectile.setVelocity(velocity);
           }
-          projectile.setVelocity(velocity);
         }
-        if (projectileDefinition.power != null && projectile instanceof Explosive) {
-          ((Explosive) projectile).setYield(projectileDefinition.power);
+        case ProjectileDefinition.BlockEntityType ce -> {
+          Location loc = player.getEyeLocation();
+          var be = BlockEntity.spawnBlockEntity(loc, definition.blockMaterial, ce.size(), velocity);
+          new BlockRunner(definition, be, player, loc);
         }
+      }
+
+      if (definition.power != null && projectile instanceof Explosive) {
+        ((Explosive) projectile).setYield(definition.power);
+      }
+      if (projectile != null) {
         projectile.setMetadata(
-            "projectileDefinition", new FixedMetadataValue(PGM.get(), projectileDefinition));
-      } finally {
-        launchingDefinition.remove();
+            "projectileDefinition", new FixedMetadataValue(PGM.get(), definition));
       }
+    } finally {
+      launchingDefinition.remove();
+    }
 
-      // If the entity implements Projectile, it will have already generated a
-      // ProjectileLaunchEvent.
-      // Otherwise, we fire our custom event.
-      if (!realProjectile) {
-        EntityLaunchEvent launchEvent = new EntityLaunchEvent(projectile, event.getPlayer());
-        match.callEvent(launchEvent);
-        if (launchEvent.isCancelled()) {
-          projectile.remove();
-          return;
-        }
+    // If the entity implements Projectile, it will have already generated a
+    // ProjectileLaunchEvent.
+    // Otherwise, we fire our custom event.
+    if (needsEvent && projectile != null) {
+      EntityLaunchEvent launchEvent = new EntityLaunchEvent(projectile, event.getPlayer());
+      match.callEvent(launchEvent);
+      if (launchEvent.isCancelled()) {
+        projectile.remove();
+        return;
       }
+    }
 
-      if (projectileDefinition.throwable) {
-        InventoryUtils.consumeItem(event);
-      }
+    if (definition.throwable) {
+      InventoryUtils.consumeItem(event);
+    }
 
-      if (projectileDefinition.coolDown != null) {
-        startCooldown(player, projectileDefinition);
-      }
+    if (definition.coolDown != null) {
+      startCooldown(player, definition);
     }
   }
 
@@ -283,5 +303,96 @@ public class ProjectileMatchModule implements MatchModule, Listener {
   public boolean isCooldownActive(Player player, ProjectileDefinition definition) {
     ProjectileCooldowns playerCooldowns = projectileCooldowns.get(player.getUniqueId());
     return (playerCooldowns != null && playerCooldowns.isActive(definition));
+  }
+
+  private class BlockRunner {
+    private final ProjectileDefinition definition;
+    private final BlockEntity blockEntity;
+    private final Player player;
+    private final Party shooterParty;
+    private final ProjectileDefinition.BlockEntityType ce;
+    private final Location currentLocation;
+    private final Vector increment;
+    private final int substeps;
+    private Vector substep;
+    private final Future<?> runner;
+    private int remainingTime;
+    private boolean collided = false;
+
+    public BlockRunner(
+        ProjectileDefinition definition,
+        BlockEntity blockEntity,
+        Player player,
+        Location spawnLocation) {
+      this.definition = definition;
+      this.blockEntity = blockEntity;
+      this.player = player;
+      this.shooterParty = Objects.requireNonNull(match.getPlayer(player)).getParty();
+      this.ce = (ProjectileDefinition.BlockEntityType) definition.projectile;
+
+      this.currentLocation = spawnLocation.clone();
+      var normalizedDirection = currentLocation.getDirection().normalize();
+      this.currentLocation.setPitch(0);
+      this.currentLocation.setYaw(0);
+
+      this.increment = normalizedDirection.clone().multiply(definition.velocity);
+      this.substeps = Math.max(1, (int) (definition.velocity / ce.size()));
+      this.substep = increment.clone().divide(new Vector(substeps, substeps, substeps));
+      if (this.substep.length() < 0.1) this.substep = normalizedDirection.clone().multiply(0.1);
+
+      this.remainingTime = (int) TimeUtils.toTicks(ce.maxTravelTime());
+      this.runner = match
+          .getExecutor(MatchScope.RUNNING)
+          .scheduleAtFixedRate(this::tick, 0L, 50L, TimeUnit.MILLISECONDS);
+    }
+
+    public void tick() {
+      if (remainingTime-- <= 0 || collided) {
+        cancel();
+        return;
+      }
+      if (definition.damage != null || ce.solidBlockCollision()) {
+        if (blockDisplayCollision(currentLocation)) {
+          collided = true;
+          return;
+        }
+      }
+
+      currentLocation.add(increment);
+      blockEntity.teleport(currentLocation);
+    }
+
+    private void cancel() {
+      this.runner.cancel(true);
+      blockEntity.remove();
+    }
+
+    private boolean blockDisplayCollision(Location location) {
+      double halfSize = 0.5 * ce.size();
+      
+      if (ce.solidBlockCollision() && NMSHacks.NMS_HACKS.collidesWithBlock(currentLocation, halfSize, increment, substeps, substep)) {
+        return true;
+      }
+      if (definition.damage != null) {
+        Entity hitEntity = NMSHacks.NMS_HACKS.collidesWithPlayer(location, halfSize, increment, this::isEnemyPlayer);
+        if (hitEntity != null) {
+          ((Player) hitEntity).damage(definition.damage, player);
+          return true;
+        }
+      }
+
+      return false;
+    }
+
+    private boolean isEnemyPlayer(Entity entity) {
+      if (!(entity instanceof Player victim)) {
+        return false;
+      }
+      var mpVictim = match.getPlayer(victim);
+      if (MatchPlayers.canInteract(mpVictim) && mpVictim.getParty() != shooterParty) {
+        return true;
+      }
+      return false;
+    }
   }
 }
--- a/core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
@@ -6,11 +6,11 @@ import java.time.Duration;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 import java.util.logging.Logger;
 import org.bukkit.entity.Arrow;
 import org.bukkit.entity.Entity;
-import org.bukkit.entity.FallingBlock;
 import org.bukkit.potion.PotionEffect;
 import org.jdom2.Document;
 import org.jdom2.Element;
@@ -25,6 +25,7 @@ import tc.oc.pgm.kits.KitParser;
 import tc.oc.pgm.util.material.BlockMaterialData;
 import tc.oc.pgm.util.xml.InvalidXMLException;
 import tc.oc.pgm.util.xml.Node;
+import tc.oc.pgm.util.xml.XMLFluentParser;
 import tc.oc.pgm.util.xml.XMLUtils;
 
 public class ProjectileModule implements MapModule<ProjectileMatchModule> {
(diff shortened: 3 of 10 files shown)


Pablete1234:
I feel like this implementation doesn't quite fit pgm. Don't get me wrong tho: i do like the idea of supporting this type of projectile! but it should be in a slightly different way.

Consider that instead of calling them "display entities" you just call them "block" projectiles.
In 1.8 they can be represented with a falling-block (only supports scale = 1, potentially 0.8 and 0.4 using normal and small armor stands with the block in their head, values of scale >1 just get ignored in legacy)
In modern they can be represented with a display entity

All the logic related to simulating the trajectory, can be shared for both, so there's little reason why that shouldn't be supported or should be gatekept/locked behind modern requirement.

Ideally this doesn't use "NMS_HACKS" at all, as none of this is actually reliant on NMS, it's just version-dependent. In fact i think you should take a look at the existance of `BlockMaterialData#spawnFallingBlock`, and just add an extra method for `spawnBlockEntity`, which returns a PGM type that is a wrapper around a "BlockEntity", in legacy the impl relies in a falling block (with no ai enabled, so it doesn't have phyisics), while in modern it relies on display entities. That way all the code in the projectile module stays clean of calls to nms hacks, it just interacts with this wrapper to rotate, move, or whatever, and the wrapper takes care of the necessary translation layer

core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
```diff
@@ -50,6 +56,31 @@ public ProjectileDefinition(
     this.throwable = throwable;
     this.precise = precise;
     this.blockMaterial = blockMaterial;
+    this.scale = scale;
+    this.solidBlockCollision = solidBlockCollision;
+    this.maxTravelTime = maxTravelTime;
+  }
+
+  public static sealed class ProjectileEntity {
```
Pablete1234: A class with no members and no functions, may as well be a sealed interface instead:
```suggestion
  public sealed interface ProjectileEntity permits RealEntity, CustomEntity {
```

core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
```diff
@@ -50,6 +56,31 @@ public ProjectileDefinition(
...
+      }
+    }
+
+    public static final class AbstractEntity extends ProjectileEntity {
+      public final AbstractEntityType entityType;
+
+      public AbstractEntity(AbstractEntityType entityType) {
+        this.entityType = entityType;
+      }
+    }
```
Pablete1234: You can probably use a record for these:

```suggestion
    record RealEntity(Class<? extends Entity> entityType) implements ProjectileEntity {}
    record CustomEntity(CustomEntityType type) implements ProjectileEntity {}
```

core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
```diff
@@ -50,6 +56,31 @@ public ProjectileDefinition(
...
+  public static sealed class ProjectileEntity {
+    public static final class RealEntity extends ProjectileEntity {
+      public final Class<? extends Entity> entityType;
+
+      public RealEntity(Class<? extends Entity> entityType) {
+        this.entityType = entityType;
+      }
+    }
+
+    public static final class AbstractEntity extends ProjectileEntity {
```
Pablete1234: `Abstract` has a very specific meaning in the java world, and given this class isn't even abstract, i'd vote against it, and instead probably use something like `CustomEntity`

core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
```diff
@@ -30,13 +33,16 @@ public ProjectileDefinition(
...
       List<PotionEffect> potion,
       Filter destroyFilter,
       Duration coolDown,
       boolean throwable,
       boolean precise,
-      BlockMaterialData blockMaterial) {
+      BlockMaterialData blockMaterial,
+      float scale,
+      boolean solidBlockCollision,
+      Duration maxTravelTime) {
```
Pablete1234: These seem to only affect custom projectile entities, so it probably makes sense that they'd go in as part of the `CustomEntity` implementation of ProjectileEntity.

Additionally `scale` should probably be `size` as that leaves it more open to future projectiles (ie: raytraced beam)

core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
```diff
@@ -50,6 +56,31 @@ public ProjectileDefinition(
...
+
+    public static final class AbstractEntity extends ProjectileEntity {
+      public final AbstractEntityType entityType;
+
+      public AbstractEntity(AbstractEntityType entityType) {
+        this.entityType = entityType;
+      }
+    }
+
+    enum AbstractEntityType {
```
Pablete1234: Same as above, avoid `Abstract` when something is not abstract

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -1,10 +1,17 @@
...
 
 import static tc.oc.pgm.util.Assert.assertTrue;
+import static tc.oc.pgm.util.nms.Packets.ENTITIES;
 
 import com.google.common.collect.ImmutableSet;
+
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.UUID;
+import java.util.concurrent.*;
```
Pablete1234: Avoid wildcard imports, configure your ide not to use them

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -88,32 +101,96 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
 
       if (this.isCooldownActive(player, projectileDefinition)) return;
 
-      boolean realProjectile = Projectile.class.isAssignableFrom(projectileDefinition.projectile);
+      boolean realProjectile = false;
+      if (projectileDefinition.projectile instanceof ProjectileDefinition.ProjectileEntity.RealEntity) {
+        realProjectile = Projectile.class.isAssignableFrom(
+            ((ProjectileDefinition.ProjectileEntity.RealEntity) projectileDefinition.projectile).entityType
+        );
+      }
```
Pablete1234: given how much `projectileDefinition.projectile` is used, not just here but in the following blocks too, you may as well extract it
```suggestion
      var projType = projectileDefinition.projectile;
      boolean realProjectile = projType instanceof ProjectileDefinition.ProjectileEntity.RealEntity re 
              && Projectile.class.isAssignableFrom(re.entityType());
```
Let's make use of java 17 features too

KaevonD:
Projectile for this feature is no longer "BlockDisplay" but "Block". In newer versions functionality remains the same and spawns a block display entity, but in lower versions it will now spawn a falling block instead.

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -88,32 +101,96 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
...
-      Entity projectile;
+      Entity projectile = null;
+      BlockEntity blockEntity = null;
       try {
         assertTrue(launchingDefinition.get() == null, "nested projectile launch");
         launchingDefinition.set(projectileDefinition);
         if (realProjectile) {
           projectile = player.launchProjectile(
-              projectileDefinition.projectile.asSubclass(Projectile.class), velocity);
+            (((ProjectileDefinition.ProjectileEntity.RealEntity) projectileDefinition.projectile).entityType)
```
Pablete1234: this is very ugly, and i'm starting to believe now that we have wrapper entities we can probably hide away these differences inside such that you can:

`projectileDefinition.projectile.launch(player, velocity)` and this logic is within there, not just for this but for the others too

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -88,32 +101,96 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
...
             projectile =
                 projectileDefinition.blockMaterial.spawnFallingBlock(player.getEyeLocation());
           } else {
-            projectile =
-                player.getWorld().spawn(player.getEyeLocation(), projectileDefinition.projectile);
+            Location loc = player.getEyeLocation();
+            if ((projectileDefinition.projectile instanceof ProjectileDefinition.ProjectileEntity.AbstractEntity) &&
+                    ((ProjectileDefinition.ProjectileEntity.AbstractEntity) projectileDefinition.projectile).entityType == ProjectileDefinition.ProjectileEntity.AbstractEntityType.BLOCK) {
+              loc.setPitch(0);
+              loc.setYaw(0);
```
Pablete1234: this seems like logic that the version-dependant impl should handle, if it needs it

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -88,32 +101,96 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
...
+              blockEntity = ENTITIES.spawnBlockEntity(loc, projectileDefinition.blockMaterial);
+              projectile = blockEntity.entity();
+            } else if (projectileDefinition.projectile instanceof ProjectileDefinition.ProjectileEntity.RealEntity) {
+              projectile =
+                      player.getWorld().spawn(loc, ((ProjectileDefinition.ProjectileEntity.RealEntity) projectileDefinition.projectile).entityType);
+            }
```
Pablete1234:
```suggestion
            } else if (projectileDefinition.projectile instanceof ProjectileDefinition.ProjectileEntity.RealEntity re) {
              projectile = player.getWorld().spawn(loc, re.entityType());
            }
```
really should use java 17's features in a bunch of these

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -88,32 +101,96 @@ && isValidProjectileAction(event.getAction(), projectileDefinition.clickAction))
...
           }
           projectile.setVelocity(velocity);
         }
         if (projectileDefinition.power != null && projectile instanceof Explosive) {
           ((Explosive) projectile).setYield(projectileDefinition.power);
         }
+        if (blockEntity != null && blockEntity.isDisplayEntity()) {
```
Pablete1234: this is a nope, you're running version-dependant logic in the common area, this is something the wrapper for your entity should handle

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -76,6 +79,9 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
         boolean throwable =
             XMLUtils.parseBoolean(projectileElement.getAttribute("throwable"), true);
         boolean precise = XMLUtils.parseBoolean(projectileElement.getAttribute("precise"), true);
+        float scale = XMLUtils.parseNumber(Node.fromChildOrAttr(projectileElement, "scale"), Float.class, 1.0f);
+        boolean solidBlockCollision = XMLUtils.parseBoolean(projectileElement.getAttribute("solid-block-collision"), true);
+        Duration maxTravelTime = XMLUtils.parseDuration(projectileElement.getAttribute("max-travel-time"), Duration.ofSeconds(1));
```
Pablete1234: as mentioned above, these seem to only matter for the block custom entity, so you can put them as part of the `parseProjectileEntity` method and have them live inside the custom entity.

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -90,13 +96,53 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+      final Class<? extends Entity> def
+    ) throws InvalidXMLException {
+      final Node node = Node.fromAttr(element, attributeName);
+      if (node == null) {
+        return new ProjectileDefinition.ProjectileEntity.RealEntity(def);
+      }
+      final String entityText = node.getValue();
+      if (!entityText.matches("[a-zA-Z0-9_]+")) {
+        throw new InvalidXMLException("Invalid entity type '" + entityText + "'", node);
+      }
```
Pablete1234: i don't see the need for this, if nothing matches it you'll find out anyways

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -90,13 +96,53 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+      if (ABSTRACT_ENTITY_MAP.containsKey(entityText.toLowerCase())) {
+        return ABSTRACT_ENTITY_MAP.get(entityText.toLowerCase());
+      }
+      return new ProjectileDefinition.ProjectileEntity.RealEntity(XMLUtils.parseEntityTypeAttribute(element, attributeName, def));
```
Pablete1234: Given custom logic will likely have to be involved, consider just using a switch, and using the fluent parser:

```suggestion
      return switch(entityText.toLowerCase(Locale.ROOT)) {
        case "entity" -> new ProjectileDefinition.BlockEntity(
                parser.parseFloat(el, "scale").optional(1.0f),
                parser.parseBool(el, "solid-block-collision").orTrue(),
                parser.parseDuration(el, "max-travel-time").optional(DEFAULT_MAX_TRAVEL);
        default -> new ProjectileDefinition.RealEntity(XMLUtils.parseEntityTypeAttribute(node));
      }
```

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -90,13 +96,53 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+          "block", new ProjectileDefinition.ProjectileEntity.AbstractEntity(
+                  ProjectileDefinition.ProjectileEntity.AbstractEntityType.BLOCK
+              )
+        );
+
+    private static ProjectileDefinition.ProjectileEntity parseProjectileEntity(
+      final Element element,
+      final String attributeName,
+      final Class<? extends Entity> def
+    ) throws InvalidXMLException {
```
Pablete1234:
```suggestion
    private static ProjectileDefinition.ProjectileEntity parseProjectileEntity(XmlFluentParser parser, Element el) throws InvalidXMLException {
```
Attribute name is *always* "projectile" and default entity is *always* arrow, there's no need to make them parameters, instead pass in the parser which will be of use

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -90,13 +96,53 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+    private static boolean acceptsBlockMaterial(final ProjectileDefinition.ProjectileEntity entity) {
+      if (entity instanceof ProjectileDefinition.ProjectileEntity.RealEntity) {
+        final Class<? extends Entity> cls = ((ProjectileDefinition.ProjectileEntity.RealEntity) entity).entityType;
+        return cls.isAssignableFrom(FallingBlock.class) || NMSHacks.NMS_HACKS.isBlockDisplayEntity(cls);
+      } else if (entity instanceof ProjectileDefinition.ProjectileEntity.AbstractEntity) {
+        return ((ProjectileDefinition.ProjectileEntity.AbstractEntity) entity).entityType ==
+                ProjectileDefinition.ProjectileEntity.AbstractEntityType.BLOCK;
+      }
+      return false;
+    }
```
Pablete1234: This seems like logic that `ProjectileEntity` should implement instead

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -38,12 +28,7 @@
 import net.minecraft.world.level.storage.LevelStorageSource;
 import net.minecraft.world.level.storage.PrimaryLevelData;
 import net.minecraft.world.level.validation.ContentValidationException;
-import org.bukkit.Bukkit;
-import org.bukkit.Chunk;
-import org.bukkit.Material;
-import org.bukkit.Nameable;
-import org.bukkit.World;
-import org.bukkit.WorldCreator;
+import org.bukkit.*;
```
Pablete1234: as mentioned above, avoid wildcard imports

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/ModernEntityPackets.java
```diff
@@ -39,6 +47,12 @@ public class ModernEntityPackets implements EntityPackets {
   private static final EntityDataAccessor<Integer> ENTITY_AIR =
       new EntityDataAccessor<>(1, EntityDataSerializers.INT);
 
+  @Override
+  public BlockEntity spawnBlockEntity(Location loc, BlockMaterialData blockMaterialData) {
+    final Entity entity = loc.getWorld().spawn(loc, BlockDisplay.class);
+    return new DisplayEntity(entity);
+  }
+
```
Pablete1234: This class is for **entity packets**, ie: fake-entities that are packet-driven
i mentioned this as an example of how you could make your impl, but this is not the place for it

util/src/main/java/tc/oc/pgm/util/nms/packets/BlockEntity.java
```diff
@@ -0,0 +1,14 @@
+package tc.oc.pgm.util.nms.packets;
```
Pablete1234: the nms -> packets package is not a good place for this, this is nethier needing to access NMS, nor is it packet based

util/src/main/java/tc/oc/pgm/util/nms/packets/BlockEntity.java
```diff
@@ -0,0 +1,14 @@
...
+import org.bukkit.entity.Entity;
+
+public interface BlockEntity {
+    boolean isDisplayEntity();
+    Entity entity();
+    // Entity must have an actual rotation of 0 before invoking this
+    // as this merely modifies the transformation matrix, otherwise it may appear offset
+    void align(float pitch, float yaw, float scale);
+    void setBlock(Material block);
+    void setTeleportationDuration(int duration);
```
Pablete1234: This is way too fine-grained of a control, the mere spawning of it should be handling this so that they're not at this level

util/src/main/java/tc/oc/pgm/util/nms/packets/BlockEntity.java
```diff
@@ -0,0 +1,14 @@
+package tc.oc.pgm.util.nms.packets;
+
+import org.bukkit.Material;
+import org.bukkit.entity.Entity;
+
+public interface BlockEntity {
+    boolean isDisplayEntity();
+    Entity entity();
```
Pablete1234: ideally you should NOT expose the entity itself, that there even is an entity is an implementation detail that should be hidden from the observer, because if for example armor stands are used, the position of the armor stand itself will not align with the position of the block entity (the armor stand feet will always be 1.8 blocks below), so if you expose the entity, that illusion breaks.

if instead you add here move methods, an armor-stand-backed impl can:
```java
private static final double OFFSET = 1.8f;

void teleport(Location loc) {
  loc = loc.clone();
  loc.setY(loc.getY() - OFFSET);
  this.entity.teleport(loc);
}
```

which is the whole idea behind hiding this in impl-dependant methods

util/src/main/java/tc/oc/pgm/util/nms/packets/BlockEntity.java
```diff
@@ -0,0 +1,14 @@
+package tc.oc.pgm.util.nms.packets;
+
+import org.bukkit.Material;
+import org.bukkit.entity.Entity;
+
+public interface BlockEntity {
+    boolean isDisplayEntity();
```
Pablete1234: Again this breaks encapsulation, the user of this api (the projectile module) should not need to know or care how the underlying magic happens

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
KaevonD: I was attempting to do this, but it was taking some time and effort to figure out. In the end, I left it and I would rather potentially come back to it later than to do it as part of this PR. I think it is ok to have here for now.

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
KaevonD: Same as what I mentioned in the other comment, I am hoping we can let this go for now and perhaps come back to add the wrapper later. I was having trouble finding a nice way to implement it to properly account for all scenarios, perhaps if you know a good way then you could add it quickly.

util/src/main/java/tc/oc/pgm/util/bukkit/entities/BlockEntity.java
```diff
@@ -0,0 +1,30 @@
...
+    return FACTORY.spawnBlockEntity(center, blockMaterialData, size, velocity);
+  }
+
+  void teleport(Location center);
+
+  void remove();
+
+  interface Factory {
+    BlockEntity spawnBlockEntity(
+        Location center, BlockMaterialData blockMaterialData, float size, Vector velocity);
```
Pablete1234, line 28: i believe `velocity` is irrelevant in all of BlockEntity itself

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -284,4 +304,95 @@ public boolean isCooldownActive(Player player, ProjectileDefinition definition)
...
+    private boolean isEnemyPlayer(Entity entity) {
+      if (!(entity instanceof Player victim)) {
+        return false;
+      }
+      var mpVictim = match.getPlayer(victim);
+      if (MatchPlayers.canInteract(mpVictim) && mpVictim.getParty() != shooterParty) {
+        return true;
+      }
+      return false;
+    }
```
Pablete1234, line 396: Doesn't properly respect rules of who can attack who, eg: friendly fire may be enabled.

Look for methods that may allow you to check this, if all else fails, nms.
You'll still want to keep the canInteract tho, that's a pretty easy check to ensure thye're not like, dead etc

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -284,4 +304,95 @@ public boolean isCooldownActive(Player player, ProjectileDefinition definition)
...
+    private boolean blockDisplayCollision(Location location) {
+      double halfSize = 0.5 * ce.size();
+      
+      if (ce.solidBlockCollision() && NMSHacks.NMS_HACKS.collidesWithBlock(currentLocation, halfSize, increment, substeps, substep)) {
+        return true;
+      }
+      if (definition.damage != null) {
+        Entity hitEntity = NMSHacks.NMS_HACKS.collidesWithPlayer(location, halfSize, increment, this::isEnemyPlayer);
+        if (hitEntity != null) {
+          ((Player) hitEntity).damage(definition.damage, player);
```
Pablete1234, line 379: This may be the only available damage value that exists in the intersect jar (ie: both 1.8 and modern), but i'm sure if you look at the bukkit apis of each version there have to be ways to provide more info, like a damage source/type style apis so that it can know its thru a projectile thrown by another player, instead of counting as more or less "player A hit player B" as if it was a punch, so that hopefully the pgm tracker can display it properly.

If you can't find anything like that, maybe you need a custom way to signal the death tracker module what kind of kill this is such that it can create appropriate messages (thinking of just using the projectile ones, like arrows/bows, ie: PlayerA was shot by PlayerB (x blocks) or similar)

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -284,4 +304,95 @@ public boolean isCooldownActive(Player player, ProjectileDefinition definition)
...
+
+      this.currentLocation = spawnLocation.clone();
+      var normalizedDirection = currentLocation.getDirection().normalize();
+      this.currentLocation.setPitch(0);
+      this.currentLocation.setYaw(0);
+
+      this.increment = normalizedDirection.clone().multiply(definition.velocity);
+      this.substeps = Math.min(10, Math.max(1, (int) (definition.velocity / Math.max(0.1, ce.size()))));
+      this.substep = increment.clone().divide(new Vector(substeps, substeps, substeps));
+      if (this.substep.length() < 0.1) this.substep = normalizedDirection.clone().multiply(0.1);
```
Pablete1234, line 341: not needed anymore, you're already calculating substeps such that it shouldn't be an issue

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -284,4 +304,95 @@ public boolean isCooldownActive(Player player, ProjectileDefinition definition)
...
+          .getExecutor(MatchScope.RUNNING)
+          .scheduleAtFixedRate(this::tick, 0L, 50L, TimeUnit.MILLISECONDS);
+    }
+
+    public void tick() {
+      if (remainingTime-- <= 0 || collided) {
+        cancel();
+        return;
+      }
+      if (definition.damage != null || ce.solidBlockCollision()) {
```
Pablete1234, line 354: these are also checked inside blockDisplayCollision, so i'd say they're redundant and you can get rid of them

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -451,4 +457,45 @@ public int getMaxWorldSize(World world) {
   public int allocateEntityId() {
     return Bukkit.getUnsafe().nextEntityId();
   }
+
+  @Override
+  public boolean collidesWithBlock(Location center, double halfSize, Vector delta, int substeps, Vector substep) {
+    Location pos = center.clone();
+    AABB AABB = new AABB(
```
Pablete1234, line 464:
```suggestion
    AABB aabb = new AABB(
```

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -451,4 +457,45 @@ public int getMaxWorldSize(World world) {
...
+          pos.getY() - halfSize,
+          pos.getZ() - halfSize,
+
+          pos.getX() + halfSize,
+          pos.getY() + halfSize,
+          pos.getZ() + halfSize
+        );
+
+        if (!world.getHandle().getLevel().noCollision(null, AABB)) {
+        return true;
```
Pablete1234, line 488:
```suggestion
          return true;
```
formatting

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -451,4 +457,45 @@ public int getMaxWorldSize(World world) {
...
+      pos.getY() - halfSize + Math.min(0, delta.getY()),
+      pos.getZ() - halfSize + Math.min(0, delta.getZ()),
+
+      pos.getX() + halfSize + Math.max(0, delta.getX()),
+      pos.getY() + halfSize + Math.max(0, delta.getY()),
+      pos.getZ() + halfSize + Math.max(0, delta.getZ())
+    );
+    
+    CraftWorld world = (CraftWorld) pos.getWorld();
+    if (!world.getHandle().getLevel().noCollision(null, AABB)) {
```
Pablete1234, line 475:
```suggestion
    var level = ((CraftWorld) pos.getWorld()).getHandle().getLevel();
    if (!level.noCollision(null, aabb)) {
```
since you need to use this in a couple places may as well declare it directly

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/impl/SpNMSHacks.java
```diff
@@ -8,21 +8,22 @@
...
 import net.minecraft.server.v1_8_R3.EntityFireworks;
 import net.minecraft.server.v1_8_R3.IBlockData;
 import net.minecraft.server.v1_8_R3.IDataManager;
+import net.minecraft.server.v1_8_R3.MovingObjectPosition;
 import net.minecraft.server.v1_8_R3.NBTTagCompound;
 import net.minecraft.server.v1_8_R3.ServerNBTManager;
+import net.minecraft.server.v1_8_R3.Vec3D;
 import net.minecraft.server.v1_8_R3.WorldData;
 import net.minecraft.server.v1_8_R3.WorldServer;
-import org.bukkit.Bukkit;
```
Pablete1234, line 21: avoid wildcard imports

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/impl/SpNMSHacks.java
```diff
@@ -229,4 +230,72 @@ public int getMaxWorldSize(World world) {
...
+      pos.getX() - halfSize + Math.min(0, delta.getX()),
+      pos.getY() - halfSize + Math.min(0, delta.getY()),
+      pos.getZ() - halfSize + Math.min(0, delta.getZ()),
+
+      pos.getX() + halfSize + Math.max(0, delta.getX()),
+      pos.getY() + halfSize + Math.max(0, delta.getY()),
+      pos.getZ() + halfSize + Math.max(0, delta.getZ())
+    );
+    
+    net.minecraft.server.v1_8_R3.World world = ((org.bukkit.craftbukkit.v1_8_R3.CraftWorld) pos.getWorld()).getHandle();
```
Pablete1234, line 248:
```suggestion
    var world = ((CraftWorld) pos.getWorld()).getHandle();
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/impl/SpNMSHacks.java
```diff
@@ -229,4 +230,72 @@ public int getMaxWorldSize(World world) {
   public int allocateEntityId() {
     return Bukkit.allocateEntityId();
   }
+
+  @Override
+  public boolean collidesWithBlock(Location center, double halfSize, Vector delta, int substeps, Vector substep) {
+    Location pos = center.clone();
+    
+    AxisAlignedBB AABB = new AxisAlignedBB(
```
Pablete1234, line 238:
```suggestion
    AxisAlignedBB aabb = new AxisAlignedBB(
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/impl/SpNMSHacks.java
```diff
@@ -229,4 +230,72 @@ public int getMaxWorldSize(World world) {
...
+        return true;
+        }
+        pos.add(substep);
+      }
+    }
+    return false;
+  }
+
+  @Override
+  public Entity collidesWithPlayer(Location center, double halfSize, Vector delta, Predicate<Entity> predicate) {
```
Pablete1234, line 271: i'm unsure why you chose to implement this in nms per-version when the bukkit API provides a proper way of getting this, via `World#getNearbyPlayers(center, xrad, yrad, zrad, predicate)`

You can do the bounding box grow and intercept here if u want but seems like the extracting of the list of players could be done outside, or even be baked into the predicate, like:

`p -> isEnemy(shooter, p) && NMS_HACKS.raytrace(p, vecStart, vecEnd, halfSize)`

then off the resulting list of entities, you pick the closest to vecStart (if any)

that way the per-version impl would be very limited, just something like:

```java
public boolean raytrace(Entity e, Vector start, Vector end, double grow) {
    Vec3D vecStart = new Vec3D(start.getX(), start.getY(), start.getZ());
    Vec3D vecEnd = new Vec3D(end.getX(), end.getY(), end.getZ());

    var bb ((CraftEntity) e).getHandle().getBoundingBox().grow(grow, grow, grow);
    return bb.a(vecStart, vecEnd) != null;
}
```

core/src/main/java/tc/oc/pgm/projectile/ProjectileMatchModule.java
```diff
@@ -284,4 +304,95 @@ public boolean isCooldownActive(Player player, ProjectileDefinition definition)
...
+    }
+
+    private boolean blockDisplayCollision(Location location) {
+      double halfSize = 0.5 * ce.size();
+      
+      if (ce.solidBlockCollision() && NMSHacks.NMS_HACKS.collidesWithBlock(currentLocation, halfSize, increment, substeps, substep)) {
+        return true;
+      }
+      if (definition.damage != null) {
+        Entity hitEntity = NMSHacks.NMS_HACKS.collidesWithPlayer(location, halfSize, increment, this::isEnemyPlayer);
```
Pablete1234, line 377: This logic seems to inherently assume that just because you collided with a block, you no longer hit a player. That's not true.

Imagine you shoot a high-speed projectile (5 blocks per second), a player is standing 2 blocks away from the wall, and you're standing 9 blocks away from the wall. You fire, first tick the projectile advances 5 blocks and ends at 4 blocks from the wall (doesn't hit the player), next tick it tries to travel 5 blocks forward, hits the wall, and is cancelled, having completely phased thru the enemy player.

For this logic to be correct, `collidesWithBlock` needs to return a vector of  *where* you collided with a block, and that should be reused for player collision checks, where effectively they're "cut short", so in the above example, on the 2nd tick, the projectile hit the wall after 4 blocks, so the start and end positions of the player collisions should be that 4-block range, instead of the 5 it would otherwise be.

Pablete1234 (changes requested):

PR #1504, opened 2025-05-15
Fix Nether/End Portals working on Modern

cswhite2000:
End portals and Nether portals work on modern.

Cancelling the teleport event when the cause is a nether or end portal fixes that.

### DIFF
--- a/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
+++ b/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
@@ -13,10 +13,12 @@ import org.bukkit.entity.FallingBlock;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event;
 import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -336,11 +336,14 @@ public void setGameRule(MatchFinishEvent event, String gameRule, boolean gameRul
...
-          && !WorldBorders.isInsideBorder(event.getTo())) {
-        event.setCancelled(true);
+    switch (event.getCause()) {
+      case PLUGIN -> {
+        if (WorldBorders.isInsideBorder(event.getFrom())
+            && !WorldBorders.isInsideBorder(event.getTo())) {
+          event.setCancelled(true);
+        }
       }
+      case NETHER_PORTAL, END_PORTAL -> event.setCancelled(true);
```
Pablete1234: If a map is set in the nether or end, will this prevent players being teleported to the match (ie: a normal cycle)?

cswhite2000: Well, I've checked the server code.

Modern handles this fine, as those causes are only used for the portal blocks.
For Sportpaper this will cause issues.

I suppose I should move this into `ModernListener`?

Pablete1234 (approved):

PR #1505, opened 2025-05-15
Add ``worldtime`` variable

zzufx:
Adds a variable that allows mapmakers to change the time of the world during a match (for instance, as an objective is completed). It is 

Example: 
```xml
<actions>
    <action id="set-daytime" scope="match" expose="true">
        <set var="worldtime" value="6000"/> <!-- Sets the world time to 6000 -->
    </action>
    <action id="set-nighttime" scope="match" expose="true">
        <set var="worldtime" value="18000"/> <!-- Sets the world time to 18000 -->
    </action>
    <action id="add-time" scope="match" expose="true">
        <set var="worldtime" value="worldtime+1000"/> <!-- Adds 1000 to the current world time -->
    </action>
    <action id="remove-time" scope="match" expose="true">
        <set var="worldtime" value="worldtime-1000"/>
    </action>
</actions>
```

I went with `worldtime` instead of `world-time` to keep in line with `timelimit` and `maxbuildheight` (and so that the default ID matched the variable type)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/variables/VariablesModule.java
+++ b/core/src/main/java/tc/oc/pgm/variables/VariablesModule.java
@@ -29,6 +29,7 @@ import tc.oc.pgm.variables.types.MaxBuildVariable;
 import tc.oc.pgm.variables.types.PlayerVariable;
 import tc.oc.pgm.variables.types.ScoreVariable;
 import tc.oc.pgm.variables.types.TimeLimitVariable;
+import tc.oc.pgm.variables.types.WorldTimeVariable;
 
 public class VariablesModule implements MapModule<VariablesMatchModule> {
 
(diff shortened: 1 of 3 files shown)


cswhite2000 (commented):
I feel like this would be better as a variable, similar to the time variable. Maybe a 'world-time' variable?

core/src/main/java/tc/oc/pgm/variables/types/WorldTimeVariable.java
```diff
@@ -0,0 +1,24 @@
...
+  @Override
+  protected double getValueImpl(Match match) {
+    long val = match.getWorld().getFullTime();
+    return (double) val;
+  }
+
+  @Override
+  protected void setValueImpl(Match match, double value) {
+    long val = (long) value;
+    match.getWorld().setFullTime(val);
```
cswhite2000: From: https://jd.papermc.io/paper/1.21.5/org/bukkit/World.html#setFullTime(long)

> void setFullTime(long time)
Sets the in-game time on the server
Note that this sets the full time of the world, which may cause adverse effects such as breaking redstone clocks and any scheduled events

Using `setTime` and correspondingly `getTime` seems like it would be safer

cswhite2000 (approved):

core/src/main/java/tc/oc/pgm/variables/VariablesModule.java
```diff
@@ -142,6 +143,7 @@ public VariablesModule parse(MapFactory factory, Logger logger, Document doc)
         features.addFeature(null, "score", ScoreVariable.INSTANCE);
         features.addFeature(null, "timelimit", TimeLimitVariable.INSTANCE);
         features.addFeature(null, "maxbuildheight", MaxBuildVariable.INSTANCE);
+        features.addFeature(null, "worldtime", WorldTimeVariable.INSTANCE);
```
Pablete1234, line 146: This is a breaking change without a proto bump

PR #1507, opened 2025-05-16
Implement ``weather`` action and allow for weather control

zzufx:
Adds a ``weather`` action that allows map authors to change the world's weather during a match (for instance, every X minutes). Aside from aesthetic uses, this can be especially useful in newer versions where rain affects things like riptide tridents.

```xml
<action id="set-rain" scope="match" expose="true">
    <weather state="rain"/>
</action>
<action id="set-thunder" scope="match" expose="true">
    <weather state="thunder"/>
</action>
<action id="set-clear" scope="match" expose="true">
    <weather state="clear"/>
</action>
```

Also fixes thunder actually being able to end on its own on existing maps, as it uses a different counter.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -445,7 +445,7 @@ public class ActionParser {
 
   @MethodParser("weather")
   public WeatherAction parseWeather(Element el, Class<?> scope) throws InvalidXMLException {
-    return new WeatherAction(
+    return WeatherAction.of(
         parser.parseEnum(WeatherMatchModule.WeatherType.class, el, "state").required());
   }
 }
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -36,6 +36,7 @@ import tc.oc.pgm.action.actions.SoundAction;
 import tc.oc.pgm.action.actions.TakePaymentAction;
 import tc.oc.pgm.action.actions.TeleportAction;
 import tc.oc.pgm.action.actions.VelocityAction;
+import tc.oc.pgm.action.actions.WeatherAction;
 import tc.oc.pgm.api.feature.FeatureValidation;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.Filterables;
@@ -50,6 +51,7 @@ import tc.oc.pgm.filters.matcher.StaticFilter;
 import tc.oc.pgm.filters.matcher.player.ParticipatingFilter;
 import tc.oc.pgm.filters.operator.AllFilter;
 import tc.oc.pgm.kits.Kit;
+import tc.oc.pgm.modules.WeatherMatchModule;
 import tc.oc.pgm.shops.ShopModule;
 import tc.oc.pgm.shops.menu.Payable;
 import tc.oc.pgm.structure.StructureDefinition;
@@ -439,5 +441,11 @@ public class ActionParser {
     var update = parser.parseBool(el, "update").orTrue();
 
     return new PasteStructureAction<>(scope, xFormula, yFormula, zFormula, structure, update);
+  }
+
+  @MethodParser("weather")
+  public WeatherAction parseWeather(Element el, Class<?> scope) throws InvalidXMLException {
+    return new WeatherAction(
+        parser.parseEnum(WeatherMatchModule.WeatherType.class, el, "state").required());
   }
 }
--- a/core/src/main/java/tc/oc/pgm/action/actions/WeatherAction.java
+++ b/core/src/main/java/tc/oc/pgm/action/actions/WeatherAction.java
@@ -1,0 +1,18 @@
+package tc.oc.pgm.action.actions;
+
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.modules.WeatherMatchModule;
+
+public class WeatherAction extends AbstractAction<Match> {
+
+  private final Enum state;
+
+  public WeatherAction(Enum state) {
+    super(Match.class);
+    this.state = state;
+  }
+
+  public void trigger(Match match) {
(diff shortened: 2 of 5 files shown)


core/src/main/java/tc/oc/pgm/action/actions/WeatherAction.java
```diff
@@ -0,0 +1,40 @@
...
+        break;
+      case WeatherType.THUNDER:
+        match.getWorld().setStorm(true);
+        match.getWorld().setThundering(true);
+        break;
+      default:
+        break;
+    }
+
+    weatherMatchModule.denyChange();
```
cswhite2000: I would prefer to have all of this logic moved into `WeatherMatchModule`

core/src/main/java/tc/oc/pgm/action/WeatherType.java
```diff
@@ -0,0 +1,7 @@
+package tc.oc.pgm.action;
+
+public enum WeatherType {
```
cswhite2000: nit:
I'd prefer for this to be inside the `WeatherMatchModule` or in that package

cswhite2000 (commented):
A couple of nits.
Pending Pablo's review as his opinion may differ

core/src/main/java/tc/oc/pgm/modules/WeatherMatchModule.java
```diff
@@ -0,0 +1,45 @@
...
+    this.shouldChange = shouldChange;
+  }
+
+  public void allowChange() {
+    shouldChange = true;
+  }
+
+  public void denyChange() {
+    shouldChange = false;
+  }
```
Pablete1234: this is not the right way to allow a weather change, by exposing the state; instead you should have a method that does all this logic

core/src/main/java/tc/oc/pgm/modules/WeatherMatchModule.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
+  public void onWeatherChange(final WeatherChangeEvent event) {
+    if (!shouldChange) {
+      event.setCancelled(true);
```
Pablete1234, line 51: should check that the world is the correct one, that was part of the "cancelAlways" method where this used to be

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -440,4 +441,9 @@ public <T extends Filterable<?>> PasteStructureAction<T> parseStructure(
 
     return new PasteStructureAction<>(scope, xFormula, yFormula, zFormula, structure, update);
   }
+
+  @MethodParser("weather")
+  public WeatherAction parseWeather(Element el, Class<?> scope) throws InvalidXMLException {
+    return new WeatherAction(XMLUtils.parseEnum(Node.fromAttr(el, "state"), WeatherType.class));
```
Pablete1234: use the new fluent parsing, it makes the code clearer, as well as fixing the issue that this current code would allow a null (ie: it is not a required attribute) which would then error out when trying to run the weather action:
```suggestion
    return new WeatherAction(parser.parseEnum(WeatherType.class, el, "style").required());
```

core/src/main/java/tc/oc/pgm/action/WeatherType.java
```diff
@@ -0,0 +1,7 @@
+package tc.oc.pgm.action;
```
Pablete1234: housing this enum standalone is probably overkill, consider just putting it inside weather action, or even weather match module

core/src/main/java/tc/oc/pgm/action/actions/WeatherAction.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  public void trigger(Match match) {
+    WeatherMatchModule weatherMatchModule = match.getModule(WeatherMatchModule.class);
+
+    weatherMatchModule.allowChange();
+
+    switch (state) {
+      case WeatherType.CLEAR:
+        match.getWorld().setStorm(false);
```
Pablete1234: extract `match.getWorld()` outside of the switch, it's being called a ton of times

core/src/main/java/tc/oc/pgm/action/actions/WeatherAction.java
```diff
@@ -0,0 +1,40 @@
...
+        match.getWorld().setStorm(true);
+        match.getWorld().setThundering(false);
+        match.getWorld().setThunderDuration(0);
+        break;
+      case WeatherType.THUNDER:
+        match.getWorld().setStorm(true);
+        match.getWorld().setThundering(true);
+        break;
+      default:
+        break;
```
Pablete1234: consider expanding the WeatherType enum to contain the storm and thundering true/false attributes, that way you can also tuck away all this logic into there where you can just call `WeatherType.RAIN.enable()` and it avoids the whole switch being needed

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/action/actions/WeatherAction.java
```diff
@@ -0,0 +1,18 @@
+package tc.oc.pgm.action.actions;
+
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.modules.WeatherMatchModule;
+
+public class WeatherAction extends AbstractAction<Match> {
+
+  private final Enum state;
```
cswhite2000: Better to use the actual type here ie `WeatherMatchModule.WeatherType` instead of `Enum`

Please update in other locations as well

cswhite2000 (dismissed):
I have no further concerns
I'll leave it up to Pablo if he wants to require the further enum change, but looks good to me 👍

Pablete1234:
Commited a few fixes myself to get this ready quicker, @cswhite2000 review my changes and i think it's ready

cswhite2000 (approved):

Pablete1234 (approved):

PR #1508, opened 2025-05-18
Update adventure and adventure-platform

calcastor:
Unblocks https://github.com/PGMDev/PGM/pull/1494

Draft because it won't actually be used on modern since Paper's copy of Adventure takes precedence, so we either need to wait until we update to a newer version of Paper or figure something else out (relocating Adventure and then solving issues that arise from that if it's possible?).


calcastor:
This is blocked until we migrate to a version of Paper which includes adventure-platform 4.4.0.

PR #1512, opened 2025-05-31
Allow default variant to be set

### LINKED ISSUE #1511: Implement `terrain` module improvements from ProjectAres
ProjectAres had a more featured terrain module, which allowed for specifying specific world folders, whether the vanilla chunk generator was used and what seed to be used in such a case, and whether pre-match physics were enabled. Additionally, ProjectAres moved the `environment` definition outside of the terrain module, preferring `dimension` directly in the map element.

PGM currently permits using the vanilla world generator and also allows for specifying a seed. However, it does not permit selecting a world folder. The only manner in which a world folder may be specified for a map is if it has a variant, which is not ideal from a mapmaking perspective when it comes to legacy and modern editions of maps.

The ability to conditionally specify maps per version is desirable for mapmakers as it allows mapmakers to avoid specifying variants for maps where the only difference between versions may "need" to be the world. For example, the map `Azure` used path blocks on 1.12, which were converted to a different block for 1.8 PGM. This is a minor aesthetic difference that should not warrant an entirely new variant.

A current example is `Polynesia`. Gameplay-wise, this map is the exact same between 1.8 and 1.21 outside of changes which may be defined per version in the `One in the Quiver` include. However, the map has a different variant for 1.21 to take advantage of new blocks and improve the look of the map. This should not necessarily have to be deemed a separate map as visible to players.

It is also already possible to define variant worlds per version.

Pablete1234:
Resolves #1457 and resolves #1511

Allows for the default variant to be specified, alongside a world:
This map is for 1.21 only (variants a, b, c) but the base map is 1.8 and uses a diff world.
```xml
<map min-server-version="1.21">
  <variant id="default" min-server-version="1.8" world="legacy"/>
  <variant id="typeA">A</variant>
  <variant id="typeB">B</variant>
  <variant id="typeB">C</variant>
</map>
```

This can also work for using a diff world when running modern or legacy:
```xml
<map>
  <variant id="default" world="${world-folder}"/>
  <variant id="typeA"  world="${world-folder}">A</variant>
  <variant id="typeB"  world="${world-folder}">B</variant>
  <variant id="typeB"  world="${world-folder}">C</variant>

  <constants>
    <if min-server-version="1.21">
      <constant id="world-folder">modern</constant>
    </if>
    <unless min-server-version="1.21">
      <constant id="world-folder" remove="true"/>
    </if>
  </constants>
</map>
```

Note: the name/override/slug attributes are NOT used for default variant, they must be specified in the map itself, just as before

### DIFF
--- a/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
+++ b/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
@@ -113,13 +113,14 @@ public class MapInfoImpl implements MapInfo {
 
   @NotNull
   private Map<String, VariantInfo> createVariantMap(Element root) throws InvalidXMLException {
-    Map<String, VariantInfo> variants = new LinkedHashMap<>();
-    variants.put(DEFAULT_VARIANT, new VariantData(root, null));
+    LinkedHashMap<String, VariantInfo> variants = new LinkedHashMap<>();
     for (Element el : root.getChildren("variant")) {
       VariantData vd = new VariantData(root, el);
       if (variants.put(vd.variantId, vd) != null)
         throw new InvalidXMLException("Duplicate variant ids are not allowed", el);
     }
+    if (!variants.containsKey(DEFAULT_VARIANT))
+      variants.putFirst(DEFAULT_VARIANT, new VariantData(root, null));
     return ImmutableMap.copyOf(variants);
   }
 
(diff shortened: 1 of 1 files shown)


cswhite2000 (approved):

calcastor:
This seems to work as intended

PR #1514, opened 2025-05-31
Update modern platform to 1.21.8

### LINKED ISSUE #1509: Items in shop GUIs on modern can be removed from chest menu
https://streamable.com/urt8hp

calcastor:
This PR updates PGM's modern platform to target 1.21.8.

There's several changes of note here:

1. From @TTtie, migrate to using a datapack for our custom legacy dimension.
   - Currently this uses the vanilla DFU, but we should look to targeting Paper's DataConverter when possible
   - This is more maintainable given Registry refactors in Paper, but breakage may be less obvious
3. Also from @TTtie, ensure we use `PluginClassLoader` to be able to keep legacy remapping. Closes https://github.com/PGMDev/PGM/issues/1509
2. In 1.21.5, `Sound` is now an interface. ~~Since PGM is a Paper plugin, this doesn't get automatically remapped so we need adjust accordingly.~~ Given the above change, we now remap but still keep some cleanup:
   - All statically defined sounds are now defined in PGM's sounds interface
   - All instances of these sounds are now played using Adventure
3. ~~Similarly, `Attribute` is now also an interface in 1.21.5.~~ Attribute changes:
   - All <1.13 attributes are parsed within PGM to support legacy XMLs, including those used by ProjectAres
   - Avoid SportPaper runtime errors by only parsing those which exist on 1.8 for that platform

Some current issues or items of note:

1. The DFU pieces of the custom dimension implementation are not extensively tested. ~~I have noticed block connection issues on certain maps at present.~~ Block connection issues fixed in recent commit.
2. ~~Boats do not get removed on 1.21.5 with the `kill-entities` action, but I do not know if they got removed on 1.21.1 either.~~
3. ~~Not certain about the change of packet in the modern implementation of `teleportEntityPacket`.~~
4. ~~Many events we extend are now protected in Paper 1.21.5. This PR changes affected events to extend `Event` rather than `PlayerEvent`, `EntityEvent`, `WorldEvent`, `BlockEvent` as such. Things seem to work to the extent that I have tested.~~ Resolved in JarIntersector.

This is buildable with the previous API intersect jar. ~~The aforementioned note regarding events came to light with a new intersect jar I created.~~ Resolved in JarIntersector.

Additionally, this PR supersedes https://github.com/PGMDev/PGM/pull/1508 and actually unblocks https://github.com/PGMDev/PGM/pull/1494 given the presence of Adventure 4.21.0 in Paper 1.21.5.


util/src/main/java/tc/oc/pgm/util/bukkit/EntityTypes.java
```diff
@@ -34,7 +34,17 @@ public class EntityTypes {
...
+      "BOAT",
+      "ACACIA_BOAT",
+      "BIRCH_BOAT",
+      "CHERRY_BOAT",
+      "DARK_OAK_BOAT",
+      "JUNGLE_BOAT",
+      "MANGROVE_BOAT",
+      "OAK_BOAT",
+      "PALE_OAK_BOAT",
+      "SPRUCE_BOAT");
```
Pablete1234: Yeah... this is not an "all of these" kinda situation, only the first to match will be considered the real mapping, so in modern the boat entity type just means acacia boat, this is why kill-entities with this entity type doesn't work.

If you want kill-entities to work you'd have to specify each individual boat in the kill-entities, or have a different/higher concept of "entity matcher" where boat can mean `*_BOAT` in modern, similar to what we do with blocks (ie: "WOOL" is taken to mean "WHITE_WOOL", "RED_WOOL", etc, as part of material matcher)

util/src/main/java/tc/oc/pgm/util/bukkit/MiscUtils.java
```diff
@@ -58,5 +58,7 @@ EntityChangeBlockEvent createEntityChangeBlockEvent(
 
   Key getSound(Sound constant);
 
+  Sound getSoundValue(String name);
```
Pablete1234: This is not really a "sound value", this is a sound.
 Consider renaming the previous method such that it's:

`Key getSoundKey(Sound constant);` 

and then renaming this one to

`Sound getSound(String name);`

util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
```diff
@@ -58,9 +64,13 @@ static Sound sound(String legacyConstant, String modernConstant, float volume, f
     // Sound.sound due to a compiler bug
     return Sound.sound(
         MISC_UTILS.getSound(
-            BukkitUtils.parse(org.bukkit.Sound::valueOf, legacyConstant, modernConstant)),
+            BukkitUtils.parse(MISC_UTILS::getSoundValue, legacyConstant, modernConstant)),
```
Pablete1234: At which point does this just become redundant? constant -> sound value -> sound key, may as well just do constant to key directly as a method (and then let the platform impl ensure that it exists)?

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
```diff
@@ -1047,9 +1047,6 @@ public static org.bukkit.attribute.Attribute parseAttribute(Node node, String te
     var attribute = Attributes.getByName(text);
     if (attribute != null) return attribute;
 
-    attribute = Attributes.getByName("generic" + text);
-    if (attribute != null) return attribute;
```
Pablete1234: i don't see why this is removable, isn't it still needed for 1.8?

util/src/main/java/tc/oc/pgm/util/attribute/Attributes.java
```diff
@@ -1,29 +1,58 @@
...
     for (String name : names) {
       BY_NAME.put(StringUtils.simplify(name), type);
     }
     return type;
   }
 
+  private static Attribute parse(boolean modernAttribute, String... names) {
+    if (modernAttribute && Platform.isModern()) return parse(names);
+    else return null;
+  }
```
Pablete1234: i don't see why we have this; if PGM does not use these attributes let's avoid giving them constants, just put the old (1.9 to 1.12) mappings in a static block for backwards-compat and we're good to go.
The method as it stands, with always `true` as param, makes no sense

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/inventory/ModernAttributeUtil.java
```diff
@@ -100,4 +103,22 @@ public void stripAttributes(ItemMeta meta) {
...
+  @Override
+  public String getAttributeName(Attribute attribute) {
+    return attribute.key().toString();
+  }
+
+  @Override
+  public Attribute getAttributeValue(String name) {
+    // Get from lower case as Paper does
+    return Registry.ATTRIBUTE.get(
+        Objects.requireNonNull(NamespacedKey.fromString(name.toLowerCase())));
```
Pablete1234: never use `toLowerCase()` for identifiers and such use `toLowerCase(Locale.ROOT)` instead

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
calcastor: It doesn't seem to be. Testing on SportPaper, a map is able to use `knockbackResistance` rather than `generic.knockbackResistance` and it works as intended without this block.

core/build.gradle.kts
```diff
@@ -22,6 +22,7 @@ tasks.named<ShadowJar>("shadowJar") {
     archiveFileName = "PGM.jar"
     archiveClassifier.set("")
     destinationDirectory = rootProject.projectDir.resolve("build/libs")
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
```
Pablete1234, line 25: seems worrying, are we duplicating classes? was this already the case before?

core/build.gradle.kts
```diff
@@ -83,6 +84,7 @@ tasks {
             expand(
                 "name" to project.name,
                 "description" to project.description,
+                "apiVersion" to "1.21.5",
```
Pablete1234: this is fake, we want remapping, meaning our api version is actually older (ie: we compile against a mostly 1.8 bukkit version, and things like class -> interface for inventory view need the remapping)

core/src/main/java/tc/oc/pgm/tnt/TNTMatchModule.java
```diff
@@ -92,7 +88,11 @@ public void handleInstantActivation(BlockPlaceEvent event) {
 
       if (callPrimeEvent(tnt, event.getPlayer())) {
         event.setCancelled(true); // Allow the block to be placed if priming is cancelled
-        world.playSound(tnt.getLocation(), FUSE_SOUND, 1, 1);
+        match.playSound(
+            Sounds.TNT_FUSE,
+            tnt.getLocation().getX(),
+            tnt.getLocation().getX(),
+            tnt.getLocation().getY());
```
Pablete1234: I'd consider adding a `playSound(Sound, Location)` to pgm's `Audience` interface, which would allow simplifying stuff like this to:

`match.playSound(Sounds.TNT_FUSE, tnt.getLocation());`

platform/platform-modern/build.gradle.kts
```diff
@@ -1,10 +1,10 @@
 plugins {
     id("buildlogic.java-conventions")
-    id("io.papermc.paperweight.userdev") version "1.7.4"
+    id("io.papermc.paperweight.userdev") version "2.0.0-beta.17"
```
Pablete1234: is there no stable release?

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/PgmBootstrap.java
```diff
@@ -31,44 +35,69 @@ public class PgmBootstrap implements PluginBootstrap {
...
+
+    // Paper 1.21.5 uses DFU to upgrade worlds, hook into that instead
+    PGMDataFixer.hookMojangDFU();
+  }
+
+  @Override
+  public JavaPlugin createPlugin(PluginProviderContext context) {
+    var ourClassLoader = (PaperPluginClassLoader) getClass().getClassLoader();
+    var pluginMeta = (PaperPluginMeta) ourClassLoader.getConfiguration();
+    var descriptor = new PluginDescriptionFile(
```
Pablete1234, line 76: is there really no alternative way of getting this?? calling a constructor with 40 params sucks, because this will break every time there's the smallest of updates.
Would be lovely if there was some method that we could call to turn PluginMeta into PluginDescriptionFile, but if that's not a thing then oh well

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/PgmBootstrap.java
```diff
@@ -31,44 +35,69 @@ public class PgmBootstrap implements PluginBootstrap {
...
+        pluginMeta.getName(),
+        pluginMeta.getName(),
+        pluginMeta.getProvidedPlugins(),
+        pluginMeta.getMainClass(),
+        null,
+        pluginMeta.getPluginDependencies(),
+        pluginMeta.getPluginSoftDependencies(),
+        pluginMeta.getLoadBeforePlugins(),
+        pluginMeta.getVersion(),
+        Collections.emptyMap(),
```
Pablete1234: `Map.of()`

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/PgmBootstrap.java
```diff
@@ -31,44 +35,69 @@ public class PgmBootstrap implements PluginBootstrap {
...
+        pluginMeta.getDescription(),
+        pluginMeta.getAuthors(),
+        pluginMeta.getContributors(),
+        pluginMeta.getWebsite(),
+        pluginMeta.getLoggerPrefix(),
+        pluginMeta.getLoadOrder(),
+        pluginMeta.getPermissions(),
+        pluginMeta.getPermissionDefault(),
+        Collections.emptySet(),
```
Pablete1234: `Set.of()`

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/PgmBootstrap.java
```diff
@@ -31,44 +35,69 @@ public class PgmBootstrap implements PluginBootstrap {
...
+        pluginMeta.getAPIVersion(),
+        Collections.emptyList());
```
Pablete1234: `List.of()`

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/dfu/PGMDataFixer.java
```diff
@@ -0,0 +1,52 @@
...
+import it.unimi.dsi.fastutil.ints.IntSortedSet;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Function;
+import net.minecraft.util.datafix.DataFixers;
+import tc.oc.pgm.util.DataVersions;
+import tc.oc.pgm.util.reflect.ReflectionUtils;
+
+public class PGMDataFixer {
```
Pablete1234: given this whole thing runs exactly once, i'd avoid static final fields that will be kept as references forever, and instead opt for having these in the method that needs them, or as non-static and just create one instance of this class that just gets GCed later

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/ModernEntityPackets.java
```diff
@@ -85,12 +85,12 @@ public Packet destroyEntitiesPacket(int... entityIds) {
 
   @Override
   public Packet teleportEntityPacket(int entityId, Location location) {
-    PacketContainer packet = PlPacket.PL.createPacket(PacketType.Play.Server.ENTITY_TELEPORT);
+    PacketContainer packet = PlPacket.PL.createPacket(PacketType.Play.Server.REL_ENTITY_MOVE_LOOK);
```
Pablete1234: You cannot make this change.
Teleport is absolute position, REL is... well, RELATIVE, which means this will break anything trying to use this

util/src/main/java/tc/oc/pgm/util/bukkit/Sounds.java
```diff
@@ -57,10 +63,13 @@ static Sound sound(String legacyConstant, String modernConstant) {
   static Sound sound(String legacyConstant, String modernConstant, float volume, float pitch) {
     // Sound.sound due to a compiler bug
     return Sound.sound(
-        MISC_UTILS.getSound(
-            BukkitUtils.parse(org.bukkit.Sound::valueOf, legacyConstant, modernConstant)),
+        BukkitUtils.parse(MISC_UTILS::getSoundKey, legacyConstant, modernConstant),
```
Pablete1234: Given in this context you know exactly one or the other and that getSoundKey methods WILL throw if not found, should probs just:
`MISC_UTILS.getSoundKey(Platform.isModern() ? modernConstant : legacyConstant)`

util/src/main/java/tc/oc/pgm/util/xml/XMLUtils.java
Pablete1234: That's only the case when both a GENERIC_* and an * version are defined for legacy and nonlegacy, if not defined, this allowed it to work regardless. I'm not sure if any attribute are like that where they aren't defined as both

Pablete1234 (changes requested):

platform/platform-modern/build.gradle.kts
calcastor: Not at present, and this is what Paper currently states in their docs to use.

core/build.gradle.kts
calcastor: SmartInvs has a duplicate `plugin.yml`, this causes issues given changes in Shadow 9.0.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/PgmBootstrap.java
TTtie: a plugin.yml file stream *could* be, in theory, fed into it, but I decided to reuse already parsed data so the same info doesn't have to be re-parsed again, even though the long call is discouraged by Paper developers themselves:

https://github.com/PaperMC/Paper/blob/2ba1675c7506cadd8a540ea452e5dafb79ae8947/paper-api/src/main/java/org/bukkit/plugin/PluginDescriptionFile.java#L270-L277

otherwise, there's no way to turn a PluginMeta into a PluginDescriptionFile, coincidentally, PaperPluginClassLoader does the same:

https://github.com/PaperMC/Paper/blob/2ba1675c7506cadd8a540ea452e5dafb79ae8947/paper-server/src/main/java/io/papermc/paper/plugin/entrypoint/classloader/PaperPluginClassLoader.java#L148-L178

core/build.gradle.kts
calcastor: This doesn't seem like it's relevant to remapping. We can't define 1.8 for paper plugins as it's older than it allows, an additionally it requires an api version to be defined. I think 1.21.5 makes sense given pieces of this PR that will not work on a modern version older than it, is there something else that is preferred?

I don't think we need to put this in the standard `plugin.yml` though, unless that also happens to be necessary.

Are we thinking of later down the line when Commodore is replaced with something else?

Pablete1234: > We can't define 1.8 for paper plugins as it's older than it allows,

You just leave it empty and it is assumed as "older than 1.13", 1.8 wasn't a version because it predates api versioning at all

core/build.gradle.kts
Pablete1234: that should probably be manually excluded then, instead of applying the more wild exclusion that could hurt us without knowing in the future

core/build.gradle.kts
calcastor: Not defining an API version for a paper plugin throws an error.

Pablete1234: Yeah nvm, looking at the docs <https://docs.papermc.io/paper/dev/plugin-yml/#api-version> it seems api version is just a way of saying "i won't run on servers older than X", not necessarily tied to the class remapping behavior. Should be fine

calcastor:
I've marked this as draft in favour of waiting until 1.21.6, as Paper will not be publishing stable 1.21.5 builds.

calcastor:
1.21.6 is still pending a couple things:

- [x] https://github.com/MrMicky-FR/FastBoard/pull/74
- [x] ProtocolLib support for 1.21.6 ~~(currently tab is broken and this seems to be the culprit)~~
- [ ] Updates to adventure-platform (actual cause of broken tab)

This drops many of the DFU changes made for 1.21.5 as they are unnecessary given Paper's DataConverter existing in 1.21.6, but maintains the custom dimension datapack, which additionally now lowers the cloud height to 128 like in 1.8.

calcastor:
This PR now uses a [PGM-specifc fork](https://github.com/calcastor/adventure-platform/commits/pgm/) of `adventure-platform` which includes a patch for 1.21.6 support.

Pablete1234 (approved):

PR #1515, opened 2025-06-04
Save block-NBT in structure module

Mew2K:
This PR extends the functionality of the `<structures>` module to allow any block-NBT data present within the structure region to be saved and restored. The implementation supports both 1.8 and 1.21.1 versions of PGM.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
+++ b/core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
@@ -2,6 +2,7 @@ package tc.oc.pgm.snapshot;
 
 import java.util.HashMap;
 import java.util.Map;
+import org.bukkit.Bukkit;
 import org.bukkit.ChunkSnapshot;
 import org.bukkit.World;
 import org.bukkit.block.Block;
@@ -14,10 +15,12 @@ import tc.oc.pgm.util.block.BlockData;
 import tc.oc.pgm.util.chunk.ChunkVector;
 import tc.oc.pgm.util.material.BlockMaterialData;
 import tc.oc.pgm.util.material.MaterialData;
+import tc.oc.pgm.util.nms.NMSHacks;
 
 public class WorldSnapshot {
   private final World world;
   private final Map<ChunkVector, ChunkSnapshot> chunkSnapshots = new HashMap<>();
+  private final Map<BlockVector, Object> savedNBT = new HashMap<>();
   private final BudgetWorldEdit worldEdit;
 
   public WorldSnapshot(World world) {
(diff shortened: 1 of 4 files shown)


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -31,19 +34,16 @@
...
 import net.minecraft.world.level.storage.LevelStorageSource;
 import net.minecraft.world.level.storage.PrimaryLevelData;
 import net.minecraft.world.level.validation.ContentValidationException;
-import org.bukkit.Bukkit;
-import org.bukkit.Chunk;
-import org.bukkit.Material;
-import org.bukkit.Nameable;
-import org.bukkit.World;
-import org.bukkit.WorldCreator;
+import org.bukkit.*;
```
cswhite2000, line 46: lets not use wildcard imports

core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
```diff
@@ -80,11 +83,36 @@ public void saveSnapshot(ChunkVector cv, @Nullable BlockState oldState) {
...
+      }
+    });
   }
 
   public void placeBlocks(Region region, BlockVector offset, boolean update) {
-    worldEdit.placeBlocks(region, offset, update);
+    Region.Static staticRegion = region.getStatic(world);
+    worldEdit.placeBlocks(staticRegion, offset, update);
+    staticRegion.getBlockVectors().forEach(pos -> {
+      Block block = world.getBlockAt(
```
cswhite2000, line 103: No need to make the block here, as most locations should have no nbt. So you're allocating lots of uneeded objects.

Probably better off moving the whole block creation into setBlockNBT
Something like `setNBTat(World world, BlockVector vec, Object nbt)`

core/src/main/java/tc/oc/pgm/snapshot/WorldSnapshot.java
```diff
@@ -80,11 +83,36 @@ public void saveSnapshot(ChunkVector cv, @Nullable BlockState oldState) {
   }
 
   public void saveRegion(Region region) {
-    region.getChunkPositions().forEach(cv -> this.saveSnapshot(cv, null));
+    Region.Static staticRegion = region.getStatic(world);
+    staticRegion.getChunkPositions().forEach(cv -> this.saveSnapshot(cv, null));
+    staticRegion.getBlockVectors().forEach(pos -> {
+      Block block = world.getBlockAt(pos.getBlockX(), pos.getBlockY(), pos.getBlockZ());
```
cswhite2000, line 89: Same here, move the block creation into the nms method.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -393,4 +390,36 @@ public int getMaxWorldSize(World world) {
   public int allocateEntityId() {
     return Bukkit.getUnsafe().nextEntityId();
   }
+
+  @Override
+  public Object getBlockNBT(Block block) {
+    ServerLevel level = ((CraftWorld) block.getWorld()).getHandle();
```
cswhite2000, line 396: We should try and see if there's an efficient way we can exit early if the location has no relevant data

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/impl/SpNMSHacks.java
```diff
@@ -229,4 +218,45 @@ public int getMaxWorldSize(World world) {
   public int allocateEntityId() {
     return Bukkit.allocateEntityId();
   }
+
+  @Override
+  public Object getBlockNBT(Block block) {
+    TileEntity tile = ((CraftWorld) block.getWorld())
```
cswhite2000, line 224: Same here, this seems like an expensive way of checking to see if data is present

cswhite2000 (commented):
Commit's not signed off correctly.

Also, this should be configurable via XML. Should probably default this to false as well, to make this not a breaking change.

And we should explore if there is a more efficient way of doing this than serializing everything to NBT and then deserializing it.

Pablete1234 (changes requested):
I think you're severely underestimating the performance impact that this will have.
Try saving and restoring a big region (eg: bed rush map) and profile the timing and memory difference between the current implementation, which avoids allocations at every opportunity, and the proposed implementation.

If this is to land in pgm, it will have to be after some severe re-thinking of the performance implications

Issue #1513, opened 2025-05-31
Adding a linking exception to the licence


Pablete1234:
PGM has been AGPL-licenced ever since it was released, however, the licence has been invalid all along, as pgm builds by linking into craftbukkit and nms code -which is proprietary- and forbidden by the GPL/AGPL licence.

The only ways around this would be:
 - For PGM to stop relying on craftbukkit entirely, depending only on the bukkit API
   - This is completely unfeasible, pgm relies too much on internals.
 - For PGM to release without platform modules
   - Would require server owners to download platform modules separately
   - Would require re-licencing or re-creating the platform modules, but it's a smaller portion, modified by way less contributors
 - For PGM to add a [linking exception for minecraft](https://gist.github.com/triphora/588f353802a3b0ea649e4fc85f75e583)
   - I believe this is by far the best option, but requires the most contributors to agree to the change.
   - Arguably it's not even a "change"; i doubt any contribution was expected not to run in the proprietary minecraft server.

Preliminary list of contributors that would be required to agree to the change (taken from github stats at time of writting):
- [x] @Pablete1234  (367 commits)
- [x]  @Electroid (179 commits)
- [x] @applenick  (117 commits)
- [x] @Pugzy (68 commits)
- [x] @KingOfSquares (36 commits)
- [x] @TheMolkaPL (34 commits)
- [x] @CoWinkKeyDinkInc (31 commits)
- [x] @calcastor (31 commits)
- [x] @cswhite2000 (31 commits)
- [x] @AustinLMayes (29 commits)
- [x] @Meeples10 (21 commits)
- [x] @OhPointFive (14 commits)
- [x] @TTtie (12 commits)
- [x] @EricZeiberg (10 commits)
- [x] ~~dependabot (9 commits)~~ - these are trivial and non-copyrigyhtable
- [x] @alexsosnovsky (8 commits)
- [x] @zzufx (8 commits)
- [x] @BGMP (7 commits)
- [x] @chatasma (6 commits)
- [x] @rmsy (6 commits)
- [x] @dentmaged (6 commits)
- [x] @botinator (6 commits)
- [x] @mrcookieunderscore13 (5 commits)
- [x] @jacobbearden (4 commits)
- [x] @lPoloBlash (4 commits
- [x] @PabloPintor (3 commits)
- [x] @iangbb (3 commits)
- [x] @chemistryx (2 commits)
- [x] @jasoryeh (2 commits)
- [x] @arcadeboss (1 commit)
[...]

PR #1517, opened 2025-06-08
Add lighting strike option to blitz module

TBG1000:
Adds a simple `lightning` attribute to the `<blitz>` module. This produces a lightning strike on the player's death location when they are eliminated (no lives left), the same way a [smoke effect](https://github.com/PGMDev/PGM/blob/8263eab3c3f0324a12b888863a0b6ae2f789dbfc/core/src/main/java/tc/oc/pgm/blitz/BlitzMatchModule.java#L168) is produced.

Implemented according to [Pablo's comment](https://github.com/PGMDev/PGM/issues/1217#issuecomment-2953632887) here. At Warzone, we intend to use this for Survival Games maps, but can of course be used for any other Blitz maps.

(Other changes in the code are due to the formatter Spotless)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blitz/BlitzConfig.java
+++ b/core/src/main/java/tc/oc/pgm/blitz/BlitzConfig.java
@@ -9,6 +9,7 @@ public class BlitzConfig {
 
   private final int lives;
   private final boolean broadcastLives;
+  private final boolean lightning;
   private final Filter filter;
   private final Filter scoreboardFilter;
   private final Filter joinFilter;
@@ -16,6 +17,7 @@ public class BlitzConfig {
   public BlitzConfig(
       int lives,
       boolean broadcastLives,
+      boolean lightning,
       Filter filter,
       Filter scoreboardFilter,
       Filter joinFilter) {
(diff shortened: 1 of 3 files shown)


Pugzy:
This feels like it should just be an attribute on the blitz element rather than a child (as this is no longer standard). i.e. 
```xml
<blitz lightning-strikes="true"/> 
<blitz lightning="true"/> <!-- or maybe just 'lightning' -->
```

Pablete1234:
Agree with pugzy, should be renamed to simply `lightning` for shortness, and should be parsed as a property (ie: either attribute or child) instead of as a child.

TBG1000:
Updated and tested, thanks for the feedback!

TBG1000:
Would love to see this merged, if what I changed was not what you were looking for or what I was supposed to do, please let me know. Happy to learn and adjust accordingly

Pablete1234:
Trying to get #1513 thru before we take in any more changes

TBG1000:
Rebased to be in agreement with the recently merged linking exception

Pablete1234 (approved):

PR #1528, opened 2025-07-20
fix(rage): determine projectile eligibility using used weapon on modern

TTtie:
In modern Minecraft, `AbstractArrow#getDamage()` (and by extension, `MiscUtils#getArrowDamage`) returns the base damage before applying any enchantment multipliers. However, we can leverage the item stack in `AbstractArrow#getWeapon()` to determine if the weapon used has a power enchantment and act accordingly.

This also enables the rage module to work with thrown tridents if they're enchanted with Power I.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rage/RageMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/rage/RageMatchModule.java
@@ -2,9 +2,9 @@ package tc.oc.pgm.rage;
 
 import static tc.oc.pgm.util.bukkit.MiscUtils.MISC_UTILS;
 
-import org.bukkit.entity.Arrow;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
+import org.bukkit.entity.Projectile;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/rage/RageMatchModule.java
```diff
@@ -28,13 +29,24 @@ public void handlePlayerDamage(EntityDamageByEntityEvent event) {
...
+        // In modern Minecraft, we have access to the item stack that fired the projectile.
+        // We can leverage that to figure out if the projectile comes from a rage weapon or not
+        var item = MISC_UTILS.getProjectileWeapon(proj);
+        return item != null && item.containsEnchantment(Enchantments.POWER);
+      } else {
+        // On 1.8 the arrow damage takes enchantments into account
+        if (damager instanceof Arrow arrow) {
+          // Arrows with damage > 2 are from power bows.
+          return MISC_UTILS.getArrowDamage(arrow) > 2.0D;
+        }
```
Pablete1234: the whole point of platform utils etc is to hide away platform-dependant code, ideally you'd just have something like:
```suggestion
      return MISC_UTILS.hasPowerEnchant(proj);
```
and let the different implementations figure out how that is done, legacy can do arrow damage > 2.0D, modern can do the get-item and check for power

Pablete1234 (approved):

PR #1531, opened 2025-07-21
Fix attributes breaking backwards compatibility

Pablete1234:
Allows `generic.whatever` to always work no matter if included or excluded, since the `generic.` prefix will always be stripped. The same goes for `generic_*`,  `player.*` or `player_*`


cswhite2000:
Forgot to run the formatter?

Pablete1234:
not just the formatter, also reworked it a bit

cswhite2000 (approved):

PR #1533, opened 2025-07-22
Add drop-flag and pickup-flag actions

ThatApplePieGuy:
Adds actions for transferring flags

```xml
<action id="pickup-red-flag" scope="player" expose="true">
    <pickup-flag flag="red-flag"/>
</action>

<action id="drop-red-flag" scope="match" expose="true">
    <drop-flag flag="red-flag"/>
</action>
```

Internally this means that flags can now transition from any state (bar Completed) to Carried. There shouldn't be any illegal transitions that become possible with this, but I could be missing something

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/state/BaseState.java
+++ b/core/src/main/java/tc/oc/pgm/flag/state/BaseState.java
@@ -21,14 +21,17 @@ import tc.oc.pgm.api.party.Party;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.flag.Flag;
+import tc.oc.pgm.flag.FlagMatchModule;
 import tc.oc.pgm.flag.Post;
 import tc.oc.pgm.flag.event.FlagCaptureEvent;
+import tc.oc.pgm.flag.event.FlagPickupEvent;
 import tc.oc.pgm.flag.event.FlagStateChangeEvent;
 import tc.oc.pgm.goals.events.GoalEvent;
 import tc.oc.pgm.spawns.events.ParticipantDespawnEvent;
 import tc.oc.pgm.teams.Team;
 import tc.oc.pgm.teams.TeamMatchModule;
 import tc.oc.pgm.util.TimeUtils;
+import tc.oc.pgm.util.bukkit.Sounds;
 
 /** Base class for all {@link Flag} states */
 public abstract class BaseState implements Runnable, State {
@@ -37,6 +40,7 @@ public abstract class BaseState implements Runnable, State {
   protected final Post post;
   protected final Instant enterTime;
   protected @Nullable Long remainingTicks;
+  protected @Nullable MatchPlayer pickingUp;
   private Future<?> task;
 
   protected BaseState(Flag flag, Post post) {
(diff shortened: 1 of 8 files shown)


core/src/main/java/tc/oc/pgm/flag/state/BaseState.java
```diff
@@ -139,7 +176,8 @@ protected void finishCountdown() {}
 
   @Override
   public boolean isCarrying(MatchPlayer player) {
-    return false;
+    // This allows CarryingFlagFilter to match and cancel the pickup before it actually happens
```
Pablete1234, line 177: This seems like a hack, the flag hasnt even been picked up yet, we can't pretend it has been picked. If you want to prevent pickups there's a filter for that

core/src/main/java/tc/oc/pgm/flag/state/BaseState.java
```diff
@@ -148,9 +186,8 @@ public boolean isCarrying(ParticipantState player) {
     return matchPlayer != null && isCarrying(matchPlayer);
   }
 
-  @Override
   public boolean isCarrying(Party party) {
-    return false;
+    return (this.pickingUp != null && party == this.pickingUp.getParty());
```
Pablete1234: Same here

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
```diff
@@ -215,29 +215,39 @@ public void tickRunning() {
 
   @Override
   public boolean isCarrying(MatchPlayer player) {
-    return this.carrier == player;
+    if (this.pickingUp == null) {
```
Pablete1234: Same here...

core/src/main/java/tc/oc/pgm/flag/state/Dropped.java
```diff
@@ -85,7 +97,7 @@ public boolean isRecoverable() {
   }
 
   @Override
-  protected boolean canPickup(MatchPlayer player) {
```
Pablete1234: Is the visibility change needed? I think all callers are now within state

core/src/main/java/tc/oc/pgm/flag/state/Completed.java
```diff
@@ -24,7 +24,7 @@ public Iterable<Location> getProximityLocations(ParticipantState player) {
   }
 
   @Override
-  protected boolean canPickup(MatchPlayer player) {
+  public boolean canPickup(MatchPlayer player) {
```
Pablete1234: Same here

core/src/main/java/tc/oc/pgm/flag/state/Uncarried.java
```diff
@@ -142,24 +107,14 @@ protected boolean inPickupRange(Player player) {
...
-
   @Override
   public void onEvent(PlayerMoveEvent event) {
     super.onEvent(event);
     MatchPlayer player = this.flag.getMatch().getPlayer(event.getPlayer());
     if (player == null || !player.canInteract() || player.getBukkit().isDead()) return;
 
     if (this.inPickupRange(player.getBukkit()) && this.canPickup(player)) {
-      this.pickupFlag(player);
+      this.flag.pickupFlag(player, this.location);
```
Pablete1234: I wouldnt route this thru flag, you have the method directly in state
```suggestion
      this.pickupFlag(player, this.location);
```

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
```diff
@@ -215,29 +215,39 @@ public void tickRunning() {
...
   }
 
-  protected void dropFlag() {
+  @Override
+  public boolean dropFlag() {
     for (Location dropLocation : this.dropLocations) {
       if (this.flag.canDrop(new PlayerQuery(null, carrier, dropLocation))) {
         this.flag.transition(new Dropped(this.flag, this.post, dropLocation, this.carrier));
-        return;
+        return true;
```
Pablete1234: Is the return value used for anything? As far as i can see the result is just ignored from the caller

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
ThatApplePieGuy: pickupFlag was also an unused boolean so I was matching that at the time, I removed the return type now to be clearer

core/src/main/java/tc/oc/pgm/flag/state/BaseState.java
ThatApplePieGuy: Discussed on discord, this matches old behavior

Pablete1234 (approved):

PR #1534, opened 2025-07-24
fix: correct wind charge behavior in modern

TTtie:
Makes it so wind charges and other non-destructive explosions:
- don't break blocks with custom drops
- don't count as a core/monument touch
- don't set off dispenser nukes

That said, it might not be 100% desirable to call a `BlockTransformEvent` for non-destructive explosions, as the non-destructive `ExplosionResult`s either won't do anything with the blocks (value of `KEEP`) or ~~will fire collateral block physics events~~ (value of `TRIGGER_BLOCK` - e. g. wind charges). Currently, they're fired without a state change.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsMatchModule.java
@@ -68,7 +68,9 @@ public class BlockDropsMatchModule implements MatchModule, Listener {
   }
 
   public static boolean causesDrops(final Event event) {
-    return event instanceof BlockBreakEvent || event instanceof EntityExplodeEvent;
+    return event instanceof BlockBreakEvent
+        || (event instanceof EntityExplodeEvent explodeEvent
+            && MISC_UTILS.isDestructiveExplosion(explodeEvent));
   }
 
   @EventHandler(priority = EventPriority.LOW)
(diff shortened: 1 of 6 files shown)


Pablete1234:
Regarding triggering BlockTransformEvent(s), i believe it makes sense to do so for things like flipping trapdoors etc, since those *are* changing the block state, and it makes sense that if a map has disabled modifying trapdoors you can't modify them with wind charges.
I don't know about blocks that won't change tho, eg: firing a windcharge against a wall of wood, where no block changes, should probably not trigger block transform events

TTtie:
OK, so it turns out I was somewhat misinformed by my logs, and the block updates can't be caught down the line. 🤦‍♀️

As for explosions with an `ExplosionResult.KEEP`, sure, it makes sense not to not fire anything but as for `ExplosionResult.TRIGGER_BLOCK` I think it might actually better to just fire it for every block because what happens when a block is "blown up" is up to the server even for regular explosions.

That said, we could map the new block state (the subset of blocks affected by wind charges is known and *relatively* small) but keeping it as a transform to the same state sounds more maintainable to me.

Marking as draft for now to refactor to ignore `ExplosionResult.KEEP` later

TTtie:
After further review of server code, an explosion event in 1.21 with an `ExplosionResult` of `KEEP` will never fire under normal circumstances as the method firing explosion events filters such results (ref: `explode()`, `interactWithBlocks()` both in `net.minecraft.world.level.ServerExplosion`). 

As such, it is safe to keep this code as-is – if it ever gets triggered with such a result, it would be better to be safe than sorry anyways.

Pablete1234 (approved):

PR #1540, opened 2025-08-04
Flags: acknowledge show-messages and show-effects; additional cleanup

calcastor:


core/src/main/java/tc/oc/pgm/flag/state/Carried.java
```diff
@@ -122,6 +123,8 @@ public void enterState() {
     SidebarMatchModule smm = this.flag.getMatch().getModule(SidebarMatchModule.class);
     if (smm != null) smm.blinkGoal(this.flag, 2, null);
 
+    if (!this.flag.hasShowOption(ShowOption.SHOW_MESSAGES)) return;
+
     if (this.flag.getDefinition().willShowRespawnOnPickup()) {
```
Pablete1234: You should merge these two ifs

core/src/main/java/tc/oc/pgm/flag/state/Respawning.java
```diff
@@ -57,6 +58,8 @@ public void enterState() {
     // Respawn is delayed
     String postName = this.post.getPostName();
 
+    if (!this.flag.hasShowOption(ShowOption.SHOW_MESSAGES)) return;
```
Pablete1234: While i'm generally a fan of early returns for pre-condition checks that affect the whole method, i would avoid it in this case where you're checking for a pretty narrow condition (if messages should be shown) on a pretty wide/generic method (enter state). (eg: if this method was called `showWhateverMessage` then the early return would be more than justified.)

As is, it can cause future bugs if someone adds more code to the end of the method that they expect runs when you enter this state, and will appear to work fine for most flags during testing, but then a map without flag messages will break

core/src/main/java/tc/oc/pgm/flag/state/Respawning.java
```diff
@@ -69,8 +72,11 @@ public void enterState() {
   }
 
   protected void respawn(@Nullable Component message) {
-    if (message != null) {
+    if (this.flag.hasShowOption(ShowOption.SHOW_EFFECTS)) {
       this.flag.playStatusSound(Sounds.FLAG_RETURN_OWN, Sounds.FLAG_RETURN);
```
Pablete1234: this introduces a behavior change; now the sound WILL play and the message won't, where as before neither would happen if message was null.

I believe this is actually never called with null so i'm unsure why it is marked nullable in the first place, and it appears the first commit to add this method (over 10 years ago) already had it marked nullable, and all 3 calls to it were never null.

So i guess either drop the nullable and the null checks, or if you wanna be extra cautious, keep the older behavior where both check for msg not null

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #1543, opened 2025-08-06
Permit disabling manual flag drop

calcastor:


Pablete1234:
I'm not sure if this is a good mechanic
it creates the false sense that by just locking it the flag can't be dropped, but there's still many reasons why the flag will drop regardless (leaving, switching team, joining obs, dying, etc), is there a good reason why you really want the player not to be able to give up the flag?

calcastor:
> I'm not sure if this is a good mechanic it creates the false sense that by just locking it the flag can't be dropped, but there's still many reasons why the flag will drop regardless (leaving, switching team, joining obs, dying, etc), is there a good reason why you really want the player not to be able to give up the flag?

Maybe the term should change in this context. Essentially I would like it to not be possible for the flag to be manually dropped for Touchdown maps.

Either way this is broken on modern at the moment, so marked as draft.

Pablete1234:
What if this was renamed to instead of `pin-to-carrier` simply be `avoid-manual-drop`? maybe that change in naming is enough to make it clear it only prevents manual dropping, could be `avoid-manual-drop`, `disable-manual-drop` or something else along those lines maybe

Pablete1234 (approved):

Pablete1234 (approved):

PR #1547, opened 2025-08-27
Add team-alias action

Javggg:
While working on a map, I thought it would be useful to dynamically change team names. I knew this was already possible in PGM using a command, so I decided it might be a good idea to add an XML action that performs the same task.

The code works, but there are still some missing checks (e.g. verifying whether the `team-id` actually corresponds to a defined team, or if the `alias` is already in use). I decided to leave those out for now and wait for feedback on the idea before refining the implementation.

To use it, simply define:

```xml
<team-alias team-id="red-team" alias="Attackers" />
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -46,7 +46,6 @@ import tc.oc.pgm.api.filter.Filterables;
 import tc.oc.pgm.api.filter.query.PartyQuery;
 import tc.oc.pgm.api.map.MapProtos;
 import tc.oc.pgm.api.map.factory.MapFactory;
-import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.party.Party;
 import tc.oc.pgm.api.player.MatchPlayer;
 import tc.oc.pgm.features.FeatureDefinitionContext;
@@ -62,6 +61,7 @@ import tc.oc.pgm.shops.ShopModule;
 import tc.oc.pgm.shops.menu.Payable;
 import tc.oc.pgm.structure.StructureDefinition;
 import tc.oc.pgm.teams.TeamFactory;
+import tc.oc.pgm.teams.TeamMatchModule;
 import tc.oc.pgm.util.MethodParser;
 import tc.oc.pgm.util.MethodParsers;
 import tc.oc.pgm.util.inventory.ItemMatcher;
@@ -419,22 +419,17 @@ public class ActionParser {
   }
 
   @MethodParser("team-alias")
-  public <T extends Filterable<?>> TeamAliasAction<T> parseTeamAliasAction(
-      Element el, Class<T> scope) throws InvalidXMLException {
-    scope = parseScope(el, scope);
+  public <T extends Filterable<?>> Action<?> parseTeamAliasAction(Element el, Class<T> scope)
+      throws InvalidXMLException {
     String alias = parser.string(el, "alias").required();
+    var action = new TeamAliasAction(alias);
+    var teamBuilder = parser.reference(TeamFactory.class, el, "team");
+    var team = scope == Party.class ? teamBuilder.orNull() : teamBuilder.required();
 
-    if (scope.getSimpleName().equals("Match")) {
-      var team = parser.reference(TeamFactory.class, el, "team").required();
-      return new TeamAliasAction.WithTeam<>(scope, alias, team);
-    } else if (scope.getSimpleName().equals("Party")) {
-      return new TeamAliasAction<>(scope, alias);
-    } else {
-      throw new InvalidXMLException(
-          "Wrong scope defined for action, scope must be " + Match.class.getSimpleName() + " or "
-              + Party.class.getSimpleName(),
-          el);
-    }
+    return team == null
+        ? action
+        : new ScopeSwitchAction<>(
+            scope, f -> f.moduleRequire(TeamMatchModule.class).getTeam(team.get()), null, action);
   }
 
   @MethodParser("take-payment")
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -19,24 +19,7 @@ import org.bukkit.enchantments.Enchantment;
 import org.bukkit.inventory.ItemStack;
 import org.jdom2.Element;
 import org.jetbrains.annotations.Nullable;
-import tc.oc.pgm.action.actions.ActionNode;
-import tc.oc.pgm.action.actions.DropFlagAction;
-import tc.oc.pgm.action.actions.EnchantItemAction;
-import tc.oc.pgm.action.actions.ExposedAction;
-import tc.oc.pgm.action.actions.FillAction;
-import tc.oc.pgm.action.actions.KillEntitiesAction;
-import tc.oc.pgm.action.actions.MessageAction;
-import tc.oc.pgm.action.actions.PasteStructureAction;
-import tc.oc.pgm.action.actions.PickupFlagAction;
-import tc.oc.pgm.action.actions.RepeatAction;
-import tc.oc.pgm.action.actions.ReplaceItemAction;
-import tc.oc.pgm.action.actions.ScopeSwitchAction;
-import tc.oc.pgm.action.actions.SetVariableAction;
-import tc.oc.pgm.action.actions.SoundAction;
-import tc.oc.pgm.action.actions.TakePaymentAction;
-import tc.oc.pgm.action.actions.TeleportAction;
-import tc.oc.pgm.action.actions.VelocityAction;
-import tc.oc.pgm.action.actions.WeatherAction;
+import tc.oc.pgm.action.actions.*;
 import tc.oc.pgm.action.replacements.Replacement;
 import tc.oc.pgm.action.replacements.ReplacementParser;
 import tc.oc.pgm.api.feature.FeatureValidation;
@@ -412,6 +395,15 @@ public class ActionParser {
         parser.filter(el, "filter").orNull(),
         parser.parseBool(el, "update").orTrue(),
         parser.parseBool(el, "events").orFalse());
+  }
+
+  @MethodParser("team-alias")
+  public TeamAliasAction parseTeamAliasAction(Element el, Class<?> scope)
+      throws InvalidXMLException {
+    String teamId = Node.fromRequiredAttr(el, "team-id").getValue();
+    String alias = Node.fromRequiredAttr(el, "alias").getValue();
+
+    return new TeamAliasAction(teamId, alias);
   }
 
   @MethodParser("take-payment")
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -19,7 +19,25 @@ import org.bukkit.enchantments.Enchantment;
 import org.bukkit.inventory.ItemStack;
 import org.jdom2.Element;
 import org.jetbrains.annotations.Nullable;
-import tc.oc.pgm.action.actions.*;
+import tc.oc.pgm.action.actions.ActionNode;
+import tc.oc.pgm.action.actions.DropFlagAction;
+import tc.oc.pgm.action.actions.EnchantItemAction;
+import tc.oc.pgm.action.actions.ExposedAction;
+import tc.oc.pgm.action.actions.FillAction;
+import tc.oc.pgm.action.actions.KillEntitiesAction;
+import tc.oc.pgm.action.actions.MessageAction;
+import tc.oc.pgm.action.actions.PasteStructureAction;
+import tc.oc.pgm.action.actions.PickupFlagAction;
+import tc.oc.pgm.action.actions.RepeatAction;
+import tc.oc.pgm.action.actions.ReplaceItemAction;
+import tc.oc.pgm.action.actions.ScopeSwitchAction;
+import tc.oc.pgm.action.actions.SetVariableAction;
+import tc.oc.pgm.action.actions.SoundAction;
+import tc.oc.pgm.action.actions.TakePaymentAction;
+import tc.oc.pgm.action.actions.TeamAliasAction;
+import tc.oc.pgm.action.actions.TeleportAction;
+import tc.oc.pgm.action.actions.VelocityAction;
+import tc.oc.pgm.action.actions.WeatherAction;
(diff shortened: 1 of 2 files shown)


core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -19,24 +19,7 @@
...
-import tc.oc.pgm.action.actions.RepeatAction;
-import tc.oc.pgm.action.actions.ReplaceItemAction;
-import tc.oc.pgm.action.actions.ScopeSwitchAction;
-import tc.oc.pgm.action.actions.SetVariableAction;
-import tc.oc.pgm.action.actions.SoundAction;
-import tc.oc.pgm.action.actions.TakePaymentAction;
-import tc.oc.pgm.action.actions.TeleportAction;
-import tc.oc.pgm.action.actions.VelocityAction;
-import tc.oc.pgm.action.actions.WeatherAction;
+import tc.oc.pgm.action.actions.*;
```
Pugzy: Avoid wildcard imports. Your editor may be doing this automatically.

Pablete1234:
Could you elaborate beyond "it would be useful" on what your use-case is here? I have a hard time finding a good use for this.

This feels like an annoying feature for players to deal with if used wrongly (eg: /join attackers now no longer works or is pointing to a different team).

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
TTtie, line 1: It might be worthwhile to make this team-scoped

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -414,6 +397,15 @@ public FillAction parseFill(Element el, Class<?> scope) throws InvalidXMLExcepti
         parser.parseBool(el, "events").orFalse());
   }
 
+  @MethodParser("team-alias")
+  public TeamAliasAction parseTeamAliasAction(Element el, Class<?> scope)
+      throws InvalidXMLException {
+    String teamId = Node.fromRequiredAttr(el, "team-id").getValue();
+    String alias = Node.fromRequiredAttr(el, "alias").getValue();
```
TTtie: Merely a style thing, but possibly fluent parsing could be used here? Not sure if using it is semantically equivalent and consistent with regular team ID + name parsing, however

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,33 @@
...
+    this.alias = alias;
+  }
+
+  @Override
+  public void trigger(Match match) {
+    List<Party> teamsList = new ArrayList<>(match.getParties());
+
+    for (Party team : teamsList) {
+      if (team instanceof Team) {
+        String teamId = ((Team) team).getId();
```
TTtie: Use pattern matching
```suggestion
      if (team instanceof Team t) {
        String teamId = t.getId();
```

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public TeamAliasAction(String teamId, String alias) {
+    super(Match.class);
+    this.teamId = teamId;
+    this.alias = alias;
+  }
+
+  @Override
+  public void trigger(Match match) {
+    List<Party> teamsList = new ArrayList<>(match.getParties());
```
TTtie: Is the defensive copy needed? I'm pretty sure you could iterate over `match.getParties()` directly

Javggg:
> Could you elaborate beyond "it would be useful" on what your use-case is here? I have a hard time finding a good use for this.
> 
> This feels like an annoying feature for players to deal with if used wrongly (eg: /join attackers now no longer works or is pointing to a different team).

Thanks for the feedback! I agree that frequent renaming during a match could be confusing, so the intended use is for more controlled scenarios.

For example:
- A mini-tournament map (like Touchdown) where each team gets a random country name.
- A multi-phase or round-based map where it makes sense to update names to reflect the current role of the team.

To reduce misuse, there could also be restrictions such as:
- Aliases cannot overlap.
- Once an alias has been used by one team, no other team can reuse it in the same match.

This way aliases act more like contextual labels for players rather than something that changes unpredictably mid-game.

Pablete1234:
Sorry for the late reply, it slipped my mind

> A mini-tournament map (like Touchdown) where each team gets a random country name.
> A multi-phase or round-based map where it makes sense to update names to reflect the current role of the team.

I am definitely not a fan of randomized team names on a map; they're already somewhat problematic because data will no longer align well in the database (ie: trying to query which team wins more often for asymmetric modes, you can't quite group the data by team name anymore because someone messed with it).

For multi-phase or round-based scenario i can see an ok usecase, which would be if the team name is modified with some qualifier/adjedctive, eg "Blue" turns "Blue Novices" or later "Blue Warriors", but swapping around "defender" and "attacker" teams for asymmetric 2-round matches is NOT something xml should be doing, that's something either PGM itself needs to support (would be nice to see round-based gamemode support in pgm itself), or a 3rd party plugin should implement as 2 pgm matches.


Regarding implementation, i'll leave a review with points that need to be addressed

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.party.Party;
+import tc.oc.pgm.teams.Team;
+
+public class TeamAliasAction extends AbstractAction<Match> {
+  private final String teamId;
```
Pablete1234: Should not be a string, this should be a reference to a team, will leave more details at parsing

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
Pablete1234: This shouldn't even be needed at all, the comment can be ignored

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
Pablete1234: same here

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,33 @@
...
+  public void trigger(Match match) {
+    List<Party> teamsList = new ArrayList<>(match.getParties());
+
+    for (Party team : teamsList) {
+      if (team instanceof Team) {
+        String teamId = ((Team) team).getId();
+        if (Objects.equals(teamId, this.teamId)) {
+          team.setName(this.alias);
+        }
+      }
```
Pablete1234:
```suggestion
    match.getFeatureContext().get(team.get().getId(), Team.class).setName(this.alias);
```

core/src/main/java/tc/oc/pgm/action/ActionParser.java
Pablete1234: +1

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -414,6 +397,15 @@ public FillAction parseFill(Element el, Class<?> scope) throws InvalidXMLExcepti
         parser.parseBool(el, "events").orFalse());
   }
 
+  @MethodParser("team-alias")
+  public TeamAliasAction parseTeamAliasAction(Element el, Class<?> scope)
+      throws InvalidXMLException {
+    String teamId = Node.fromRequiredAttr(el, "team-id").getValue();
```
Pablete1234: Should be parsing a feature reference to a team factory, that ensures that the team must exist in the map, and if it doesn't will throw corresponding xml parsing error:

```suggestion
    var team = parser.reference(TeamFactory.class, el, "team").required();
```
Also note, we do not use "team-id" anywhere in pgm, things are never suffixed "-id", so it should just be `"team"`

Also note, team could be optional and have two implementations:
 - if team is informed, the current match-scoped action runs
 - if team is not informed, a team-scoped action runs instead

As an example you can see how `SetVariableAction` parsing also has a `SetVariableAction.Indexed` variant, this is all an implementation detail, towards the user it's just as if this action (team-alias) works either with or without a specified team.

core/src/main/java/tc/oc/pgm/action/ActionParser.java
Pablete1234: +1, could use fluent parsing

```suggestion
    String alias = parser.string(el, "alias").required();
```

Pablete1234 (changes requested):

Javggg:
Sorry for the long delay in addressing the reviews.

I’ve made the suggested changes. The action now works with scope taken into account: if the scope is team, the attribute is not required; if it’s match, then it is required.

I also updated the behavior of the TeamAliasAction class to make it simpler.

One important thing that should be improved: I used magic strings in ActionParser to distinguish between the scopes. I’m aware this is not the best approach, but I couldn’t find another way to make it work.

```java
if (scope.getSimpleName().equals("Match")) {
    var team = parser.reference(TeamFactory.class, el, "team").required();
    return new TeamAliasAction.WithTeam<>(scope, alias, team);
} else if (scope.getSimpleName().equals("Party")) {
    return new TeamAliasAction<>(scope, alias);
}
```

I’m sure there is a better solution... please point me to a part of the code where something similar is handled, or feel free to commit a fix directly. Thanks!

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,45 @@
+package tc.oc.pgm.action.actions;
+
+import tc.oc.pgm.api.feature.FeatureReference;
+import tc.oc.pgm.filters.Filterable;
+import tc.oc.pgm.match.MatchImpl;
+import tc.oc.pgm.match.PartyImpl;
+import tc.oc.pgm.teams.Team;
+import tc.oc.pgm.teams.TeamFactory;
+
+public class TeamAliasAction<T extends Filterable<?>> extends AbstractAction<T> {
```
Pablete1234:
```suggestion
public class TeamAliasAction extends AbstractAction<Party> {
```
If this is the team scoped one, the scope should just be party
This change should greatly simplify the rest of the impl of this class as your methods are not just generic T, but party directly.

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,45 @@
...
+    this.alias = alias;
+  }
+
+  @Override
+  public void trigger(T scope) {
+    if (scope instanceof PartyImpl) {
+      ((PartyImpl) scope).getParty().setName(this.alias);
+    }
+  }
+  ;
```
Pablete1234:
```suggestion
```
remove leftover redundant `;` s

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,45 @@
...
+import tc.oc.pgm.match.MatchImpl;
+import tc.oc.pgm.match.PartyImpl;
+import tc.oc.pgm.teams.Team;
+import tc.oc.pgm.teams.TeamFactory;
+
+public class TeamAliasAction<T extends Filterable<?>> extends AbstractAction<T> {
+  protected final String alias;
+
+  public TeamAliasAction(Class<T> scope, String alias) {
+    super(scope);
```
Pablete1234:
```suggestion
    super(Party.class);
```

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,45 @@
...
+    this.alias = alias;
+  }
+
+  @Override
+  public void trigger(T scope) {
+    if (scope instanceof PartyImpl) {
+      ((PartyImpl) scope).getParty().setName(this.alias);
+    }
+  }
```
Pablete1234:
```suggestion
  public void trigger(Party party) {
    party.setName(this.alias);
  }
```

core/src/main/java/tc/oc/pgm/action/actions/TeamAliasAction.java
```diff
@@ -0,0 +1,45 @@
...
+    public void trigger(T scope) {
+      if (scope instanceof MatchImpl) {
+        ((MatchImpl) scope)
+            .getFeatureContext()
+            .get(team.get().getId(), Team.class)
+            .setName(this.alias);
+      }
+    }
+    ;
+  }
```
Pablete1234: There's no reason for this to exist, it can just be handled at construction time as a switch-scope action

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -414,6 +418,25 @@ public FillAction parseFill(Element el, Class<?> scope) throws InvalidXMLExcepti
...
+      var team = parser.reference(TeamFactory.class, el, "team").required();
+      return new TeamAliasAction.WithTeam<>(scope, alias, team);
+    } else if (scope.getSimpleName().equals("Party")) {
+      return new TeamAliasAction<>(scope, alias);
+    } else {
+      throw new InvalidXMLException(
+          "Wrong scope defined for action, scope must be " + Match.class.getSimpleName() + " or "
+              + Party.class.getSimpleName(),
+          el);
+    }
```
```suggestion
    var action = new TeamAliasAction(alias);    
    var teamBuilder = parser.reference(TeamFactory.class, el, "team");
    var team = scope == Party.class ? teamBuilder.orNull() : teamBuilder.required();
    
    return team == null ? action : 
            new SwitchScopeAction(scope, f -> f.moduleRequire(TeamMatchModule.class).getTeam(team.get()), null, action);
```

This should make this action effectively speaking work on all scopes.
- If its party scoped, team is optional (it can still be defined, and then it will work for that team)
- If its in any other scope, team is required, and that team will be used when running the action

Pablete1234 (changes requested):

Pablete1234 (dismissed):
Code LGTM now, much simpler and streamlined than before

You'll need to resolve the signoff etc tho

Javggg:
Ready, all commits signed

Pablete1234 (approved):

PR #1551, opened 2025-09-14
Implement `damager` filter

### LINKED ISSUE #1443: Implement `damager` filter
ProjectAres had a `damager` filter, which OCN used to filter damage from fireworks in its global include. This would be helpful ~~for doing the same for modern PGM, and~~ for other applications.

calcastor:
Would close https://github.com/PGMDev/PGM/issues/1443

I'm not sure if I have adapted this correctly from ProjectAres. This has been tested with denying damage from arrows:

```xml
<damage>
    <deny>
        <damager>
            <entity>arrow</entity>
        </damager>
    </deny>
</damage>
```

Something like this does not work:

```xml
<damage>
    <deny>
        <damager>
            <entity>player</entity>
        </damager>
    </deny>
</damage>
```

The intended use case of this feature is filtering projectiles for Touchdown maps. This permits Touchdown maps to have bows while also maintaining the ability to pass the flag to others via actions and a snowball consumable without awkward workarounds, by filtering for only damage from snowballs.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/matcher/damage/DamagerQueryModifier.java
+++ b/core/src/main/java/tc/oc/pgm/filters/matcher/damage/DamagerQueryModifier.java
@@ -10,9 +10,9 @@ import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.filters.modifier.QueryModifier;
 import tc.oc.pgm.tracker.info.EntityInfo;
 
-public class DamagerFilter extends QueryModifier<DamageQuery, EntityTypeQuery> {
+public class DamagerQueryModifier extends QueryModifier<DamageQuery, EntityTypeQuery> {
 
-  public DamagerFilter(Filter child) {
+  public DamagerQueryModifier(Filter child) {
     super(child, DamageQuery.class, EntityTypeQuery.class);
   }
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/tc/oc/pgm/filters/matcher/damage/DamagerQueryModifier.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public DamagerQueryModifier(Filter child) {
+    super(child, DamageQuery.class, EntityTypeQuery.class);
+  }
+
+  @Nullable
+  @Override
+  protected EntityTypeQuery transformQuery(DamageQuery query) {
+    var damager = query.getDamageInfo().getDamager();
+    if (damager instanceof EntityInfo) {
```
Pablete1234:
```suggestion
    if (damager instanceof EntityInfo entityInfo) {
```

core/src/main/java/tc/oc/pgm/filters/matcher/damage/DamagerQueryModifier.java
```diff
@@ -0,0 +1,43 @@
...
+      return new EntityTypeQuery() {
+        @Override
+        public Class<? extends Entity> getEntityType() {
+          return ((EntityInfo) damager).getEntityClass();
```
Pablete1234:
```suggestion
          return entityInfo.getEntityClass();
```

Pablete1234 (changes requested):
slightly better use of java 17 features, also needs an `@override` in `GenericDamageInfo` for the getDamager method that already existed in that class, but now overrides from parent

Pablete1234 (approved):

PR #1553, opened 2025-09-15
Implement `gliding` filter

### LINKED ISSUE #1503: Implement `gliding` filter for modern

calcastor:
Closes https://github.com/PGMDev/PGM/issues/1503

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/FilterModule.java
+++ b/core/src/main/java/tc/oc/pgm/filters/FilterModule.java
@@ -18,6 +18,7 @@ import tc.oc.pgm.filters.matcher.block.VoidFilter;
 import tc.oc.pgm.filters.matcher.match.MatchPhaseFilter;
 import tc.oc.pgm.filters.matcher.player.CanFlyFilter;
 import tc.oc.pgm.filters.matcher.player.FlyingFilter;
+import tc.oc.pgm.filters.matcher.player.GlidingFilter;
 import tc.oc.pgm.filters.matcher.player.GroundedFilter;
 import tc.oc.pgm.filters.matcher.player.ParticipatingFilter;
 import tc.oc.pgm.filters.matcher.player.PlayerMovementFilter;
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/filters/matcher/player/GlidingFilter.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import org.bukkit.event.Event;
+import tc.oc.pgm.api.filter.query.PlayerQuery;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.util.event.PlayerCoarseMoveEvent;
+
+public class GlidingFilter extends ParticipantFilter {
+  public static final GlidingFilter INSTANCE = new GlidingFilter();
```
Pablete1234:
```suggestion
  public static final GlidingFilter INSTANCE = Platform.isModern() ? new GlidingFilter() : StaticFilter.DENY;
```
Probably makes sense to just optimize it away to a deny for legacy

core/src/main/java/tc/oc/pgm/filters/FilterModule.java
```diff
@@ -95,6 +96,7 @@ public FilterModule parse(MapFactory factory, Logger logger, Document doc)
         features.addFeature(null, "sprinting", PlayerMovementFilter.SPRINTING);
         features.addFeature(null, "grounded", GroundedFilter.INSTANCE);
         features.addFeature(null, "flying", FlyingFilter.INSTANCE);
+        features.addFeature(null, "gliding", GlidingFilter.INSTANCE);
```
Pablete1234: cannot be added here without a proto bump, otherwise any map who was using this as an id for a map feature will break

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #1555, opened 2025-09-25
Show proper flag name in inventory on modern

Javggg:
This fixes the name shown for the flag item in the inventory.
From this:
Captura de pantalla 2025-09-25 a la(s) 12 51 42 p m
To this:
Captura de pantalla 2025-09-25 a la(s) 12 52 06 p m

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/state/Carried.java
+++ b/core/src/main/java/tc/oc/pgm/flag/state/Carried.java
@@ -25,7 +25,6 @@ import org.bukkit.event.inventory.InventoryClickEvent;
 import org.bukkit.event.player.PlayerDropItemEvent;
 import org.bukkit.event.player.PlayerMoveEvent;
 import org.bukkit.inventory.ItemStack;
-import org.bukkit.inventory.meta.ItemMeta;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.filter.query.Query;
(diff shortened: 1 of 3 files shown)


TTtie (changes requested):
I think a more clean fix for this would be to render the translations in `ModernColorUtils` (see below) instead (this is done in 1.8, but not in modern)

Edit: appended blocks showing inconsistency between modern and 1.8:

https://github.com/PGMDev/PGM/blob/0e8279cd61b10efa19733e64c0b7ac1389b01dad/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/material/ModernColorUtils.java#L120-L123

https://github.com/PGMDev/PGM/blob/0e8279cd61b10efa19733e64c0b7ac1389b01dad/platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpColorUtils.java#L91-L94

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
```diff
@@ -110,8 +111,14 @@ public void enterState() {
     kit = this.flag.getDefinition().getCarryKit();
     if (kit != null) carrier.applyKit(kit, false);
 
+    // Add name to the flag item for modern
+    ItemStack bannerItem = this.flag.getBannerItem();
+    ItemMeta itemMeta = bannerItem.getItemMeta();
+    itemMeta.setDisplayName(this.flag.getColoredName());
+    bannerItem.setItemMeta(itemMeta);
```
Pablete1234: This should be elsewhere, the `flag.getBannerItem()` should already be getting a renamed flag, as that's done once instead of every time

go look at the constructor for Flag and figure out why its a problem there

TTtie (changes requested):
You're still trying to fix what isn't broken. The flag code is, in fact, completely a-OK and needs no changes. 

The issue here is that on modern, translatable components are set directly and passed over onto the client without any other processing; they must be server-side rendered before they are sent over. On 1.8, this is done already – as setting chat components as an item name isn't possible there. See [previous review](https://github.com/PGMDev/PGM/pull/1555#pullrequestreview-3268995268).

Javggg:
Sorry, that made a lot of sense and I don't know why I didn't check it earlier when you pointed it out.
Now it works and the code is where it belongs.

TTtie (approved):

Pablete1234 (approved):

PR #1556, opened 2025-09-26
Support item matcher in itemmods

Pablete1234:
Allows for item-mods to have a `matcher` subelement for `match` which is an item matcher definition, such as those found in holding/carrying filters or replace-item/enchant-item actions.

eg:

```xml
<item-mods>
    <rule>
        <match ignore-metadata="true">
            <matcher material="bow" amount="1"/>
        </match>
        <modify show-enchantments="false" unbreakable="true">
            <enchantment level="2">infinity</enchantment>
        </modify>
    </rule>
</item-mods>
```

Do note: item-mods still apply JUST ONCE whenever the item is first observed (eg: when a chest is opened, or when the kit is created), so things like trying to modify the item when it gets to a certain durability, or if it gets enchanted with something after the fact is not and will not be supported.

This is more of a prototype because it was requested by a mapdev, but i'm not convinced this actually solves their intended use-case which was modifying punch bows in chests, as itemmods can only add not remove stuff. Also not convinced on the xml name for it, but we can't really use `material` or `item` as either of those are already interpreted as being materials for the material matcher.


eliminate123:
Good work pablete. this is a certified hood classic 👍

PR #1561, opened 2025-10-26
Disable leaving after a match ends

slayersource:
Certain players like to leave matches after they end but before Stats pops up so they have a Obs name tag when it shows up.

This PR disables leaving after a match ends.


Pablete1234 (commented):
You'll need to run the formater for the build to pass

Pablete1234 (approved):

PR #1565, opened 2025-11-15
Pull and display player component based on what team that player spent the most time on

slayersource:
In the stats roll up, show player's team color based on what team they spent the most time on

In a long CTW defense match, there are instances where a defender disconnects cause people have to touch grass. But that defender might have had the top kill streak or top kills. When the match ends their name shows as disconnected. It would be nice to see their team color so we know where their contributions helped most.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -381,11 +381,12 @@ public class StatsMatchModule implements MatchModule, Listener {
   }
 
   private Component getPlayerComponent(UUID uuid) {
+    var mp = match.getPlayer(uuid);
     var player = player(uuid, NameStyle.VERBOSE);
 
-    if (player != PlayerComponent.UNKNOWN
-        && match.getPlayer(uuid) != null
-        && !match.getPlayer(uuid).isObserving()) return player;
+    if (mp != null && mp.getCompetitor() != null && player != PlayerComponent.UNKNOWN) {
+      return player;
+    }
     return stats.column(uuid).values().stream()
         .max(Comparator.comparing(PlayerStats::getTimePlayed))
         .map(PlayerStats::getPlayerComponent)
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -382,15 +382,17 @@ public class StatsMatchModule implements MatchModule, Listener {
 
   private Component getPlayerComponent(UUID uuid) {
     var mp = match.getPlayer(uuid);
-    var player = player(uuid, NameStyle.VERBOSE);
-    if (mp != null && mp.getCompetitor() != null && player != PlayerComponent.UNKNOWN) {
-      return player;
+    Component player = null;
+
+    if (mp != null && mp.getCompetitor() != null) {
+      player = player(uuid, NameStyle.VERBOSE);
+      if (player != PlayerComponent.UNKNOWN) return player;
     }
 
     return stats.column(uuid).values().stream()
         .max(Comparator.comparing(PlayerStats::getTimePlayed))
         .map(PlayerStats::getPlayerComponent)
-        .orElse(player);
+        .orElse(player != null ? player : player(uuid, NameStyle.VERBOSE));
   }
 
   /** Formats raw damage to damage relative to the amount of hearths the player would have broken */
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -382,7 +382,10 @@ private Component credit(Set<UUID> players) {
 
   private Component getPlayerComponent(UUID uuid) {
     var player = player(uuid, NameStyle.VERBOSE);
-    if (player != PlayerComponent.UNKNOWN) return player;
+
+    if (player != PlayerComponent.UNKNOWN
+        && match.getPlayer(uuid).getBukkit().isOnline()
+        && !match.getPlayer(uuid).isObserving()) return player;
```
slayersource: I'm not exactly sure if this is the best way to check if a player is online or not due to my lack of familiarity with anything minecraft code related

slayersource:
@Pablete1234 Since we discussed on discord I can't re-request review in GitHub so link if there's anything else I should do for the PR

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -382,7 +382,10 @@ private Component credit(Set<UUID> players) {
 
   private Component getPlayerComponent(UUID uuid) {
     var player = player(uuid, NameStyle.VERBOSE);
-    if (player != PlayerComponent.UNKNOWN) return player;
+
+    if (player != PlayerComponent.UNKNOWN
+        && match.getPlayer(uuid) != null
+        && !match.getPlayer(uuid).isObserving()) return player;
```
Pablete1234: avoid the duplicate lookup on match.getPlayer, you can also optimize away the creating player component if we know they're offline or in obs.


Additionally `isObserving` will return `true` for players before and after match ends, because even if they're on blue/red team, they're effectively speaking observating, not participating (they're in creative mode, flying around, not able to interact with the map, etc).

You can see `MatchPlayer#isObserving` gets their party and calls if `Party#isObserving`, which for `Team` is  implemented as `!match.isRunning()`
 
The more correct alternative is using either `mp.getParty() instanceof Competitor` or `mp.getCompetitor() != null`, which leaves the change as:

```java
    var mp = match.getPlayer(uuid);
    if (mp != null && mp.getCompetitor() != null) {
      var player = player(uuid, NameStyle.VERBOSE);
      if (player != PlayerComponent.UNKNOWN) return player;
    }
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -381,8 +381,12 @@ private Component credit(Set<UUID> players) {
   }
 
   private Component getPlayerComponent(UUID uuid) {
+    var mp = match.getPlayer(uuid);
     var player = player(uuid, NameStyle.VERBOSE);
-    if (player != PlayerComponent.UNKNOWN) return player;
+    if (mp != null && mp.getCompetitor() != null && player != PlayerComponent.UNKNOWN) {
+      return player;
+    }
+
```
Pablete1234, line 385: the player component creation can be skipped often and isn't too cheap of a call, you should put it inside the if so it's only considered for online, non-obs players
```java
    var mp = match.getPlayer(uuid);
    if (mp != null && mp.getCompetitor() != null) {
      var player = player(uuid, NameStyle.VERBOSE);
      if (player != PlayerComponent.UNKNOWN) return player;
    }
```

slayersource: We need the player for Line 393 tho
```
return stats.column(uuid).values().stream()
        .max(Comparator.comparing(PlayerStats::getTimePlayed))
        .map(PlayerStats::getPlayerComponent)
        .orElse(player);
```

Unless I should do `match.getPlayer(uuid)` where the `.orElse(...)` is

Pablete1234: You can store it as a variable and still only compute it if needed:
```java
    var mp = match.getPlayer(uuid);
    Component player;
    if (mp != null && mp.getCompetitor() != null) {
      player = player(uuid, NameStyle.VERBOSE);
      if (player != PlayerComponent.UNKNOWN) return player;
    }

    // [...]
       .orElseGet(ps -> player != null ? player : player(uuid, NameStyle.VERBOSE));
```

Pablete1234 (approved):
Fixed it up and simplified it, as if player is online it is guaranteed the component won't be unknown

PR #1566, opened 2025-11-16
feat: allow quoting inline filters using single quotes

TTtie:
This PR allows using single quotes to quote a filter during inline filter parsing:
```xml
<filter id="all('variable = value', ...)" />
<filter id='all(&apos;variable = value&apos;, ...)' />
```

in addition to the currently supported:
```xml
<filter id="all(&quot;variable = value&quot;, ...)" />
<filter id="all(variable\ =\ value, ...)" />
<filter id='all("variable = value", ...)' /> <!-- legal, but very few maps use single quotes for XML attribute values -->
```

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/parser/StringReader.java
+++ b/util/src/main/java/tc/oc/pgm/util/parser/StringReader.java
@@ -22,8 +22,8 @@ public class StringReader {
    * @return The next word read in full.
    */
   public String readText() {
-    int quoteChar = pollFor('"', '\'');
-    boolean quoted = quoteChar != -1;
+    char quoteChar = pollQuote();
+    boolean quoted = quoteChar != 0;
 
     for (int i = nextIdx; i < string.length(); i++) {
       char ch = string.charAt(i);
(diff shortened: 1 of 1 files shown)


util/src/main/java/tc/oc/pgm/util/parser/StringReader.java
```diff
@@ -89,6 +91,27 @@ public boolean pollIf(char ch) {
...
   }
 
+  /**
+   * Peeks at the next char ignoring whitespace, if it's one of {@param chs}, polls it and returns
+   * the matching character
+   *
+   * @param chs The characters to expect next
+   * @return one of {@param chs} if it is the next char, and it was skipped. -1 otherwise.
+   */
+  public int pollFor(char... chs) {
```
Pablete1234: atp i'd rather have a pollQuote that is hardcoded to poll for " or ' instead of an arbitrary array of chars, and returns a char instead of an int, which may be `"`, `'` or a NULL char (ie: char 0) if none found

Pablete1234 (approved):

PR #1567, opened 2025-11-17
Add normalization to mutable cuboid region

Javggg:
While working on the XML for a map, I used a `cuboid` region variable that moves to random positions and applies a `<fill/>` action. This worked fine.

However, when applying the same logic in another map, it stopped working entirely. The region appeared to be set correctly when using `/region`, but the automatic glass fill that occurs when running that command also didn’t work. This indicated that the issue was not with `<fill/>` itself, the region behaved as if it contained zero blocks.

After inspecting the code and running several tests, I found that `Region.getBlockVectorIterator()` was returning an empty iterator. When comparing the dynamically assigned coordinates to the working map, it seemed that normalizing would make it work. The underlying issue seems to be that the mutable cuboid can end up with inverted bounds (max < min). Neither the iterator nor `contains()` operate correctly this way.

By normalizing the bounds before iterating and before checking containment, the region once again produces valid block positions and behaves correctly.

To avoid altering behavior of existing static regions, the normalization is applied only inside `CuboidRegion.Mutable`, by overriding `contains()` and `getBlockVectorIterator()`. Based on my testing, this fully resolves the issue without impacting other region types.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/regions/Bounds.java
+++ b/core/src/main/java/tc/oc/pgm/regions/Bounds.java
@@ -42,6 +42,21 @@ public class Bounds implements Cloneable {
     return new Bounds(this);
   }
 
+  public Bounds normalizedClone() {
+    Vector min = this.getMin();
+    Vector max = this.getMax();
+
+    double minX = Math.min(min.getX(), max.getX());
+    double minY = Math.min(min.getY(), max.getY());
+    double minZ = Math.min(min.getZ(), max.getZ());
+
+    double maxX = Math.max(min.getX(), max.getX());
+    double maxY = Math.max(min.getY(), max.getY());
+    double maxZ = Math.max(min.getZ(), max.getZ());
+
+    return new Bounds(new Vector(minX, minY, minZ), new Vector(maxX, maxY, maxZ));
+  }
+
   public static Bounds unbounded() {
     return new Bounds(
         new Vector(Double.NEGATIVE_INFINITY, Double.NEGATIVE_INFINITY, Double.NEGATIVE_INFINITY),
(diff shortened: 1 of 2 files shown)


Pablete1234:
Works as intended, a region with minX=10 and maxX=5 is simply an empty region (technically negative space?). If your actions create a region different to what you wanted, change your set actions so that they do this logic, but pgm should continue to behave like it does for this

Javggg:
I get that inverted bounds are considered an empty region, but the issue here is that this behavior is inconsistent with how other regions work. In XML you can define corners in any order and they are interpreted normally, but CuboidRegion.Mutable is the only one that silently behaves differently.

If the intended behavior is that mutable cuboids must always be strictly min→max, then there should at least be some validation or warning. Right now the region just acts as empty with no feedback at all, which makes it look like a bug when two nearly identical region definitions behave differently.

I’m not arguing for changing the model, only pointing out that the current silent failure makes debugging extremely confusing. A check or warning would already solve the problem.

Pablete1234:
I see what you mean, and yes xml defined regions that are incorrect get fixed at parse time, variable regions are ketting you "peek in" and access the internal model, and create empty regions that you otherwise wouldn't get, and some way of warning would be good but i dont see what that would be without making it look like an error when it could be intended behavior by the map author

Javggg:
One possible middle ground could be to make this behavior explicit and opt-in.
For example, adding an optional attribute such as:

```xml
<cuboid allow-inverted="false" />
```

PGM then would automatically reorder min/max, matching what authors naturally expect based on how static regions behave.
If a map author intentionally wants the current behavior (an empty or inverted cuboid), they can allow it explicitly.

The behavior difference can be confusing for map authors, especially when both XML styles appear symmetrical.

Pablete1234:
for backwards compat reasons it'd have to be opt-in, not opt-out, at which point it kinda loses its point

PR #1569, opened 2025-12-03
Cleanup some flag and match messages

calcastor:
1. Only drop flag if it's removed as helmet if the current state is `Carried`; when we don't do this you can see that the flag is "dropped" twice in KotF: once when the match ends and then again after that, which is noticeable as duplicate messages.
2. Don't bother sending the "Flag will respawn in n seconds" message if the match is over, and also don't bother showing the flag particle beam either, if the flag is in the `Respawning` state when the match is over.
3. Don't send the flag pickup message to console twice; `ChatManager.broadcastMessage` and friends already do this.
4. Send match winner message to console.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/state/Carried.java
+++ b/core/src/main/java/tc/oc/pgm/flag/state/Carried.java
@@ -74,7 +74,7 @@ public class Carried extends Spawned implements Missing {
 
   @Override
   public boolean isRecoverable() {
-    return true;
+    return this.flag.getMatch().isRunning();
   }
 
   @Override
--- a/core/src/main/java/tc/oc/pgm/flag/state/Respawning.java
+++ b/core/src/main/java/tc/oc/pgm/flag/state/Respawning.java
@@ -10,6 +10,7 @@ import net.kyori.adventure.text.TranslatableComponent;
 import net.kyori.adventure.text.format.NamedTextColor;
 import net.kyori.adventure.text.format.TextColor;
 import org.bukkit.Location;
+import org.bukkit.entity.Player;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.party.Party;
 import tc.oc.pgm.flag.Flag;
@@ -54,7 +55,8 @@ public class Respawning extends Spawned implements Returning {
   public void enterState() {
     super.enterState();
 
-    if (Duration.ZERO.equals(respawnTime)) return;
+    if (Duration.ZERO.equals(respawnTime) || !this.flag.getMatch().isRunning()) return;
+
     // Respawn is delayed
     String postName = this.post.getPostName();
 
@@ -82,6 +84,11 @@ public class Respawning extends Spawned implements Returning {
     }
 
     this.flag.transition(new Returned(this.flag, this.post, this.respawnTo));
+  }
+
+  @Override
+  protected boolean canSeeParticles(Player player) {
+    return this.flag.getMatch().isRunning();
   }
 
   @Override
--- a/core/src/main/java/tc/oc/pgm/goals/TouchableGoal.java
+++ b/core/src/main/java/tc/oc/pgm/goals/TouchableGoal.java
@@ -24,7 +24,6 @@ import tc.oc.pgm.channels.ChatManager;
 import tc.oc.pgm.goals.events.GoalCompleteEvent;
 import tc.oc.pgm.goals.events.GoalTouchEvent;
 import tc.oc.pgm.spawns.events.ParticipantDespawnEvent;
-import tc.oc.pgm.util.Audience;
 
 /**
  * A {@link Goal} that may be 'touched' by players, meaning the player has made some tangible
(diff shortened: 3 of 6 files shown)


core/src/main/java/tc/oc/pgm/flag/state/Carried.java
```diff
@@ -213,7 +213,7 @@ public void tickRunning() {
     }
 
     if (!isFlag(carrier.getInventory().getHelmet())) {
-      this.dropFlag();
+      if (isCurrent()) this.dropFlag();
     }
```
Pablete1234, line 217: i find it somewhat concerning that `tickRunning` as a whole is called if its not the current state, how is this happening?

core/src/main/java/tc/oc/pgm/flag/state/Respawning.java
```diff
@@ -54,7 +56,9 @@ protected Duration getDuration() {
   public void enterState() {
     super.enterState();
 
-    if (Duration.ZERO.equals(respawnTime)) return;
+    if (Duration.ZERO.equals(respawnTime) || this.flag.getMatch().getPhase() == MatchPhase.FINISHED)
```
Pablete1234:
```suggestion
    if (Duration.ZERO.equals(respawnTime) || this.flag.getMatch().isFinished())
```
Also, should it be `!match.isRunning()`  instead? i guess it can't happen before the match so its prob not relevant

core/src/main/java/tc/oc/pgm/flag/state/Respawning.java
```diff
@@ -84,6 +88,11 @@ protected void respawn(Component message) {
     this.flag.transition(new Returned(this.flag, this.post, this.respawnTo));
   }
 
+  @Override
+  protected boolean canSeeParticles(Player player) {
+    return this.flag.getMatch().getPhase() != MatchPhase.FINISHED;
```
Pablete1234:
```suggestion
    return !this.flag.getMatch().isFinished();
```

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
calcastor: I'm not sure but this isn't the only place we guard `dropFlag` like this; we also do so in `onEvent (InventoryClickEvent event)` in `Carried` but with an additional guard of delegating it to a `ScheduledExecutorService` only for `MatchPhase.RUNNING`.

Looking at usages of `tickRunning`, it's called only if the match is running in `BaseState` and the rest of the usages are just `super`s in classes extending `BaseState`.

calcastor:
Newest commit at least works around https://github.com/PGMDev/PGM/issues/1580, but I don't know if it's the root issue

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
Pablete1234: tickRunning is called by `run()` in BaseState, which is a ticking task started on `enterState` and ended on `leaveState`, meaning that if tickRunning is called, it's because the state must be the current state.
This check is redundant

Pablete1234:
Get rid of the workaround and the comment, i've investigated and have a fix for the root cause, but it's quite extensive and merits its own pr.

core/src/main/java/tc/oc/pgm/goals/TouchableGoal.java
```diff
@@ -189,19 +188,18 @@ public boolean shouldShowTouched(@Nullable Competitor team, Party viewer) {
...
+      Component message = getTouchMessage(toucher, false);
+
+      if (shouldShowTouched(toucher.getParty())) {
+        if (showEnemyTouches()) {
+          ChatManager.broadcastMessage(message, viewer -> !toucher.isPlayer(viewer));
+        } else {
+          ChatManager.broadcastPartyMessage(
+              message, toucher.getParty(), viewer -> !toucher.isPlayer(viewer));
+        }
+      }
```
Pablete1234: Does this mean if the toucher is null no message at all will show? that does not sound working as intended

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
```diff
@@ -142,15 +142,17 @@ public void leaveState() {
     SidebarMatchModule smm = this.flag.getMatch().getModule(SidebarMatchModule.class);
     if (smm != null) smm.stopBlinkingGoal(this.flag);
 
-    if (this.flag.hasShowOption(ShowOption.SHOW_MESSAGES)) this.carrier.sendActionBar(empty());
+    if (this.carrier.isParticipating()) {
```
Pablete1234: get rid of this, i'll fix it in a different pr

core/src/main/java/tc/oc/pgm/goals/TouchableGoal.java
calcastor: Need to reevaluate this one later, honestly don't know why I did this change or if I meant to commit it.

core/src/main/java/tc/oc/pgm/flag/state/Carried.java
calcastor: Replaced by ensuring flags only can be dropped if `state` is not already `Dropped`, and also by making flags only recoverable if the match is running.

Pablete1234 (approved):

PR #1570, opened 2025-12-05
Generate more fitting biomes for different dimensions in modern

zzufx:
In modern, maps set in the Nether or The End generate empty chunks with the Plains biome. This means you need to leave more chunks than you normally would when pruning a map to prevent very noticeable sky color changes (specifically in the Nether)

p_hell
p_plains

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
+++ b/util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
@@ -38,10 +38,18 @@ public class NullChunkGenerator extends ChunkGenerator {
 
     ChunkData chunkData = super.createChunkData(world);
 
-    // For everyblock in the chunk set the biome to plains
+    Biome newBiome;
+
+    switch (world.getEnvironment()) {
+      case NETHER -> newBiome = Biome.valueOf("NETHER_WASTES");
+      case THE_END -> newBiome = Biome.valueOf("THE_END");
+      default -> newBiome = Biome.PLAINS;
+    }
+
+    // For every block in the chunk set the biome
     for (int x = 0; x < 16; x++) {
       for (int z = 0; z < 16; z++) {
-        biome.setBiome(x, z, Biome.PLAINS);
+        biome.setBiome(x, z, newBiome);
       }
     }
     return chunkData;
--- a/util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
+++ b/util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
@@ -31,25 +31,24 @@ public class NullChunkGenerator extends ChunkGenerator {
 
   @Override
   public ChunkData generateChunkData(
-      World world, Random random, int chunkX, int chunkZ, BiomeGrid biome) {
+      World world, Random random, int chunkX, int chunkZ, BiomeGrid biomeGrid) {
     if (SKIP_BIOMES) {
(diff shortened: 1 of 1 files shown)


util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
```diff
@@ -38,12 +38,17 @@ public ChunkData generateChunkData(
...
 
-    // For everyblock in the chunk set the biome to plains
+    // For every block in the chunk set the biome
     for (int x = 0; x < 16; x++) {
       for (int z = 0; z < 16; z++) {
-        biome.setBiome(x, z, Biome.PLAINS);
+        switch (world.getEnvironment()) {
+          case NETHER -> biome.setBiome(x, z, Biome.valueOf("NETHER_WASTES"));
+          case THE_END -> biome.setBiome(x, z, Biome.valueOf("THE_END"));
+          default -> biome.setBiome(x, z, Biome.PLAINS);
```
Pablete1234: The biome to use should be computed outside the loop to avoid it being looked up 256 times per generated chunk, as well as the unnecessary branching inside the loop

Pablete1234 (changes requested):

util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
```diff
@@ -38,10 +38,18 @@ public ChunkData generateChunkData(
...
     ChunkData chunkData = super.createChunkData(world);
 
-    // For everyblock in the chunk set the biome to plains
+    Biome newBiome;
+
+    switch (world.getEnvironment()) {
+      case NETHER -> newBiome = Biome.valueOf("NETHER_WASTES");
+      case THE_END -> newBiome = Biome.valueOf("THE_END");
+      default -> newBiome = Biome.PLAINS;
+    }
```
Pablete1234:
```suggestion
    Biome biome = switch (world.getEnvironment()) {
      case NETHER -> Biome.valueOf("NETHER_WASTES");
      case THE_END -> Biome.valueOf("THE_END");
      default -> Biome.PLAINS;
    };
```

util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
Pablete1234:
```suggestion
      World world, Random random, int chunkX, int chunkZ, BiomeGrid biomeGrid) {
```

Pablete1234:
```suggestion
      return super.generateChunkData(world, random, chunkX, chunkZ, biomeGrid);
```

util/src/main/java/tc/oc/pgm/util/chunk/NullChunkGenerator.java
```diff
@@ -38,10 +38,18 @@ public ChunkData generateChunkData(
...
+
+    // For every block in the chunk set the biome
     for (int x = 0; x < 16; x++) {
       for (int z = 0; z < 16; z++) {
-        biome.setBiome(x, z, Biome.PLAINS);
+        biome.setBiome(x, z, newBiome);
```
Pablete1234:
```suggestion
        biomeGrid.setBiome(x, z, biome);
```

Pablete1234 (changes requested):

Pablete1234 (approved):

cswhite2000 (approved):

PR #1573, opened 2025-12-06
Add `riptiding` filter

zzufx:
Adds a riptiding filter based on the gliding filter.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Modules.java
+++ b/core/src/main/java/tc/oc/pgm/api/Modules.java
@@ -135,6 +135,7 @@ import tc.oc.pgm.tnt.TNTMatchModule;
 import tc.oc.pgm.tnt.TNTModule;
 import tc.oc.pgm.tntrender.TNTRenderMatchModule;
 import tc.oc.pgm.tracker.TrackerMatchModule;
+import tc.oc.pgm.util.platform.Platform;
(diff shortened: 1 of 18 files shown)


core/src/main/java/tc/oc/pgm/filters/matcher/player/RiptidingFilter.java
```diff
@@ -0,0 +1,28 @@
...
+import tc.oc.pgm.util.event.PlayerCoarseMoveEvent;
+import tc.oc.pgm.util.platform.Platform;
+
+public class RiptidingFilter extends ParticipantFilter {
+  public static final FilterDefinition INSTANCE =
+      Platform.isModern() ? new RiptidingFilter() : StaticFilter.DENY;
+
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    return Collections.singleton(PlayerCoarseMoveEvent.class);
```
Pablete1234, line 17: This should probably be using the riptide event that exists in modern, but it's annoying to get it here.

Maybe modern filter impls need to be re-thought so that they reside more in the modern side, but that's out of scope for this pr.

cswhite2000 (approved):

Pablete1234:
Gliding filter has been moved to platform modules which opens up for this pr being cleaned up after that approach

PR #1575, opened 2025-12-10
Add `riptiding` filter

zzufx:
Adds a riptiding filter based on the gliding filter.
Was having trouble with the old PR so opening this one

### DIFF
--- a/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/filters/RiptidingFilter.java
+++ b/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/filters/RiptidingFilter.java
@@ -1,0 +1,24 @@
+package tc.oc.pgm.platform.modern.filters;
+
+import java.util.Collection;
+import java.util.Collections;
+import org.bukkit.event.Event;
+import org.bukkit.event.player.PlayerRiptideEvent;
+import tc.oc.pgm.api.filter.FilterDefinition;
+import tc.oc.pgm.api.filter.query.PlayerQuery;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.filters.matcher.player.ParticipantFilter;
+
+public class RiptidingFilter extends ParticipantFilter {
+  public static final FilterDefinition INSTANCE = new RiptidingFilter();
+
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    return Collections.singleton(PlayerRiptideEvent.class);
+  }
+
+  @Override
+  protected boolean matches(PlayerQuery query, MatchPlayer player) {
+    return player.getBukkit().isRiptiding();
+  }
+}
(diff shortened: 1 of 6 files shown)


cswhite2000 (approved):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/filters/RiptidingFilter.java
```diff
@@ -0,0 +1,24 @@
...
+import tc.oc.pgm.api.filter.query.PlayerQuery;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.filters.matcher.player.ParticipantFilter;
+
+public class RiptidingFilter extends ParticipantFilter {
+  public static final FilterDefinition INSTANCE = new RiptidingFilter();
+
+  @Override
+  public Collection<Class<? extends Event>> getRelevantEvents() {
+    return Collections.singleton(PlayerRiptideEvent.class);
```
Pablete1234, line 17: This will make the filter dynamically ALLOW when you start riptiding, but nothing will make it check again when done riptiding, so it will be unreliable

cswhite2000: That is a concern.

And listening to a coarse move might not stop it either if you say crash into a corner. Although it would probably fix most of the cases

zzufx: I'm not sure if this is something else, but using a simple map with
```xml
<actions>
    <trigger filter="riptiding" scope="player">
        <action scope="player">
            <message text="`aYou're riptiding!"/>
        </action>
    </trigger>
    <trigger filter="not(riptiding)" scope="player">
        <action scope="player">
            <message text="`cYou're NOT riptiding!"/>
        </action>
    </trigger>
    <trigger filter="gliding" scope="player">
        <action scope="player">
            <message text="`aYou're gliding!"/>
        </action>
    </trigger>
    <trigger filter="not(gliding)" scope="player">
        <action scope="player">
            <message text="`cYou're NOT gliding!"/>
        </action>
    </trigger>
</actions>
```
has always worked well - when you hit a block, you can still riptide for a little bit depending on the angle, and the messages pop up at the appropriate time

Same with /pgm:filter `riptiding`/`gliding` - sometimes the messages are a bit delayed when respawning, but the command always gives the correct value

Pablete1234: The `/pgm:filter` command will always be correct, because that checks the filter at that instant. The issue is with dynamic filters (eg: triggers).

Having a map with both riptiding and gliding (which reacts to coarse move event) will hide most of the issue, but if you make a map with JUST riptiding, you'll notice it never triggers `not(riptiding)`.
Additionally the gliding filter has a similar issue, where if you just stand on one block it will never change dynamic state, only once crossing a block.

The only way to fix these issues properly is to either have a listener that will create some sort of RiptideEndEvent (listens for riptide start, creates a scheduled task to check if player isn't riptiding anymore, and once that happens sends the event and cancels the task), or turning the filter into something similar to MonostableFilter where it has a reactor that keeps track of players and checks their status until it ends

PR #1577, opened 2025-12-11
Update modern platform to 1.21.11

calcastor:
~~Not certain about what I have done for game rules here.~~


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
```diff
@@ -107,14 +110,21 @@ public void onPlayerTeleport(final PlayerTeleportEvent event) {
...
-        .runTask(
-            PGM.get(), () -> ((CraftPlayer) event.getPlayer()).getHandle().setClientLoaded(true));
+    // ServerGamePacketListenerImpl#waitingForRespawn is set after the event.
+    Bukkit.getScheduler().runTask(PGM.get(), () -> {
+      try {
+        WAITING_FOR_RESPAWN.set(((CraftPlayer) event.getEntity()).getHandle().connection, false);
+      } catch (IllegalAccessException e) {
+        throw new RuntimeException(e);
+      }
+    });
```
Pablete1234: There's a `markClientUnloadedAfterDeath` and a `restartClientLoadTimerAfterRespawn`, maybe just use that to mark the player as respawned right after dead instead of reflection

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/gamerules/ModernRemappedGameRule.java
```diff
@@ -0,0 +1,45 @@
...
+      } else {
+        return Result.err(new IllegalStateException(
+            "Don't know how to parse value for " + handle.getType().getName()));
+      }
+
+      try {
+        return Result.ok(handle.getType().cast(parsedValue));
+      } catch (ClassCastException e) {
+        return Result.err(e);
+      }
```
Pablete1234:
```suggestion
      return Result.ok(handle.getType().cast(parsedValue));
```
given it's already wrapped in a catch for throwable that has the exact same handling, seems these would be identical behaviors.

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/modules/gamerules/SpGameRules.java
```diff
@@ -0,0 +1,106 @@
...
+import net.minecraft.server.v1_8_R3.GameRules;
+import org.jetbrains.annotations.Unmodifiable;
+import org.jspecify.annotations.NullMarked;
+import tc.oc.pgm.util.bukkit.GameRule;
+import tc.oc.pgm.util.platform.Supports;
+
+@Supports(SPORTPAPER)
+@NullMarked
+public class SpGameRules implements tc.oc.pgm.util.bukkit.GameRules.GameRuleRegistry {
+  protected static final @Unmodifiable Map<String, String> DEFAULTS;
```
Pablete1234: Seems like DEFAULTS should just be a field populated in the GameRule (ie SpKVGameRule or whatever) object with whatever its default is

util/src/main/java/tc/oc/pgm/util/bukkit/GameRule.java
```diff
@@ -0,0 +1,59 @@
...
+   */
+  void set(World world, @Nullable T value);
+
+  /**
+   * Sets the value of the game rule in the given world from a string representation
+   *
+   * @param world The world to set the game rule in
+   * @param value The string representation of the value to set, or null to reset to default
+   */
+  void setFromString(World world, @Nullable String value);
```
Pablete1234: I think this method should go, strings for gamerules are a thing of the past, our api should just pretend they're always `T` and legacy can do whatever it needs to to turn them back into strings internally if needed

util/src/main/java/tc/oc/pgm/util/bukkit/GameRule.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Tries to parse the given string into a value through the implementation's parsing logic
+   *
+   * @param value The string value to parse
+   * @return A Result containing the parsed value or an error if parsing failed
+   */
+  Result<T, ?> tryParse(String value);
```
Pablete1234: I question the use of `Result<T, ?>`, it's very functional-based but in java the more traditional approach would simply be:

`T tryParse(String value) throws X` and make the caller try-catch for it, instead of expecting them to `switch` over the result.

I'm fine with either, just just leaving it as food for thought.

core/src/main/java/tc/oc/pgm/gamerules/GameRulesModule.java
```diff
@@ -30,33 +35,59 @@ public GameRulesMatchModule createMatchModule(Match match) {
...
           String value = gameRuleElement.getValue();
 
           if (value == null) {
-            throw new InvalidXMLException("Missing value for gamerule " + rule, gameRuleElement);
-          } else if (gameRules.containsKey(rule)) {
-            throw new InvalidXMLException(rule + " has already been specified", gameRuleElement);
-          }
+            throw new InvalidXMLException(
+                "Missing value for game rule " + ruleName, gameRuleElement);
+          } else {
```
Pablete1234: consider using an early return to avoid the whole method getting yet another indent from the `else` block:
```java
if (value == null)
  throw new InvalidXMLException("Missing value for game rule " + ruleName, gameRuleElement);

GameRule<?> rule = GameRules.getByName(ruleName);
// ...
```

Pablete1234 (changes requested):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
TTtie: None of the methods specified do the thing I want it to do.

- `markClientUnloadedAfterDeath` is the very method blocking interaction (see `ServerPlayer#die(DamageSource)`)
- `restartClientLoadTimerAfterRespawn` unmarks `ServerGamePacketListenerImpl#waitingForRespawn` but also adds the 60-tick client world load timeout (that will never happen as respawn screens are blocked), reintroducing #1593.

core/src/main/java/tc/oc/pgm/gamerules/GameRulesModule.java
```diff
@@ -78,12 +77,12 @@ public GameRulesModule parse(MapFactory factory, Logger logger, Document doc)
...
-            case Result.Err<?, ?>(Throwable err) ->
-              throw new InvalidXMLException(
-                  "Failed to parse game rule value for " + rule.name() + ": " + err.getMessage(),
-                  gameRuleElement);
+          try {
+            gameRules.put(rule, rule.tryParse(value));
+          } catch (Throwable e) {
+            throw new InvalidXMLException(
+                "Failed to parse game rule value for " + rule.name() + ": " + e.getMessage(),
+                gameRuleElement);
```
Pablete1234: You're missing passing `e` as the cause for the invalid xml exception
```suggestion
            throw new InvalidXMLException(
                "Failed to parse game rule value for " + rule.name() + ": " + e.getMessage(),
                gameRuleElement,
                e);
```

Pablete1234 (approved):

PR #1578, opened 2025-12-11
Improvements to `gliding` and `riptiding` filters

zzufx:
This PR should fix some discussed issues in both the `gliding` and `riptiding` filters by adding some custom events - I also thought about comparing Poses for riptiding instead, would have to store the old one.


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
```diff
@@ -56,6 +59,33 @@ public void onPlayerAttackEntity(PrePlayerAttackEntityEvent event) {
     handleCall(pgmEvent, event);
   }
 
+  @EventHandler(ignoreCancelled = true)
+  public void onGlideToggle(EntityToggleGlideEvent event) {
```
Pablete1234: why do we need to replicate bukkit's entity toggle glide event? should be able to use that directly

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
```diff
@@ -56,6 +59,33 @@ public void onPlayerAttackEntity(PrePlayerAttackEntityEvent event) {
...
+      PlayerGlideToggleEvent pgmEvent = new PlayerGlideToggleEvent(player, player.isGliding());
+      Bukkit.getScheduler().runTask(PGM.get(), () -> handleCall(pgmEvent, event));
+    }
+  }
+
+  @EventHandler(ignoreCancelled = true)
+  public void onRiptideStart(org.bukkit.event.player.PlayerRiptideEvent event) {
+    Player player = event.getPlayer();
+
+    Bukkit.getScheduler()
```
Pablete1234: pgm has its own executor under `PGM.get().getExecutor()` that allows scheduling and cancelling futures.
You can see an example in `MatchTabManager`'s `RenderTask`, where it schedules itself as a future in the constructor then the run method can cancel that future, or in `MapDevCommand`'s `DisplayRunner` which also creates the future and cancels it itself in the run method

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/util/event/player/PlayerGlideToggleEvent.java
```diff
@@ -0,0 +1,28 @@
+package tc.oc.pgm.platform.modern.util.event.player;
+
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.player.PlayerEvent;
+
+public class PlayerGlideToggleEvent extends PlayerEvent {
```
Pablete1234: not needed, just use bukkit's

Pablete1234 (changes requested):

zzufx:
Not really sure on how to use the player in EntityToggleGlideEvent event without creating my own PlayerToggleGlideEvent - tried a few things but nothing worked. Can just leave the PR at riptiding if necessary

Pablete1234:
I gave it a try and it worked fine for me, i believe your issue was probably that dynamic filters were not picking up the `Entity getEntity()` method of the event as one of the possible actors to filter on.

Ontop of that, i found that modern has an `EntityPoseChangeEvent` which triggers when your pose changes (riptiding and gliding are also poses that trigger this), and is able to replace all of these.

Pablete1234 (approved):

PR #1579, opened 2025-12-11
THE crawling filter

eliminate123:
It's over for walkcels. We are now entering the crawling revolution. 
Think you can bypass filters by crawling? Think again! This PR introduces THE `<crawling/>` filter
Ex:
```xml
<filters>
    <crawling id="the-crawler"/>
</filters>
<actions>
    <trigger filter="the-crawler">
        <action>
            <message text="`lTHE CRAWLER IS HERE!"/>
        </action>
    </trigger>
</actions>


Pablete1234:
Doesn't even build, and suffers from the same issue as gliding where if not moving this won't update

Pablete1234:
Feel free to reopen if you fix the filter

PR #1581, opened 2025-12-26
Voted pool minor improvements

Pablete1234:
Makes a few relatively minor fixes to voted pools:
 - The map scores/length will always persist under the base id of the map (the id of the default variant)
   - This fixes a situation where a variant may have played (staff set it,or it was sponsored), and then PGM gives another variant of the same map in the votebook.
 - Map selection for voted pools now takes into account the map that just played.
   - Before, pgm would weigh following options of the vote if they were of the same gamemode, so 2nd option was less likely to be a repeated gamemode of 1st option. Now this list of "previous" contains the currently played map too. So if a CTW just finished playing, the vote is less likely to contain CTW options. Same for rage, bridge, or any other gamemode.


cswhite2000 (approved):

eliminate123:
Christmas miracle

PR #1583, opened 2025-12-27
Allow using regions in `teleport` action

zzufx:
Allows specifying regions as destinations for `<teleport/>` actions. This is useful if you've already defined those regions for other things elsewhere in the code, or if you're not using formulas and just want to define a destination as a point with the usual `"x,y,z"` format instead of the unique `x="" y="" z=""` of the teleport action (either manually or when using constants).

```xml
<teleport region="first-checkpoint"/>
<teleport region="second-checkpoint" yaw="90" pitch="0"/>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -36,6 +36,7 @@ import tc.oc.pgm.action.actions.SoundAction;
 import tc.oc.pgm.action.actions.TakePaymentAction;
 import tc.oc.pgm.action.actions.TeamAliasAction;
 import tc.oc.pgm.action.actions.TeleportAction;
+import tc.oc.pgm.action.actions.TeleportRegionAction;
 import tc.oc.pgm.action.actions.VelocityAction;
 import tc.oc.pgm.action.actions.WeatherAction;
 import tc.oc.pgm.action.replacements.Replacement;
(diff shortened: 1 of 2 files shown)


cswhite2000 (commented):
I feel like these should be separate classes, or subclasses

Something like TeleportAction and TeleportRegionAction
or TeleportAction.Point and TeleportAction.Region

cswhite2000 (approved):

Pablete1234 (changes requested):
Before there was no good definition of what should happen if overlapping region and xyz were defined, which was a problem. That since has been fixed.

However, the way i conceived solving it was a bit different. IMO, it'd make sense that you're able to define a region, as well as an x/y/z (but not all 3), and pgm will pick a point in the region, then modify the x/y/z appropriately using the formula(s).

This would allow for a style where you're able to define a region + an offset in a very convenient way.

zzufx:
> Before there was no good definition of what should happen if overlapping region and xyz were defined, which was a problem. That since has been fixed.
> 
> However, the way i conceived solving it was a bit different. IMO, it'd make sense that you're able to define a region, as well as an x/y/z (but not all 3), and pgm will pick a point in the region, then modify the x/y/z appropriately using the formula(s).
> 
> This would allow for a style where you're able to define a region + an offset in a very convenient way.

That kind of sounds like something that could be added in a further PR - this PR's intent was just adding region support, without necessarily adding "extra" stuff. Being able to have all 3 + region could make sense too, right? Would be able to choose a starting region and modify x/y/z using things like team score and such.

Pablete1234:
I was considering absolute positions only, if relative need to work tho, that becomes complicated as you effectively need to perform two teleports, one for the region such that the player.x,y & z variables update, and then the formulas for x,y,z would be updated.
My idea was actually that a single teleport location would happen, and it'd be the combination of finding a location in the region, then applying the x/y/z formulas to override up to 2 of the locations, then perform the teleport just once to that final destination

zzufx:
I feel like absolute positions could make usage a bit confusing - either way I feel like that's something that could be added in a later PR for new functionality

Pablete1234:
Since push by contributors to your branch hasn't been enabled i'll supersede the pr instead

PR #1584, opened 2025-12-28
Extra variables for map vote modifiers

ThatApplePieGuy:
Adds `participants`, `observers`, and `match_duration` variables that can be used in the vote modifier config. These could be useful for scaling back map sizes based on predicted player churn after a long match


core/src/main/java/tc/oc/pgm/rotation/vote/MapVotePicker.java
```diff
@@ -157,18 +166,28 @@ private double getRepeatedGamemodes(List<MapInfo> selected, MapInfo map) {
...
       super(
           Map.of(
               "score", score,
               "same_gamemode", sameGamemode,
               "mapsize", mapsize,
-              "players", players),
+              "players", players,
+              "participants", participants,
+              "observers", observers,
+              "match_duration", matchDuration),
```
Pablete1234: `play_minutes` may be more appropriate as there's precedent (see VotingPool.Context), and it specifies the unit

Pablete1234 (approved):

PR #1585, opened 2025-12-29
Fix spawn state transitions delaying events

### LINKED ISSUE #1580: Flags do not drop on disconnect when drop-kit has a team-colored item
Flags that have a `drop-kit` that contains a team colored item will result in an exception when the flag carrier disconnects. The flag will not drop on the ground and soft-locks that objective.

Image

Image

```xml
<kits>
    <kit id="drop-kit">
        <chestplate material="leather chestplate" team-color="true"/>
    </kit>
</kits>
<flags name="Flag" drop-kit="drop-kit">
    <flag id="blue-flag" color="blue" owner="blue">
        <post yaw="-90" recover-time="20s" respawn-time="8s">-34,9,-8</post>
    </flag>
</flags>
```

Sample map: [flag_drop_kit.zip](https://github.com/user-attachments/files/24231870/flag_drop_kit.zip)

Console: https://gist.github.com/mitchts/e79a33924b2fe9b0afb5b05b2ad63d63

Pablete1234:
Taken from 1.9+ PGM, this fixes an issue in how state transitions are currently handled, and it's that due to a transition being able to cause another, all events were queued up for later instead of happening in place.
This cause that leaving the `Alive` state would queue a `ParticipantDespawnEvent`, that would happen *after* that participant had left the team, leading to consequences like a flag's kit being applied to someone who wasn't on a team anymore. Should fix #1580 .

Now state transitions are allowed to overlap, and they'll be simply queued up and processed one after another, allowing events to be thrown at the appropriate time.


cswhite2000 (approved):
No major issues, but the event naming is problematic

These are essentially the same name, but the name doesn't differentiate which is before the change and which is after

PlayerChangePartyEvent
PlayerPartyChangeEvent

any of these would be more clear:
PlayerPostPartyChangeEvent
PlayerPartyChangedEvent
PlayerChangedPartyEvent

Or if you want to differentiate the other way
PrePlayerChangePartyEvent
PlayerPreChangePartyEvent
etc.

Pablete1234:
Issue with event names is any name change will be breaking for 3rd party plugins relying on them, which i'm trying to avoid.
I've followed the naming that 1.9+ pgm used:
reference:
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/events/PlayerChangePartyEvent.java
https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/events/PlayerPartyChangeEventBase.java

cswhite2000:
Isn't `PlayerChangePartyEvent` new for this version of pgm? Shouldn't there be nothing relying on at least that one?

Pablete1234:
Yea one of the events is new, but it's not really a *pre* event it's a wrapping event, happens around the event, you can have the handler run code both before and after the side effects

cswhite2000:
That's a bit of a pain. Well I'll leave it up to you. Feel free to merge

Pablete1234:
I'll merge this as-is, and i'll work on more touch-ups in a later pr

PR #1589, opened 2026-01-03
Admin chat audit message improvements

cswhite2000:
Adds:
- admin chat audit messages for `/end` and `/join -f`
- map info click to set-next admin chat audit messages

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapOrderCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapOrderCommand.java
@@ -2,6 +2,8 @@ package tc.oc.pgm.command;
 
 import static net.kyori.adventure.text.Component.text;
 import static net.kyori.adventure.text.Component.translatable;
+import static net.kyori.adventure.text.event.ClickEvent.runCommand;
+import static net.kyori.adventure.text.event.HoverEvent.showText;
 import static tc.oc.pgm.api.Permissions.DEV;
 import static tc.oc.pgm.api.map.Phase.DEVELOPMENT;
 import static tc.oc.pgm.command.util.ParserConstants.CURRENT;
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/command/MapOrderCommand.java
```diff
@@ -86,7 +88,10 @@ public void setNext(
   }
 
   public static void sendSetNextMessage(@NotNull MapInfo map, CommandSender sender) {
-    Component mapName = text(map.getName(), NamedTextColor.GOLD);
+    Component mapName = text(map.getName(), NamedTextColor.GOLD)
+        .hoverEvent(showText(translatable(
+            "command.maps.hover", NamedTextColor.GRAY, map.getStyledName(MapNameStyle.COLOR))))
+        .clickEvent(runCommand("/map " + map.getName()));
```
Pablete1234, line 94: Should this be part of the maps' name component? ie, doesn't map.getName(FANCY) or similar give you a component with these things already?

cswhite2000: The current [`MapNameStyle`](https://github.com/PGMDev/PGM/blob/dev/util/src/main/java/tc/oc/pgm/util/named/MapNameStyle.java) doesn't have an option to do this

You can see an example where we do something similar [here](https://github.com/PGMDev/PGM/blob/dev/core/src/main/java/tc/oc/pgm/command/MapCommand.java#L114)

Overall, we definitely should have this be part of the map name component.

Mostly, I just don't like the current part where we essentially have random boolean flags in an enum for formatting, and I didn't want to add another such flag and more enum options. I'd like a builder more, but that also seemed somewhat overkill for this.
So I ended up doing the simple thing we do elsewhere.

Pablete1234: IMO the boolean flags grouped in an enum is a model that makes alot of sense, it just breaks down if you want every single different use of the component to have one specific combination of such boolean flags.

Ideally you should have a few styles that are then consistently reused across the codebase, but it seems half of the enum styles aren't even used and we apply custom stuff after the fact in a few places.
I guess that's something to fix up in the future.

Pablete1234 (approved):

PR #1591, opened 2026-01-04
Refactor MapNameStyle usage

cswhite2000:
Refactoring MapNameStyle usage

Based on the discussion from https://github.com/PGMDev/PGM/pull/1589

Also switched most usages of map names in chat to be clickable to run `/map [name name]`

I'm not super happy with the enum naming and such, but what I have here is workable

### DIFF
--- a/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
+++ b/core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
@@ -254,34 +254,29 @@ public class MapInfoImpl implements MapInfo {
 
     if (style.isColor) name.color(NamedTextColor.GOLD);
     if (style.isHighlight) name.decoration(TextDecoration.UNDERLINED, true);
+
+    Component component = name.build();
+
     if (style.showAuthors) {
-      Component component = translatable(
+      component = translatable(
           "misc.authorship",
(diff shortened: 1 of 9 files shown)


Pablete1234:
IMO there is no need to keep on having a combination of all the possible flags in the enum.

Get rid of `_INTERACTION` suffix, just make the color and with author options always include the clickable /map

core/src/main/java/tc/oc/pgm/map/MapInfoImpl.java
```diff
@@ -253,7 +255,7 @@ public Component getStyledName(MapNameStyle style) {
     if (style.isColor) name.color(NamedTextColor.GOLD);
     if (style.isHighlight) name.decoration(TextDecoration.UNDERLINED, true);
     if (style.showAuthors) {
-      return translatable(
+      Component component = translatable(
```
Pablete1234: I'd flatten this out to something like:
```suggestion
    Component component = name.build();
    if (style.showAuthors) {
      component = translatable(...);
    }
    if (style.isInteractive) {
      component = component.hoverEvent(...).clickEvent(...);
    }
    return component;
```

Pablete1234 (approved):

PR #1592, opened 2026-01-04
Make player class components more interactive

cswhite2000:
Adds more interactivity to Class messages

For `/classes`
- Makes it so each entry can be clicked to select that class

For `/class` and the class message from PickerMatchModule
- Adds a hover in the format: `Class Name: Description`
- No click action

Also standardized the response message for class selection. (The coloring was flipped)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/classes/PlayerClass.java
+++ b/core/src/main/java/tc/oc/pgm/classes/PlayerClass.java
@@ -7,15 +7,17 @@ import static tc.oc.pgm.util.Assert.assertNotNull;
 import com.google.common.collect.ImmutableSet;
 import java.util.Set;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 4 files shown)


core/src/main/java/tc/oc/pgm/classes/PlayerClass.java
```diff
@@ -54,6 +60,19 @@ public String getFamilyName() {
     return this.longdescription;
   }
 
+  public Component getComponent() {
```
Pablete1234: tip, you can make PlayerClass implement `ComponentLike` which allows you to append it and stuff

cswhite2000: Well, that's convenient

Pablete1234 (approved):

PR #1596, opened 2026-01-11
Implement ShiftClick bulk purchase in shops

ElBuenAnvita:
Allows the ability to purchase full stacks of items in shops using `Shift + Click` on an Icon in the villager shop menu.

Some notes:
* Had to update `SmartInvs` api dependency to a JitPack snapshot to fix the upstream bug where Shift-Clicks were not being detected (see https://github.com/MinusKube/SmartInvs/pull/185)
* Only `KitNode` (Items) Actions can perform bulk executions.
* Implemented `purchaseStack` which calculates the maximum affordable amount based on inventory, currency and item stackability.


Pablete1234:
Just because the type of action is a kit node doesn't mean it can be bought multiple times.
A kit node can contain armor, potion effects, and even actions. Any of those can be problematic if ran multiple times.

You need to mark the shop icons that are "simple" items (they do not define any custom kit, and it uses a default kit based on the icon) at parse time to make this work.

buildSrc/src/main/kotlin/buildlogic.java-conventions.gradle.kts
```diff
@@ -30,7 +31,7 @@ dependencies {
     api("org.incendo:cloud-minecraft-extras:2.0.0-beta.13")
     api("me.lucko:commodore:2.2")
     api("fr.mrmicky:fastboard:2.1.5")
-    api("fr.minuskube.inv:smart-invs:1.2.7") { exclude("*") }
+    api("com.github.MinusKube:SmartInvs:master-SNAPSHOT") { exclude("*") }
```
calcastor:
```suggestion
    api("com.github.MinusKube:SmartInvs:9c9dbbe") { exclude("*") }
```

calcastor (changes requested):

ElBuenAnvita:
Hi @Pablete1234, thanks for the feedback. Made the changes you told me, please check them!

core/src/main/java/tc/oc/pgm/shops/menu/Payment.java
```diff
@@ -54,4 +54,17 @@ public boolean matches(ItemStack item) {
...
+
+    int totalCurrency = 0;
+    for (ItemStack item : inventory.getContents()) {
+      if (item != null && matches(item)) {
+        totalCurrency += item.getAmount();
+      }
+    }
+    // Whole division: If you have 10 gold and if it costs 3, return 3.
+    return totalCurrency / price;
+  }
```
Pablete1234, line 71: These 2 methods (`hasPayment` and `getAffordableAmount`) should probably be changed such that `getAffordableAmount` gets a limit param (and once limit is found, it stops searching for more(PlayerInventory inventory, int max)` method such that `hasPayment` is simply implemented as:

```java
public boolean hasPayment(PlayerInventory inventory) {
  return getAffordableAmount(inventory, 1) > 0;
}
```

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -46,4 +47,54 @@ public void purchase(Icon icon, MatchPlayer buyer) {
...
+      } else {
+        break;
+      }
+    }
+
+    if (successfulCharges > 0) {
+      triggerBulk(icon.getAction(), buyer, successfulCharges);
+      buyer.getBukkit().updateInventory();
+      buyer.playSound(Sounds.SHOP_PURCHASE); // Only one sound
+    }
```
Pablete1234, line 63: You should be going thru the `Payable` interface, not going straight into the different payments and calculating it yourself here.

Consider simply making `Payable`'s `canPurchase` and `takePayment` methods support a multiplier (for up to how many times you want to take the payment) and return an `int` for how many times it is possible to take that payment.

That would cleanup this code to be simply:
```java
int purchases = icon.takePayment(buyer, desiredPurchases);
if (purchases <= 0) {
  buyer.sendWarning(translatable("shop.currency.insufficient"));
  return;
}
for (int i = purchases; i > 0; i--) 
  action.trigger(player);

buyer.getBukkit().updateInventory();
buyer.playSound(Sounds.SHOP_PURCHASE); // Only one sound
```

core/src/main/java/tc/oc/pgm/shops/menu/ShopMenu.java
```diff
@@ -267,7 +267,13 @@ private ClickableItem getPurchasableItem(Icon icon) {
     meta.addItemFlags(ItemFlag.values());
     item.setItemMeta(meta);
 
-    return ClickableItem.of(item, c -> shop.purchase(icon, getViewer()));
+    return ClickableItem.of(item, c -> {
+      if (c.isShiftClick()) {
+        shop.purchaseStack(icon, getViewer());
+      } else {
+        shop.purchase(icon, getViewer());
+      }
```
Pablete1234: These two methods could probably just be the same, with an additional param to know if u want a stack or not:
`shop.purchase(icon, getViewer(), c.isShiftClick())`

then you have:
```java
void purchase(Icon icon, MatchPlayer viewer, boolean stack) {
  int desiredPurchases = 1;
  if (stack && icon.isStackable()) {
    desiredPurchases = stackSize / amountPerPurchase; // do the actual calculation you already do
  }
  // rest of the logic
}
```

Pablete1234 (commented):
Hey, generally speaking the feature looks good, but i think there's a few spots that could use some refactoring so that codepaths are shared, instead of being duplicated with slight changes, leading to cleaner and more reusable overall code.

core/src/main/java/tc/oc/pgm/shops/menu/ShopMenu.java
ElBuenAnvita: Done. Merged both methods into one, code looks cleaner now, thx for the tips

core/src/main/java/tc/oc/pgm/shops/menu/Payable.java
```diff
@@ -14,17 +14,33 @@ default boolean isFree() {
...
+  default int canPurchase(MatchPlayer buyer, int max) {
+    if (!buyer.getMatch().isRunning() || !buyer.isParticipating()) return 0;
+    if (isFree()) return max;
+
+    int affordable = max;
+    for (Payment payment : getPayments()) {
+      affordable = payment.getAffordableAmount(buyer.getInventory(), affordable);
+      if (affordable <= 0) break;
+    }
+    return affordable;
```
Pablete1234, line 31:
```suggestion
    int affordable = max;
    var inv = buyer.getInventory();
    for (Payment payment : getPayments()) {
      affordable = Math.min(affordable, payment.getAffordableAmount(inv, affordable));
      if (affordable <= 0) break;
    }
    return affordable;
```
must get the one you can afford the least of

core/src/main/java/tc/oc/pgm/shops/menu/Payment.java
```diff
@@ -39,19 +39,31 @@ public ChatColor getColor() {
...
         : item.getType() == currency;
   }
+
+  public int getAffordableAmount(PlayerInventory inventory, int max) {
+    if (price <= 0) return max;
+
+    int totalCurrency = 0;
+    int targetCurrency = max * price; // total desired
+
+    for (ItemStack item : inventory.getContents()) {
```
Pablete1234: `inventory.getContents()` is problematic as in modern and legacy it returns different things (in legacy it only counts inventory while in modern it includes armor and offhand).

PGM has the `Slot` api to workaround this and function the same in modern and legacy:
```suggestion
    for (var slot : Slot.Storage.storage().toList()) {
      ItemStack item = slot.getItem(inventory);
```
Also the version in where payment is taken (ie: `for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {`) should be updated too

core/src/main/java/tc/oc/pgm/shops/menu/Payable.java
```diff
@@ -14,17 +14,33 @@ default boolean isFree() {
...
+  default int takePayment(MatchPlayer buyer, int max) {
+    int affordable = canPurchase(buyer, max);
+    if (affordable <= 0) return 0;
 
     if (!isFree()) {
       PlayerInventory inventory = buyer.getInventory();
       for (Payment payment : getPayments()) {
-        int remaining = payment.getPrice();
+        int remaining = payment.getPrice() * affordable;
         for (int slot = 0; slot < inventory.getSize() && remaining > 0; slot++) {
```
Pablete1234:
```suggestion
        for (var slot : Slot.Storage.storage().toList()) {
```
Gotta use pgm's slot api, gotta change get/set items to use `slot.getItem(inventory)` and `slot.setItem(inventory, item)`

core/src/main/java/tc/oc/pgm/shops/Shop.java
```diff
@@ -46,4 +46,31 @@ public void purchase(Icon icon, MatchPlayer buyer) {
       buyer.sendWarning(translatable("shop.currency.insufficient"));
     }
```
Pablete1234: This purchase method isn't redirecting to the new one, should be:

```suggestion
  public void purchase(Icon icon, MatchPlayer buyer) {
    purchase(icon, buyer, false);
  }
```
I believe it should be identical behavior with stack = false

Pablete1234 (changes requested):
Just a few minor things remaining, looks much better now

core/src/main/java/tc/oc/pgm/shops/menu/Payable.java
ElBuenAnvita: `payment.getAffordableAmount(inventory, max)` already caps it at `max` (in this case, `affordable`), so why the `Math.min`? Or I'm I missing something? 😵‍💫

https://github.com/PGMDev/PGM/pull/1596/changes#diff-b472517e962e7eac4c727d1b80196582e58602f0e0e4d7d8463bed138bb0d1a7R61-R63

Pablete1234: Sorry i missed this msg, and yes you're right, if the impl of getAffordableAmount is correct it should never return a number > affordable, but it's a bit safer to put the min in anyways, up to you tho

Pablete1234 (commented):
Looks very good to me, one last thing i have noticed however:

When buying a stack, if your inventory is full, the OverflowWarningKit would apply N times, spamming your chat.
I'm not too sure on what the best way to fix that would be, maybe ensure the player has an empty inventory slot before allowing them to buy in bulk and early-fail before taking the payment?

Pablete1234 (approved):

PR #1597, opened 2026-01-19
Miscellaneous cleanup and dependency updates

calcastor:
Some of this is taken out of https://github.com/PGMDev/PGM/pull/1577

1. Update Spotless, Shadow, JUnit
2. Update Adventure, JGit, Cloud; use `isTransitive = false` for SmartInvs and JGit
3. Add some additional formatting to Spotless, forbidding wildcard imports and formatting annotations as well as trimming trailing whitespace
4. ~~Exclude JSpecify annotations as well others since Adventure, as well as Paper, are generally migrating to it~~
5. Avoid Adventure deprecation in MessageAction, and Java deprecation in username resolver
6. Largely remove wildcard imports, but don't enforce and keep some where it makes sense to do so
7. Migrate some code to use modern Java features

Other stuff worth mentioning separately:

1. ~~Use JetBrains Annotations in place of JSpecify: should we actually look at migrating to JSpecify as well, given Paper and others doing the same?~~ Begin a gradual migration to JSpecify where possible
2. ~~Don't define unused mappings statically in many Bukkit utils; this also removes some technically unneeded statically defined names such as in PotionEffectTypes that were used and instead uses Bukkit's directly.~~


core/src/main/java/tc/oc/pgm/action/actions/MessageAction.java
```diff
@@ -71,11 +71,13 @@ private Component replaceClickEvents(
       Component component, Function<MatchResult, String> replacer) {
     var click = component.clickEvent();
     if (click != null) {
-      var matcher = PATTERN.matcher(click.value());
+      var payload = ((ClickEvent.Payload.Text) click.payload());
```
Pablete1234: It's no longer safe to assume the payload will be text, should be made part of the if check

```java
if (click != null && click.payload() instanceof ClickEvent.Payload.Text payload) {
  var matcher = PATTERN.matcher(payload.value());

  // ...
```

core/src/main/java/tc/oc/pgm/listeners/PGMListener.java
```diff
@@ -283,8 +283,8 @@ public void announceDynamicMapPoolChange(MapPoolAdjustEvent event) {
       // No limit
       Component forced = translatable("pool.change.force", poolName, staffName);
       if (event.getTimeLimit() != null) {
-        Component time = TemporalComponent.briefNaturalApproximate(event.getTimeLimit())
-            .color(NamedTextColor.GREEN);
+        Component time =
+            TemporalComponent.duration(event.getTimeLimit()).color(NamedTextColor.GREEN);
```
Pablete1234: We generally prefer static imports for chat components, ie:

```suggestion
            duration(event.getTimeLimit()).color(NamedTextColor.GREEN);
```

core/src/main/java/tc/oc/pgm/spawns/states/Dead.java
```diff
@@ -26,9 +27,9 @@ public class Dead extends Spawning {
   private static final PotionEffect CONFUSION =
       new PotionEffect(PotionEffects.NAUSEA, 100, 0, true, false);
   private static final PotionEffect BLINDNESS_SHORT =
-      new PotionEffect(PotionEffects.BLINDNESS, 21, 0, true, false);
```
Pablete1234, line 29: The point of introducing our own enum wrappers (ie: `PotionEffects`) is to stop relying on bukkits' (ie: `PotioneffectType`), it get smore confusing when some use the bukkit one and some use the pgm one.

core/src/main/java/tc/oc/pgm/death/DeathMessageBuilder.java
```diff
@@ -214,8 +218,8 @@ boolean block(BlockInfo blockInfo) {
   private static final Set<EntityType> IGNORED_ENTITIES = ImmutableSet.of(
       EntityTypes.COMPLEX_PART,
       // Modern servers have a translation for the ender crystal
-      Platform.isModern() ? EntityTypes.UNKNOWN : EntityTypes.ENDER_CRYSTAL,
-      EntityTypes.UNKNOWN);
+      Platform.isModern() ? EntityType.UNKNOWN : EntityTypes.ENDER_CRYSTAL,
+      EntityType.UNKNOWN);
```
Pablete1234: When we have a PGM wrapper for the enum, its better to use it consistently instead of using the bukkit EntityType sometimes and pgm's EntityTypes others

util/src/main/java/tc/oc/pgm/util/bukkit/BukkitUtils.java
```diff
@@ -123,29 +126,29 @@ static String potionEffectTypeName(final PotionEffectType type) {
   }
 
   Map<PotionEffectType, String> POTION_EFFECT_MAP = ImmutableMap.<PotionEffectType, String>builder()
-      .put(PotionEffects.BLINDNESS, "Blindness")
-      .put(PotionEffects.NAUSEA, "Nausea")
-      .put(PotionEffects.RESISTANCE, "Resistance")
+      .put(PotionEffectType.SPEED, "Speed")
```
Pablete1234: again, avoid mixing bukkit and pgm's enums

util/src/main/java/tc/oc/pgm/util/bukkit/EntityTypes.java
```diff
@@ -11,35 +11,32 @@ public class EntityTypes {
...
-      parse("MINECART_MOB_SPAWNER", "SPAWNER_MINECART");
-  public static final EntityType PIG_ZOMBIE = parse("PIG_ZOMBIE", "ZOMBIFIED_PIGLIN");
-  public static final EntityType MUSHROOM_COW = parse("MUSHROOM_COW", "MOOSHROOM");
-  public static final EntityType SNOWMAN = parse("SNOWMAN", "SNOW_GOLEM");
   public static final EntityType ENDER_CRYSTAL = parse("ENDER_CRYSTAL", "END_CRYSTAL");
-  public static final EntityType SPLASH_POTION = parse("SPLASH_POTION", "POTION");
-  public static final EntityType FISHING_HOOK = parse("FISHING_HOOK", "FISHING_BOBBER");
-  public static final EntityType LIGHTNING = parse("LIGHTNING", "LIGHTNING_BOLT");
   public static final EntityType COMPLEX_PART = parse("COMPLEX_PART", "UNKNOWN");
-  public static final EntityType UNKNOWN = EntityType.UNKNOWN;
```
Pablete1234, line 42: we should keep them as constants to use

util/src/main/java/tc/oc/pgm/util/bukkit/PotionEffects.java
```diff
@@ -15,29 +15,15 @@ public class PotionEffects {
...
-  public static final PotionEffectType REGENERATION = PotionEffectType.REGENERATION;
-  public static final PotionEffectType SLOWNESS = parse("SLOW", "slowness");
-  public static final PotionEffectType MINING_FATIGUE = parse("SLOW_DIGGING", "mining_fatigue");
-  public static final PotionEffectType SPEED = PotionEffectType.SPEED;
-  public static final PotionEffectType WATER_BREATHING = PotionEffectType.WATER_BREATHING;
-  public static final PotionEffectType WEAKNESS = PotionEffectType.WEAKNESS;
-  public static final PotionEffectType WITHER = PotionEffectType.WITHER;
-  public static final PotionEffectType HEALTH_BOOST = PotionEffectType.HEALTH_BOOST;
-  public static final PotionEffectType ABSORPTION = PotionEffectType.ABSORPTION;
-  public static final PotionEffectType SATURATION = PotionEffectType.SATURATION;
```
Pablete1234, line 40: same here

Pablete1234 (commented):
Generally speaking, some of the pgm "enum like" are *purely* for XML compatibility, and PGM doesn't use them at all, and those are fine if incomplete or only containing entries that changed. However, for those enums that pgm actually uses in its own code (eg: effects) it makes sense to contain all the entries, just be a 1-1 mapping of bukkit's but with multiversion support

Pablete1234 (approved):

PR #1599, opened 2026-01-22
Treewide cleanup and use of modern Java features

calcastor:
This PR selectively applies many IntelliJ suggestions that are less disruptive and/or less breaking:

1. Use switch expressions where possible
2. Use pattern variables where possible
3. Migrate some internal classes to records where possible, and use record patterns where applicable
4. Add many missing annotations
5. Cleanup a lot of logic oddities, and remove some unused variables in classes going back to ProjectAres
6. Migrate all modified classes to JSpecify following #1597 

Some future considerations:
1. FileUtils seems to be quite legacy and cites a source site that no longer exists, maybe worth using modern Java APIs for this
2. There are a lot of classes that we could migrate to being records, but many getters are used outside of PGM so perhaps maintaining those while deprecating is worth doing.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
+++ b/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
@@ -6,6 +6,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -100,11 +101,20 @@ public class ClassModule implements MapModule<ClassMatchModule> {
       if (family == null)
         throw new InvalidXMLException("Unable to determine family for classes", doc);
 
+      Set<String> usedNames = Sets.newHashSet();
+
       ImmutableMap.Builder<String, PlayerClass> builder = ImmutableMap.builder();
       PlayerClass defaultClass = null;
 
       for (Element classEl : classElements) {
         PlayerClass cls = parseClass(classEl, factory.getKits(), family);
+
+        usedNames.add(cls.getName().toLowerCase());
+        if (usedNames.contains(cls.getName().toLowerCase())) {
+          throw new InvalidXMLException(
+              "Class already registered to \" + cls.getName() + \"; skipping second instance",
+              classEl);
+        }
 
         String classFamily = classEl.getAttributeValue("family");
         if (!family.equals(classFamily)) {
--- a/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
+++ b/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
@@ -6,7 +6,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -15,7 +14,7 @@ import java.util.Set;
 import java.util.logging.Logger;
 import org.jdom2.Document;
 import org.jdom2.Element;
-import org.jetbrains.annotations.Nullable;
+import org.jspecify.annotations.Nullable;
 import tc.oc.pgm.api.map.MapModule;
 import tc.oc.pgm.api.map.MapTag;
 import tc.oc.pgm.api.map.factory.MapFactory;
@@ -86,8 +85,7 @@ public class ClassModule implements MapModule<ClassMatchModule> {
       for (Element classEl : classElements) {
         String classFamily = classEl.getAttributeValue("family");
         if (classFamily != null) {
-          Integer num = familyFrequency.get(classFamily);
-          familyFrequency.put(classFamily, (num != null ? num.intValue() : 0) + 1);
+          familyFrequency.compute(classFamily, (k, num) -> (num != null ? num : 0) + 1);
         }
       }
 
@@ -102,28 +100,17 @@ public class ClassModule implements MapModule<ClassMatchModule> {
       if (family == null)
         throw new InvalidXMLException("Unable to determine family for classes", doc);
 
-      Set<String> usedNames = Sets.newHashSet();
       ImmutableMap.Builder<String, PlayerClass> builder = ImmutableMap.builder();
       PlayerClass defaultClass = null;
 
       for (Element classEl : classElements) {
         PlayerClass cls = parseClass(classEl, factory.getKits(), family);
 
-        if (usedNames.contains(cls.getName().toLowerCase())) {
-          throw new InvalidXMLException(
-              "Class already registered to \" + cls.getName() + \"; skipping second instance",
-              classEl);
-        }
-
         String classFamily = classEl.getAttributeValue("family");
-        if (family == null) {
-          family = classFamily;
-        } else {
-          if (!family.equals(classFamily)) {
-            throw new InvalidXMLException(
-                "Family was determined to be '" + family + "' but class specified '" + classFamily,
-                classEl);
-          }
+        if (!family.equals(classFamily)) {
+          throw new InvalidXMLException(
+              "Family was determined to be '" + family + "' but class specified '" + classFamily,
+              classEl);
         }
 
         if (XMLUtils.parseBoolean(classEl.getAttribute("default"), false)) {
--- a/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
+++ b/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
@@ -109,12 +109,13 @@ public class ClassModule implements MapModule<ClassMatchModule> {
       for (Element classEl : classElements) {
         PlayerClass cls = parseClass(classEl, factory.getKits(), family);
 
-        usedNames.add(cls.getName().toLowerCase());
         if (usedNames.contains(cls.getName().toLowerCase())) {
           throw new InvalidXMLException(
               "Class already registered to \" + cls.getName() + \"; skipping second instance",
               classEl);
         }
+
+        usedNames.add(cls.getName().toLowerCase());
 
         String classFamily = classEl.getAttributeValue("family");
         if (!family.equals(classFamily)) {
--- a/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
+++ b/core/src/main/java/tc/oc/pgm/classes/ClassModule.java
@@ -24,6 +24,7 @@ import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.kits.Kit;
 import tc.oc.pgm.kits.KitModule;
 import tc.oc.pgm.kits.KitParser;
+import tc.oc.pgm.util.StringUtils;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.material.ItemMaterialData;
 import tc.oc.pgm.util.xml.InvalidXMLException;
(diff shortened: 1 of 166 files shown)


core/src/main/java/tc/oc/pgm/command/util/CommandKeys.java
```diff
@@ -9,11 +9,11 @@
...
-  public static final CloudKey<Match> MATCH = CloudKey.of("_pgm_match_", new TypeToken<Match>() {});
+  public static final CloudKey<Match> MATCH = CloudKey.of("_pgm_match_", new TypeToken<>() {});
 
   public static final CloudKey<SettingKey> SETTING_KEY =
-      CloudKey.of("_pgm_setting_key_param_", new TypeToken<SettingKey>() {});
+      CloudKey.of("_pgm_setting_key_param_", new TypeToken<>() {});
 
   public static final CloudKey<MapPoolType> POOL_TYPE =
-      CloudKey.of("_pgm_pool_type_param_", new TypeToken<MapPoolType>() {});
+      CloudKey.of("_pgm_pool_type_param_", new TypeToken<>() {});
```
TTtie: Consider passing the class directly, there's an overload for this

```suggestion
  public static final CloudKey<Match> MATCH = CloudKey.of("_pgm_match_", Match.class);

  public static final CloudKey<SettingKey> SETTING_KEY =
      CloudKey.of("_pgm_setting_key_param_", SettingKey.class);

  public static final CloudKey<MapPoolType> POOL_TYPE =
      CloudKey.of("_pgm_pool_type_param_", MapPoolType.class);
```

core/src/main/java/tc/oc/pgm/db/CacheDatastore.java
```diff
@@ -25,29 +26,27 @@ public class CacheDatastore implements Datastore {
...
-          }
-        });
-    this.settings = CacheBuilder.newBuilder().build(new CacheLoader<UUID, Settings>() {
+    this.usernames = CacheBuilder.newBuilder().softValues().build(new CacheLoader<>() {
       @Override
-      public Settings load(UUID id) {
+      public Username load(@NonNull UUID id) {
+        return datastore.getUsername(id);
+      }
+    });
```
TTtie: May be worthwhile to use `CacheLoader.from` instead of constructing new anonymous objects, e. g.
```
    this.usernames = CacheBuilder.newBuilder()
            .softValues()
            .build(CacheLoader.from(datastore::getUsername));
```

core/src/main/java/tc/oc/pgm/filters/operator/FilterNode.java
```diff
@@ -59,15 +59,11 @@ public QueryResponse query(Query query) {
 
   public static FilterNode allow(Filter child) {
     return new FilterNode(
-        Collections.<Filter>emptyList(),
-        Collections.singletonList(child),
-        Collections.<Filter>emptyList());
+        Collections.emptyList(), Collections.singletonList(child), Collections.emptyList());
```
TTtie: `List.of(...)` is shorter and immutable like `Collections.empty/singletonList` is
```suggestion
        List.of(), List.of(child), List.of());
```

core/src/main/java/tc/oc/pgm/filters/operator/FilterNode.java
```diff
@@ -59,15 +59,11 @@ public QueryResponse query(Query query) {
...
   }
 
   public static FilterNode deny(Filter child) {
     return new FilterNode(
-        Collections.<Filter>emptyList(),
-        Collections.<Filter>emptyList(),
-        Collections.singletonList(child));
+        Collections.emptyList(), Collections.emptyList(), Collections.singletonList(child));
```
TTtie: Analogically
```suggestion
        List.of(), List.of(), List.of(child));
```

core/src/main/java/tc/oc/pgm/points/SpreadPointProvider.java
```diff
@@ -55,7 +55,7 @@ public Location getPoint(Match match, @Nullable Entity entity) {
       }
     }
 
-    int index = (int) Math.floor(match.getRandom().nextInt(bestPoints.size()));
+    int index = (int) (double) match.getRandom().nextInt(bestPoints.size());
```
TTtie: The double-cast is redundant
```suggestion
    int index = match.getRandom().nextInt(bestPoints.size());
```

core/src/main/java/tc/oc/pgm/classes/ClassModule.java
```diff
@@ -102,30 +100,19 @@ public ClassModule parse(MapFactory factory, Logger logger, Document doc)
       if (family == null)
         throw new InvalidXMLException("Unable to determine family for classes", doc);
 
-      Set<String> usedNames = Sets.newHashSet();
```
Pablete1234, line 105: Nothing controls duplicate class names?
this seems like an earlier bug (used names not being added to) that is just being covered up by straight up removing it here

core/src/main/java/tc/oc/pgm/db/CacheDatastore.java
```diff
@@ -25,32 +25,13 @@ public class CacheDatastore implements Datastore {
...
-        return datastore.getMapActivity(name);
-      }
-    });
+    this.usernames =
+        CacheBuilder.newBuilder().softValues().build(CacheLoader.from(datastore::getUsername));
+    this.settings =
+        CacheBuilder.newBuilder().softValues().build(CacheLoader.from(datastore::getSettings));
+    this.skins = CacheBuilder.newBuilder().softValues().build(CacheLoader.from(datastore::getSkin));
+    this.activities =
+        CacheBuilder.newBuilder().softValues().build(CacheLoader.from(datastore::getMapActivity));
```
Pablete1234: You're adding `softValues()` to things that intentionally did not have them before, and that's problematic.

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
```diff
@@ -62,17 +62,8 @@ public void onParticipantLeave(PlayerPartyChangeEvent event) {
...
-          }
-          break;
-        case DELETE:
-          enderchest.clear();
-          break;
-        default:
-          break;
+      if (fallback == DropoffFallback.DELETE) {
+        enderchest.clear();
       }
```
Pablete1234, line 76: this is a behavior change

core/src/main/java/tc/oc/pgm/filters/operator/AllowFilter.java
```diff
@@ -20,11 +20,9 @@ public boolean respondsTo(Class<? extends Query> queryType) {
 
   @Override
   public QueryResponse query(Query query) {
-    switch (filter.query(query)) {
```
Pablete1234, line 23: Tbh i'd keep these as switches even if they have just two cases, they read more clearly as switches

core/src/main/java/tc/oc/pgm/filters/operator/DenyFilter.java
```diff
@@ -18,11 +18,9 @@ public boolean respondsTo(Class<? extends Query> queryType) {
 
   @Override
   public QueryResponse query(Query query) {
-    switch (filter.query(query)) {
```
Pablete1234, line 21: same here

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/ModernEntityPackets.java
```diff
@@ -134,8 +134,6 @@ public Packet entityHeadEquipment(int entityId, ItemStack helmet) {
   @Override
   public Packet entityMetadataPacket(int entityId, Entity entity, boolean complete) {
     var data = ((CraftEntity) entity).getHandle().getEntityData().packAll();
-    return data == null
-        ? Packet.of()
-        : new ModernPacket<>(new ClientboundSetEntityDataPacket(entityId, data));
```
Pablete1234, line 139: this is a behavior change, i assume `packAll()` is probably promising a nonnull result, but i'm not sure if pgm should trust that

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/ModernTabPackets.java
```diff
@@ -54,7 +54,6 @@ public PlayerInfo createPlayerInfoPacket(EnumPlayerInfoAction action) {
   @Override
   public Packet spawnPlayerPacket(int entityId, UUID uuid, Location loc, Player player) {
     var data = ((CraftEntity) player).getHandle().getEntityData().packAll();
-    if (data == null) return Packet.of();
```
Pablete1234, line 57: same here

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/ModernTabPackets.java
```diff
@@ -135,7 +134,7 @@ public void addPlayerInfo(
         String name,
         int ping,
         @Nullable Skin skin,
-        @Nullable net.kyori.adventure.text.Component displayName) {
```
Pablete1234, line 138: why did these stop being nullable?

util/src/main/java/tc/oc/pgm/util/tablist/TabManager.java
```diff
@@ -95,8 +95,8 @@ public TabEntry getPlayerEntry(Player player) {
     return this.playerEntries.get(player);
   }
 
-  public TabEntry removePlayerEntry(Player player) {
-    return this.playerEntries.remove(player);
+  public void removePlayerEntry(Player player) {
```
Pablete1234: even if pgm doesn't use the return type from this, its technically part of the public api and i see no reason to get rid of it

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/enderchest/EnderChestMatchModule.java
calcastor: We already return early in this method when `dropoffs.isEmpty()` is true.

Pablete1234: Then that's a bug, because DELETE wouldn't actually delete unless some dropoffs are defined

calcastor: Removed the early return

core/src/main/java/tc/oc/pgm/classes/ClassModule.java
calcastor: Ensured we're adding `cls.getName().toLowerCase()` to `usedNames`

core/src/main/java/tc/oc/pgm/classes/ClassModule.java
```diff
@@ -115,15 +115,13 @@ public ClassModule parse(MapFactory factory, Logger logger, Document doc)
               classEl);
         }
 
+        usedNames.add(cls.getName().toLowerCase());
```
Pablete1234:
```suggestion
if (!usedNames.add(StringUtils.normalize(cls.getName()))) {
throw new InvalidXMLException(
"Class already registered to \" + cls.getName() + \"; skipping second instance",
classEl);
}
```
Can be simplified, and we can improve the normalization

core/src/main/java/tc/oc/pgm/db/CacheDatastore.java
```diff
@@ -25,32 +25,10 @@ public class CacheDatastore implements Datastore {
...
-        return datastore.getSkin(id);
-      }
-    });
-    this.activities = CacheBuilder.newBuilder().build(new CacheLoader<String, MapActivity>() {
-      @Override
-      public MapActivity load(String name) {
-        return datastore.getMapActivity(name);
-      }
-    });
+    this.usernames = CacheBuilder.newBuilder().build(CacheLoader.from(datastore::getUsername));
```
Pablete1234: You've lost softValues on usernames

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/ModernTabPackets.java
```diff
@@ -161,7 +162,7 @@ public void addPlayerInfo(
...
```
Pablete1234, line 164: no longer nullable?

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #1605, opened 2026-02-02
Migrate several classes to records

calcastor:


calcastor:
I'll come back to this at a later date

PR #1608, opened 2026-02-04
Set custom-meta-applied for all item rules

calcastor:


core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
```diff
@@ -19,22 +19,17 @@
...
 
   public boolean applyRules(ItemStack stack) {
     if (stack == null || stack.getType() == Material.AIR || APPLIED.has(stack)) {
       return false;
     } else {
-      APPLIED.set(stack, true);
       for (ItemRule rule : rules) {
         if (rule.matches(stack)) {
           rule.apply(stack);
+          APPLIED.set(stack, true);
```
Pablete1234, line 32: why?
The whole point of applied is it serves as an optimization so that we don't need to check this item again

calcastor: Not doing this breaks filters like `carrying` on modern without `ignore-metadata="true"` when item-mods are in use. This can be observed in Touchdown maps on modern with OCC's `legacy-pvp` include (or any touchdown map with an item-mod added); you will receive two snowballs on flag pickup.

This is also what was done in ProjectAres: https://github.com/OvercastNetwork/ProjectAres/blob/master/PGM/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java#L67

Pablete1234 (approved):
I think this is more of a bandaid fix than a real one, and we should look into fixing item matchers to generally ignore locked/unsharable status, but its good enough for now

PR #1611, opened 2026-02-08
Reduce boilerplate code using Lombok

calcastor:


Pugzy:
I've found using lombok in projects that weren't a small/quick hack together makes things harder overall.

Debugging is a big one, harder to find constructor usages, getters/setters etc. I spend most of my time in PGM live debugging and stepping through code so this would be a net negative. It's a larger learning curve too.

For a project like PGM with newer developers and a large stable codebase, this seems like a bad idea.

calcastor:
there has been a Negotiation

PR #1614, opened 2026-02-08
Open shop action

Javggg:
This PR introduces a new action for opening shops.
By exposing shop access as an action, shops can now be triggered dynamically, such as through consumables, rather than only through static NPCs.

Usage:
```xml
<open-shop shop="shop-id" />
```


core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -388,6 +390,14 @@ public KillEntitiesAction parseKillEntities(Element el, Class<?> scope)
     return new KillEntitiesAction(parser.filter(el, "filter").required());
   }
 
+  @MethodParser("open-shop")
+  public OpenShop parseOpenShop(Element el, Class<?> scope) throws InvalidXMLException {
+    Attribute shopAttr = XMLUtils.getRequiredAttribute(el, "shop");
+    String shopId = shopAttr.getValue();
+
+    return new OpenShop(shopId);
```
Pablete1234: This is not how you want to parse things; this will make makes no validation attempt at map load time, and puts all the burden into matchtime trying to resolve the shop.

You should be parsing to an `XmlReference<Shop>`, that pgm will validate at the end of map load to ensure that a feature with that id of type shop exists in the feature store

Javggg: Perfect. I'll make the changes shortly

Javggg: I tried to do this before and it didn't work. I don't know what I did wrong then lol.

Pablete1234 (approved):

PR #1616, opened 2026-02-10
Use records where least disruptive

calcastor:


Pablete1234:
While i don't oppose merging of this, i believe its going to be impossible to merge in its current state, and to make merging viable needs to be split into smaller PRs for different systems, which can be reviewed/merged separately

PR #1620, opened 2026-02-12
Add upgrade strategy for items & blocks

Pablete1234:
It's not perfect, but its the closest to perfect I could do for now.

There's still some things that won't exactly behave the same in modern and legacy, mainly stairs and slabs.
Eg: in modern `stone_slab:0` would match only bottom half smooth stone slabs, in modern it will match any smooth stone slab (the alternative would be it only matching non-waterlogged bottom-block smooth stone slabs, which is probably not wanted behavior, and gets worse with other things in this category).


cswhite2000 (approved):
this is pain

PR #1621, opened 2026-02-14
Avoid using PlayerSpawnLocationEvent for RecipeUnlocker

calcastor:


Pablete1234:
I believe something like this was my first approach, but it caused all the new unlocked recipes to flash as toasts for like several minutes, which is why i pivoted to making sure it ran on the "recipebook initialization" phase instead of a subsequent add.

Has that behavior changed?

kashike:
> Has that behavior changed?

Setting the second parameter (`notification`) to `false` in `ClientboundRecipeBookAddPacket.Entry(RecipeDisplayEntry contents, boolean notification, boolean highlight)` prevents the toast notification from being displayed as of [23w07a](https://minecraft.wiki/w/Java_Edition_23w07a#General).

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/RecipeUnlocker.java
```diff
@@ -1,25 +1,44 @@
...
+    Bukkit.getScheduler().runTask(PGM.get(), () -> {
+      var player = ((CraftPlayer) event.getPlayer()).getHandle();
+
+      // Send completed recipe book to client
+      var packet = new ClientboundRecipeBookAddPacket(ENTRIES, true);
+      player.connection.send(packet);
+
+      // Update known recipes on server
+      player.getRecipeBook().known.addAll(RECIPE_MANAGER.recipes.byKey.keySet());
+    });
```
Pablete1234: I'd suggest a few changes: 
 - make `RecipeUnlocker` implement `PacketSender` to inherit those packet sending utils
 - utilize pgm's executor
 - check for isConnected, a player may leave before the end of the tick, causing an exception when sending them a packet
 - extract Player out of the event so that is the only param retained for the lambda's class

```suggestion
    var player = event.getPlayer();
    PGM.get().getExecutor().submit(() -> {
      if (!player.isConnected()) return;
      
      send(new ClientboundRecipeBookAddPacket(ENTRIES, true), player);
      player.getRecipeBook().known.addAll(RECIPE_MANAGER.recipes.byKey.keySet());
    });
```

Pablete1234 (approved):

PR #1622, opened 2026-02-14
Use AsyncPlayerSpawnLocationEvent for SpawnMatchModule on modern

calcastor:
Adds a shim of `AsyncPlayerSpawnLocationEvent` to handle Paper's `AsyncPlayerSpawnLocationEvent` on modern and `PlayerSpawnLocationEvent` on legacy. This avoids early creation of the player object on modern by use of the deprecated `PlayerSpawnLocationEvent`.

Prior to Spigot support, PGM used to use `PlayerInitialSpawnEvent` for SportPaper; perhaps we should use this again given that we have limited our supported platforms to SportPaper and modern Paper?

### DIFF
--- a/core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
@@ -26,7 +26,6 @@ import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.PlayerDeathEvent;
 import org.bukkit.event.inventory.InventoryClickEvent;
 import org.jspecify.annotations.Nullable;
-import org.spigotmc.event.player.PlayerSpawnLocationEvent;
 import tc.oc.pgm.api.PGM;
 import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.query.Query;
@@ -55,6 +54,7 @@ import tc.oc.pgm.spawns.states.State;
 import tc.oc.pgm.teams.Team;
 import tc.oc.pgm.util.event.PlayerItemTransferEvent;
 import tc.oc.pgm.util.event.player.PlayerAttackEntityEvent;
+import tc.oc.pgm.util.event.player.PlayerSpawnLocationEvent;
 
 @SuppressWarnings("UnstableApiUsage")
 @ListenerScope(MatchScope.LOADED)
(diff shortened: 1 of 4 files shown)


calcastor:
Draft until I actually test on SportPaper

calcastor:
This also works on SportPaper provided we only make the PGM event asynchronous on modern which isn't very surprising. I've renamed the event accordingly.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernListener.java
```diff
@@ -123,4 +124,11 @@ public void onPlayerDeath(PlayerDeathEvent event) {
       conn.handleAcceptPlayerLoad(PLAYER_LOADED_PACKET);
     });
   }
+
+  @EventHandler(priority = EventPriority.MONITOR)
+  public void onInitialSpawn(io.papermc.paper.event.player.AsyncPlayerSpawnLocationEvent event) {
+    var pgmEvent = new PlayerSpawnLocationEvent(event.getSpawnLocation());
+    handleCall(pgmEvent, event);
+    event.setSpawnLocation(pgmEvent.getSpawnLocation());
```
Pablete1234, line 132: it is likely unsafe to query pgm async for a spawn location, it probably needs to run in the main thread, at which point, why are we even trying to replace the old event that did just that

calcastor: PlayerSpawnLocationEvent is deprecated by Paper, hence. 

If we don't care about that right now then I'll just suppress the deprecation warning for this specific event and call it a day. This seems like it could matter more in the future though.

Pablete1234: i'm aware of the deprecation, but it still is the event that fits us best, they've not added a 1:1 replacement as far as i know while the deprecated one works as intended

calcastor:
While I don't think this event is going to be removed in the near future, I don't think we should expect a one-to-one replacement when it is, so I have made some changes that should make it safer.

core/src/main/java/tc/oc/pgm/spawns/SpawnMatchModule.java
```diff
@@ -64,6 +64,7 @@ public class SpawnMatchModule implements MatchModule, Listener, Tickable {
 
   private final Match match;
   private final SpawnModule module;
+  private final Location initialSpawnLocation;
```
Pablete1234:
```suggestion
  private final Location worldSpawn;
```

Pablete1234 (approved):

Pablete1234 (approved):

PR #1627, opened 2026-02-18
Dynamic nametag visibility

Javggg:
After iterating on the idea, we settled on the following syntax:
```xml
<!-- FFA: controls visibility for all players -->
<players name-tags-filter="some-filter" />

<!-- Teams: control allies and enemies visibility independently -->
<teams name-tags-allies-filter="some-filter-a" name-tags-enemies-filter="some-filter-b">
  <team .../>
</teams>
```
Both attributes are incompatible with `show-name-tags`. The filters are party-scoped.


core/src/main/java/tc/oc/pgm/action/actions/NametagVisibilityAction.java
```diff
@@ -0,0 +1,40 @@
...
+      var team = teamModule.get().getTeam(party.getNameLegacy());
+
+      if (team != null) {
+        team.setNameTagVisibility(this.visibility);
+        Objects.requireNonNull(party.getMatch().getModule(ScoreboardMatchModule.class))
+            .updatePartyScoreboardTeam(party);
+      }
+    }
+    if (ffaModule.isPresent()) {
+      party.getMatch().getCompetitors().forEach(competitor -> {
```
cswhite2000: With this, you're setting the visibility manually for all players in the match. You should only be setting it for the team of the individual competitor.

core/src/main/java/tc/oc/pgm/action/actions/NametagVisibilityAction.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public void trigger(Party party) {
+    var teamModule = party.getMatch().moduleOptional(TeamMatchModule.class);
+    var ffaModule = party.getMatch().moduleOptional(FreeForAllMatchModule.class);
+
+    if (teamModule.isPresent()) {
+      var team = teamModule.get().getTeam(party.getNameLegacy());
+
+      if (team != null) {
```
cswhite2000:
```suggestion
   if (party instanceof Team team) {
```

core/src/main/java/tc/oc/pgm/action/actions/NametagVisibilityAction.java
Javggg: I’m not exactly sure what you mean.
As far as I understand, in FFA every player is assigned a scoreboard team. So to make this work, I _have_ to loop through every competitor, right?

Pablete1234: The action affects the current scooed team/competitor/party.
If a single player in ffa gets visibility changed, then only that competitor should get the change.
If you want everyone in the match to be affected, you make your action match scoped instead and the switch-scope is what will take care of this "iterating thru all the teams/tributes" for you.

Ontop of that tho, for ffa this gets complicated because players joining later than the change will not get updated.

Consider that for something that is a state of the match, maybe a one-time action isnt the solution. Yes actions are easy to just code a new one for, but it may not be the right tool here.

Consider an expansion to the scoreboard module (the one that handles the actual scoreboard team creation) that parses xml for some match scoped dynamic filter on when visibility should change.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/action/actions/NametagVisibilityAction.java
Javggg: Perfect, I'll work on that. Thanks for the feedback!

Javggg:
Sorry for the delay on this.
Considering the feedback, I decided to transform this into a filter-controlled behavior instead of actions, addressing the concerns about state consistency and players joining after a change.
It works by adding new attributes to <players> and <team> that accept a dynamic filter, incompatible with show-name-tags:
```xml
<!-- FFA: controls visibility for everyone -->
<players name-tags-filter="some-filter" />

<!-- Teams: control allies and enemies visibility independently -->
<team ... name-tags-allies-filter="some-filter-a" name-tags-enemies-filter="some-filter-b" />
```
When the filter changes state, the scoreboard team is updated immediately for all current players. Players joining after the change will also get the correct visibility.

core/src/main/java/tc/oc/pgm/teams/TeamModule.java
```diff
@@ -114,8 +117,22 @@ private static TeamFactory parseTeamDefinition(Element el, MapFactory factory)
 
     ChatColor color = XMLUtils.parseChatColor(Node.fromAttr(el, "color"), ChatColor.WHITE);
     DyeColor dyeColor = XMLUtils.parseDyeColor(el.getAttribute("dye-color"), null);
-    NameTagVisibility nameTagVisibility = XMLUtils.parseNameTagVisibility(
-        Node.fromAttr(el, "show-name-tags"), NameTagVisibility.ALWAYS);
+
+    NameTagVisibility nameTagVisibility =
+        XMLUtils.parseNameTagVisibility(Node.fromAttr(el, "show-name-tags"), null);
+    Filter nameTagAlliesFilter =
+        parser.filter(el, "name-tags-allies-filter").dynamic(Match.class).orNull();
```
cswhite2000: We should see if we can come up with a shorter concise name for these

core/src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
```diff
@@ -145,6 +149,44 @@ public void load() {
...
   }
 
+  @EventHandler
+  public void onMatchLoad(MatchLoadEvent event) {
+    var fmm = match.needModule(FilterMatchModule.class);
+    var smm = match.needModule(ScoreboardMatchModule.class);
+
+    for (Team team : teams) {
+      TeamFactory info = team.getInfo();
+      boolean[] alliesState = {true};
```
cswhite2000: I guess you're using an array to get around variables needing to be effectively final when used in a lambda. Generally, your ide would suggest using an AtomicBoolean instead.

But overall, this is not the right way to write this part of the logic.

For this I'd probably use a HashMap for storing the state, but I'll defer to Pablo on this for what you should do

Pablete1234: Yea, storing state this way is dirty, and the only reason this is being done is because you have 2 filters that are controlling one of 4 states of the same variable. I think you may be able to avoid this state all together, if you consider the matrix of states:

|  allies \ enemies  | true              | false                |
|-------|-------------------|----------------------|
| true  | ALWAYS            | HIDE_FOR_OTHER_TEAMS |
| false | HIDE_FOR_OWN_TEAM | NEVER                |

Each of the two filters, only needs to move in one of the axis, so given that you're able to get the current value you can make the switch to the other without needing to store the boolean values at all.

eg: if enemies response = false
ALWAYS turns into HIDE_FOR_OTHER_TEAMS
HIDE_FOR_OWN_TEAM turns into never

Enemies filter uses the horizontal direction (move right for false, move left for true), and allies filter uses the vertical direction

core/src/main/java/tc/oc/pgm/teams/TeamModule.java
Javggg: I agree, but I wanted to keep some consistency with the original `show-name-tags` attribute.
Could be `allies-tags-filter / enemies-tags-filter`.
Other suggestions are welcome.

cswhite2000: Your existing `name-tags-allies-filter` is probably better than `allies-tags-filter`. The tag is a bit long, but I don't want to lose meaning. A shorter name would be better, but only if we can come up with a good one that doesn't lose necessary info

cswhite2000 (dismissed):

core/src/main/java/tc/oc/pgm/ffa/FreeForAllMatchModule.java
```diff
@@ -124,13 +133,31 @@ public void setMaxPlayers(@Nullable Integer maxPlayers, @Nullable Integer maxOve
...
+  public void onMatchLoad(MatchLoadEvent event) {
+    var fmm = match.needModule(FilterMatchModule.class);
+
+    fmm.onChange(Match.class, options.nameTagVisibilityFilter, ((filterable, response) -> {
+      setNameTagVisibility(response ? NameTagVisibility.ALWAYS : NameTagVisibility.NEVER);
+
+      var smm = match.needModule(ScoreboardMatchModule.class);
+      tributes.forEach((uuid, tribute) -> {
+        smm.updatePartyScoreboardTeam(tribute);
+      });
```
Pablete1234: Feels like this logic should be part of setNameTagVisibility, since if anything else was to want to change that it'd probably also want to update existing tributes

core/src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
```diff
@@ -145,6 +149,46 @@ public void load() {
...
+        fmm.onChange(Match.class, info.getNameTagAlliesFilter(), (filterable, response) -> {
+          NameTagVisibility current = team.getNameTagVisibility();
+          team.setNameTagVisibility(
+              response
+                  ? (current == NameTagVisibility.HIDE_FOR_OWN_TEAM
+                      ? NameTagVisibility.ALWAYS
+                      : NameTagVisibility.HIDE_FOR_OTHER_TEAMS)
+                  : (current == NameTagVisibility.ALWAYS
+                      ? NameTagVisibility.HIDE_FOR_OWN_TEAM
+                      : NameTagVisibility.NEVER));
```
Pablete1234: a switch here may make this nicer to read, something along the lines of:
```java
team.setNameTagVisibility(switch(current) {
  case ALWAYS, HIDE_FOR_OWN_TEAM -> response ? ALWAYS : HIDE_FOR_OWN_TEAM;
  case HIDE_FOR_OTHER_TEAMS, NEVER -> response ? HIDE_FOR_OTHER_TEAMS : NEVER;
})
```
(please do double-check which goes where, )

core/src/main/java/tc/oc/pgm/teams/TeamMatchModule.java
```diff
@@ -145,6 +149,46 @@ public void load() {
...
+          smm.updatePartyScoreboardTeam(team);
```
Pablete1234: As mentioned earlier, maybe updating the visibility should be a method that then calls for smm to update? ie: `TeamMatchModule#setNameTagVisibility(Team team, NameTagVisibility visibility)` that updates it on the team and then updates it in smm?

core/src/main/java/tc/oc/pgm/teams/TeamModule.java
```diff
@@ -114,8 +117,22 @@ private static TeamFactory parseTeamDefinition(Element el, MapFactory factory)
...
+    Filter nameTagEnemiesFilter =
+        parser.filter(el, "name-tags-enemies-filter").dynamic(Match.class).orNull();
```
Pablete1234: I do wonder if it would make more sense for these to be party/team-scoped instead of match-scoped, and apply to whatever team they respond true/false to separately, eg:

```xml
<all>
  <time>10m</time>
  <team>blue</team>
</all>
```
would make it so blue after 10mins starts passing the filter, while the rest of the teams never would.

Which also allows for things like:
```xml
<score>5..</score>
```
which makes it so whatever team(s) get to 5 score would start showing their names.


Of course if that were to happen, you would need a single filter instead of one per team, i do wonder if that is a cleaner and more flexible approach, it may be

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/teams/TeamModule.java
Javggg: I decided to leave these as match-scoped because of the reasoning you previously said of this being a "match state". But yeah, it absolutely makes sense to make it party-scoped instead.

Pablete1234 (approved):

Pablete1234:
Good work on the pr, looks good to me now.
Please do update your original comment with the new syntax/examples, and confirm that testing shows this all working as expected. :+1:

Javggg:
Testing confirms dynamic updates work correctly when filters change state, and players joining after a change get the correct visibility.

PR #1628, opened 2026-02-19
⏰ THE Wait action ⏰

eliminate123:
This PR introduces a new action type. This action type is called:
```xml 
<wait duration="2s"></wait>
```
This is useful in scenarios where an after filter is not practical, due to not having a filter to trigger from. One use case is as follows:
```xml
<action id="wait-action" scope="match" expose="true">
    <message text="Message one"/>
    <sound preset="alert" volume="0.75" pitch="1.5"/>
    <wait duration="2s">
        <message text="Message two"/>
        <sound preset="alert" volume="0.75" pitch="1.5"/>
    </wait>
    <wait duration="5s">
        <message text="Message three"/>
        <sound preset="alert" volume="0.75" pitch="1.5"/>
    </wait>
</action>
```

Without the wait action, this code would look like this:
    
```xml
<filters>
    <after duration="2s" id="wait-2" filter="some-filter"/>
    <after duration="5s" id="wait-3" filter="some-filter"/>
</filters>
<actions>
    <trigger filter="some-filter">
        <action>
            <message text="Message one"/>
            <sound preset="alert" volume="0.75" pitch="1.5"/>
        </action>
    </trigger>
    <trigger filter="wait-2">
        <action>
            <message text="Message two"/>
            <sound preset="alert" volume="0.75" pitch="1.5"/>
        </action>
    </trigger>
    <trigger filter="wait-3">
        <action>
            <message text="Message three"/>
            <sound preset="alert" volume="0.75" pitch="1.5"/>
        </action>
    </trigger>
</actions>
```
This is an action that is very new player friendly, and a function that already exists in most programming languages.


Pablete1234 (changes requested):
This implementation is too simplistic, and its a liability. Nothing prevents an action scheduled in one map to run in a future one.

Pablete1234 (dismissed):

Pablete1234:
Cleaned up and fixed a few things:
 - Made it so the duration must be between 0s and 60s, if over a minute of delay is needed this is not the mechanism for it.
 - Separated player-scoped into a specialized action so that not all of them need to do the party logic
 - Generally cleaner code

Still undecided:
 - `schedule` may be a better name than `wait`, as wait implies it blocks the action.
For example, this is how "wait" (eg: `Time.sleep()`) would work:
```xml
<action scope="match">
  <message text="3"/>

  <wait duration="1s"/>
  <message text="2"/>

  <wait duration="1s"/>
  <message text="1"/>

  <wait duration="1s"/>
  <message text="Go!"/>
</action>
```
(do note, this would be very complicated to implement as we can't just sleep the main thread in pgm, we'd need actions to effectively implement a continuation api of sorts).


So, maybe the syntax should be `schedule` and `after` for the duration:
```xml
<action id="wait-action" scope="match" expose="true">
    <message text="Message one"/>
    <sound preset="alert" volume="0.75" pitch="1.5"/>
    <schedule after="2s">
        <message text="Message two"/>
        <sound preset="alert" volume="0.75" pitch="1.5"/>
    </wait>
    <schedule after="5s">
        <message text="Message three"/>
        <sound preset="alert" volume="0.75" pitch="1.5"/>
    </wait>
</action>
```

cswhite2000:
I definitely prefer
```xml
<schedule after="5s>
```
over
```xml
<wait duration=5s>
```

eliminate123:
I think schedule is fine

Pablete1234 (dismissed):

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -514,4 +517,20 @@ public PickupFlagAction parsePickupFlag(Element el, Class<?> scope) throws Inval
...
+  @SuppressWarnings("unchecked")
+  public <B extends Filterable<?>> Action<?> parseSchedule(Element el, Class<B> scope)
+      throws InvalidXMLException {
+    scope = parseScope(el, scope);
+    var action = parseAction(el, scope);
+    var duration = parser.duration(el, "duration").between(WAIT_RANGE).required();
+
+    return MatchPlayer.class.isAssignableFrom(scope)
+        ? new ScheduleAction.Player(duration, (Action<? super MatchPlayer>) action)
+        : new ScheduleAction<>(scope, duration, action);
```
```suggestion
    var after = parser.duration(el, "after").between(WAIT_RANGE).required();

    return MatchPlayer.class.isAssignableFrom(scope)
        ? new ScheduleAction.Player(after, (Action<? super MatchPlayer>) action)
        : new ScheduleAction<>(scope, after, action);
```

core/src/main/java/tc/oc/pgm/action/actions/ScheduleAction.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public void trigger(MatchPlayer mp) {
+      var party = mp.getParty();
+      schedule(mp, () -> {
+        if (mp.getParty() == party) child.trigger(mp);
+      });
+    }
+  }
```
Pablete1234:
```suggestion
  protected final Duration after;
  protected final Action<? super B> child;

  public ScheduleAction(Class<B> scope, Duration after, Action<? super B> child) {
    super(scope);
    this.after = after;
    this.child = child;
  }

  @Override
  public void trigger(B t) {
    schedule(t, () -> child.trigger(t));
  }

  protected void schedule(B t, Runnable r) {
    t.getMatch()
        .getExecutor(MatchScope.RUNNING)
        .schedule(r, after.toMillis(), TimeUnit.MILLISECONDS);
  }

  /** Specialization for running with a player, requires that the player did not change teams. */
  public static class Player extends ScheduleAction<MatchPlayer> {
    public Player(Duration after, Action<? super MatchPlayer> child) {
      super(MatchPlayer.class, after, child);
    }

    @Override
    public void trigger(MatchPlayer mp) {
      var party = mp.getParty();
      schedule(mp, () -> {
        if (mp.getParty() == party) child.trigger(mp);
      });
    }
  }
```

Pablete1234 (approved):

cswhite2000 (approved):

PR #1630, opened 2026-02-23
Implement `cleanupPlayer` on modern

calcastor:


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -417,7 +418,8 @@ public void cleanupWorld(World world) {
 
   @Override
   public void cleanupPlayer(Player player) {
-    // no-op
+    var nmsPlayer = ((CraftPlayer) player).getHandle();
+    nmsPlayer.getBukkitEntity().setKiller(null);
```
TTtie: would be better to act on the player object directly IMO
```suggestion
    player.setKiller(null);
```

TTtie (changes requested):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
calcastor: Lol oops

Pablete1234 (approved):

PR #1631, opened 2026-02-23
Implement component application via item mods

calcastor:


Pablete1234 (changes requested):
You're parsing every time an item needs to modify, that isnt acceptable.
Parsing needs to happen ahead of time, you cant hold references to Node(s) after parse time because that holds the whole document and prevents xml being unloaded after pgm is done parsing.
You basically need a concept of "item component applicator" of sorts, that gets build at parse time, then called with the item at runtime.

util/src/main/java/tc/oc/pgm/util/inventory/InventoryUtils.java
```diff
@@ -198,6 +198,10 @@ default void setUnbreakable(ItemStack item, boolean unbreakable) {
 
     boolean isViewable(Inventory inventory);
 
+    default void applyComponents(ItemStack itemStack, String components) {
+      // Default no-op for legacy
+    }
+
     default void applyComponents(ItemStack itemStack, Node components) throws InvalidXMLException {
       // Default no-op for legacy
     }
```
Pablete1234: These should be the same method, just build a component applicator:

```
    default ComponentApplicator applyComponents(Material type, Node components) throws InvalidXMLException {
      return itemStack -> {};
    } 
```

building the applicator can throw an invalid xml exception and it will be applied to the node

core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
```diff
@@ -52,10 +55,18 @@ public static class Factory implements MapModuleFactory<ItemModifyModule> {
...
+        Element elModify = XMLUtils.getRequiredUniqueChild(elRule, "modify");
         PotionMeta meta = (PotionMeta) Bukkit.getItemFactory().getItemMeta(Material.POTION);
-        factory.getKits().parseItemMeta(XMLUtils.getRequiredUniqueChild(elRule, "modify"), meta);
+        factory.getKits().parseItemMeta(elModify, meta);
+
+        Node components = Node.fromChildOrAttr(elModify, "components");
+        ComponentApplicator applicator = null;
+        if (components != null) {
+          String componentsStr = components.getValueNormalize();
+          applicator = stack -> INVENTORY_UTILS.applyComponents(stack, componentsStr);
```
Pablete1234: This "applicator" is still parsing the component every time.

This should be built inside applyComponents (should rename), see followup comment

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/inventory/ModernInventoryUtil.java
```diff
@@ -109,16 +109,24 @@ public boolean isViewable(Inventory inventory) {
...
+    }
+  }
+
+  public void applyComponents(ItemStack itemStack, Node components) throws InvalidXMLException {
+    try {
+      applyComponents(itemStack, components.getValueNormalize());
+    } catch (RuntimeException e) {
+      throw new InvalidXMLException("Failed to parse components", components, e.getCause());
     }
   }
```
Pablete1234, line 127: These need to do the heavy lifting, should be just 1 function that returns the applicator:

Something a bit like:
```java
    var key = ITEMS.getKey(nmsStack.getItem());
    if (key == null) throw new IllegalStateException("Invalid item stack: " + nmsStack);

    var str = new StringReader(key.toShortString() + "[" + components.getValueNormalize() + "]");
    var components = ITEM_PARSER.parse(str).components();
    return is -> CraftItemStack.unwrap(is).applyComponents(components);
```

The issue is you need to get a `Material` for the item to apply, which for itemmods (because it can apply to diff items) you'll need to probably have some way of getting the first of them, and suffer any potential consequences that may have.

Pablete1234 (changes requested):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/inventory/ModernInventoryUtil.java
```diff
@@ -109,14 +110,14 @@ public boolean isViewable(Inventory inventory) {
...
+  public ComponentApplicator buildComponentApplicator(Node components) throws InvalidXMLException {
     try {
-      var key = ITEMS.getKey(nmsStack.getItem());
-      if (key == null) throw new IllegalStateException("Invalid item stack: " + nmsStack);
-
-      var str = new StringReader(key.toShortString() + "[" + components.getValueNormalize() + "]");
-      nmsStack.applyComponents(ITEM_PARSER.parse(str).components());
+      // We don't need the actual material we're applying the component to, just something to parse
+      // the components against
+      var str = new StringReader("minecraft:stone[" + components.getValueNormalize() + "]");
```
Pablete1234: This is not true, as the components that are allowed depend on the material; some components are not combinable (eg: durability and max count).

So if we always use stone (which by default has a max_count of 64), it means durability won't be possible to set (it will throw a parsing exception, because it tries to make a patch ontop of stone's base components).

For the purposes of itemmods, you need to extract a material from the material matcher and use that as your representative "sample". Its not perfect, but if you're itemmoding blocks, its likely they're all blocks, and if youre itemmoding swords or tools, its likely they're all tools, and so these kind of conflicts are less likely to happen

Pablete1234 (approved):

PR #1633, opened 2026-02-26
Ignore armour in `carrying` and fix-up item matching

calcastor:
1. Ignore armour in `carrying`: While PGM didn't promise this before, it's how it behaved in legacy by default and I think moving away from such an expectation would warrant a proto bump.
2. Fix-up ItemMatcher
3. ~~Ignore `custom-meta-applied` when checking if items are similar Apply item mods in KitParser; this makes comparisons in `Materials.itemsSimilar` accurate when comparing against kit-given items.~~

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/matcher/player/CarryingItemFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/matcher/player/CarryingItemFilter.java
@@ -40,7 +40,9 @@ public class CarryingItemFilter extends ParticipantItemFilter {
 
   @Override
   protected Stream<ItemStack> getItems(MatchPlayer player) {
-    Stream<ItemStack> inventory = Slot.Player.player().map(s -> s.getItem(player));
+    Stream<ItemStack> inventory = Stream.concat(
+            Slot.Storage.storage(), Slot.OffHand.offHand().stream())
+        .map(s -> s.getItem(player));
 
     // Potentially add the crafting grid if that's the currently open inventory
     InventoryView invView = player.getBukkit().getOpenInventory();
--- a/util/src/main/java/tc/oc/pgm/util/inventory/ItemMatcher.java
+++ b/util/src/main/java/tc/oc/pgm/util/inventory/ItemMatcher.java
@@ -42,15 +42,24 @@ public class ItemMatcher {
       // Strip all meta, then re-add if needed
       newItem.setItemMeta(null);
 
-      // Restore name or enchants
-      if (!ignoreName) newItem.getItemMeta().setDisplayName(item.getItemMeta().getDisplayName());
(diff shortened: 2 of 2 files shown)


util/src/main/java/tc/oc/pgm/util/material/Materials.java
```diff
@@ -125,6 +126,18 @@ static boolean itemsSimilar(ItemStack first, ItemStack second, boolean skipDur)
         || (!skipDur && first.getDurability() != second.getDurability())) {
       return false;
     }
+
+    // Ignore custom-meta-applied tag
+    ItemTag<Boolean> APPLIED = ItemTag.newBoolean("custom-meta-applied");
+    if (APPLIED.has(first)) {
+      first = first.clone();
```
calcastor: This isn't great; it's a band-aid fix and will break with https://github.com/PGMDev/PGM/pull/1633

calcastor: Just dropping it

core/src/main/java/tc/oc/pgm/filters/matcher/player/CarryingItemFilter.java
```diff
@@ -40,7 +40,8 @@ public Collection<Class<? extends Event>> getRelevantEvents() {
 
   @Override
   protected Stream<ItemStack> getItems(MatchPlayer player) {
-    Stream<ItemStack> inventory = Slot.Player.player().map(s -> s.getItem(player));
+    Stream<ItemStack> inventory =
+        Slot.Player.player().filter(s -> !(s instanceof Slot.Armor)).map(s -> s.getItem(player));
```
Pablete1234: Instead of filtering the stream just use `Stream.concat(Slot.Storage.storage(), Slot.OffHand.offhand().stream())` to create a stream with just what you want

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -463,6 +464,9 @@ public ItemMatcher parseItemMatcher(Element parent, String childName) throws Inv
     if (stack == null)
       throw new InvalidXMLException("Child " + childName + " element expected", parent);
 
+    ItemModifyModule imm = factory.getModule(ItemModifyModule.class);
```
Pablete1234: I dont believe this will be effective; any module can request parsing an item matcher, and unless they've specified a dependency on itemmods theres no order guarantee, so this may simply return null even if item mods are present, just because they loaded later.

Its the reason item parsing itself doesnt apply the itemmods directly.

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #1634, opened 2026-03-01
Implement slot groups for carrying, replace-item and enchant-item

Pablete1234:
TLDR of the changes:
 - Added the concept of slot groups
 - `slots` optional attribute for `carrying`, `holding` and `wearing` filters
 - `slots` optional attribute for `replace-item` and `enchant-item` actions.
 - `total-amount` optional xml attribute for `carrying`, `holding` and `wearing` filters, if used, pgm matches your whole inventory instead of per-itemstack

# Slot groups

Slot groups are here! what are slot groups? they're a way for xml to define a set of slots something should work in, these make sense in the context of carrying filter, or replace-item actions. Let's see an example:

```xml
<carrying slots="hotbar">
  <item material="diamond"/>
</carrying>
```
While previously carrying would include everything in your inventory (except armor), this would specify that you only are going to look at hotbar slots.

This can go further, and check individual slots, or a combination of slots and groups too:
```xml
<carrying slots="hotbar.0, hotbar.1, horbar.2">...</carrying>
<carrying slots="armor.chest,hands">...</carrying>
<carrying slots="hands,armor">...</carrying>
```

Slot groups are a comma separated list of any of the following:
 - `all` every single slot in a players' inventory. Includes `storage`, `offhand`, `armor`, `cursor` and `crafting`
 - `carrying` everything except armor. Includes `storage`, `offhand`, `cursor`, and `crafting`
 - `storage` main inventory space. Includes `hotbar`, and `pockets`
   -  `hotbar` slots 0-8, the slots you have always visible in your hotbar
   -  `pockets` slots 9-36, the rows inside your inventory
 - `equipment` additional equipment slots. Includes `offhand` and `armor`, does NOT include `mainhand`.
   - `armor` the 4 armor pieces slots  
   - `hands` both hands, Includes `mainhand` and `offhand`
     - `mainhand` the currently selected item in your hotbar, also aliased as `hand`.
       - Be aware: it is incompatible to use `mainhand` and any `hotbar` slot together, as they overlap.
     - `offhand` the offhand slot, in legacy this is just an always empty slot
- `cursor` represents the item currently held at your cursor; ie: while having your inventory open, if you click once on an item it will then is in your cursor slot.
- `crafting` represents the crafting slots (2x2 or 3x3 crafting grid)
- Any slot definition is also allowed (eg: `hotbar.0`, `0`, `armor.chest`) 

Note for `cursor` and `crafting`: these exist to avoid exploits where a user can effectively "hide" an item from your carrying filter by just picking it up and holding it in their cursor, or putting it in the crafting grid.

The `holding`, `wearing` and `carrying` filters are now the exact same, just using a different slotgroup by default (`hands`, `armor` and `carrying`, respectively).

# Matching across slots, instead of stack by stack

It was often the case that when a mapmaker wants to see if you're carrying at least 10 of something, just because you split it into 2 stacks of 5 doesn't mean you're no longer carrying that. Historically the filters have been very item-stack based, that meaning, they check if any individual slot matches the item in the filter, including the amount, not if the sum of all items does. Now mapdevs have the option of picking either behavior.

Note, both `amount` and `total-amount` are incompatible, you must pick one.
```xml
<carrying total-amount="5.."/> <!-- checks if you have 5 or more arrows in total -->
  <item material="arrow"/>
</carrying>
<carrying amount="5.."/> <!-- checks if you have any stack with 5 or more arrows -->
  <item material="arrow"/>
</carrying>
```


util/src/main/java/tc/oc/pgm/util/inventory/InventoryUtils.java
```diff
@@ -216,6 +216,11 @@ default void setUnbreakable(ItemStack item, boolean unbreakable) {
 
     boolean isViewable(Inventory inventory);
 
+    default Collection<Class<? extends Event>> getRelevantEvents(SlotGroup group) {
+      // Legacy needs to additional events
```
cswhite2000:
```suggestion
      // Legacy needs no additional events
```

util/src/main/java/tc/oc/pgm/util/inventory/SlotGroup.java
```diff
@@ -0,0 +1,142 @@
...
+  public static final SlotGroup EQUIPMENT = register("equipment", Slot.Equipment.equipment());
+  public static final SlotGroup OFFHAND = register("offhand", Slot.OffHand.offHand().stream());
+  public static final SlotGroup ARMOR = register("armor", Slot.Armor.armor());
+
+  public static final SlotGroup HANDS = register("hands", Slot.Equipment.hands());
+  public static final SlotGroup MAINHAND =
+      register("mainhand", Stream.of(Slot.MainHand.mainHand()));
+
+  public static final SlotGroup CURSOR = register("cursor", Stream.of(Slot.Cursor.cursor()));
+  public static final SlotGroup CRAFTING = register("crafting", Slot.Crafting.crafting());
```
cswhite2000, line 46: If we're adding this for the 3x3 crafting grid, should we be adding this for every other block inventory that doesn't store items?

for example:
```
EnchantingInventory
AnvilInventory
GrindstoneInventory
LoomInventory
CartographyInventory
StonecutterInventory
SmithingInventory
```

cswhite2000 (dismissed):
Overall looks good 👍

util/src/main/java/tc/oc/pgm/util/inventory/SlotGroup.java
Pablete1234: I only did it because it was preexising behavior, and its very common. I dont know that doing it for all those others is necessary

cswhite2000: probably not, but if the point is to avoid hiding items then the same logic applies. I'll leave it up to you

cswhite2000 (dismissed):

cswhite2000 (approved):

PR #1637, opened 2026-03-08
fix: clamp excessive velocity settings

TTtie:
Clamps excessive velocity settings to avoid excessive overshooting on modern versions as 1.8 clamps any entity velocity over the absolute value of 3.9 per axis.

For the time being, the following are affected on legacy maps:
- Velocity applications (PGM will issue a warning if such a velocity is detected)
- Velocity actions (PGM will clamp the value at runtime. PGM will also try to detect constant expressions that set the value way too high and issue a warning if so.)
- Double jump with custom power settings (double jump power over 15 triggers a warning, which still fits within the limits -- the maximum I've encountered on OCC is 10)

Ref: https://github.com/PGMDev/Website/issues/307
Ref: https://github.com/OvercastCommunity/PublicMaps/commit/49890320390fce9893c24dd5db17d43d00ed988d


core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -467,13 +468,41 @@ public Action<? super MatchPlayer> parseTakePayment(Element el, Class<?> scope)
...
+      velocity = formula.applyAsDouble(null);
+    } catch (Throwable e) {
+      // If trying to evaluate a formula results in an exception, it likely depends on player/match
+      // data, and thus we can't determine if it is excessive or not at parse time
+      velocity = 0;
+    }
+
+    if (Math.abs(velocity) > 3.9) {
+      LoggingUtils.warn(
+          factory.getLogger(),
```
Pablete1234: feels like maybe `factory` should be the one to allow logging warnings for maps?

```suggestion
      factory.warn("bad things happening", node);
```

TTtie: Moved the warning method under the map factory.

Not quite sure if I want to add a warning callback to the [material adapter](https://github.com/PGMDev/PGM/blob/5051474cb33d3a432547fcb4179c558c471c0046/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/material/ModernMaterialUtils.java#L179-L181) which uses the same pattern but with the global game logger, or leave it up to the technical debt overlords, however…

Pablete1234 (approved):

Issue #1041, opened 2022-08-06
Implement use & break filters in maxbuildheight


Pablete1234:
Currently max build height prevents block placing but does not prevent breaking, or using (eg: placing boats). It would be nice to support such things in a similar way to <apply>:
```xml
  <maxbuildheight place="never" break="always" use="always">32</maxbuildheight> <!-- defaults -->
  <maxbuildheight use="never">32</maxbuildheight> <!-- deny use above height limit (opening chests or placing boats) -->
```

PR #1638, opened 2026-03-08
feat: filters to control place/block/use on maxbuildheight

kashike:
#1041

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/matcher/block/MaxBuildFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/matcher/block/MaxBuildFilter.java
@@ -1,16 +1,25 @@
 package tc.oc.pgm.filters.matcher.block;
 
+import tc.oc.pgm.api.filter.Filter;
 import tc.oc.pgm.api.filter.query.BlockQuery;
 import tc.oc.pgm.api.filter.query.Query;
+import tc.oc.pgm.filters.matcher.StaticFilter;
 import tc.oc.pgm.filters.matcher.WeakTypedFilter;
+import tc.oc.pgm.filters.operator.SingleFilterFunction;
 import tc.oc.pgm.regions.RegionMatchModule;
 
 /** Matches blocks that are above the max building height */
-public class MaxBuildFilter implements WeakTypedFilter<BlockQuery> {
+public class MaxBuildFilter extends SingleFilterFunction implements WeakTypedFilter<BlockQuery> {
+  private static final MaxBuildFilter DENY = new MaxBuildFilter(StaticFilter.DENY);
 
-  public static final MaxBuildFilter INSTANCE = new MaxBuildFilter();
+  public static MaxBuildFilter of(Filter filter) {
+    if (filter == StaticFilter.DENY) return DENY;
+    return new MaxBuildFilter(filter);
+  }
 
-  private MaxBuildFilter() {}
+  private MaxBuildFilter(final Filter filter) {
+    super(filter);
+  }
 
   @Override
   public Class<? extends BlockQuery> queryType() {
@@ -30,12 +39,12 @@ public class MaxBuildFilter implements WeakTypedFilter<BlockQuery> {
             .map(RegionMatchModule::getMaxBuildHeight)
             .filter(maxBuild -> query.getBlock().getY() >= maxBuild)
             .isPresent()
-        ? QueryResponse.DENY
+        ? filter.query(query)
         : QueryResponse.ABSTAIN;
   }
 
   @Override
   public String toString() {
-    return "MaxBuildFilter{}";
+    return "MaxBuildFilter{filter=" + filter + "}";
   }
 }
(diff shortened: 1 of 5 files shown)


core/src/main/java/tc/oc/pgm/regions/RegionFilterApplicationParser.java
```diff
@@ -83,18 +84,48 @@ public void parseLane(Element el) throws InvalidXMLException {
             RFAScope.BLOCK_PLACE, new NegativeRegion(region), filter, message, false));
   }
 
-  public Integer parseMaxBuildHeight(Element el) throws InvalidXMLException {
+  // <maxbuildheight place="never" break="always" use="always">38</maxbuildheight>
```
Pablete1234:
```suggestion
```
leftover comment

core/src/main/java/tc/oc/pgm/regions/RegionFilterApplicationParser.java
```diff
@@ -83,18 +84,48 @@ public void parseLane(Element el) throws InvalidXMLException {
...
+    Filter placeFilter = StaticFilter.DENY;
+    Filter breakFilter = StaticFilter.ABSTAIN;
+    Filter useFilter = null;
+    Integer y = null;
+    if (el != null) {
+      placeFilter = filterParser.parseProperty(el, "place", placeFilter);
+      breakFilter = filterParser.parseProperty(el, "break", breakFilter);
+      useFilter = filterParser.parseProperty(el, "use", useFilter);
+      y = XMLUtils.parseNumber(el, Integer.class);
+    }
```
Pablete1234: These could be merged together given defaults, as null will just use that default:
```suggestion
    Filter placeFilter = filterParser.parseProperty(el, "place", StaticFilter.DENY);
    Filter breakFilter = filterParser.parseProperty(el, "break", StaticFilter.ABSTAIN);
    Filter useFilter = filterParser.parseProperty(el, "use", StaticFilter.ABSTAIN);
```

core/src/main/java/tc/oc/pgm/regions/RegionFilterApplicationParser.java
```diff
@@ -83,18 +84,48 @@ public void parseLane(Element el) throws InvalidXMLException {
...
+    Component message = translatable("match.maxBuildHeight");
```
Pablete1234: could be moved to a constant outside, but up to you

core/src/main/java/tc/oc/pgm/regions/RegionFilterApplicationParser.java
```diff
@@ -83,18 +84,48 @@ public void parseLane(Element el) throws InvalidXMLException {
...
+    if (useFilter != null) {
+      prepend(
+          el,
+          new RegionFilterApplication(
+              RFAScope.USE,
+              EverywhereRegion.INSTANCE,
+              new MaxBuildFilter(useFilter),
+              message,
+              false));
+    }
```
Pablete1234, line 109: all 3 of these can now simplify to "if filter is not abstain":
```java
    if (xxxFilter != StaticFilter.ABSTAIN) {
      prepend(
          el,
          new RegionFilterApplication(
              RFAScope.XXX,
              EverywhereRegion.INSTANCE,
              new MaxBuildFilter(xxxFilter),
              message,
              false));
    }
```
I also wonder if it makes sense for `MaxBuildFilter` to have a constant for when the filter is explicitly deny, as that's pretty common.

Could do with a `MaxBuildFilter.of(xxxFilter)` style method and make that check if the xxxFilter is deny, return the constant (what preexisted as `MaxBuildFilter.INSTANCE`) otherwise return a new one for the specific filter

core/src/main/java/tc/oc/pgm/regions/RegionFilterApplicationParser.java
```diff
@@ -83,18 +84,48 @@ public void parseLane(Element el) throws InvalidXMLException {
...
 
-    return el == null ? null : XMLUtils.parseNumber(el, Integer.class);
+    return y;
```
Pablete1234:
```suggestion
    return XMLUtils.parseNumber(el, Integer.class, null);
```
can just use a defaulted parsing method here too

core/src/main/java/tc/oc/pgm/filters/matcher/block/MaxBuildFilter.java
```diff
@@ -1,16 +1,17 @@
...
 /** Matches blocks that are above the max building height */
-public class MaxBuildFilter implements WeakTypedFilter<BlockQuery> {
-
-  public static final MaxBuildFilter INSTANCE = new MaxBuildFilter();
-
-  private MaxBuildFilter() {}
+public class MaxBuildFilter extends SingleFilterFunction implements WeakTypedFilter<BlockQuery> {
+  public MaxBuildFilter(final Filter filter) {
+    super(filter);
+  }
```
Pablete1234: Given the rest of the comments, this would become a private constructor, the way to create it would be thru the static method, similar to how AllFilter etc do it so that they can apply these type of savings

Pablete1234 (changes requested):

Pablete1234 (approved):

PR #1642, opened 2026-03-12
Support material matchers in item matchers

Pablete1234:
Currently item matchers are locked to matching a single material type, with this, they are allowed to match any arbitrary amount of types, via a material matcher:
```xml
<wearing id="wearing-protection">
  <item enchantment="protection:1">
    <match>
      <all-items/>
    </match>
  </item>
</wearing>

<holding id="holding-block">
  <item amount="5" name="Block!">
    <match>
      <material>stone</material>
      <material>dirt</material>
    </match>
  </item>
</holding>
```

The main change is that the `item` no longer gets a `material="stone"` attribute, and instead gets a `<match>` child which parses as a material matcher, similar to how itemmods' `match`, items' `can-place-on` work.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
+++ b/core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
@@ -52,13 +52,15 @@ public class ItemModifyModule implements MapModule<ItemModifyMatchModule> {
 
       List<ItemRule> rules = new ArrayList<>();
       for (Element el : XMLUtils.flattenElements(doc.getRootElement(), "item-mods", "rule")) {
-        var items = parser.node(XMLUtils::parseMaterialMatcher, el, "match")
+        var items = parser
+            .node(XMLUtils::parseMaterialMatcher, el, "match")
             .child()
             .validate(MaterialMatcher.NOT_EMPTY)
             .required();
 
         var elModify = XMLUtils.getRequiredUniqueChild(el, "modify");
-        var applicator = factory.getKits().parseItemMeta(items.getRepresentativeMaterial(), elModify, true);
+        var applicator =
+            factory.getKits().parseItemMeta(items.getRepresentativeMaterial(), elModify, true);
 
         ItemRule rule = new ItemRule(items, applicator);
         rules.add(rule);
--- a/core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
+++ b/core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
@@ -59,8 +59,7 @@ public class ItemModifyModule implements MapModule<ItemModifyMatchModule> {
             .required();
 
         var elModify = XMLUtils.getRequiredUniqueChild(el, "modify");
-        var applicator =
-            factory.getKits().parseItemMeta(items.getRepresentativeMaterial(), elModify, true);
+        var applicator = factory.getKits().parseItemMeta(items.getSample(), elModify, true);
 
         ItemRule rule = new ItemRule(items, applicator);
         rules.add(rule);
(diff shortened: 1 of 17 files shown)


util/src/main/java/tc/oc/pgm/util/material/matcher/AllMaterialMatcher.java
```diff
@@ -35,6 +35,11 @@ public Set<Material> getMaterials() {
     return EnumSet.allOf(Material.class);
   }
 
+  @Override
+  public Material getRepresentativeMaterial() {
+    return Material.STONE;
```
cswhite2000, line 40: I don't see any bugs with this current implementation, but I feel like this makes it really easy for someone else to mess this up if they touch this code.

core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
```diff
@@ -51,12 +52,13 @@ public static class Factory implements MapModuleFactory<ItemModifyModule> {
...
-            parser.node(XMLUtils::parseMaterialMatcher, el, "match").child().required();
-        var material = items.getMaterials().iterator().next();
+        var items = parser.node(XMLUtils::parseMaterialMatcher, el, "match")
+            .child()
+            .validate(MaterialMatcher.NOT_EMPTY)
+            .required();
 
         var elModify = XMLUtils.getRequiredUniqueChild(el, "modify");
-        var applicator = factory.getKits().parseItemMeta(material, elModify, true);
+        var applicator = factory.getKits().parseItemMeta(items.getRepresentativeMaterial(), elModify, true);
```
cswhite2000: When using `<all-materials/>` I think the old material chosen would have been air, whereas now it would be stone.
Although I don't think this will cause any issues, but its a little difficult to tell

cswhite2000 (commented):
formatter is failing

core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
Pablete1234: Using `air` to represent "all materials" imo is weird, because if i'm trying to check that you have "at least 5 of anything" i doubt you'd be ok with "oh yea, this player indeed has 5 air" on their inventory.

Regarding the implicit behavior change, this is only affecting itemmods that use item components, which has only been possible since very recently, so i doubt it can even create an issue. The only other thing that parses to allow something like `all-materials` is can-break/can-place-on, which won't get queried for representative materials. The other queries for representative materials are only done on things that are parsed as material matchers but really aren't allowed to use wildcards anyways

cswhite2000: Makes sense. And yeah, I had come to the conclusion that this would work with all the current code paths.

I guess my point is moreso that `getRepresentativeMaterial` is a bit misleading for some of the MaterialMatcher implementations if you just read the api, without looking at the usage. Previously, that behaviour was just a bit more hidden.

But yeah, should be fine given you're a codeowner and can block changes if someone else messes with this.

cswhite2000 (dismissed):

core/src/main/java/tc/oc/pgm/itemmeta/ItemModifyModule.java
Pablete1234: I'm open to alternative names, but `getMaterial` seemed a bit too vague given a matcher could be N materials, which is why i chose this name

cswhite2000: maybe `getSampleMaterial`?

Pablete1234: updated to `getSample`

cswhite2000 (approved):

eliminate123:
I've tested this PR and it works. Here's an example of what this xml would look like in a map:

```xml
<filters>
    <carrying id="sword-in-cursor" ignore-metadata="true" slots="cursor">
        <item>
            <match>
                <material>wood sword</material>
                <material>stone sword</material>
                <material>gold sword</material>
                <material>iron sword</material>
            </match>
        </item>
    </carrying>
</filters>
<actions>
    <trigger filter="sword-in-cursor" scope="player">
        <action>
            <enchant-item enchantment="sharpness" ignore-metadata="true" level="1" slots="cursor">  <!-- Apply enchantment to sword in cursor slot.  If using holding filter slots="cursor" is unnecessary-->
                <find>
                    <match>
                        <material>wood sword</material>
                        <material>stone sword</material>
                        <material>gold sword</material>
                        <material>iron sword</material>
                    </match>
                </find>
            </enchant-item>
        </action>
    </trigger>
</actions>
```

PR #1643, opened 2026-03-12
fix: stereo global sounds on 1.14+

imkunet:
With the inclusion of https://github.com/PGMDev/PGM/pull/1635, which introduced "global sounds" (sounds with more desirable characteristics for what they were being used for) based on strategically playing loud sounds far away from the player comes a bug: [MC-146721](https://bugs.mojang.com/browse/MC/issues/MC-146721). This manifests in the form of things like quiet [portal sounds](https://github.com/PGMDev/PGM/pull/1635#issuecomment-4035905537) playing really loudly on modern game versions.

This pull request remedies the aforementioned issue. To do so, it takes a list of all stereo sounds from the game version `1.21.11` (which also has their `1.8` equivalents as keys) and if you are on `1.14`+, checks if a global sound playing is in that list. If those conditions are met, it will try to then play the sound as best as it can:
- If you are unaffected by [MC-138832](https://bugs.mojang.com/browse/MC/issues/MC-138832) i.e. outside of the range of `1.15.2`-`1.16.5`, it will fall back to playing the sound at your feet (the old way) since there is no better way to play the sound. I've gotten it to happen as low as `1.14.4`, though so I suspect there is some packet translation error in a Via plugin or the bug has reached that far back in versions, so if your game client is at or below `1.16.5` and is at or above `1.14`, it will fall back to the adventure old way.
- If the server is on a modern platform, [adventure-platform](https://github.com/PaperMC/adventure-platform) supports attaching the sound to the target player (there is no legacy compatibility). The downside to this approach though is that ears are interpolated on the client and update position every frame, while the sound emitter is at the uninterpolated position of the player, so moving can cause the sound to be heard more in one ear than the other.
- Otherwise, it will fall back to the most legacy way of playing the sound, which is to just play it at the player's feet (the old way).

A side effect of fixing it at the Audience is that the fix also comes for free to default PGM sounds. Nice!

## Validation
Performed by listening for expected behavior, and in development by using "printf debugging". Platforms have ViaVersion `5.7.2`, ViaBackwards `5.7.2`, and ViaRewind `4.0.15`. The sound tested was `portal.travel` (on 1.8.9) or `block.portal.travel` (on modern). Remember, the below table only applies to problematic stereo sounds!

|Server Version|Client Version|Notes|
|-|-|-|
|`1.21.10`|`1.21.11`|✅ Falls back to adventure emitters|
|`1.21.10`|`1.16.5`|✅ Falls back to adventure|
|`1.21.10`|`1.8.9`|✅ Uses the omnipresent method|
|||
|`1.8.9`|`1.21.11`|✅ Falls back to adventure emitters|
|`1.8.9`|`1.16.5`|✅ Falls back to adventure|
|`1.8.9`|`1.8.9`|✅ Uses the omnipresent method|
|||
|`1.21.10`|`1.14.4`|✅ Falls back to adventure (@arcadeboss' case tested specifically)|
|`1.8.9`|`1.14.4`|✅ Falls back to adventure|

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/Audience.java
+++ b/util/src/main/java/tc/oc/pgm/util/Audience.java
@@ -21,7 +21,6 @@ import org.bukkit.command.CommandSender;
 import org.jetbrains.annotations.NotNull;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 import tc.oc.pgm.util.bukkit.Sounds;
-import tc.oc.pgm.util.bukkit.ViaUtils;
 import tc.oc.pgm.util.text.ComponentRenderer;
 
 /** Receiver of chat messages, sounds, titles, and other media. */
@@ -64,9 +63,7 @@ public interface Audience extends ForwardingAudience.Single {
     if (global) {
       if (player.isEmpty()) break omnipresent;
       // account for MC-146721 only on affected clients for affected sounds
-      var version = ViaUtils.getProtocolVersion(player.get());
-      if (Sounds.MODERN_GLOBAL_SOUNDS.contains(sound.name().value())
-          && version >= ViaUtils.VERSION_1_14) break omnipresent;
+      if (Sounds.MODERN_GLOBAL_SOUNDS.contains(sound.name().value())) break omnipresent;
       var location = player.get().getEyeLocation();
       var realVolume =
           soundDistance / (16f * (1f - Math.max(0f, Math.min(maxVolume, sound.volume()))));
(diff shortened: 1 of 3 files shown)


imkunet:
I've been instructed to not rely on server version or do all the fancy stuff to account for the weird differences in  certain versions. The behavior has been simplified:
- If your client is on `1.14+` and the sound that plays is a known stereo sound, it will play using the old adventure fallback method.

I have not gone in to validate it (since my previous validation steps demonstrate that this should work), but the behavior should look like for buggy stereo sounds:

|Server Version|Client Version|Notes|
|-|-|-|
|`1.21.10`|`1.21.11`|Falls back to adventure|
|`1.21.10`|`1.16.5`|Falls back to adventure|
|`1.21.10`|`1.8.9`|Uses the omnipresent method|
|||
|`1.8.9`|`1.21.11`|Falls back to adventure|
|`1.8.9`|`1.16.5`|Falls back to adventure|
|`1.8.9`|`1.8.9`|Uses the omnipresent method|
|||
|`1.21.10`|`1.14.4`|Falls back to adventure|
|`1.8.9`|`1.14.4`|Falls back to adventure|

imkunet:
I've been instructed to remove all version checking, so now the code will just check to see if the sound is in the stereo list and then play it through adventure's method if it is.

Pablete1234 (approved):

PR #1645, opened 2026-03-20
Match modern material names on legacy

calcastor:


Pablete1234 (changes requested):
We've been down the path of xseries in the past and it made the jar bigger and things harder on PGM. We opted to have our own abstractions because they suited us better, and is the reason we have pgm's BlockMaterialData/ItemMaterialData and MaterialMatcher with per-platform implementations.

I suspect the reason for this change is for legacy server support of post-flattening names (ie: gray_carpet instead of carpet:7), and if that's the case i think i'd rather maintain our own lightweight remapping for the cases that are relevant, like `color_material` remapping to `material:color`.

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,371 @@
...
+    put("SKELETON_SKULL", 0, "SKULL_ITEM");
+    put("SKELETON_WALL_SKULL", 0, "SKULL_ITEM");
+    put("WITHER_SKELETON_SKULL", 1, "SKULL_ITEM");
+    put("WITHER_SKELETON_WALL_SKULL", 1, "SKULL_ITEM");
+    put("ZOMBIE_HEAD", 2, "SKULL_ITEM");
+    put("ZOMBIE_WALL_HEAD", 2, "SKULL_ITEM");
+    put("PLAYER_HEAD", 3, "SKULL_ITEM");
+    put("PLAYER_WALL_HEAD", 3, "SKULL_ITEM");
+    put("CREEPER_HEAD", 4, "SKULL_ITEM");
+    put("CREEPER_WALL_HEAD", 4, "SKULL_ITEM");
```
Pablete1234: In legacy we have `SKULL` and `SKULL_ITEM`, if you're matching blocks in the world you probably want skull, if you're trying to give as part of a kit you want skull item, i have my doubts on how that all will work with this, like if you deny breaking `CREEPER_HEAD` it will actually do nothing cause the item isnt the block

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,371 @@
...
+    put("OAK_STAIRS", 0, "WOOD_STAIRS");
+    put("OAK_BOAT", 0, "BOAT");
+    put("OAK_DOOR", 0, "WOOD_DOOR");
+    put("OAK_FENCE", 0, "FENCE");
+    put("OAK_FENCE_GATE", 0, "FENCE_GATE");
+    put("OAK_BUTTON", 0, "WOOD_BUTTON");
+    put("OAK_PRESSURE_PLATE", 0, "WOOD_PLATE");
+    put("OAK_TRAPDOOR", 0, "TRAP_DOOR");
+    put("OAK_SIGN", 0, "SIGN");
+    put("OAK_WALL_SIGN", 0, "WALL_SIGN");
```
Pablete1234: These feel like they can be compacted too with a loop that goes thru the diff wood types just like colors

You can `i % 4 ` the metadata for *_LOG and *_LEAVES, and use `12 + (i % 4)` for _WOOD (as that's the all-sided-wood), as well as appending the "_2" as long as i >= 4:

```java
private static final String[] WOODS = {
    "OAK", "SPRUCE", "BIRCH", "JUNGLE", "ACACIA", "DARK_OAK"
  };
```

```suggestion
    for (int i = 0; i < WOODS.length; i++) {
      var type = WOODS[i];
      
      var suffix = i >= 4 ? "_2" : "";
      put(type + "_LOG", i % 4, "LOG" + suffix);
      put(type + "_WOOD", 12 + (i % 4), "LOG" + suffix);
      put(type + "_LEAVES", i % 4, "LEAVES" + suffix);
      
      put(type + "_PLANKS", i, "WOOD");
      put(type + "_SAPLING", i, "SAPLING");
      put(type + "_SLAB", i, "WOOD_STEP");
      
      if (i > 0) {
        put(type + "_STAIRS", 0, prefix + "_WOOD_STAIRS");
        put(type + "_DOOR", 0, prefix + "_DOOR_ITEM");
      }
      
      // Map them all to oak; other variants don't exist in 1.8
      put(type + "_BOAT", 0, "BOAT");
      put(type + "_BUTTON", 0, "WOOD_BUTTON");
      put(type + "_PRESSURE_PLATE", 0, "WOOD_PLATE");
      put(type + "_TRAPDOOR", 0, "TRAP_DOOR");
      put(type + "_SIGN", 0, "SIGN");
      put(type + "_WALL_SIGN", 0, "WALL_SIGN");
    }
    // Oak stuff that needs remapping to generic version    
    put("OAK_STAIRS", 0, "WOOD_STAIRS");
    put("OAK_DOOR", 0, "WOOD_DOOR");      
    put("OAK_FENCE", 0, "FENCE");
    put("OAK_FENCE_GATE", 0, "FENCE_GATE");
```

Keep in mind tho, this has a similar problem to the skull thing, remapping `OAK_DOOR`/`SPRUCE_DOOR` to `WOOD_DOOR`/`SPRUCE_DOOR_ITEM` (both of those are the item form) will not allow these to be used in block-based filters, as they would be searching for the wrong item.
If you're building your map for legacy, you'd have to know the distinction between item and block and use it appropriately, this takes that away from you.

Maybe you need a reverse-flattening like the 1.21 one did for material matchers (like, if you add `wool` it actually adds all 16 diff wool materials into the material matcher on modern), so that if you add `SPRUCE_DOOR` to a material matcher it actually adds both item and block form.

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,371 @@
...
+    put("VOID_AIR", 0, "AIR");
+    put("WALL_TORCH", 0, "TORCH");
+    put("WATER", 0, "STATIONARY_WATER");
+    put("WHEAT", 0, "CROPS");
+    put("WHEAT_SEEDS", 0, "SEEDS");
+  }
+
+  private static void put(String modern, int data, String legacyName) {
+    Material m = Material.getMaterial(legacyName);
+    if (m != null) {
```
Pablete1234:
```suggestion
    if (m == null) throw new IllegalArgumentException();
```
If m is null i'd throw an exception in this scenario, it means bad code referencing a non existing material.

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+      put(color + "_STAINED_GLASS", "STAINED_GLASS", i);
+      put(color + "_STAINED_GLASS_PANE", "STAINED_GLASS_PANE", i);
+      put(color + "_TERRACOTTA", "STAINED_CLAY", i);
+      // Black and white swapped for the below items
+      put(color + "_DYE", "INK_SACK", 15 - i);
+      put(color + "_BANNER", "BANNER", 15 - i);
+      put(color + "_WALL_BANNER", "WALL_BANNER", 15 - i);
+    }
+
+    put("INK_SAC", "INK_SACK"); // Black dye
```
Pablete1234:
```suggestion
    put("INK_SAC", "INK_SACK:0"); // Black dye
```
I believe modern's ink sac should only map to ink sack:0 (ie: black dye) in legacy, not to all dyes

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("DIAMOND_HORSE_ARMOR", "DIAMOND_BARDING");
+
+    // Stone variants
+    put("GRANITE", "STONE:1");
+    put("POLISHED_GRANITE", "STONE:2");
+    put("DIORITE", "STONE:3");
+    put("POLISHED_DIORITE", "STONE:4");
+    put("ANDESITE", "STONE:5");
+    put("POLISHED_ANDESITE", "STONE:6");
+    put("STONE_BRICKS", "SMOOTH_BRICK");
```
Pablete1234:
```suggestion
    put("STONE_BRICKS", "SMOOTH_BRICK:0");
```
Same rationale as dyes, `STONE_BRICKS` material in modern shouldn't include mossy/cracked/chiseled if used in a material matcher

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+
+    // Sand
+    put("RED_SAND", "SAND:1");
+
+    // Sandstone
+    put("CHISELED_SANDSTONE", "SANDSTONE:1");
+    put("SMOOTH_SANDSTONE", "SANDSTONE:2");
+    put("SANDSTONE_SLAB", "STEP:1");
+    put("CHISELED_RED_SANDSTONE", "RED_SANDSTONE:1");
+    put("RED_SANDSTONE_SLAB", "STONE_SLAB2");
```
Pablete1234:
```suggestion
    // Sandstone
    put("CHISELED_SANDSTONE", "SANDSTONE:1");
    put("SMOOTH_SANDSTONE", "SANDSTONE:2"); // Modern-only
    put("CUT_SANDSTONE", "SANDSTONE:2");
    put("CHISELED_RED_SANDSTONE", "RED_SANDSTONE:1");
    put("SMOOTH_RED_SANDSTONE", "RED_SANDSTONE:2"); // Modern-only
    put("CUT_RED_SANDSTONE", "RED_SANDSTONE:2");
```
What in 1.8 we knew as "smooth (red) sandstone" in modern is actually "cut (red) sandstone", 
Modern's "smooth sandstone" didn't really exist in 1.8, but i think its a sane remap to keep both.
Also i'd move the slabs together with the rest of slabs

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+
+    // Slabs
+    put("STONE_SLAB", "STEP");
+    put("COBBLESTONE_SLAB", "STEP:3");
+    put("BRICK_SLAB", "STEP:4");
+    put("NETHER_BRICK_SLAB", "STEP:6");
+    put("QUARTZ_SLAB", "STEP:7");
```
Pablete1234, line 222:
```suggestion
    // Slabs
    put("SMOOTH_STONE_SLAB", "STEP:0");
    put("SANDSTONE_SLAB", "STEP:1");
    put("PETRIFIED_OAK_SLAB", "STEP:2");
    put("COBBLESTONE_SLAB", "STEP:3");
    put("BRICK_SLAB", "STEP:4");
    put("STONE_BRICK_SLAB", "STEP:5");
    put("NETHER_BRICK_SLAB", "STEP:6");
    put("QUARTZ_SLAB", "STEP:7");
    put("RED_SANDSTONE_SLAB", "STEP2:0");

    // Modern-only slabs that reasonably remap backwards
    put("STONE_SLAB", "STEP:0"); // remap to smooth stone
    put("SMOOTH_SANDSTONE_SLAB", "STEP:1"); // remap to sandstone slab
    put("CUT_SANDSTONE_SLAB", "STEP:1"); // remap to sandstone slab
    put("SMOOTH_RED_SANDSTONE_SLAB", "STEP2:0"); // remap to red sandstone slab
    put("CUT_RED_SANDSTONE_SLAB", "STEP2:0"); // remap to red sandstone slab
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+
+    // Cobblestone walls
+    put("COBBLESTONE_WALL", "COBBLE_WALL");
+    put("MOSSY_COBBLESTONE_WALL", "COBBLE_WALL:1");
+
+    // Sponge
+    put("WET_SPONGE", "SPONGE:1");
+
+    // Infested blocks
+    put("INFESTED_STONE", "MONSTER_EGGS");
```
Pablete1234:
```suggestion
    put("INFESTED_STONE", "MONSTER_EGGS:0");
```
Similar rationale as in other cases, we don't want it to match different blocks

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("INFESTED_COBBLESTONE", "MONSTER_EGGS:1");
+    put("INFESTED_STONE_BRICKS", "MONSTER_EGGS:2");
+    put("INFESTED_MOSSY_STONE_BRICKS", "MONSTER_EGGS:3");
+    put("INFESTED_CRACKED_STONE_BRICKS", "MONSTER_EGGS:4");
+    put("INFESTED_CHISELED_STONE_BRICKS", "MONSTER_EGGS:5");
+
+    // Flowers
+    put("DANDELION", "YELLOW_FLOWER");
+    put("POPPY", "RED_ROSE");
```
Pablete1234:
```suggestion
    put("POPPY", "RED_ROSE:0");
```
poppy should refer only to the first red_rose state, same rationale as ink

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("ALLIUM", "RED_ROSE:2");
+    put("AZURE_BLUET", "RED_ROSE:3");
+    put("RED_TULIP", "RED_ROSE:4");
+    put("ORANGE_TULIP", "RED_ROSE:5");
+    put("WHITE_TULIP", "RED_ROSE:6");
+    put("PINK_TULIP", "RED_ROSE:7");
+    put("OXEYE_DAISY", "RED_ROSE:8");
+
+    // Double-tall plants
+    put("SUNFLOWER", "DOUBLE_PLANT");
```
Pablete1234:
```suggestion
    put("SUNFLOWER", "DOUBLE_PLANT:0");
```
same here

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("LILAC", "DOUBLE_PLANT:1");
+    put("TALL_GRASS", "DOUBLE_PLANT:2");
+    put("LARGE_FERN", "DOUBLE_PLANT:3");
+    put("ROSE_BUSH", "DOUBLE_PLANT:4");
+    put("PEONY", "DOUBLE_PLANT:5");
+
+    // Short plants
+    put("SHORT_GRASS", "LONG_GRASS:1");
+    put("FERN", "LONG_GRASS:2");
+    put("DEAD_BUSH", "LONG_GRASS");
```
Pablete1234:
```suggestion
    put("DEAD_BUSH", "LONG_GRASS:0");
    put("SHORT_GRASS", "LONG_GRASS:1");
    put("FERN", "LONG_GRASS:2");
```
same here

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+
+    // Stems
+    put("ATTACHED_MELON_STEM", "MELON_STEM:7");
+    put("ATTACHED_PUMPKIN_STEM", "PUMPKIN_STEM:7");
+
+    // Fish
+    put("COD", "RAW_FISH");
```
Pablete1234:
```suggestion
    put("COD", "RAW_FISH:0");
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("MOOSHROOM_SPAWN_EGG", "MONSTER_EGG:96");
+    put("OCELOT_SPAWN_EGG", "MONSTER_EGG:98");
+    put("HORSE_SPAWN_EGG", "MONSTER_EGG:100");
+    put("RABBIT_SPAWN_EGG", "MONSTER_EGG:101");
+    put("VILLAGER_SPAWN_EGG", "MONSTER_EGG:120");
+
+    // Misc
+    put("TERRACOTTA", "HARD_CLAY");
+    put("CHARCOAL", "COAL:1");
+    put("PETRIFIED_OAK_SLAB", "WOOD_STEP:43");
```
Pablete1234: this ain't right, i've added it correctly in the slab section
```suggestion
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("WOODEN_SLAB", "WOOD_STEP");
```
Pablete1234: "wooden slab" isn't even a thing in modern, is it?
```suggestion
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("BEEF", "RAW_BEEF");
+    put("BREWING_STAND", "BREWING_STAND_ITEM", "BREWING_STAND");
+    put("BRICK", "CLAY_BRICK");
+    put("BRICKS", "BRICK");
```
Pablete1234: This is going to be either problematic or useless.

if someone types `BRICK` do they refer to 1.8's brick (ie: the block, not CLAY_BRICK which would be the item) or modern's `BRICK` (which would be the item).

If 1.8 material parsing takes precedence, then the remap is useless, if the remap takes precedence, you're breaking maps.

Either way i'd consider just removing the first remap and keeping just bricks -> brick (the block), and if you want to refer to the item, you'll have to use clay_brick

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("BROWN_MUSHROOM_BLOCK", "HUGE_MUSHROOM_1");
+    put("CAKE", "CAKE_BLOCK");
+    put("CARROT", "CARROT_ITEM");
+    put("CARROTS", "CARROT");
```
Pablete1234: Similar scenario as bricks, "CARROT" refers to the 1.8 block, if you remap "CARROT" to "CARROT_ITEM" you're breaking block matching

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("END_PORTAL", "ENDER_PORTAL");
+    put("END_PORTAL_FRAME", "ENDER_PORTAL_FRAME");
+    put("END_STONE", "ENDER_STONE");
+    put("EXPERIENCE_BOTTLE", "EXP_BOTTLE");
+    put("FARMLAND", "SOIL");
+    put("FILLED_MAP", "MAP");
+    put("FIREWORK_ROCKET", "FIREWORK");
+    put("FIREWORK_STAR", "FIREWORK_CHARGE");
+    put("FIRE_CHARGE", "FIREBALL");
+    put("FLOWER_POT", "FLOWER_POT_ITEM", "FLOWER_POT");
```
Pablete1234, line 377: In modern flower pots have many diff materials now (eg: `POTTED_POPPY`, `POTTED_CACTUS` etc), unsure if you wanna bother remapping those

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("GLISTERING_MELON_SLICE", "SPECKLED_MELON");
+    put("GUNPOWDER", "SULPHUR");
+    put("HEAVY_WEIGHTED_PRESSURE_PLATE", "IRON_PLATE");
+    put("IRON_BARS", "IRON_FENCE");
+    put("IRON_DOOR", "IRON_DOOR_BLOCK");
+    put("LAVA", "STATIONARY_LAVA");
+    put("LEAD", "LEASH");
+    put("LIGHT_WEIGHTED_PRESSURE_PLATE", "GOLD_PLATE");
+    put("LILY_PAD", "WATER_LILY");
+    put("MAP", "EMPTY_MAP");
```
Pablete1234: Similar issue as the bricks problem, you have `filled_map -> map` and `map -> empty_map`, in 1.8 if someone gave a `map` they'd expect what modern calls a filled map, not an empty map

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("MELON", "MELON_BLOCK");
+    put("MELON_SLICE", "MELON");
```
Pablete1234: again the brick problem

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("MYCELIUM", "MYCEL");
+    put("NETHER_BRICK", "NETHER_BRICK_ITEM");
+    put("NETHER_PORTAL", "PORTAL");
+    put("NETHER_QUARTZ_ORE", "QUARTZ_ORE");
+    put("NETHER_WART", "NETHER_WARTS");
+    put("PISTON", "PISTON_BASE");
+    put("PISTON_HEAD", "PISTON_EXTENSION");
+    put("PORKCHOP", "PORK");
+    put("POTATO", "POTATO_ITEM");
+    put("POTATOES", "POTATO");
```
Pablete1234, line 399: Brick problem again

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("SNOWBALL", "SNOW_BALL");
+    put("SPAWNER", "MOB_SPAWNER");
+    put("STICKY_PISTON", "PISTON_STICKY_BASE");
+    put("STONE_PRESSURE_PLATE", "STONE_PLATE");
+    put("SUGAR_CANE", "SUGAR_CANE_BLOCK");
+    put("TNT_MINECART", "EXPLOSIVE_MINECART");
+    put("VOID_AIR", "AIR");
+    put("WALL_TORCH", "TORCH");
+    put("WATER", "STATIONARY_WATER");
+    put("WHEAT", "CROPS");
```
Pablete1234: WHEAT is the item, CROPS is the block, and in legacy wheat already refers to the former

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("REDSTONE_ORE", "GLOWING_REDSTONE_ORE");
+    put("REDSTONE_TORCH", "REDSTONE_TORCH_ON");
+    put("RED_MUSHROOM_BLOCK", "HUGE_MUSHROOM_2");
+    put("REPEATER", "DIODE");
+    put("REPEATING_COMMAND_BLOCK", "COMMAND");
+    put("SNOWBALL", "SNOW_BALL");
+    put("SPAWNER", "MOB_SPAWNER");
+    put("STICKY_PISTON", "PISTON_STICKY_BASE");
+    put("STONE_PRESSURE_PLATE", "STONE_PLATE");
+    put("SUGAR_CANE", "SUGAR_CANE_BLOCK");
```
Pablete1234: `SUGAR_CANE` refers to the item in 1.8, while `SUGAR_CANE_BLOCK` is only the block

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,394 @@
...
+    put("MUSIC_DISC_MELLOHI", "RECORD_7");
+    put("MUSIC_DISC_STAL", "RECORD_8");
+    put("MUSIC_DISC_STRAD", "RECORD_9");
+    put("MUSIC_DISC_WARD", "RECORD_10");
+    put("MUSIC_DISC_11", "RECORD_11");
+    put("MUSIC_DISC_WAIT", "RECORD_12");
+  }
+
+  private static SpMaterialData parse(String name) {
+    int split = name.indexOf(':');
```
Pablete1234:
```suggestion
    int split = name.indexOf(':', 2);
```
lets split only in 2 parts (avoids stuff like `a:2:c` passing as valid)

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialData.java
```diff
@@ -15,18 +15,28 @@
 class SpMaterialData implements LegacyMaterialData, ItemMaterialData, BlockMaterialData {
   private final Material material;
   private final short damage;
+  private final boolean hasData;
```
Pablete1234: This is only relevant to modern remapping, and should live outside of SpMaterialData

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialParser.java
```diff
@@ -75,20 +110,47 @@ public MaterialData visit(Material material, short data) {
...
+      public SpMaterialData visit(ModernMaterialNames.MaterialMapping mapping) {
+        return mapping.block();
+      }
     };
 
     T visit(Material material);
 
     T visit(Material material, short data);
+
+    default T visit(ModernMaterialNames.MaterialMapping mapping) {
```
Pablete1234: I'd not include a default impl, make each adapter handle it itself as it sees fit

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialParser.java
```diff
@@ -15,11 +15,39 @@ public static MaterialData parseBukkit(Node node) throws InvalidXMLException {
...
+  public static SpMaterialData parseItem(String text, Node node) throws InvalidXMLException {
+    return parse(text, node, false, Adapter.PGM_ITEM);
+  }
+
+  public static SpMaterialData parseBlock(String text, Node node) throws InvalidXMLException {
+    return parse(text, node, false, Adapter.PGM_BLOCK);
   }
 
   public static Material parseMaterial(String text, Node node) throws InvalidXMLException {
+    return parse(text, node, true, Adapter.PGM_BLOCK).getItemType();
```
Pablete1234: This codepath can be used for either blocks or items (at least one that i know, shop currencies, gets here), so may need to reconsider using PGM_BLOCK adapter in this case, or maybe needs a param or two separate impls for parseItemMaterial and parseBlockMaterial, idk.


Update: given what i've said about how single vs multi-mapping can be handled, this may be a good place for that; could make an `Adapter.PGM_MATERIAL` that will only succesfully if it's a single modern material mapping that is found, and that the metadata is empty for that remap.

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialParser.java
```diff
@@ -28,37 +56,44 @@ public static Material parseMaterial(String text, Node node) throws InvalidXMLEx
...
     }
 
-    text = text.toUpperCase(Locale.ROOT).replaceAll("\\s+", "_").replaceAll("\\W", "");
+    var material = Material.getMaterial(text);
+    if (material != null) return material;
 
-    // At some point prior to legacy, this rename happened
-    if (text.equals("SNOWBALL")) return Material.SNOW_BALL;
+    var modern = ModernMaterialNames.get(text);
+    if (modern != null) return modern.block().getItemType();
```
Pablete1234: should not default to block, if i'm not mistaken this is only being used for parsing of `material:1` kinda stuff (within parse), maybe the behavior should be lifted there so that you can call the right adapter method with the modern material mapping

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialParser.java
```diff
@@ -28,37 +56,44 @@ public static Material parseMaterial(String text, Node node) throws InvalidXMLEx
...
+    // This permits materials with distinct item and block variants
+    // in legacy to be matched to the correct variant always.
+    if (modern != null
+        && (material == null || modern.item().getItemType() != modern.block().getItemType())) {
+      return adapter.visit(modern);
     }
+
+    if (material != null) return adapter.visit(material);
+
+    throw new InvalidXMLException("Could not find material '" + text + "'.", node);
```
Pablete1234: As explained thru "the brick problem" the "correct variant always" is probably not true

```suggestion
    var material = Material.getMaterial(text);
    if (material != null) return adapter.visit(material);
    
    var modern = ModernMaterialNames.get(text);
    if (modern != null) return adapter.visit(modern);
    
    throw new InvalidXMLException("Could not find material '" + text + "'.", node);
```

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialUtils.java
```diff
@@ -170,6 +170,19 @@ public MaterialMatcher.Builder visit(Material material, short data) {
...
 
+    @Override
+    public MaterialMatcher.Builder visit(ModernMaterialNames.MaterialMapping mapping) {
+      var item = mapping.item();
+      var block = mapping.block();
+
+      addMaterialData(item);
+      if (item.getItemType() != block.getItemType() || item.getData() != block.getData()) {
+        addMaterialData(block);
+      }
```
Pablete1234:
```suggestion
      if (mapping.single()) {
        var type = mapping.type();
        mapping.data().ifPresentOrElse(d -> visit(type, d), () -> visit(type));
      } else {
        var bl = mapping.blockType();
        mapping.blockData().ifPresentOrElse(d -> visit(bl, d), () -> visit(bl));
        var it = mapping.itemType();
        mapping.itemData().ifPresentOrElse(d -> visit(it, d), () -> visit(it));
      }
```
Something along these lines would be nice, because its now a few places that are checking if itemType != blockType etc.

This assumes the MaterialMapping record has gotten a few changes to make it more robust, mainly:

Fields become:
`boolean single, Material blockType, Optional<Short> blockData, Material itemType, Optional<Short> itemData`

Preconditions:
blockData and itemData must be nonnull (use empty optionals instead)
blockType and itemType cannot both be null
if blockType is null, blockData cannot be a non-empty optional
if itemType is null, itemData cannot be a non-empty optional
if `single` is true, blockType+blockData must be equal to itemType+itemData, or one of the two can be null.
if `single` is false, both `blockType` and `itemType` must be non-null, and the blocktype+blockData and itemType+itemData pairs must be different (either the type is different, or the optionals are).

And then a couple methods can be trivially added for:
`type` if !single throws, otherwise returns either blockType or itemType, whichever is notnull
`data` if !single throws, otherwise returns either blockData or itemData, whichever type field is notnull.

What this kind of model simplifies is that if a mapping exists for just item, or just block, or both but its the same mapping, you can just get it via `.type()` and `.data()`, if it's known to be different mappings, you can chose to handle the two separately. And this allows other places where this is used to maybe work fine with single-mapping materials, but if there's a multi-mapping material throw an exception and say "sorry, ambiguity for what this means" or similar

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/SpMaterialParser.java
```diff
@@ -28,37 +56,44 @@ public static Material parseMaterial(String text, Node node) throws InvalidXMLEx
...
-    if (material == null) {
-      throw new InvalidXMLException("Could not find material '" + text + "'.", node);
-    }
-    return material;
+    throw new InvalidXMLException("Could not find material '" + text + "'.", node);
   }
 
-  public static <T> T parse(String text, @Nullable Node node, boolean matOnly, Adapter<T> adapter)
+  @SuppressWarnings("deprecation")
+  private static <T> T parse(String text, Node node, Adapter<T> adapter)
```
Pablete1234: We have a few too many `parse` methods, maybe clarify that this one parses just and only materials without meta:

```suggestion
  private static <T> T parseMaterial(String text, Node node, Adapter<T> adapter)
```

Pablete1234 (changes requested):

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/material/ModernMaterialNames.java
```diff
@@ -0,0 +1,437 @@
...
+      @Nullable Short itemDamage,
+      Material blockMaterial,
+      @Nullable Short blockDamage) {
+
+    MaterialMapping(Material material, @Nullable Short damage) {
+      this(material, damage, material, damage);
+    }
+
+    boolean single() {
+      return itemMaterial == blockMaterial && Objects.equals(itemDamage, blockDamage);
```
Pablete1234: This is still missing one key point, even if the item and block are not the same, it can still be a single material mapping (ie: one is null and the other is not).

There needs to be a distinction between `item` and `block` material, even if they're not different (one of them can just be null).

Eg:
`put("WOODEN_AXE", "WOOD_AXE");`
Should effectively become:
`new MaterialMapping(Material.WOOD_AXE, null, null, null);`
because it has an item form, but it surely has no block form.
This should happen for all the materials we remap-into that are not blocks.

And the same can happen the other way around too, things like water, or moving piston exist as a block but not as an item, and those mappings should still be considered "single" mappings.

What this allows is that if you try to make a core's material `WOODEN_AXE` it knows that maps to nothing and has potential to complain about it being invalid, instead of just parsing the block material as WOODEN_AXE and pretending everything is fine.


The actual way to implement it is that the `put(String modern, String legacy)` method should look if the legacy parsed material is an item and if its a block (`Material#isBlock` and check what  `SpMaterialUtils#validateItem` does to know if it's an item)

Pablete1234 (approved):

Pablete1234:
@calcastor @cswhite2000 i've done a last pass of cleanup and added a bunch of additional validation/exceptions to avoid edge-cases popping-up and hunting us in the future. Please doublecheck what i've pushed and see if you can spot anything else i may have missed.

Note, a couple mappings were caught to be wrong or bad with the additional validations, mainly:
- `DEAD_BUSH` apparently in legacy both dead bush and longgrass with type=dead exist, and they're distinct blocks/items with different behavior (dead bush can only be placed on sand, longgrass(dead) can only be placed on grass). I've removed the mapping since the legacy mapping is already ok.
- `LAVA`/`WATER` -> `STATIONARY_*` would mean its impossible to make a filter targeting "moving water", so i've removed the mappings. If making a map for legacy mapdevs will continue to need to know the difference between water and stationary_water, i don't see a way around this.
- `NETHER_STALK`/`NETHER_WARTS` the order was backwards, now item and block are in the right parameters

the rest of the mappings seem not to interfere with legacy material names so they're ok to stay, there is now a validation ran on startup that checks this is the case.

cswhite2000 (approved):

PR #1647, opened 2026-04-03
Use Mojang's API for username fetching

calcastor:


calcastor:
Will be superseded by https://github.com/PGMDev/PGM/issues/1673

PR #1648, opened 2026-04-04
Migrate modern platform to 26.2

calcastor:
to-do:

- ~~Switch to tagged Cloud release when existent~~
- ~~There are likely side-effects to each match world being loaded as a separate dimension to look into (time desync is one; time seems to "cycle" despite gamerule until a match ends?? not sure)~~ ~~pending Paper TO-DOs~~
- intersect jar
- ~~https://github.com/PaperMC/adventure-platform/pull/241 so PGM doesn't need to rely on plugin remapping; eventually we need adventure-platform to support Adventure 5.0 when that is tagged~~


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -327,75 +326,26 @@ public World createWorld(WorldCreator creator) {
...
-    // Abort if level.dat is empty
-    if (!levelStorageAccess.hasWorldData()) {
-      throw new UnsupportedOperationException("Cannot use PGM createWorld with an empty level.dat");
-    }
-
-    Dynamic<?> dataTag;
-    LevelSummary summary;
+  private static int readDataVersion(Path worldFolder) {
+    Path levelDat = worldFolder.resolve("level.dat");
+    if (!Files.exists(levelDat)) return 0;
```
TTtie: This is already implemented as MiscUtils#getWorldDataVersion (with less checking and slightly different return values)

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernNMSHacks.java
```diff
@@ -235,87 +214,104 @@ public World createWorld(WorldCreator creator) {
...
+
+    LevelStem configuredStem =
+        console.registryAccess().lookupOrThrow(Registries.LEVEL_STEM).getValue(actualDimension);
+    if (configuredStem == null) {
+      throw new IllegalStateException("Missing configured level stem " + actualDimension);
+    }
+
+    // PGM: We need to read data version from level.dat pre-migration
+    int dataVersion =
+        MISC_UTILS.getWorldDataVersion(server.getWorldContainer().toPath().resolve(name));
```
TTtie: The util method takes a path to level.dat itself instead of the path to the world directory, btw.
```suggestion
        MISC_UTILS.getWorldDataVersion(server.getWorldContainer().toPath().resolve(name).resolve("level.dat"));
```

buildSrc/src/main/kotlin/buildlogic.java-conventions.gradle.kts
```diff
@@ -18,27 +18,39 @@ repositories {
...
 }
 
 dependencies {
     api("org.jdom:jdom2:2.0.6.1")
     api("net.kyori:adventure-api:4.26.1")
     api("net.kyori:adventure-text-serializer-plain:4.26.1")
-    api("net.kyori:adventure-platform-bukkit:4.4.1")
+    // adventure-platform fork with ViaVersion and 1.21.11+ fixes
+    // https://github.com/OvercastCommunity/adventure-platform
+    api("com.github.OvercastCommunity.adventure-platform:adventure-platform-bukkit:04de657e85")
```
cswhite2000: I guess your upstream pr was ignored?
Is that repo just not particularly responsive?

calcastor: `adventure-platform` is largely unmaintained; numerous PRs have been left dormant there despite addressing relevant concerns for projects which use it. Whether it is going to be updated for Adventure 5.x or see any updates at all was up in the air last I spoke to a maintainer about it. It seems like platform-native Adventure implementations are the preferred/supported way forward, but that doesn't work for us given the platforms we support and how we take advantage of `adventure-platform` to pass our own component renderer.

The set of changes in the fork are related to ViaVersion 5.x support, as well as the PR I made which fixes vote books not opening. I also added a fix for translations in 1.16+ when ViaFacet is used in https://github.com/OvercastCommunity/adventure-platform/commit/53c838e42190f802b505281754baeebc6b0bf2b1. The fork also drops usage of GSONComponentSerializer that are removed in Adventure 5.x, and I additionally have a change elsewhere that migrates `adventure-platform` to Adventure 5.x when that becomes relevant (currently Paper 26.1.2 is still on 4.26.1 so we don't need that yet).

core/src/main/java/tc/oc/pgm/action/actions/MessageAction.java
```diff
@@ -75,14 +75,23 @@ private Component replaceClickEvents(
...
     var children = new ArrayList<>(component.children());
     children.replaceAll(child -> replaceClickEvents(child, replacer));
     return component.children(children);
   }
 
+  @SuppressWarnings({"rawtypes", "unchecked", "RedundantSuppression"})
+  private ClickEvent replaceTextClickEvent(
+      ClickEvent.Action action, ClickEvent.Payload.Text payload) {
+    if (!action.supports(payload)) {
+      throw new IllegalArgumentException("Action " + action + " does not support text payloads");
```
Pablete1234: i'm not sure that this can ever happen; we're only replacing clicks who's payloads were Text meaning they must be strings after the change too.
The only thing you seem to need this for is generics, but you're not using them either.

consider just a method like:
```java
private <T extends ClickEvent.Payload.String> ClickEvent<T> replaceEvent(ClickEvent<T> ev, Function<MatchResult, String> replacer) {
  var matcher = PATTERN.matcher(payload.value());
  var result = new StringBuilder();
  while (matcher.find()) matcher.appendReplacement(result, replacer.apply(matcher));
  matcher.appendTail(result);
  var resultPayload = ClickEvent.Payload.string(result.toString());
  return ClickEvent.clickEvent(click.action(), resultPayload);
}
```
and only call it inside the if section of that previous method, where you know the click is of the right type

util/src/main/java/tc/oc/pgm/util/platform/Platform.java
```diff
@@ -25,16 +25,29 @@ public abstract class Platform {
...
+    return new Version(leadingInt(parts, 0), leadingInt(parts, 1), leadingInt(parts, 2));
+  }
+
+  private static int leadingInt(String[] parts, int idx) {
+    if (idx >= parts.length) return 0;
+    String s = parts[idx];
+    int end = 0;
+    while (end < s.length() && Character.isDigit(s.charAt(end))) end++;
+    return end > 0 ? Integer.parseInt(s.substring(0, end)) : 0;
+  }
```
Pablete1234: You're better off just parsing a regex:
```
private static final Pattern VERSION_MATCHER = Pattern.compile("(\\d{1,2})\\.(\\d{1,2})\\.(\\d{1,2})");`
```
just call find on the string, and parse the 3 groups

core/src/main/java/tc/oc/pgm/match/MatchPlayerImpl.java
```diff
@@ -332,12 +333,15 @@ public void setFrozen(boolean yes) {
...
-        ENTITIES.spawnFreezeEntity(bukkit, FROZEN_VEHICLE_ENTITY_ID, isLegacy()).send(bukkit);
-        ENTITIES.entityMount(bukkit.getEntityId(), FROZEN_VEHICLE_ENTITY_ID).send(bukkit);
+        ENTITIES
+            .spawnFreezeEntity(bukkit, getFrozenVehicleId(world), isLegacy())
+            .send(bukkit);
+        ENTITIES.entityMount(bukkit.getEntityId(), getFrozenVehicleId(world)).send(bukkit);
       } else {
-        ENTITIES.destroyEntitiesPacket(FROZEN_VEHICLE_ENTITY_ID).send(bukkit);
+        ENTITIES.destroyEntitiesPacket(getFrozenVehicleId(world)).send(bukkit);
       }
```
Pablete1234, line 342: this is NOT going to work, frozen vehicle used to be a constant, so spawning/deleting using the same constant is referring to spawning/deleting that same constant.
With this change it is now calling the method which allocates a new entity id every time

util/src/main/java/tc/oc/pgm/util/platform/Platform.java
```diff
@@ -21,20 +22,32 @@
...
   public static final Variant VARIANT;
 
+  private static Version parseServerVersion(final @NonNull String versionName) {
+    var matcher = VERSION_MATCHER.matcher(versionName);
+    return matcher.find()
+        ? new Version(
+            Integer.parseInt(matcher.group(1)),
+            Integer.parseInt(matcher.group(2)),
+            Integer.parseInt(matcher.group(3)))
+        : new Version(0, 0, 0);
```
Pablete1234: any reason to manually parse instead of using TextParser.parseVersion, besides just the part where you're deciding what to grab? imo would be cleaner to use a regex to find the potential version here and pass that to the parser to get a real version, instead of defaulting to 0,0,0

util/build.gradle.kts
```diff
@@ -4,8 +4,8 @@ plugins {
 }
 
 dependencies {
-    compileOnly("dev.pgm.paper:paper-api:1.8_1.21.11-SNAPSHOT")
-    testImplementation("io.papermc.paper:paper-api:1.21.11-R0.1-SNAPSHOT")
+    compileOnly("dev.pgm.paper:paper-api:1.8_26.1.2-SNAPSHOT")
```
Pablete1234: still targeting 26.1?

calcastor: Waiting on non-alpha for the moment

calcastor:
Marking as draft pending https://github.com/retrooper/packetevents/pull/1505; things work fine on modern Paper but this is not the case for legacy.

calcastor:
> Marking as draft pending [retrooper/packetevents#1505](https://github.com/retrooper/packetevents/pull/1505); things work fine on modern Paper but this is not the case for legacy.

Given that upstream packetevents is working fine on modern (as Paper is providing the Adventure dependency), and that packetevents is optional for legacy, I've undrafted this PR.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/util/ModernWorldLoader.java
```diff
@@ -0,0 +1,343 @@
...
+        console,
+        Util.backgroundExecutor(),
+        console.storageSource,
+        worldGenSettings,
+        dimensionKey,
+        customStem,
+        false, // PGM: force isDebug false
+        biomeZoomSeed,
+        ImmutableList.of(), // PGM: no spawners
+        true, // PGM: force tickTime true
```
TTtie, line 199:
```suggestion
        true,
```
extremely tiny nit but this being `true` comes from Paper rather than from us

TTtie (approved):
Seems okay to me from a glance.

Might want to consider using [Markdown doc comments](https://openjdk.org/jeps/467) for the new code going forward now that we'll be targeting Java 25 once merged but this is merely a thing of taste.

PR #1650, opened 2026-04-07
Add hover event for top 10 player stats in StatsMatchModule.getMessage()

nicoliee:
I updated getMessage to add a hover tooltip that shows the Top 10 players for that stat, skips entries with value 0, and displays tied entries on the same line using TextFormatter.list.

The objective is to make it easier for people to see top <stat> without the use of a website or going into the stats menu for each team and comparing them manually.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -20,7 +20,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
@@ -79,6 +78,7 @@ import tc.oc.pgm.stats.menu.items.VerboseStatsMenuItem;
 import tc.oc.pgm.teams.Team;
 import tc.oc.pgm.tracker.TrackerMatchModule;
 import tc.oc.pgm.tracker.info.ProjectileInfo;
+import tc.oc.pgm.util.collection.RankedSet;
 import tc.oc.pgm.util.named.NameStyle;
 import tc.oc.pgm.util.text.TextFormatter;
 import tc.oc.pgm.util.usernames.UsernameResolvers;
@@ -106,6 +106,7 @@ public class StatsMatchModule implements MatchModule, Listener {
   private List<MenuItem> teams;
 
   private static final int TOP_HOVER_LIMIT = 10;
+  private static final int MAX_PLAYERS = 15;
 
   public StatsMatchModule(Match match, List<StatType.OfFormula> formulaStats) {
     this.match = match;
@@ -499,38 +500,43 @@ public class StatsMatchModule implements MatchModule, Listener {
   }
 
   private Component buildTopHover(AggStat<?> agg) {
-    var top = allPlayerStats.entrySet().stream()
-        .map(e -> {
-          Number val = getStatValue(agg.type, match.getPlayer(e.getKey()), e.getValue());
-          return val != null && val.doubleValue() > 0
-              ? Map.entry(e.getKey(), val.doubleValue())
-              : null;
-        })
-        .filter(Objects::nonNull)
-        .sorted(Map.Entry.<UUID, Double>comparingByValue().reversed())
-        .toList();
+    RankedSet<Map.Entry<UUID, Double>> ranked =
+        new RankedSet<>(Comparator.<Map.Entry<UUID, Double>, Double>comparing(Map.Entry::getValue)
+            .reversed());
 
-    if (top.size() <= 1) return null;
+    for (Map.Entry<UUID, PlayerStats> e : allPlayerStats.entrySet()) {
+      Number val = getStatValue(agg.type, match.getPlayer(e.getKey()), e.getValue());
+      if (val != null && val.doubleValue() > 0)
+        ranked.add(Map.entry(e.getKey(), val.doubleValue()));
+    }
 
-    var lines = new ArrayList<Component>();
-    for (int i = 0; i < top.size() && lines.size() < TOP_HOVER_LIMIT; ) {
-      double value = top.get(i).getValue();
-      int j = i;
-      while (j < top.size() && Double.compare(top.get(j).getValue(), value) == 0) {
-        j++;
+    if (ranked.size() <= 1) return null;
+
+    List<Component> lines = new ArrayList<>();
+    int playersShown = 0;
+    int rank = 1;
+
+    for (Set<Map.Entry<UUID, Double>> rankSet : ranked.ranksView()) {
+      if (lines.size() >= TOP_HOVER_LIMIT || playersShown >= MAX_PLAYERS) break;
+
+      List<Map.Entry<UUID, Double>> entries = new ArrayList<>(rankSet);
+      int rankSize = entries.size();
+
+      if (rankSize <= 3) {
+        List<Component> names =
+            entries.stream().map(e -> getPlayerComponent(e.getKey())).toList();
+        lines.add(rankLine(
+            rank, agg, list(names, NamedTextColor.GRAY), entries.getFirst().getValue()));
+        playersShown += rankSize;
+      } else {
+        for (Map.Entry<UUID, Double> entry : entries) {
+          if (lines.size() >= TOP_HOVER_LIMIT || playersShown >= MAX_PLAYERS) break;
+          lines.add(rankLine(rank, agg, getPlayerComponent(entry.getKey()), entry.getValue()));
+          playersShown++;
+        }
       }
 
-      var tiedPlayers = top.subList(i, j).stream()
-          .map(entry -> getPlayerComponent(entry.getKey()))
-          .toList();
-
-      lines.add(text((lines.size() + 1) + ". ")
-          .color(NamedTextColor.WHITE)
-          .append(list(tiedPlayers, NamedTextColor.GRAY))
-          .append(text(" - ").color(NamedTextColor.GRAY))
-          .append(agg.type.makeNumber(value)));
-
-      i = j;
+      rank += rankSize;
     }
 
     return Component.join(
@@ -539,5 +545,13 @@ public class StatsMatchModule implements MatchModule, Listener {
                 Stream.of(translatable("match.stats.top", agg.type.component(empty()))),
                 lines.stream())
             .toList());
+  }
+
+  private Component rankLine(int rank, AggStat<?> agg, Component who, double value) {
+    return text(rank + ". ")
+        .color(NamedTextColor.WHITE)
+        .append(who)
+        .append(text(" - ").color(NamedTextColor.GRAY))
+        .append(agg.type.makeNumber(value));
   }
 }
--- a/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
@@ -23,7 +23,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.JoinConfiguration;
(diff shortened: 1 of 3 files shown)


core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +494,44 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
         player != null ? formulaStats.formula().apply(player) : null;
     };
   }
+
+  private Component buildTop10Hover(AggStat<?> agg) {
+    Map<Double, List<UUID>> byValue = new java.util.TreeMap<>(Comparator.reverseOrder());
+    Component header = text("Top ").append(agg.type.component(empty()));
```
cswhite2000: Should be using a translatable component

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +494,44 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
         player != null ? formulaStats.formula().apply(player) : null;
     };
   }
+
+  private Component buildTop10Hover(AggStat<?> agg) {
+    Map<Double, List<UUID>> byValue = new java.util.TreeMap<>(Comparator.reverseOrder());
```
cswhite2000: import this

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +494,44 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+          .color(NamedTextColor.WHITE)
+          .append(namesPart)
+          .append(text(" - ").color(NamedTextColor.GRAY))
+          .append(agg.type.makeNumber(val));
+
+      lines.add(line);
+      rank += tied.size();
+    }
+
+    if (lines.isEmpty()) return empty();
```
cswhite2000: lines can't be empty as you always add a header

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +494,44 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+      if (val == null) return;
+      double d = val.doubleValue();
+      if (d <= 0) return;
+      byValue.computeIfAbsent(d, k -> new ArrayList<>()).add(uuid);
+    });
+
+    List<Component> lines = new ArrayList<>();
+    lines.add(header);
+    int rank = 1;
+    for (Map.Entry<Double, List<UUID>> entry : byValue.entrySet()) {
```
cswhite2000: I'm not convinced this approach is better than just having a stream, sorting it and then taking the top N results, then mapping to component.

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +494,44 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+      if (rank > 10) break;
+      double val = entry.getKey();
+      List<UUID> tied = entry.getValue();
+
+      List<Component> names =
+          tied.stream().map(this::getPlayerComponent).collect(Collectors.toList());
+      Component namesPart = TextFormatter.list(names, NamedTextColor.WHITE);
```
cswhite2000: if everyone ties then this could be every player in the match.

Older clients crash if they recieve messages that are too long.

cswhite2000 (commented):
10 might be too many names for a hover.
Could you add a screenshot so we can see what this looks like

Also, your commit is not properly signed off. See https://github.com/PGMDev/PGM/blob/dev/docs/CONTRIBUTING.md

nicoliee:
This is how the hover looks when there are tied entries.
I included many ties to demonstrate how it behaves when too many players appear on a single line.

I'm not sure if this is ideal, or if it would be better to limit it to something like the top 5. I'm also unsure how to handle cases where a large number of players are tied (e.g., 5+ players on the same line), since limiting the number of displayed players on a single line would defeat the purpose of the hover.

# Preview:
screenshot-2026-04-08_11-22-46

Pablete1234:
I wonder if it'd make more sense to only show one player per row regardless of ties, and just show them with the same ranking, ie:
```
1. PlayerA - 50
2. PlayerB - 45
2. PlayerC - 45
2. PlayerD - 45
5. PlayerE - 40
6. PlayerF - 38
...
```
This somewhat guarantees that there will always only be 10 players displayed at most, regardless of N-way ties

nicoliee:
I was thinking of listing them only when there are up to 4 ties; otherwise just show it like you suggested. What do you think? It might be a bit inconsistent, but it helps present ties more cleanly (and still shows more players on hover) while handling cases with 5+ ties, which are pretty rare.

Pablete1234:
imo 4+ ties are already pretty long and hard to read, i think 3 is the max that really fits reasonably in something like this, and i'd be fine with that, up to 3-way ties can be merged into one line, more need to be split

i am still somewhat concerned about older clients being kicked by the worst-case of this (ie: up to 30 player names may display)

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +497,47 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+
+    if (top.size() <= 1) return null;
+
+    var lines = new ArrayList<Component>();
+    for (int i = 0; i < top.size() && lines.size() < TOP_HOVER_LIMIT; ) {
+      double value = top.get(i).getValue();
+      int j = i;
+      while (j < top.size() && Double.compare(top.get(j).getValue(), value) == 0) {
+        j++;
+      }
```
Pablete1234: You should make use of pgm's `RankedSet` for this, feel free to expand it to expose the an iterator over ranks if you want

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +498,60 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+    if (ranked.size() <= 1) return null;
+
+    List<Component> lines = new ArrayList<>();
+    int playersShown = 0;
+    int rank = 1;
+
+    for (Set<Map.Entry<UUID, Double>> rankSet : ranked.ranksView()) {
+      if (lines.size() >= TOP_HOVER_LIMIT || playersShown >= MAX_PLAYERS) break;
+
+      List<Map.Entry<UUID, Double>> entries = new ArrayList<>(rankSet);
```
Pablete1234, line 522: you should be able to operate on the `rankSet` directly, no need for the list

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +498,60 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+          lines.add(rankLine(rank, agg, getPlayerComponent(entry.getKey()), entry.getValue()));
+          playersShown++;
+        }
+      }
+
+      rank += rankSize;
+    }
+
+    return Component.join(
+        JoinConfiguration.newlines(),
```
Pablete1234, line 543: i believe 1.7 does not handle `\n` well, at the very least in chat it doesn't
double-check if it works for hovers, and if it doesn't, consider skipping sending this hover to 1.7 clients

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +498,60 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+        Stream.concat(
+                Stream.of(translatable("match.stats.top", agg.type.component(empty()))),
+                lines.stream())
+            .toList());
+  }
+
+  private Component rankLine(int rank, AggStat<?> agg, Component who, double value) {
```
Pablete1234, line 550:
```suggestion
  private Component rankLine(int rank, AggStat<?> agg, Component who, Number value) {
```
you don't actually know that the type of this stat is a double, some are ints (eg: kills are ints), and the rest of the code at least tries to respect that distinction, this should try too

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +498,60 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+
+    for (Set<Map.Entry<UUID, Double>> rankSet : ranked.ranksView()) {
+      if (lines.size() >= TOP_HOVER_LIMIT || playersShown >= MAX_PLAYERS) break;
+
+      List<Map.Entry<UUID, Double>> entries = new ArrayList<>(rankSet);
+      int rankSize = entries.size();
+
+      if (rankSize <= 3) {
+        List<Component> names =
+            entries.stream().map(e -> getPlayerComponent(e.getKey())).toList();
```
Pablete1234, line 527: You may want to look at how `credit` function does this, and maybe even reuse it!
mainly talking about using: 
`list(Collections2.transform(players, this::getPlayerComponent), null);`
since the transformed list avoids the allocation cost of the copy, as well as the stream

in your case it would look like:
`var names = Collections2.transform(entries, e -> getPlayerComponent(e.getKey()));`

core/src/main/java/tc/oc/pgm/stats/StatsMatchModule.java
```diff
@@ -492,4 +498,60 @@ private static Number getStatValue(StatType<?> statType, MatchPlayer player, Sta
...
+  private Component buildTopHover(AggStat<?> agg) {
+    RankedSet<Map.Entry<UUID, Double>> ranked =
+        new RankedSet<>(Comparator.<Map.Entry<UUID, Double>, Double>comparing(Map.Entry::getValue)
+            .reversed());
+
+    for (Map.Entry<UUID, PlayerStats> e : allPlayerStats.entrySet()) {
+      Number val = getStatValue(agg.type, match.getPlayer(e.getKey()), e.getValue());
+      if (val != null && val.doubleValue() > 0)
+        ranked.add(Map.entry(e.getKey(), val.doubleValue()));
+    }
```
Pablete1234, line 511: Instead of using a `RankedSet<Map.Entry<UUID, Double>>`, you can use a `RankedSet<Map.Entry<UUID, PlayerStats>>` directly, that just takes in `allPlayerStats` (call with just one addAll or similar) and use a comparator to actually decide ordering (ie: `Comparator.comparing(e -> getStatValue(agg.type, match.getPlayer(e.getKey()), e.getValue())`) and make sure to do nulls last and the right order etc.

```suggestion
    var ranked =
        new RankedSet<Map.Entry<UUID, PlayerStats>>(
            Comparator.nullsLast(
              Comparator.comparing(e -> getStatValue(agg.type, match.getPlayer(e.getKey()), e.getValue()))
            .reversed()));
    ranked.addAll(allPlayerStats);
```


However, this feels like you're re-creating the same logic already done for figuring out the top stats (the whole agg stat class exists for that!), just re-calculating it a second time for top ranks instead of just the max player.

Notice that AggStat already has a `Set<UUID>` with the top#1 players, maybe that's what needs to become the ranked set, and the main stat aggregation logic step would be the one doing the heavy lifting for you; look for this piece of code a bit higher up:
```java
    allPlayerStats.forEach((uuid, s) -> {
      MatchPlayer player = match.getPlayer(uuid);
      stats.replaceAll(stat -> stat.track(uuid, player, s));
    });
```
It's already computing all the agg stats, for all stats, iterating for all players. You could just expand on that aggStat's `track` method to not just keep the top 1 players, but to keep all players, ranked by their stat (ie: a ranked set)

Pablete1234 (changes requested):
After looking into this a bit more, i believe the calculation should be a bit more baked into our existing aggregation logic (ie: `AggStat` and its `track` method).

The display/design side of it seems to have been resolved tho, with the limit of max 10 lines, max 15 players, and max of 3 players in a line, which all seem reasonable to me.

PR #1652, opened 2026-04-11
Random ballot voting option

Pablete1234:
Still in active development, but opening as draft for the time being

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMConfig.java
+++ b/core/src/main/java/tc/oc/pgm/PGMConfig.java
@@ -195,7 +195,7 @@ public final class PGMConfig implements Config {
     }
 
     this.allowExtraVotes = parseBoolean(config.getString("votes.allow-extra-votes", "true"));
-    this.maxExtraVotes = parseInteger(config.getString("votes.max-extra-votes", "5"));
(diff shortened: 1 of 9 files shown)


cswhite2000:
Thoughts on some sort of weighted random ballot as an intermediate option between this and the current winner takes all?

Basically, normalizing the scores, raising it to some configurable power, and then using the new weights?

---

Something like so:

- Map A: 50% of votes
- Map B: 30% of votes
- Map C: 20% of votes

Then:

- Map A: 0.5^2 = score of 0.25
- Map B: 0.3^2 = score of 0.09
- Map C: 0.2^2 = score of 0.04

Total 0.38
- Map A: 0.25 / 0.38 = ~66% chance
- Map B: 0.09 / 0.38 = ~24% chance
- Map C: 0.04 / 0.38 = ~10% chance

---

Then you could make it follow the vote to a lesser or greater degree by adjusting the power its raised to

Pablete1234:
Currently what i've added is a min and max threshold, so we can avoid "unfair" outcomes. Eg:
- if most voted map has >66% of the votes, its a direct win (>max threshold).
- if a map has <10% of the votes, it is not part of the randomization (< min threshold).

With those it should be possible to exclude options that very few people want, but if 3 maps have 30-50% each, then they all get a chance based on their votes.

PR #1655, opened 2026-04-14
Migrate to PacketEvents

calcastor:


core/src/main/java/tc/oc/pgm/api/integration/NickIntegration.java
```diff
@@ -1,16 +1,27 @@
...
+
+  /**
+   * Get the skin for a player per viewer
+   *
+   * @param player The viewed player
+   * @param viewer The viewing player
+   * @return The skin shown to a viewer for a player
+   */
+  @Nullable
+  Skin getPlayerSkin(Player player, Player viewer);
```
Pablete1234: For backwards-compat reasons, include a default impl of this that just returns null

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/PePacket.java
```diff
@@ -0,0 +1,44 @@
...
+  private final PacketWrapper<?> wrapper;
+
+  public PePacket(PacketWrapper<?> wrapper) {
+    this.wrapper = wrapper;
+  }
+
+  @Override
+  public void send(Player viewer) {
+    if (viewer.isOnline()) {
+      PacketEvents.getAPI().getPlayerManager().sendPacket(viewer, wrapper);
```
Pablete1234:
```suggestion
      PE.sendPacket(viewer, wrapper);
```
and have a `public static final PlayerManager PE = PacketEvents.getAPI().getPlayerManager()` of sorts
also should be done on the rest of uses

cswhite2000 (dismissed):

Pablete1234 (approved):

PR #1657, opened 2026-04-15
--hideDefaults flag for /variables

eliminate123:
This PR adds a ``-d (--hideDefaults)`` flag for ``/variables``. The -d flag hides all default player. variables. 

player. variables take up 3 pages when doing /variables.

On connect 4 when using ``/variables`` 
image

Now when using ``/variables -ad``
image

### DIFF
--- a/core/src/main/java/tc/oc/pgm/command/MapDevCommand.java
+++ b/core/src/main/java/tc/oc/pgm/command/MapDevCommand.java
@@ -45,6 +45,7 @@ import tc.oc.pgm.util.material.MaterialData;
 import tc.oc.pgm.util.text.TextFormatter;
 import tc.oc.pgm.variables.Variable;
 import tc.oc.pgm.variables.VariablesMatchModule;
+import tc.oc.pgm.variables.types.PlayerVariable;
 
 public class MapDevCommand {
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/command/MapDevCommand.java
```diff
@@ -61,9 +62,11 @@ public void showVariables(
       @Argument("target") @Default(CURRENT) MatchPlayer target,
       @Argument("page") @Default("1") int page,
       @Flag(value = "query", aliases = "q") String query,
+      @Flag(value = "hideDefaults", aliases = "d") boolean hideDefaults,
       @Flag(value = "all", aliases = "a") boolean all) {
 
     List<Map.Entry<String, Variable<?>>> variables = vmm.getVariables()
+        .filter(e -> !(hideDefaults && e.getValue() instanceof PlayerVariable))
```
cswhite2000, line 69: Your implementation is to hide all player variables but the name of the flag is to hide defaults?

these are not quite the same thing

Pablete1234:
Imo it should be default behavior that default unused variables get hidden, and can be shown with `-a` (all). However that requires keeping track of what formulas/expressions are using what variables, across all modules that use them, and not just filtering "player." variables

PR #1660, opened 2026-04-21
Track groundedness server-side

calcastor:
`isOnGround` for players is deprecated, at least in Paper 1.21.11, since it is client-side; this PR moves to a server-side calculation that should be more accurate. This should also resolve some of the kill credit issues on modern, since we are no longer trying to track `isOnGround` with `EntityPoseChangeEvent`.


calcastor:
The main intent behind this PR was to ensure modern had a more suitable way of checking onGround than EntityPoseChangeEvent might provide; something packet-level would work better for this.

PR #1661, opened 2026-04-23
Match `flag-captured` to respawning flags where `wasCaptured` is true

calcastor:
While `flag-captured` and all other flag filters have tended to refer to the explicit state of a flag, it is helpful to be able to filter a flag by whether or not it has been captured, even if its actual state is `Respawning`. 

Alternatively, we could implement a `flag-respawning` filter with a `captured="true"` option.


calcastor:
Closing as there isn't a current use case, though if we ever decide to implement it I think it would make sense as part of `flag-respawning`.

PR #1662, opened 2026-04-23
Don't show particles for dropped flags when recover time is 0

calcastor:
On maps where the flag recover time is 0, the flag particle beam never shows for a flag in the `Dropped` state unless the match ends while someone is carrying the flag, so this PR makes `Dropped` only show the particle beam if the recover time isn't 0.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/flag/state/Dropped.java
+++ b/core/src/main/java/tc/oc/pgm/flag/state/Dropped.java
@@ -9,6 +9,7 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
 import net.kyori.adventure.text.format.TextColor;
 import org.bukkit.Location;
+import org.bukkit.entity.Player;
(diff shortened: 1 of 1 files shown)


core/src/main/java/tc/oc/pgm/flag/state/Dropped.java
```diff
@@ -83,6 +84,11 @@ public void leaveState() {
     super.leaveState();
   }
 
+  @Override
+  protected boolean canSeeParticles(Player player) {
+    return !Duration.ZERO.equals(this.getDuration());
+  }
+
```
Pablete1234, line 91: This is giving me a deja-vu vibe

We already had to add a thing similar to this in `Respawning`, and in that case it was about only showing particles if the match is running

Should we just make that the overall rule? just stop showing flag particles after match end in general?

calcastor: I think it makes sense to show them if the banner block physically exists.

Pablete1234 (approved):

PR #1663, opened 2026-04-24
Proto 1.5.1, register inner flag posts

### LINKED ISSUE #1538: Composite posts do not register child posts by id
The following is an excerpt from the include for the Touchdown gamemode:

```xml
<flags flag-proximity-metric="none" net-proximity-metric="none">
    <flag id="flag" name="Flag" color="${flag-color}" shared="true" pickup-kit="${flag-carrier-kit}" carry-kit="${flag-carrier-effects}" carry-message="`f`lYou are the `b`lFlag `f`lcarrier!">
        <post yaw="${flag-yaw}" respawn-time="${flag-respawn}" recover-time="0s" pickup-filter="any(flag-posts,all(player_state=${carry},alive),carrying-football)">
            <post id="default-post">${flag-post}</post>
            <if constant="team-flag-posts" constant-value="true">
                <post id="team-one-post" respawn-filter="never">${team-one-flag-post}</post>
                <post id="team-two-post" respawn-filter="never">${team-two-flag-post}</post>
            </if>
        </post>
        <unless constant="team-flag-posts" constant-value="true">
            <net points="${score-points}" region="team-one-portal" capture-filter="team-two"/>
            <net points="${score-points}" region="team-two-portal" capture-filter="team-one"/>
        </unless>
        <if constant="team-flag-posts" constant-value="true">
            <net points="${score-points}" region="team-one-portal" capture-filter="team-two" post="team-two-post"/>
            <net points="${score-points}" region="team-two-portal" capture-filter="team-one" post="team-one-post"/>
        </if>
    </flag>
</flags>
```

When cycling to a map which implements the relevant features from this piece of the include, an error is thrown citing: `'post' attribute of 'flag' > 'net' element @ line 804: No post with ID 'team-two-post'`

Pablete1234:
Fixes #1538 by registering inner flag posts, but it requires a proto bump as they may have ids that collide with other features.
Additionally given the required proto bump, i've made it so a few more inline filters/variables can make it as singleton ids: filters `gliding` and `riptiding`, and the variable `worldtime`


cswhite2000 (commented):
Overall looks good to me 👍
I do wonder if there is anything else we should add to this as well given we are bumping the proto

build is failing due to formatter

core/src/main/java/tc/oc/pgm/api/map/MapProtos.java
```diff
@@ -46,4 +46,10 @@ public interface MapProtos {
 
   // Several fixes to actions & scopes
   Version ACTION_REVAMP = new Version(1, 5, 0);
+
+  // Single flag posts inside composite posts also get registered
+  Version INNER_POST_IDS = new Version(1, 5, 1);
+
+  // Make more singletons have built-in default ids
+  Version FEATURE_SINGLETON_IDS_2 = new Version(1, 5, 1);
```
TTtie, line 54: I think it would be worthwhile to consider reserving something like a namespace of some sort (just like every decent programming language does with `std::...`, `java.*`, etc.) for builtins now that we're bumping the map proto here so we can add new ones without having to proto bump again in the future.

Otherwise, all OK by me

cswhite2000 (approved):

core/src/main/java/tc/oc/pgm/api/map/MapProtos.java
Pablete1234: i have mixed feelings about this, because if we start reserving the `pgm.` namespace, feels like new features will have to be pushed down into "pgm." prefix, and then when a new proto drops they can stop needing to add that prefix.

i still believe if this continues to happen, the proper solution will be to have "fallback ids" in the sense that pgm can register the names, but if the map defines it to something else, then the maps' decision wins.

Yes, this also brings potential confusion if a map overrides a base id with their own that does something completely different, its kind of a similar problem to old JS having writtable undefined, hence someone could `undefined = true` causing unexpected behavior everywhere, but i'm willing to hope that mapdevs are not just randomly assigning "maxbuildheight" to something completely arbitrary just to mess with the actual max build height varaible, and it would only affect their own map/xml.

Pablete1234:
> I do wonder if there is anything else we should add to this as well given we are bumping the proto

yea i'm wondering that too, these were the few that i could think of including, i'm open to suggestions if anyone else knows anything else minor that should be included

PR #1664, opened 2026-05-06
Add support for Spawning LivingEntities to xml spawners

cswhite2000:
Add support for spawning LivingEntities via xml spawners.

So you can do something like so:
```xml
<spawners>
    <spawner spawn-region="south-mob-point" player-region="south-mob" delay="4s">
        <mob type="Zombie" custom-name="Boss" custom-name-visible="true" baby="true"
             health="40" max-health="40" fire-ticks="600" can-pickup-items="false"/>
    </spawner>
</spawners>
```
Or for say modern:
```xml
<spawners>
    <spawner spawn-region="north-mob-point" player-region="north-mob" delay="4s">
        <mob type="Bee" silent="true" glowing="true" invulnerable="true" ai="true" anger="100" has-nectar="true" cannot-enter-hive-ticks="600"/>
    </spawner>
</spawners>
```

With this you should be able to spawn essentially any living entity with xml spawners. 

As for xml parameters, I made it so that you can use essentially any setter in the bukkit api for a living entity with a single parameter that is either a primitive type or an enum.

I've specifically excluded being able to spawn the classes: Player, Wither, and Ender Dragon

### DIFF
--- a/core/src/main/java/tc/oc/pgm/entity/MobProperties.java
+++ b/core/src/main/java/tc/oc/pgm/entity/MobProperties.java
@@ -1,0 +1,154 @@
+package tc.oc.pgm.entity;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import org.bukkit.DyeColor;
+import org.bukkit.entity.Ageable;
+import org.bukkit.entity.ArmorStand;
+import org.bukkit.entity.Bat;
+import org.bukkit.entity.Creeper;
+import org.bukkit.entity.Damageable;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.Guardian;
+import org.bukkit.entity.Horse;
+import org.bukkit.entity.IronGolem;
+import org.bukkit.entity.LivingEntity;
+import org.bukkit.entity.Ocelot;
+import org.bukkit.entity.Pig;
+import org.bukkit.entity.PigZombie;
+import org.bukkit.entity.Rabbit;
+import org.bukkit.entity.Sheep;
+import org.bukkit.entity.Slime;
+import org.bukkit.entity.Tameable;
+import org.bukkit.entity.Wolf;
+import org.bukkit.entity.Zombie;
+import org.bukkit.material.Colorable;
+import org.jdom2.Element;
+import org.jetbrains.annotations.Nullable;
+import tc.oc.pgm.entity.MobProperty.Parser;
+import tc.oc.pgm.util.platform.Platform;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+import tc.oc.pgm.util.xml.Node;
+import tc.oc.pgm.util.xml.XMLUtils;
+
+public abstract class MobProperties {
+  protected static final Parser<Boolean> BOOL = XMLUtils::parseBoolean;
+  protected static final Parser<Integer> INT = n -> XMLUtils.parseNumber(n, Integer.class);
+  protected static final Parser<Double> DOUBLE = n -> XMLUtils.parseNumber(n, Double.class);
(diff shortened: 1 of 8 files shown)


Pablete1234:
This supports a ton of properties, but it seems to ignore items or other "kits" entirely.

Would it also make sense to have extensions for armor pieces? Similar to a kit's syntax but supporting only the armor pieces, a mainhand and an offhand (modern only) item?

Note that this may depend on entities: eg zombies take in helmet, chestplate, etc while llamas or pigs take saddle, and horses take a "body" armor piece as well as a saddle.

Also thinking about potion effects, like giving the entity invisibility seems like a pretty useful usecase.

Maybe a light rework of some kits such that they support applying to LivingEntity (item, armor, potions) ?

cswhite2000:
Yeah that would be the next step, I just didn't want to make this pr too big.

core/src/main/java/tc/oc/pgm/entity/MobProperty.java
```diff
@@ -0,0 +1,25 @@
...
+  public interface Parser<V> {
+    V parse(Node node) throws InvalidXMLException;
+  }
+
+  public record Applied<E, V>(MobProperty<E, V> property, V value) {
+    @SuppressWarnings("unchecked")
+    public void applyTo(Entity entity) {
+      property.setter().accept((E) entity, value);
+    }
+  }
```
Pablete1234: i think this should be just a `Consumable<Entity>`, and not take in the whole MobProperty but rather just the BiConsumer + Value and turn it into a Consumer; in other words, decouple the parsing part (mob property having the xml name, parser, etc) with the applying part (which requires just value and setter)

core/src/main/java/tc/oc/pgm/entity/MobProperties.java
```diff
@@ -0,0 +1,155 @@
...
+        throw new InvalidXMLException(
+            "Attribute '" + attr.getName() + "' is not a valid property for "
+                + type.getSimpleName(),
+            el);
+      }
+    }
+    return applied;
+  }
+
+  private record Entry<E, V>(Class<E> ownerType, MobProperty<E, V> property) {}
```
Pablete1234: Seems like an overdo of composition? why not put owner type directly in MobProperty ?

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnableMob.java
```diff
@@ -0,0 +1,42 @@
...
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.entity.MobProperty;
+import tc.oc.pgm.spawner.Spawnable;
+import tc.oc.pgm.spawner.Spawner;
+
+public class SpawnableMob implements Spawnable {
+
+  private final Class<? extends LivingEntity> entityType;
+  private final List<MobProperty.Applied<?, ?>> properties;
```
Pablete1234: Maybe these can be bundled together in some kind of SpawnableEntity record with just some `spawn(location)` method that applies all the properties.
Ideally, it should use the spawn method that sets the properties before the entity is spawned, not after, but that only exists in modern paper. Would be cool if sportpaper got it too and we could use it as part of intersect jar.

Reason i say this is cause it feels like the "mob property" system is what's separated out of spawners, but i think "spawn this entity" is what should be, as a whole, extracted

core/src/main/java/tc/oc/pgm/entity/MobProperty.java
cswhite2000: Makes sense, will do

core/src/main/java/tc/oc/pgm/spawner/objects/SpawnableMob.java
cswhite2000: That would be nice to be able to set the properties before the entity is spawned. But yeah I don't think I'll be doing that as part of this pr. 

I'll see what I can do with a SpawnableEntity record though.

Pablete1234 (approved):

PR #1665, opened 2026-05-10
Platform-specific listeners for block physics and BlockTransformListener

calcastor:
Permits for platform-specific `BlockTransformListener` extends where events that are unique to specific platforms can be also be listened on. Additionally adds listeners for  `BlockExplodeEvent` and `LeavesDecayEvent` for both modern and SportPaper.

Also implements `pre-match-physics` from ProjectAres with a default of false as was the case in ProjectAres, fixing things like water flowing in undesirable scenarios in legacy maps on the modern platform (covered by `BlockFromToEvent`). Additionally we cancel `BlockDestroyEvent` on modern when pre-match physics is disabled to avoid legacy maps with unsupported blocks like redstone wire from breaking, but we update to block to maintain its correct block connections.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/PGMPlugin.java
+++ b/core/src/main/java/tc/oc/pgm/PGMPlugin.java
@@ -48,7 +48,6 @@ import tc.oc.pgm.db.SQLDatastore;
 import tc.oc.pgm.db.SqlUsernameResolver;
 import tc.oc.pgm.integrations.SimpleVanishIntegration;
 import tc.oc.pgm.listeners.AntiGriefListener;
-import tc.oc.pgm.listeners.BlockTransformListener;
 import tc.oc.pgm.listeners.FormattingListener;
 import tc.oc.pgm.listeners.InitialMatchLoader;
 import tc.oc.pgm.listeners.JoinLeaveAnnouncer;
@@ -216,7 +215,7 @@ public class PGMPlugin extends JavaPlugin implements PGM, Listener {
     final File[] worldDirs = getServer().getWorldContainer().listFiles();
     if (worldDirs != null) {
       for (File dir : worldDirs) {
-        if (dir.isDirectory() && dir.getName().startsWith("match")) {
+        if (dir.isDirectory() && Match.isMatchWorld(dir.getName())) {
           FileUtils.delete(dir);
         }
       }
@@ -413,7 +412,6 @@ public class PGMPlugin extends JavaPlugin implements PGM, Listener {
     registerEvents(new PlayerMoveListener());
     registerEvents(new ItemTransferListener());
     registerEvents(new TNTMinecartPlacementListener());
-    new BlockTransformListener(this).registerEvents();
     registerEvents(matchManager);
     inventoryManager.init();
     registerEvents(afkTracker);
--- a/core/src/main/java/tc/oc/pgm/listeners/BlockPhysicsListener.java
+++ b/core/src/main/java/tc/oc/pgm/listeners/BlockPhysicsListener.java
@@ -1,0 +1,33 @@
+package tc.oc.pgm.listeners;
+
+import org.bukkit.World;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.Listener;
+import org.bukkit.event.block.BlockFromToEvent;
+import org.bukkit.event.block.BlockPhysicsEvent;
+import org.jspecify.annotations.NullMarked;
+import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.match.Match;
+
+@NullMarked
+public class BlockPhysicsListener implements Listener {
+
+  protected boolean allowPhysics(World world) {
+    Match match = PGM.get().getMatchManager().getMatch(world);
+    if (match == null) return !Match.isMatchWorld(world);
+    if (match.isFinished()) return false;
+    if (match.isRunning()) return true;
+    return match.getMap().getWorld().initialPhysics();
+  }
+
+  @EventHandler(priority = EventPriority.LOWEST)
+  public void onBlockPhysics(BlockPhysicsEvent event) {
+    if (!allowPhysics(event.getBlock().getWorld())) event.setCancelled(true);
+  }
+
+  @EventHandler(priority = EventPriority.LOWEST)
+  public void onBlockFromTo(BlockFromToEvent event) {
+    if (!allowPhysics(event.getBlock().getWorld())) event.setCancelled(true);
+  }
+}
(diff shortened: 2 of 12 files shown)


cswhite2000 (dismissed):

core/src/main/java/tc/oc/pgm/listeners/BlockTransformListener.java
```diff
@@ -400,6 +415,16 @@ private void finishEntityExplode(
     }
   }
 
+  protected void finishBlockExplode(
```
cswhite2000: This is basically a duplicate of `finishEntityExplode`. why not for this impl refactor out a helper method that both call?

core/src/main/java/tc/oc/pgm/listeners/BlockPhysicsListener.java
```diff
@@ -0,0 +1,44 @@
...
+    this.plugin = plugin;
+  }
+
+  public void registerEvents() {
+    plugin.getServer().getPluginManager().registerEvents(this, plugin);
+  }
+
+  protected boolean allowPhysics(World world) {
+    Match match = PGM.get().getMatchManager().getMatch(world);
+    if (match == null) return !world.getName().startsWith("match");
```
cswhite2000: I feel like we should have a util instead of always using `startsWith("match");` to check if its a match world

cswhite2000 (dismissed):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernBlockPhysicsListener.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @EventHandler(priority = EventPriority.LOWEST)
+  public void onFluidLevelChange(FluidLevelChangeEvent event) {
+    if (!allowPhysics(event.getBlock().getWorld())) event.setCancelled(true);
+  }
+
+  @EventHandler(priority = EventPriority.LOWEST)
+  public void onBlockDestroy(BlockDestroyEvent event) {
+    if (allowPhysics(event.getBlock().getWorld())) return;
```
Pablete1234: I'm not convinced that the update shape logic should be gated behind allowPhysics:
Before match start, physics are (usually, but not always) disabled, during the match however, they're enabled. Does this mean that if the map is large enough that it doesn't all get loaded on match start, later when players get close to the floating redstone it will all pop-out and pgm will be ok with it?

To me it seems like this event should somehow be related/gated to chunk load events, where if the chunk is being loaded in this tick we apply this logic to prevent redstone breaking and still connect it

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernBlockTransformListener.java
```diff
@@ -0,0 +1,173 @@
...
+          event,
+          event.getBlock().getState(),
+          event.getNewState(),
+          entity == null ? null : Trackers.getOwner(entity));
+    }
+  }
+
+  @EventWrapper
+  public void onSpongeAbsorb(final SpongeAbsorbEvent event) {
+    for (BlockState newState : event.getBlocks()) {
```
Pablete1234, line 165: does this mean that if *any* of the absorbed blocks is denied, the whole event is denied? should it instead work like explosions where only the denied blocks don't go away but the undenied are still removed?

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernBlockTransformListener.java
```diff
@@ -0,0 +1,173 @@
...
+  @EventWrapper
+  public void onMoistureChange(final MoistureChangeEvent event) {
+    callEvent(new BlockTransformEvent(event, event.getBlock().getState(), event.getNewState()));
+  }
+
+  @EventWrapper
+  public void onBlockFertilize(final BlockFertilizeEvent event) {
+    for (BlockState newState : event.getBlocks()) {
+      callEvent(event, newState.getBlock().getState(), newState, event.getPlayer());
+    }
```
Pablete1234, line 183: same here, should it allow some and block some without being an all-or-nothing?

calcastor: I'm not sure; we didn't do this for onStructureGrow before.

Given that we didn't seem to handle that event in ProjectAres it may make sense to make it behave like that too.

core/src/main/java/tc/oc/pgm/PGMPlugin.java
```diff
@@ -413,7 +413,8 @@ private void registerListeners() {
     registerEvents(new PlayerMoveListener());
     registerEvents(new ItemTransferListener());
     registerEvents(new TNTMinecartPlacementListener());
-    new BlockTransformListener(this).registerEvents();
+    PlatformListenerFactory.FACTORY.createBlockPhysicsListener(this).registerEvents();
+    PlatformListenerFactory.FACTORY.createBlockTransformListener(this).registerEvents();
```
Pablete1234: any real need for these to be explicit and part of a diff factory?
you have `Platform.MANIFEST.onEnable(this)` just a few lines above, you could register version-specifid listeners there

calcastor: I was under the impression that ordering may matter, but if it does not then I can simplify it.

Pablete1234: for some listeners it does matter, for others it doesn't, and finding out which is which may not be trivial

arcadeboss:
@calcastor - TT pointed me to this PR when we were discussing newly broken redstone in maps.  I noticed that the minecarts in Red Rover are no longer moving, which is an aesthetic bummer.  And Fall Guys seems to have also suffered from some game-breaking redstone mechanics failures.

I don't want to have to add a bunch of stuff to fix, as seemed to be the suggestion from TT/Javg.   Maybe it would be nice to have an opt-out flag, as that could allow for an easy resolution from the XML side, if that is possible.

calcastor:
> @calcastor - TT pointed me to this PR when we were discussing newly broken redstone in maps. I noticed that the minecarts in Red Rover are no longer moving, which is an aesthetic bummer. And Fall Guys seems to have also suffered from some game-breaking redstone mechanics failures.
> 
> I don't want to have to add a bunch of stuff to fix, as seemed to be the suggestion from TT/Javg. Maybe it would be nice to have an opt-out flag, as that could allow for an easy resolution from the XML side, if that is possible.

The issue doesn't seem to come from this PR; I can reproduce the Red Rover issue on the dev branch.

Pablete1234 (approved):

PR #1666, opened 2026-05-10
Implement variable duration configs

Pablete1234:
Allows multiple configs to support either a fixed duration, or a list of different durations based on the current playerount or match length.

Allows for longer cycle after a long match, shorter cycles on lower playercount times, etc

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/VariableDuration.java
+++ b/core/src/main/java/tc/oc/pgm/api/VariableDuration.java
@@ -1,0 +1,53 @@
+package tc.oc.pgm.api;
+
+import static tc.oc.pgm.util.text.TextParser.parseDuration;
+
+import java.time.Duration;
+import org.bukkit.configuration.ConfigurationSection;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.util.VariableDurationImpl;
+import tc.oc.pgm.util.text.TextException;
+
+/** A duration that can vary based on the number of players and the length of the match. */
+public interface VariableDuration {
+  /**
+   * Gets the duration for the given match.
+   *
+   * @param match The match.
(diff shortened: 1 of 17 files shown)


core/src/main/resources/config.yml
```diff
@@ -48,11 +48,21 @@ map:
 #  "1m5s" = 1 minute and 5 seconds
 #  "0s"   = immediately
 #  -1     = disabled
```
cswhite2000, line 50: `-1 = disabled` is broken by this code change, at least for start time. It uses the default of 30s

interestingly `"-1"` does work, while `-1` does not

core/src/main/java/tc/oc/pgm/api/VariableDuration.java
```diff
@@ -0,0 +1,53 @@
...
+   * @param object The object to parse (String or ConfigurationSection).
+   * @param def The default duration string if the object is null.
+   * @return The parsed variable duration.
+   * @throws TextException If the object is not a valid variable duration.
+   */
+  static VariableDuration parse(Object object, String def) throws TextException {
+    return switch (object) {
+      case ConfigurationSection cs -> VariableDurationImpl.parse(cs);
+      case String str -> VariableDurationImpl.of(parseDuration(str));
+      case null, default -> VariableDurationImpl.of(parseDuration(def));
```
cswhite2000: I think this is the part where the `-1` defaults get ignored, as I believe the object is an integer, not a configuration section

Pablete1234: makes sense, will rework

cswhite2000 (dismissed):

cswhite2000 (approved):

PR #1667, opened 2026-05-10
Emulate PlayerOnGroundEvent in modern from movement packets

calcastor:
`EntityPoseChangeEvent` isn't necessarily an ideal replacement for SportPaper's `PlayerOnGroundEvent`, but it's not really possible to replicate that one-to-one without modifying Paper, and given the age of that API it seems like there would be a reason it is not in Spigot or Paper by now. This PR combines a PacketEvents movement packets listener with Bukkit event listeners to try to approximate the behaviour of `PlayerOnGroundEvent` more closely.

This should see some more focused testing in production-level scenarios, but the goal is to attempt to address incorrect kill credits in modern.

### DIFF
--- a/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/OnGroundListener.java
+++ b/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/OnGroundListener.java
@@ -1,0 +1,146 @@
+package tc.oc.pgm.platform.modern.packets;
+
+import com.destroystokyo.paper.event.player.PlayerJumpEvent;
+import com.github.retrooper.packetevents.event.PacketListenerCommon;
+import com.github.retrooper.packetevents.event.PacketListenerPriority;
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.wrapper.play.client.WrapperPlayClientPlayerFlying;
+import io.papermc.paper.event.player.PlayerFailMoveEvent;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import org.bukkit.Bukkit;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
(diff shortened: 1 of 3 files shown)


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/OnGroundListener.java
```diff
@@ -0,0 +1,148 @@
...
+@SuppressWarnings("deprecation")
+public class OnGroundListener implements Listener {
+
+  private final PacketListenerCommon listener;
+  private final Map<UUID, Boolean> pendingOnGround = new ConcurrentHashMap<>();
+  private final Map<UUID, Boolean> lastOnGround = new ConcurrentHashMap<>();
+
+  public OnGroundListener() {
+    Bukkit.getOnlinePlayers()
+        .forEach(player -> lastOnGround.put(player.getUniqueId(), player.isOnGround()));
```
cswhite2000: Won't this run before there are any players on the server?

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/OnGroundListener.java
```diff
@@ -0,0 +1,148 @@
...
+    Player player = Bukkit.getPlayer(playerId);
+    if (player == null) {
+      removeAll(playerId);
+      return;
+    }
+
+    callIfCurrent(player, onGround);
+  }
+
+  private void callIfCurrent(Player player, boolean onGround) {
```
cswhite2000: nit: name is a bit unclear, something like `emitIfChanged` would be more clear

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/packets/OnGroundListener.java
calcastor: It was there in the event of a hard reload but I suppose that isn't something Paper supports anyway.

cswhite2000 (dismissed):

Pablete1234:
For reference, i'm still researching this one to see if we can get an impl that does not depend on packet listening which is why i'm holding on reviewing for now

PR #1668, opened 2026-05-10
Add kits for mobs

cswhite2000:
Updated:
Switched to applying regular kits insead of making a new mob-kits

Sample xml:

```xml
<kits>
    <kit id="boss-skeleton">
        <helmet material="iron_helmet" drop-chance="0"/>
        <chestplate material="iron_chestplate" drop-chance="0"/>
        <item slot="weapon.mainhand" material="iron_sword" unbreakable="true" drop-chance="0.7"/>
        <effect duration="9999">strength</effect>
        <attribute operation="add" amount="10">generic.max_health</attribute>
    </kit>
</kits>

<spawners>
    <spawner spawn-region="..." player-region="..." delay="4s">
        <mob type="Skeleton" custom-name="Captain" kit="boss-skeleton"/>
    </spawner>

    <spawner spawn-region="..." player-region="..." delay="4s">
        <mob type="WitherSkeleton" custom-name="King" kit="boss-skeleton">
            <kit>
                <effect duration="9999" amplifier="2">speed</effect>
                <attribute operation="add" amount="5">generic.attack_damage</attribute>
            </kit>
        </mob>
    </spawner>

    <spawner spawn-region="..." player-region="..." delay="4s">
        <mob type="Spider" custom-name="Swift">
            <kit attribute="generic.movement_speed:multiply:1.5;generic.attack_damage:add:3"/>
        </mob>
    </spawner>
</spawners>
```

Supports:

- Referencing a shared kit via `kit="..."` on `<mob>`, and/or inline `<kit>` children
- Equipment: `<helmet>`, `<chestplate>`, `<leggings>`, `<boots>`, and
  `<item slot="weapon.mainhand|weapon.offhand">`, each with an optional `drop-chance`
- `<effect>`, `<attribute>`, `<health>` and `<max-health>`
- Mob-specific gear directly on `<mob>` (properties, not kit content): `<saddle>`,
  `<horse-armor>`, `<llama-decor>`


core/src/main/java/tc/oc/pgm/entity/SpawnableEntity.java
```diff
@@ -1,31 +1,54 @@
...
+import tc.oc.pgm.entity.kits.MobKitParser;
 import tc.oc.pgm.util.xml.InvalidXMLException;
 import tc.oc.pgm.util.xml.Node;
 import tc.oc.pgm.util.xml.XMLUtils;
 
 public record SpawnableEntity(
-    Class<? extends LivingEntity> entityType, List<Consumer<Entity>> properties) {
+    Class<? extends LivingEntity> entityType,
+    List<Consumer<Entity>> properties,
+    List<MobKit> kits) {
```
Pablete1234: Realistically, what's the difference between these two?

They're all a bunch of `Consumer<Entity>` one way or another, you could say properties are a kind of kit, or kits are a kind of property

cswhite2000: Right, I kept it separate for post validation of references, but yeah, for application they can be the same thing.

TheRealPear:
I understand why it was done, but I'm not a big fan of the whole mob-kits module being a separate thing

cswhite2000:
Could you go into more detail about why you think integrating it into `<kits>` would be better?
This isn't currently merged, so its not set in stone

core/src/main/java/tc/oc/pgm/entity/SpawnableEntity.java
```diff
@@ -1,31 +1,99 @@
...
+      kits.add(kitParser.parse(kitEl));
+    }
+
+    // Kit contents can't be inspected until references resolve, so mob compatibility
+    // is validated through the feature context, which defers until after resolution
+    FeatureValidation<KitDefinition> validation = (def, node) -> def.validateMob(type, node);
+    for (Kit kit : kits) {
+      if (kit instanceof XMLKitReference ref) factory.getFeatures().validate(ref, validation);
+      else if (kit instanceof KitDefinition def)
+        factory.getFeatures().validate(def, validation, new Node(el));
```
Pablete1234: It appears the logic of:
if reference -> call one validate
if definition -> call another validate
else -> throw

exists in at least 3 places (FeatureRegionParser, FeatureFilterParser, and now here).

Feels like `validate` with a FeatureDefinition, in in feature definition context, should check if that definition implements xml reference, and if so, consider it as a reference instead of as a definition, such that it has to call resolve and get before validating.

core/src/main/java/tc/oc/pgm/entity/SpawnableEntity.java
```diff
@@ -1,31 +1,99 @@
...
+  }
+
+  private static void parseHorseEquipment(
+      Element el,
+      Class<? extends LivingEntity> type,
+      KitParser kitParser,
+      List<Consumer<Entity>> properties)
+      throws InvalidXMLException {
+    Element saddleEl = el.getChild("saddle");
+    Element armorEl = el.getChild("horse-armor");
```
Pablete1234:
```suggestion
    Element armorEl = el.getChild("body");
```
In modern, this is regarded as the "armor.body" slot, and i think it fits better (more than just horses are able to have stuff set in that equipment slot, like wolves or nautilus).

core/src/main/java/tc/oc/pgm/entity/SpawnableEntity.java
```diff
@@ -1,31 +1,99 @@
...
+      }
+      ItemStack saddle = saddleEl == null ? null : kitParser.parseItem(saddleEl, false);
+      ItemStack armor = armorEl == null ? null : kitParser.parseItem(armorEl, false);
+      properties.add(e -> {
+        Horse horse = (Horse) e;
+        if (saddle != null) horse.getInventory().setSaddle(saddle);
+        if (armor != null) horse.getInventory().setArmor(armor);
+      });
+    }
+    Element decorEl = el.getChild("llama-decor");
```
Pablete1234: llamas use the "body" slot for the carpet too

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -234,8 +234,14 @@ private ArmorKit.ArmorItem parseArmorItem(Element el) throws InvalidXMLException
...
     boolean locked = XMLUtils.parseBoolean(el.getAttribute("locked"), false);
+    Float dropChance = parseDropChance(el);
 
-    return new ArmorKit.ArmorItem(stack, locked);
+    return new ArmorKit.ArmorItem(stack, locked, dropChance);
+  }
+
+  protected @Nullable Float parseDropChance(Element el) throws InvalidXMLException {
+    var attr = el.getAttribute("drop-chance");
+    return attr == null ? null : XMLUtils.parseNumber(attr, Float.class, Range.closed(0f, 1f));
```
```suggestion
    return parser.parseFloat(el, "drop-chance").between(Range.closed(0f, 1f)).orNull();
```
Also should probably make that range a constant outside of this

core/src/main/java/tc/oc/pgm/kits/EquipmentKit.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override
+  public boolean mobCompatible() {
+    return true;
+  }
+
+  @Override
+  public void validateMob(Class<? extends LivingEntity> mobType, Node node)
+      throws InvalidXMLException {
+    if (slots.containsKey(Slot.Saddle.saddle()) && !Horse.class.isAssignableFrom(mobType)) {
```
Pablete1234: There's alot more mobs that support saddles, mainly:
 - Horses (including mules/donkeys),
 - camels
 - pigs
 - nautilus
 - strider
 - copper golem (it can hold a red rose in the antenna if given by an iron golem, and that's the saddle slot)

Seems like it should be a version-specific handler for who can use this slot, for legacy it needs a couple branches, while for modern it can probably just set the slot directly via the api

core/src/main/java/tc/oc/pgm/kits/EquipmentKit.java
```diff
@@ -0,0 +1,73 @@
...
+      throw new InvalidXMLException(
+          "saddle requires a horse, got " + mobType.getSimpleName(), node);
+    }
+    if (slots.containsKey(Slot.Body.body())
+        && !Horse.class.isAssignableFrom(mobType)
+        && !EntityEquipmentUtil.EQUIPMENT.isLlama(mobType)) {
```
Pablete1234: Body can be used by, at least (may be forgetting some more):
- horse (horse armor)
- llama (carpet)
- nautilus (nautilus armor)
- wolf (wolf armor) (modern only) 
- happy ghast (harness) (yes, the harness goes in the body slot not the saddle one, don't ask me why)
- sulfur cube (held block)

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -228,29 +229,50 @@ public FlyKit parseFlyKit(Element parent) throws InvalidXMLException {
...
-    for (Slot.Armor armorSlot : Slot.Armor.armor().toList()) {
-      var armorItem = parseArmorItem(el.getChild(armorSlot.armorTypeName()));
-      if (armorItem != null) armor.put(armorSlot, armorItem);
+  private EquipmentKit.EquipmentItem parseEquipmentItem(Element el, Slot slot)
+      throws InvalidXMLException {
+    ItemStack stack = parseEquipmentStack(el);
+    Float dropChance = parseDropChance(el);
+    if (dropChance != null && slot instanceof Slot.MobEquipment) {
+      throw new InvalidXMLException("drop-chance is not supported for " + slot, el);
     }
```
Pablete1234: don't know about legacy, but in modern nms [all EquipmentSlots can get a drop chance](https://mcsrc.dev/1/26.2/net/minecraft/world/entity/DropChances#L9)

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -261,6 +283,11 @@ public ItemKit parseItemKit(Element el) throws InvalidXMLException {
       ItemStack item = this.parseItemStack(itemEl);
 
       if (item != null) {
+        if (itemEl.getAttribute("drop-chance") != null) {
```
Pablete1234: may as well not parse this, it should show up as an unused node

core/src/main/java/tc/oc/pgm/features/FeatureDefinitionContext.java
```diff
@@ -161,10 +161,34 @@ public <T extends FeatureDefinition> void validate(
...
-      FeatureReference<T> reference, FeatureValidation<T> validation) throws InvalidXMLException {
+      FeatureReference<? extends T> reference, FeatureValidation<T> validation)
+      throws InvalidXMLException {
     validations.add(new PendingValidation<>(reference, validation, reference.getNode()));
   }
 
+  /** Validate a feature that may be either a definition or an unresolved reference */
+  public <T extends FeatureDefinition> void validate(
+      Class<T> type, Object feature, FeatureValidation<T> validation, Node node)
+      throws InvalidXMLException {
```
Pablete1234: I don't believe the new method to be needed, i think the existing method that takes in the definition should be the one having this behavior.

This whole thing is only an issue in cases where you have something like "Filter" which is then implemented by either "FilterDefinition" or "XmlFilterRefernce". The thing passed as param can be a `Filter` and just needs to be checked for being a reference.


```java
  @SuppressWarnings("unchecked")
  public <T extends FeatureDefinition> void validate(
      Object definition, FeatureValidation<T> validation, Node node) throws InvalidXMLException {
    if (definition instanceof XMLFeatureReference<?> ref) {
      validations.add(new PendingValidation<>((XMLFeatureReference<T>) ref, validation, node));
    } else if (definition instanceof FeatureDefinition fd) {
      validations.add(new PendingValidation<>((T) fd, validation, node));
    } else {
      throw new IllegalStateException("...");
    }
  }
```
Potentially try to catch the cast, but i assume due to erasure that's never going to actually throw until you actually try to run the validation

core/src/main/java/tc/oc/pgm/entity/SpawnableEntity.java
```diff
@@ -1,31 +1,65 @@
...
+    if (kitAttr != null) {
+      kits.add(kitParser.parseReference(new Node(kitAttr), kitAttr.getValue()));
+    }
+    for (Element kitEl : el.getChildren("kit")) {
+      kits.add(kitParser.parse(kitEl));
+    }
+    var equipment = kitParser.parseEquipmentKit(el);
+    if (equipment != null) {
+      kits.add(equipment);
+    }
```
Pablete1234: this looks problematic, i think the kit itself needs to be parsing this, not parsing separately, otherwise a reference will fail, eg:

```xml
<kits>
  <kit id="horse-kit">
    <saddle material="saddle"/>
    <body material="diamond horse armor"/>
  </kit>
</kits>

<spawners>
  <spawner ...>
    <mob type="horse" kit="horse-kit"/>
  </spawner>
</spanwers>
```


Also generally seems like this should just be using the XmlFluentParser to support a kit "property" (which is then parsed as either a reference or an inline child):
`parser.kit(el, "kit").optional(KitNode.EMPTY)`

this will give you a single kit (which internally is a kit node, with a list of however many individual kits) which should also make it easier to apply,

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -228,29 +229,51 @@ public FlyKit parseFlyKit(Element parent) throws InvalidXMLException {
...
     }
+    return new EquipmentKit.EquipmentItem(stack, dropChance);
+  }
 
-    if (!armor.isEmpty()) {
-      return new ArmorKit(armor);
-    } else {
-      return null;
-    }
+  private ItemStack parseEquipmentStack(Element el) throws InvalidXMLException {
```
Pablete1234: imo this should work not just for "equipment" stacks, but for parsing items in general

it makes stuff like shops, killrewards, spawners have more flexibility for using all of these things in their items

eg: spawner requires `<item ...>` so it benefits too from allowing other special parsers to trigger even for "item"

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/entity/SpawnableEntity.java
```diff
@@ -7,25 +7,41 @@
...
+
+    List<Consumer<Entity>> properties =
+        MobProperties.MOB_PROPERTIES.parseAttributes(type, el, "kit");
+
+    Kit kit = factory.getParser().kit(el, "kit").optional(KitNode.EMPTY);
+
+    // Kit contents can't be inspected until references resolve, so mob compatibility
+    // is validated through the feature context, which defers until after resolution
+    FeatureValidation<KitDefinition> validation = (def, node) -> def.validateMob(type, node);
+    factory.getFeatures().validate(kit, validation, new Node(el));
```
Pablete1234, line 42: Would be nice to add as a .validate in kits, which could be used for parsing dynamic kits in the future too, ie:

```suggestion
    Kit kit = factory.getParser().kit(el, "kit").mob(type).optional(KitNode.EMPTY);
```

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -228,29 +229,35 @@ public FlyKit parseFlyKit(Element parent) throws InvalidXMLException {
...
+    return new EquipmentKit.EquipmentItem(stack, dropChance);
+  }
 
-    if (!armor.isEmpty()) {
-      return new ArmorKit(armor);
-    } else {
-      return null;
-    }
+  protected @Nullable Float parseDropChance(Element el) throws InvalidXMLException {
+    return parser.parseFloat(el, "drop-chance").between(DROP_CHANCE_RANGE).orNull();
```
Pablete1234, line 260: given this now has a single usecase, could be inlined

core/src/main/java/tc/oc/pgm/kits/KitParser.java
```diff
@@ -228,29 +229,35 @@ public FlyKit parseFlyKit(Element parent) throws InvalidXMLException {
...
-    for (Slot.Armor armorSlot : Slot.Armor.armor().toList()) {
-      var armorItem = parseArmorItem(el.getChild(armorSlot.armorTypeName()));
-      if (armorItem != null) armor.put(armorSlot, armorItem);
+  private EquipmentKit.EquipmentItem parseEquipmentItem(Element el, Slot slot)
+      throws InvalidXMLException {
+    ItemStack stack = parseItem(el, true);
+    Float dropChance = parseDropChance(el);
+    if (dropChance != null && !slot.supportsDropChance()) {
+      throw new InvalidXMLException(
+          "drop-chance for " + slot + " is not supported on this server version", el);
```
Pablete1234, line 254: I wonder if it'd be better to no-op for unsupported than failing, and this goes for a few things in general (offhand, saddles on specific mobs, etc), i'm not sure if it makes sense to more silently fail than for pgm to complain about everything

util/src/main/java/tc/oc/pgm/util/inventory/Slot.java
```diff
@@ -49,6 +52,29 @@ public abstract class Slot {
...
+
+  /** Lazy holder, so the {@link Slot} class initializer never references its subclasses */
+  private static final class EquipmentTags {
+    static final Map<String, Slot> BY_TAG = build();
+
+    private static Map<String, Slot> build() {
+      var tags = new LinkedHashMap<String, Slot>();
+      Armor.armor().forEach(armor -> tags.put(armor.armorTypeName(), armor));
+      tags.put("mainhand", MainHand.mainHand());
+      tags.put("offhand", OffHand.offHand().orElse(null));
```
Pablete1234, line 71: Should it jsut be ignored if null (ie: legacy) instead of keeping the offhand -> null key?

Pablete1234 (approved):
Minor suggestions, but its approved as they aren't really that important to the overall pr

core/src/main/java/tc/oc/pgm/kits/KitParser.java
cswhite2000: Yeah I wasn't sure either

PR #1671, opened 2026-05-13
Mark all legacy leaves as persistent at conversion

calcastor:
Leaf blocks with certain data values in legacy worlds don't decay, but when these worlds are upgraded on modern many blocks which did not decay in legacy do not maintain persistency. It's generally safe to assume all leaf blocks in PGM maps are not meant to decay, so mark all legacy leaves as persistent when we upgrade the world.


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/PgmBootstrap.java
```diff
@@ -34,6 +36,7 @@ public class PgmBootstrap implements PluginBootstrap {
 
   private static final String NAMESPACE = "pgm";
   private static final String PATH = "legacy_overworld";
+  private static final int V1500 = 1500;
```
Pablete1234: put this in DataVersions and rename it to whatever snapshot or version it happens to correspond with

Pablete1234 (approved):

PR #1672, opened 2026-05-14
Resolve several participant-observer barrier issues on modern

calcastor:
1. Prevent observers from triggering sculk sensors and sculk shriekers
2. Cancel knockback from explosions for observers
3. Remove match participants as mob targets if they leave the team and go to observers, or once the match ends

### DIFF
--- a/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/ModernMobsMatchModule.java
+++ b/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/ModernMobsMatchModule.java
@@ -1,0 +1,61 @@
+package tc.oc.pgm.platform.modern.modules;
+
+import java.util.Collection;
+import java.util.List;
+import org.bukkit.entity.Mob;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.Listener;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchModule;
+import tc.oc.pgm.api.match.MatchScope;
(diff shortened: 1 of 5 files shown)


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/ModernMobsMatchModule.java
```diff
@@ -0,0 +1,74 @@
...
+        event.getMatch().getParticipants().stream()
+            .map(player -> player.getBukkit().getUniqueId())
+            .collect(Collectors.toSet()));
+  }
+
+  private void clearMobTargets(World world, Player target) {
+    clearMobTargets(world, Set.of(target.getUniqueId()));
+  }
+
+  private void clearMobTargets(World world, Collection<UUID> targetUUIDs) {
```
Pablete1234: you only ever pass either one target uuid, or every online player.
Change it so you pass a single UUID, and if null then assume all targets should be reset

cswhite2000 (dismissed):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/ModernMobsMatchModule.java
```diff
@@ -0,0 +1,63 @@
...
+    clearMobTargets(event.getMatch(), null);
+  }
+
+  private void clearMobTargets(Match match, @Nullable Player targetPlayer) {
+    for (var entity : match.getWorld().getLivingEntities()) {
+      if (entity instanceof Mob mob
+          && mob.getTarget() instanceof Player target
+          && (targetPlayer == null
+              ? match.getParticipantState(target) != null
+              : targetPlayer.equals(target))) {
```
Pablete1234:
```suggestion
          && (targetPlayer == null || targetPlayer.equals(target))) {
```

Pablete1234 (approved):

PR #1673, opened 2026-05-17
Add support for multiple username APIs

Pablete1234:
Instead of relying on just electroid's api for username resolving, pgm adopts 3 different supported apis for resolving.
Support for ratelimiting is not tied specifically to mojang's api and can work for any ratelimited api.
Could offer support for config-based apis, or at least allow configuring if you want to enable/disable a specific resolver.

### DIFF
--- a/util/src/main/java/tc/oc/pgm/util/usernames/ApiUsernameResolver.java
+++ b/util/src/main/java/tc/oc/pgm/util/usernames/ApiUsernameResolver.java
@@ -4,6 +4,7 @@ import static tc.oc.pgm.util.Assert.assertNotNull;
 import static tc.oc.pgm.util.Assert.assertTrue;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import java.io.IOException;
 import java.net.NoRouteToHostException;
@@ -19,13 +20,13 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-import java.util.logging.Level;
-import org.bukkit.Bukkit;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import org.bukkit.plugin.Plugin;
 import tc.oc.pgm.util.bukkit.BukkitUtils;
 
 /** Utility to resolve Minecraft usernames from an external HTTP API. */
-public abstract class ApiUsernameResolver extends AbstractBatchingUsernameResolver {
+public class ApiUsernameResolver extends AbstractBatchingUsernameResolver {
   private static final int HTTP_OK = 200;
   private static final int HTTP_TOO_MANY_REQUESTS = 429;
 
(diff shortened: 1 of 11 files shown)


calcastor:
What's the rationale behind keeping code around for Electroid's API? If you input [`jcane`'s UUID](https://api.ashcon.app/mojang/v2/user/f610e558-a3c8-44fb-8622-68cc49dbc18f) as an example, you see his previous username, skin, etc., which was a [previously reported issue](https://github.com/Electroid/mojang-api/issues/84) before Electroid gave end-of-life notice.

Pablete1234:
rationale is old info is better than no info, i'll move it down the list to be under mojang's api

applenick:
Would it be possible to make the Electroid resolver's URL configurable or add an additional configurable resolver? We have a local Mojang API running on the box that provides the same output as Electroid's (but up to date). Have been using it for Community and various other plugins, and it's worked great. It would be nice to switch PGM to using it as well for OCC's use case.

Also, since his service is end-of-life, it makes more sense to have a configurable option anyway, in my opinion.

Edit: Whoops, I read the code before the PR description and saw that you already mentioned a config option 😅. So this is more me throwing in my support for having a configurable one. Thanks!

util/src/main/java/tc/oc/pgm/util/usernames/ApiUsernameResolver.java
cswhite2000: You updated the other call path to `complete(id, UsernameResponse.of(name, now, getClass()));`. Should this be updated as well?

util/src/main/java/tc/oc/pgm/util/usernames/ApiUsernameResolver.java
```diff
@@ -97,19 +116,40 @@ protected void process(List<UUID> uuids) {
...
+  private String resolveSync(UUID id, int maxRetries) throws Exception {
+    var response = httpClient.send(buildRequest(id), HttpResponse.BodyHandlers.ofString());
+
+    return switch (response.statusCode()) {
+      case HTTP_OK -> getUsername(GSON.fromJson(response.body(), JsonObject.class));
+      case HTTP_TOO_MANY_REQUESTS -> {
+        if (maxRetries > 1) {
+          backoffMs = backoff(2d, true);
+          var jitter = (Math.random() * 0.5d);
+          Thread.sleep(backoff(1d + jitter, false));
```
cswhite2000, line 143: Won't this sleep on the common pool threads if we get a 429 from ElectroidApi or PlayerDB?

util/src/main/java/tc/oc/pgm/util/usernames/ApiUsernameResolver.java
Pablete1234: yea i'll update the class

util/src/main/java/tc/oc/pgm/util/usernames/ApiUsernameResolver.java
Pablete1234: yea, i'll migrate all api consumers to default to virtual thread executors and make mojang singlethreaded

cswhite2000 (approved):

PR #1674, opened 2026-05-17
Add cylinder and sphere variables

eliminate123:
This PR introduces two new region variables: 
```xml 
<cylinder id="" base="" radius="" height=""/>
<sphere id="" origin="" radius=""/>
```

These variables function similarly to the existing cuboid variable. CylindricalRegion and SphereRegion have both been updated to support runtime updates via a Mutable inner class. It is important to note that in 1.8 doing /region some-cylinder or /region some-sphere will throw an error in chat, this isn't an issue in modern.

Both variables act as array where each value has a corresponding index.
**For cylinders:**
``[0] = BASE_X``
``[1] = BASE_Y``
``[2] = BASE_Z``
``[3] = RADIUS``
``[4] = HEIGHT``

**For spheres:**
``[0] = ORIGIN_X``
``[1] = ORIGIN_Y``
``[2] = ORIGIN_Z``
``[3] = RADIUS``

It's worth noting that all region variables automatically floor ``player.x player.y player.z`` variables. I'm not sure how to fix this, but it would be nice if this wasn't something that automatically happened. 

image
image


core/src/main/java/tc/oc/pgm/regions/CylindricalRegion.java
```diff
@@ -72,4 +72,35 @@ public String toString() {
...
+  public static class Mutable extends CylindricalRegion {
+    public Mutable(Vector base, double radius, double height) {
+      super(base, radius, height);
+    }
+
+    public Vector getMutableBase() {
+      return base;
+    }
+
+    public double getMutableRadius() {
```
Pablete1234: primitives can't be mutated, this makes no sense

core/src/main/java/tc/oc/pgm/regions/CylindricalRegion.java
```diff
@@ -7,10 +7,10 @@
 import tc.oc.pgm.api.region.RegionDefinition;
 
 public class CylindricalRegion implements RegionDefinition.HardStatic {
-  private final Vector base;
```
Pablete1234, line 10: vector can be mutated, so it doesn't need to stop being final

core/src/main/java/tc/oc/pgm/variables/types/CylindricalVariable.java
```diff
@@ -0,0 +1,172 @@
...
+import tc.oc.pgm.features.StateHolder;
+import tc.oc.pgm.filters.FilterMatchModule;
+import tc.oc.pgm.filters.Filterable;
+import tc.oc.pgm.regions.Bounds;
+import tc.oc.pgm.regions.CylindricalRegion;
+import tc.oc.pgm.variables.Variable;
+import tc.oc.pgm.variables.VariablesMatchModule;
+
+public class CylindricalVariable extends AbstractVariable<Match>
+    implements Variable.Indexed<Match>, StateHolder<CylindricalRegion.Mutable>, RegionDefinition {
```
Pablete1234: these classes are copy-pasted versions of the cuboid variable class, with the only difference being the components available, and the type of region, should be trivial to make a generic abstract parent and have 3 minimal impls that extend:


```suggestion
public class RegionVariable<R extends Region, C extends Component<R>> extends AbstractVariable<Match>
    implements Variable.Indexed<Match>, StateHolder<T>, RegionDefinition {
```
and have some kind of interface for component:

```
interface Component<R extends Region> {
  ToDoubleFunction<R> getter();
  ObjDoubleConsumer<R> setter();
}
```
and the constructor of the generic parent takes in the array of components

Pablete1234 (commented):
Just a few minor details, and the potential to avoid high duplicity of code.

However i don't quite get what you mean by all region types flooring the values of player.x/y/z, and that "in 1.8 doing /region some-cylinder or /region some-sphere will throw an error in chat, this isn't an issue in modern.", what error is it?

eliminate123:
> Just a few minor details, and the potential to avoid high duplicity of code.
> 
> However i don't quite get what you mean by all region types flooring the values of player.x/y/z, and that "in 1.8 doing /region some-cylinder or /region some-sphere will throw an error in chat, this isn't an issue in modern.", what error is it?

Literally i had evil sportpaper 1.8 localhost, ignore what i said

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public boolean isBlockBounded() {
+    return initial.isBlockBounded();
+  }
+
+  @Override
+  public double getValue(Filterable<?> context, int index) {
+    return components[index].getter().applyAsDouble(getState(getAncestor(context)));
```
Pablete1234: you've lost bound checks

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,103 @@
...
+  public int size() {
+    return components.length;
+  }
+
+  @Override
+  public boolean isDynamic() {
+    return Variable.Indexed.super.isDynamic();
+  }
+
+  protected Variable<Match> getComponent(tc.oc.pgm.regions.Component<R> component) {
```
Pablete1234: use imports

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,103 @@
...
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.region.Region;
+import tc.oc.pgm.api.region.RegionDefinition;
+import tc.oc.pgm.features.StateHolder;
+import tc.oc.pgm.filters.Filterable;
+import tc.oc.pgm.regions.Bounds;
+import tc.oc.pgm.regions.Component;
+import tc.oc.pgm.variables.Variable;
+
+public abstract class RegionVariable<R extends RegionDefinition.Mutable & Region.Static>
```
Pablete1234: I see no reason to have such a specific wildcard

core/src/main/java/tc/oc/pgm/variables/types/CuboidVariable.java
```diff
@@ -1,176 +1,55 @@
...
-
-  public double getComponent(Filterable<?> match, Component component) {
-    return component.getter.applyAsDouble(component.vector.apply(match.state(this)));
-  }
-
-  public void setComponent(Filterable<?> match, Component component, double value) {
-    component.setter.accept(component.vector.apply(match.state(this)), value);
+  protected CuboidRegion.Mutable createState() {
+    return initial.asMutableCopy();
   }
```
Pablete1234: can be moved to the abstract parent class, given you already have added a MutableSource interface, just need to make it generic:
```java
interface MutableSource<T extends Region.Mutable> {
  T asMutableCopy()
}
```
alternatively, you can just drop both MutableSource and Mutable given they don't really add anything if you still have the specific createState method for each impl; all the parent needs is an initial region (which, should be generic and protected so that the children can access it), and a state region, which should also be a generic that is created from createState()

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public int size() {
+    return components.length;
+  }
+
+  @Override
+  public boolean isDynamic() {
+    return Variable.Indexed.super.isDynamic();
```
Pablete1234: Please take CuboidVariable as the example of how these interface methods should be implemented, they had proper implementations

Pablete1234 (commented):
Overall the abstraction is very much a step in the right direction, but try to keep the details that were present in CuboidVariable also present in the generic impl, they were there for a reason.

Also as mentioned, the Mutable and MutableSource interfaces seem very redundant

eliminate123:
For block/point vars the xml looks like this:

Both follow the same general structure where indicies:
``[0] = X``
``[1] = Y``
``[2] = Z``

Setting their values in an action looks something like this:|
```xml
<actions>
    <trigger filter="some-filter" scope="player">
        <action>
            <set var="point_var" index="0" value="player.x"/>
            <set var="point_var" index="1" value="player.y"/>
            <set var="point_var" index="2" value="player.z"/>
        </action>
    </trigger>
</actions>
```

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+import tc.oc.pgm.variables.Variable;
+
+public abstract class RegionVariable<
+        R extends RegionDefinition.Mutable, I extends Region & RegionDefinition.MutableSource<R>>
+    extends AbstractVariable<Match>
+    implements Variable.Indexed<Match>, StateHolder<R>, RegionDefinition {
+
+  private final Component<R>[] components;
+  protected final I initial;
+  private final Map<Match, R> states = new WeakHashMap<>();
```
Pablete1234: This is not how we hold per-match state, please look at the original implementation of CuboidVariable for reference.

Note: as-is this is a memory leak problem

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+  protected RegionVariable(Component<R>[] components, I initial) {
+    super(Match.class);
+    this.components = components;
+    this.initial = initial;
+  }
+
+  protected R createState() {
+    return initial.asMutableCopy();
+  }
+
```
Pablete1234: This is not needed anymore, it was only created for subclasses to implement their own

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public void load(Match match) {
+    states.put(match, createState());
```
Pablete1234: again, this is not how you hold per-match state, see original impl for reference

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private R getState(Match match) {
+    return states.get(match);
+  }
```
Pablete1234: again, this is not how you hold per-match state, see original impl for reference

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  public Region.Static getStaticImpl(Match match) {
+    return (Region.Static) getState(match);
```
Pablete1234: again, this is not how you hold per-match state, see original impl for reference

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  public double getValue(Filterable<?> context, int index) {
+    if (index < 0 || index >= components.length) {
+      String msg =
+          String.format("Index %d out of bounds for %s", index, getClass().getSimpleName());
+      PGM.get().getGameLogger().log(Level.SEVERE, msg);
+      return 0;
+    }
+    return components[index].getter().applyAsDouble(getState(getAncestor(context)));
```
Pablete1234: This was tidy in a checkBounds function before (reused in both get and set value), which intentionally defaulted to the first component instead of defaulting to 0, this is a breaking change
Also it lost warning for what region id in particular got the warning, making it impossible to debug for mapdevs.

Also getAncestor is not needed, as any Filterable can  get the state of the StateHolder

Please reference the original impl.

core/src/main/java/tc/oc/pgm/variables/types/RegionVariable.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Override
+  public void setValue(Filterable<?> context, int index, double value) {
+    if (index < 0 || index >= components.length) {
+      String msg =
+          String.format("Index %d out of bounds for %s", index, getClass().getSimpleName());
+      PGM.get().getGameLogger().log(Level.SEVERE, msg);
+      return;
+    }
+    components[index].setter().accept(getState(getAncestor(context)), value);
```
Pablete1234: same as previous comment, please reference original impl

core/src/main/java/tc/oc/pgm/variables/VariableParser.java
```diff
@@ -122,6 +127,28 @@ public Variable<MatchPlayer> parsePlayerLocation(Element el) throws InvalidXMLEx
...
+  @MethodParser("point")
+  public Variable<Match> parsePoint(Element el) throws InvalidXMLException {
+    String baseId = FeatureDefinitionContext.parseId(el);
+    var variable = new PointVariable(factory.getRegions().parsePoint(el));
+    for (PointVariable.Component component : PointVariable.Component.values()) {
+      var subId = baseId + "." + component.name().toLowerCase(Locale.ROOT);
+      factory.getFeatures().addFeature(el, subId, variable.getComponent(component));
+    }
+    return variable;
+  }
```
Pablete1234, line 135: this is a bunch of copy-paste, you can reuse most of it:
eg:
```
  @MethodParser("point")
  public Variable<Match> parsePoint(Element el) throws InvalidXMLException {
    return registerComponents(new PointVariable(factory.getRegions().parsePoint(el)));
  }
  
  private <R extends RegionDefinition.Mutable, T extends RegionVariable<R, ?>> void registerComponents(Element el, T reg) {
    String baseId = FeatureDefinitionContext.parseId(el);
    for (Component<R> component : reg.getComponents()) {
      var subId = baseId + "." + component.name().toLowerCase(Locale.ROOT);
      factory.getFeatures().addFeature(el, subId, variable.getComponent(component));
    }
    return reg;
  }
```
expose a getComponents() method in RegionVariable and add a `name()` method to Component, and you should be good to go

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/variables/types/BlockVariable.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class BlockVariable extends RegionVariable<BlockRegion.Mutable, BlockRegion> {
+
+  public BlockVariable(BlockRegion initial) {
+    super(Component.values(), initial);
+  }
+
+  public Variable<Match> getComponent(Component component) {
+    return super.getComponent(component);
+  }
```
Pablete1234: this is completely redundant code

core/src/main/java/tc/oc/pgm/variables/types/BlockVariable.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.function.ToDoubleFunction;
+import tc.oc.pgm.regions.BlockRegion;
+
+public class BlockVariable extends RegionVariable<BlockRegion.Mutable, BlockRegion> {
+
+  public BlockVariable(BlockRegion initial) {
+    super(Component.values(), initial);
+  }
+
+  public enum Component implements tc.oc.pgm.regions.Component<BlockRegion.Mutable> {
```
Pablete1234:
```suggestion
  enum BlockComponents implements Component<BlockRegion.Mutable> {
```
Use imports, if that means you have to rename the component enum then go ahead, also seems like it doesn't need to be public anymore right?

core/src/main/java/tc/oc/pgm/variables/types/CuboidVariable.java
```diff
@@ -1,176 +1,47 @@
...
-  public void setComponent(Filterable<?> match, Component component, double value) {
-    component.setter.accept(component.vector.apply(match.state(this)), value);
+  public CuboidVariable(CuboidRegion initial) {
+    super(Component.values(), initial);
   }
 
   public Variable<Match> getComponent(Component component) {
-    return new ComponentVariable(component);
+    return super.getComponent(component);
   }
```
Pablete1234: this is redundant, just re-calling parent exactly as-is

core/src/main/java/tc/oc/pgm/variables/types/BlockVariable.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.function.ObjDoubleConsumer;
+import java.util.function.ToDoubleFunction;
+import tc.oc.pgm.regions.BlockRegion;
+
+public class BlockVariable extends RegionVariable<BlockRegion.Mutable, BlockRegion> {
+
+  public BlockVariable(BlockRegion initial) {
+    super(Component.values(), initial);
+  }
```
Pablete1234, line 14:
```suggestion
  private static final Component<T>[] COMPONENTS = BlockComponents.values();

  public BlockVariable(BlockRegion initial) {
    super(COMPONENTS, initial);
  }
```
Enum.values() in java creates a new instance every time, reusing it can be heavily beneficial as each instance of BlockVariable will now be reusing the same array of components instead of allocating a new one

Pablete1234 (commented):
Just a few last minor tweaks, make sure to apply the suggested fixes across all the diff new region variables you've added, but otherwise LGTM, good work!

Pablete1234 (approved):

PR #1675, opened 2026-05-20
Implement a tracker for end crystals

calcastor:
Hooks end crystals into PGM's pre-existing trackers to credit the player who detonates and end crystal via any means (projectile, melee hit, etc.), falling back on the player who placed the end crystal on modern only, since end crystals can't be placed on legacy.

This also resolves an issue where players could not damage themselves with end crystals even if `self="false"` on `entity explosion` in `<disabledamage/>`; unfortunately this uses reflection, and is a similar issue to the one we work around for TNT in modern's `TntListener`.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/tracker/TrackerMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/tracker/TrackerMatchModule.java
@@ -9,6 +9,7 @@ import org.bukkit.event.entity.EntityDamageByEntityEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.projectiles.BlockProjectileSource;
 import org.bukkit.projectiles.ProjectileSource;
+import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.api.match.MatchModule;
@@ -27,6 +28,7 @@ import tc.oc.pgm.tracker.trackers.CactiTracker;
 import tc.oc.pgm.tracker.trackers.CombatLogTracker;
 import tc.oc.pgm.tracker.trackers.DeathTracker;
 import tc.oc.pgm.tracker.trackers.DispenserTracker;
+import tc.oc.pgm.tracker.trackers.EndCrystalTracker;
 import tc.oc.pgm.tracker.trackers.EntityTracker;
 import tc.oc.pgm.tracker.trackers.FallTracker;
 import tc.oc.pgm.tracker.trackers.FallingBlockTracker;
@@ -36,6 +38,7 @@ import tc.oc.pgm.tracker.trackers.ProjectileTracker;
 import tc.oc.pgm.tracker.trackers.SpleefTracker;
 import tc.oc.pgm.tracker.trackers.TNTTracker;
 
+@NullMarked
 public class TrackerMatchModule implements MatchModule {
 
   private final EntityTracker entityTracker;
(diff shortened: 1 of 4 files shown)


cswhite2000 (dismissed):

core/src/main/java/tc/oc/pgm/tracker/TrackerMatchModule.java
```diff
@@ -44,6 +47,7 @@ public class TrackerMatchModule implements MatchModule {
   private final FireTracker fireTracker;
   private final FallingBlockTracker fallingBlockTracker;
   private final CactiTracker cactiTracker;
+  private final EndCrystalTracker endCrystalTracker;
```
Pablete1234: only damage resolvers get added here and reused, in your case you can just create it on load() directly like dispense or tnt tracker does

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/tracker/ModernTrackerMatchModule.java
```diff
@@ -0,0 +1,81 @@
...
+   * disabled. We have to do something similar for TNT, see {@link TntListener#onTntSpawn}; however,
+   * as causingEntity is private and final, we must use reflection.
+   */
+  @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+  public void onCrystalDamage(EntityDamageByEntityEvent event) {
+    if (!(event.getEntity() instanceof EnderCrystal)) return;
+    if (tmm.getEntityTracker().getOwner(event.getDamager()) == null) return;
+
+    var handle = ((CraftDamageSource) event.getDamageSource()).getHandle();
+    ReflectionUtils.setField(handle, null, CAUSING_ENTITY);
```
Pablete1234: obviously not a fan of this one, best bet for a workaround may be to, if the damaged entity is an end crystal, and the event isn't cancelled, and the damage source's causingEntity is a player, then cancel this event (which will cancel this end crystal explosion), and just call to damage the entity directly with a new damage source, created to mimic the original but excluding the player.
You may call the NMS method directly to avoid it deriving the damage source differently, ie:
`EndCrystal#hurtServer(level, damageSource, amount)`

You may also need to manually call the track entity here cause the other event listener (the multi-version one) probably never picks this up properly.

Pablete1234 (changes requested):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/tracker/ModernTrackerMatchModule.java
```diff
@@ -0,0 +1,90 @@
...
+    // Only intercept when a player has damaged the end crystal
+    if (!(event.getDamageSource().getCausingEntity() instanceof Player)) return;
+
+    ParticipantState owner = tmm.getEntityTracker().getOwner(event.getDamager());
+    if (owner == null) return;
+
+    var nmsCrystal = ((CraftEnderCrystal) crystal).getHandle();
+    var handle = ((CraftDamageSource) event.getDamageSource()).getHandle();
+    var damageSource = new DamageSource(
+        handle.typeHolder(), handle.getDirectEntity(), null, handle.sourcePositionRaw());
```
Pablete1234, line 72: does this work for when direct entity is a player? eg: player hitting the crystal with a sword

calcastor: Yes this works as expected; additionally we don't seem to need to call the tracker here redundantly because the event listener is lower priority than the version agnostic one; I can maintain that though.

Pablete1234 (approved):

Issue #1600, opened 2026-01-23
Crossbows loaded with fireworks shoot arrows if modifybowprojectile has a pickup-filter


mitchts:
https://github.com/user-attachments/assets/3ce6f80c-4c47-4c53-9747-3379f6b18437

[modifybowprojectile.zip](https://github.com/user-attachments/files/24816365/modifybowprojectile.zip)

Regardless of if the `pickup-filter` changes to allow picking up arrows, crossbows loaded with fireworks will shoot an arrow if a pickup-filter is present on the `<modifybowprojectile/>` module. Only exception I've found is when using `always` as the filter.

PR #1676, opened 2026-05-23
Don't replace non-arrows when modifying bow projectiles

calcastor:
This may be relevant to https://github.com/PGMDev/PGM/issues/1600, but I have not tested that specific case yet

### DIFF
--- a/core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
+++ b/core/src/main/java/tc/oc/pgm/modules/ModifyBowProjectileMatchModule.java
@@ -58,8 +58,11 @@ public class ModifyBowProjectileMatchModule implements MatchModule, Listener {
     Plugin plugin = PGM.get();
     Entity newProjectile;
 
-    if (this.cls == Arrow.class && event.getProjectile() instanceof Arrow) {
-      // Don't change the projectile if it's an Arrow and the custom entity type is also Arrow
+    // Don't replace non-arrow projectiles
+    if (!(event.getProjectile() instanceof Arrow)) return;
+
+    if (this.cls.isInstance(event.getProjectile())) {
+      // Don't change the projectile if it matches the custom entity type
       newProjectile = event.getProjectile();
     } else {
       // Replace the projectile


Pablete1234:
I don't believe this is the behavior that modifybowprojectile should have, and the issue relating to #1600 is a "parsing" issue of assuming "replace with arrow" when in the xml you have not specified wanting any replacement.

If there is a wish to extend the module further such that only X projectile gets replaced with Y (eg: arrow -> enderpearl, firework -> primed tnt), or only certain items are affected (eg: bow -> replace with enderpearl, crossbow -> unaffected), i think those should be part of its own feature implementation.

Is there a specific usecase wanting to do something like this?

Pablete1234:
See https://github.com/PGMDev/PGM/pull/1679 for the minimal fix of #1600, leaving this one open in case you want to expand it into supporting alternative options. 
The simplest i can think of is a `bow-filter="bows-only"` (which filters the `EntityShootBowEven's#getBow` so that, on modern, you can say modifybowprojectile should affect only bows and not crossbows) and/or a `projectile-filter="arrows-only"` which filters for the projectile shot.

Both of those would also mean modifybowprojectile gets the ability to enable/disable itself over the match (eg: imagine it enabling after 5 minutes or whatever).

The more "extensive" rework would mean you would need to allow for multiple modifybowprojectile "rules" and pick which one applies, similar to the rework the `<respawn>` got a while back, but i think that may be overkill for what you want to do here.

calcastor:
> I don't believe this is the behavior that modifybowprojectile should have, and the issue relating to #1600 is a "parsing" issue of assuming "replace with arrow" when in the xml you have not specified wanting any replacement.
> 
> 
> 
> If there is a wish to extend the module further such that only X projectile gets replaced with Y (eg: arrow -> enderpearl, firework -> primed tnt), or only certain items are affected (eg: bow -> replace with enderpearl, crossbow -> unaffected), i think those should be part of its own feature implementation.
> 
> 
> 
> Is there a specific usecase wanting to do something like this?
> 
> 
> 
> 

It was mostly aiming to address a seemingly false (at least more noticeably so now) assumption with this module that the projectile shot from a bow-like weapon by default is always an arrow. 

#1679 seems more appropriate for the intent.

PR #1678, opened 2026-05-23
Resolve material data parsing inconsistencies

calcastor:
This became noticeable when we loaded The Funky Queen to OCC; things defined like "stained glass:15" weren't parsed on modern correctly while being supported on SportPaper. It probably makes sense to support this given that `damage` in modern tends to refer to actual durability status of the item when relevant.

### DIFF
--- a/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/material/ModernMaterialParser.java
+++ b/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/material/ModernMaterialParser.java
@@ -73,9 +73,29 @@ class ModernMaterialParser {
 
   public static ItemMaterialData parseItem(String text, short dmg, Node node)
       throws InvalidXMLException {
-    var res = Adapter.PGM_ITEM.visit(parseLegacyMaterial(text, node), dmg);
-    validateItem(res.getItemType(), node);
-    return res;
+    String[] pieces = text.split(":");
+    if (pieces.length > 2) {
+      throw new InvalidXMLException("Invalid material pattern '" + text + "'.", node);
+    }
+
+    Material material = parseLegacyMaterial(pieces[0], node);
+    short mdData = pieces.length == 2 ? XMLUtils.parseNumber(node, pieces[1], Short.class) : 0;
+    if (mdData != dmg && mdData != 0 && dmg != 0) {
+      throw new InvalidXMLException(
+          "Mismatching damage, parsed '" + text + ":" + dmg + "' but should be '" + text + ":"
+              + mdData + "'",
+          node);
+    }
+    short data = dmg != 0 ? dmg : mdData;
+
+    if (data != 0 && !material.isLegacy() && !canUseMeta(material)) {
+      throw new InvalidXMLException(
+          "Material '" + material + "' cannot have a damage/meta value of " + data, node);
+    }
+
+    var md = Adapter.PGM_ITEM.visit(material, data);
+    validateItem(md.getItemType(), node);
+    return md;
   }
 
   public static Material parseMaterial(String text, Node node) throws InvalidXMLException {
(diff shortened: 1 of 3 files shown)


Pablete1234:
For context, the issue with dmg and `:15` is that they are two places in which pgm allows defining damage values:

```xml
<!-- Allowed forms -->
<item material="stained glass:15"/> <!-- should work, meta embedded in material -->
<item material="stained glass" damage="15"/> <!-- should work, meta specified as damage -->
<item material="stained glass:15" damage="15"/> <!-- Redundant, but because they're the same we'll allow it -->

<!-- Disallowed forms -->
<item material="stained glass:15" damage="14"/> <!-- Not allowed: make up your mind, 14 or 15? -->
<item material="black stained glass" damage="15"/> <!-- Cannot mix modern name with damage -->
<item material="black stained glass:15" /> <!-- Cannot mix modern name with damage -->
<item material="black stained glass:15" damage="15" /> <!-- Cannot mix modern name with both damages -->
```
However, note that the last 3 have an exception, and its items with durability (ie: tools etc), and possibly potions.


So i think that `parseLegacyMaterial` needs to distinguish between legacy (which requires a follow-up `upgrade(legacy, meta)`) and modern parsing, which doesn't require upgrading and should complain about unusable damage values. Note that `upgrade` already checks to only need to upgrade legacy materials. And if a damage (or a `:meta` is included alongside a non-legacy material it probably needs warning as long as the material is a non-damageable (ie: not tools etc) material.


TLDR:
- Split the `material:value`
- If the material is non-legacy and has `:meta`, complain (exception)
- If the damage param is nonzero, and is different to the `:meta`, complain (exception)
- If the material is legacy, pass it thru `upgrade(material, damage)`.
- When calling `visit` do so with whichever of the two metas is nonzero

calcastor:
> For context, the issue with dmg and `:15` is that they are two places in which pgm allows defining damage values:
> 
> ```
> 
> <item material="stained glass:15"/> 
> <item material="stained glass" damage="15"/> 
> <item material="stained glass:15" damage="15"/> 
> 
> 
> <item material="stained glass:15" damage="14"/> 
> <item material="black stained glass" damage="15"/> 
> <item material="black stained glass:15" /> 
> <item material="black stained glass:15" damage="15" /> 
> ```
> 
> However, note that the last 3 have an exception, and its items with durability (ie: tools etc), and possibly potions.
> 
> So i think that `parseLegacyMaterial` needs to distinguish between legacy (which requires a follow-up `upgrade(legacy, meta)`) and modern parsing, which doesn't require upgrading and should complain about unusable damage values. Note that `upgrade` already checks to only need to upgrade legacy materials. And if a damage (or a `:meta` is included alongside a non-legacy material it probably needs warning as long as the material is a non-damageable (ie: not tools etc) material.
> 
> TLDR:
> 
> * Split the `material:value`
> * If the material is non-legacy and has `:meta`, complain (exception)
> * If the damage param is nonzero, and is different to the `:meta`, complain (exception)
> * If the material is legacy, pass it thru `upgrade(material, damage)`.
> * When calling `visit` do so with whichever of the two metas is nonzero



> For context, the issue with dmg and `:15` is that they are two places in which pgm allows defining damage values:
> 
> ```
> 
> <item material="stained glass:15"/> 
> <item material="stained glass" damage="15"/> 
> <item material="stained glass:15" damage="15"/> 
> 
> 
> <item material="stained glass:15" damage="14"/> 
> <item material="black stained glass" damage="15"/> 
> <item material="black stained glass:15" /> 
> <item material="black stained glass:15" damage="15" /> 
> ```
> 
> However, note that the last 3 have an exception, and its items with durability (ie: tools etc), and possibly potions.
> 
> So i think that `parseLegacyMaterial` needs to distinguish between legacy (which requires a follow-up `upgrade(legacy, meta)`) and modern parsing, which doesn't require upgrading and should complain about unusable damage values. Note that `upgrade` already checks to only need to upgrade legacy materials. And if a damage (or a `:meta` is included alongside a non-legacy material it probably needs warning as long as the material is a non-damageable (ie: not tools etc) material.
> 
> TLDR:
> 
> * Split the `material:value`
> * If the material is non-legacy and has `:meta`, complain (exception)
> * If the damage param is nonzero, and is different to the `:meta`, complain (exception)
> * If the material is legacy, pass it thru `upgrade(material, damage)`.
> * When calling `visit` do so with whichever of the two metas is nonzero

Addressed this in both modern and SportPaper when using modern names.

Pablete1234 (approved):

PR #1680, opened 2026-05-26
Add config for current map vote book weighting

Pugzy:
PR [#1581](https://github.com/PGMDev/PGM/pull/1581) introduced `prevMap` to `MapVotePicker.getMaps()`, prepending the current map to the "previous maps" list so that maps sharing its gamemode receive reduced weight in the vote. This was always-on with no way to disable it.

Forcing gamemode diversity sounds good on paper, but some gamemodes are more popular than others and penalising them in the vote takes away player choice. The vote picker already tries to pick varied maps, so a bit of gamemode overlap between the current map and the next vote is not really a problem.

Some pool types don't need this logic at all either. In arcade-style pools, a map's gamemode tag is often a poor reflection of how it actually plays, so gamemode-based exclusion does more harm than good.

Adds `excludeCurrentGamemode` to `VoteConstants`, backed by config key `exclude-current-gamemode` (default: `true`). The check lives inside `MapVotePicker.getMaps()`, which owns `VoteConstants` and all other weighting rules. `VotingPool` always passes `match.getMap()` and the picker decides whether to use it.

```yaml
exclude-current-gamemode: true   # default, reduces weight for maps sharing a gamemode with the current map
exclude-current-gamemode: false  # current map's gamemode is not considered, gamemode weighting still applies across vote candidates
```

Creating this PR as a start for discussion, it would be nice to run tests with this disabled for some period.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/rotation/pools/VotingPool.java
+++ b/core/src/main/java/tc/oc/pgm/rotation/pools/VotingPool.java
@@ -177,7 +177,7 @@ public class VotingPool extends MapPool {
       Formula<Match> scoreAfterPlay,
       int minCooldown,
       int minutesPerDay,
-      boolean excludeUsingCurrentMap) {
+      boolean useCurrentMap) {
     private VoteConstants(ConfigurationSection section, int mapAmount) {
       this(
           section.getInt("vote-options", MapVotePicker.MAX_VOTE_OPTIONS), // Show 5 maps
@@ -192,7 +192,7 @@ public class VotingPool extends MapPool {
               .map(m -> new Context(m.getDuration())),
           section.getInt("cooldown.min-length", 30),
           section.getInt("cooldown.minutes-per-day", 30),
-          section.getBoolean("exclude-using-current-map", true));
+          section.getBoolean("score.use-current-map", true));
     }
 
     public double afterVoteScore(double score) {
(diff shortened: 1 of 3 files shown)


Pablete1234:
> Some pool types don't need this logic at all either. In arcade-style pools, a map's gamemode tag is often a poor reflection of how it actually plays, so gamemode-based exclusion does more harm than good.

That can already be accounted for with vote weight formula, so its not an added feature, unless you mean only that gamemode *should* be relevant just not for the current map, which in an arcade pool i dont believe matters at all.

> Adds excludeCurrentGamemode to VoteConstants, backed by config key exclude-current-gamemode

This gives the impression that the only thing changing is the gamemode, what changes tho is if the current map is assumed to be in the vote as a fake "first option", i think the config name should not talk about gamemode and should reflect map instead, eg "exclude-using-current-map"

Pugzy:
@Pablete1234 updated name/references. Wasn't sure if best to name it current or previous.. as they're both kinda true.

Pablete1234:
branch missing dco but otherwise lgtm

Pugzy:
> branch missing dco but otherwise lgtm

Signed, but I think the current name is now confusing.

`exclude-using-current-map` does this exclude the use of the current map i.e no map used? or does it exclude using the current map?

Pablete1234:
maybe `exclusion-uses-current-map` ?

PR #1681, opened 2026-05-29
Team Classes - Max FIlter

technodono:
For ranked maps such as classbowl, classes are currently disabled for the map as pgm has currently no way to limit the amount of players on a specific class per team. The new max filter sets a team-scoped limit on the amount of players that are able to pick a specific class (with the default class for now ignoring this limit until a better implementation is found).

This is a very crude patch (and I imagine could be implemented much better implemented by a dev) but works for the most part. Still need to implement a check for when players swap teams since it can be used to bypass this limit.

A sample xml snippet of the feature can be found below (sets the limit per class to 1 player maximum)
```xml
<classes family="Classes" max="1">
    <class name="Defender" default="true" description="`a`lMovement| █`8░░ `n`6`lCombat| ██`8░ `n`d`lHealth| ██`8░ `n`b`l Ability: `7Geared-Up" icon="iron chestplate">
        <kit parents="base-kit" id="defender" filter="has-defender">
            <action id="reset-gearup-cooldown"/>
            <item slot="0" unbreakable="true" material="stone sword"/>
            <chestplate unbreakable="true" locked="true" material="gold chestplate"/>
            <walk-speed>${lvl-1-speed}</walk-speed>
            <knockback-reduction>0.25</knockback-reduction>
        </kit>
    </class>
        <class name="Assasin" description="`a`lMovement| ███ `n`6`lCombat| ███ `n`d`lHealth| █`8░░ `n`b`l Ability: `7Play Dirty" icon="blaze powder">
            <kit parents="base-kit" id="assasin" filter="has-assasin">
                <item slot="0" unbreakable="true" material="stone sword" enchantment="sharpness:1"/>
                <chestplate unbreakable="true" color="FF0C7F" locked="true" material="leather chestplate"/>
                <shield health="${shield-health}" delay="${shield-cooldown}"/>
                <walk-speed>${lvl-3-speed}</walk-speed>
            </kit>
        </class>
... [Add more classes here]
</classes>
```


core/src/main/java/tc/oc/pgm/classes/ClassMatchModule.java
```diff
@@ -135,6 +136,26 @@ public Set<UUID> getClassMembers(PlayerClass cls) {
...
+    int count = 0;
+    for (MatchPlayer player : this.match.getPlayers()) {
+      if (party.equals(player.getParty())) {
+        if (getPlayingClass(player.getId()).equals(cls)) {
+          count++;
+        }
+      }
+    }
+    return count;
+  }
```
TTtie: The whole thing can be simplified down to
```suggestion
  public long getPartyClassCount(Party party, PlayerClass cls) {
    return party.getPlayers().stream()
        .filter(player -> getPlayingClass(player.getId()).equals(cls))
        .count();
  }
```

core/src/main/java/tc/oc/pgm/classes/ClassMatchModule.java
```diff
@@ -163,10 +184,28 @@ public PlayerClass setPlayerClass(UUID userId, PlayerClass cls) {
...
+      if (!cls.equals(currentPlaying) && !cls.equals(currentSelected)) {
+        int currentCount = this.getPartyClassCount(matchPlayer.getParty(), cls);
+
+        if (currentCount >= cls.getMax()) {
+          matchPlayer.sendWarning(net.kyori.adventure.text.Component.text(
+              "That class is full! Maximum limit reached for your team. (" + cls.getMax() + " max)",
+              net.kyori.adventure.text.format.NamedTextColor.RED
+          ));
+
+          throw exception("That class is full! Maximum limit reached for your team.");
```
TTtie: The `sendWarning` call is redundant and should go into the exception text thrown below

The argument for `exception` is a translation key for localization purposes, see e. g. 81fcde56709b384c74a9eef685bc8e771eb15671 to get a brief glimpse of how to do so.

technodono: Ok cool will do

technodono:
@TTtie https://github.com/PGMDev/PGM/pull/1681/commits/8595141a7463b1fb448c9cc5b0dc7589ce79ec15
Have the WIP changes alongside your implemented suggestions for today.

## How the system should work:
Red/Blue observing teams (30s match loading)
- Cannot select a class if the max limit for the team is reached
- When swapping to a new team and is full set the class back to the default

During match:
- Upon dying players can select any class so that they can change class and swap teams without fuss
- If a person respawns on any team and breaks the class limit they are reverted to their previous class
- If a person respawns on any team breaks the class limit and the reverted class is full they get sent back to the default class
- 
## Bugs
Struggling to get these features working during the first 30s of the match (i.e. when red/blue teams are technically observers). I've purposely disabled the limit for the observer team (not red/blue I think) but a better approach may need to be that players pick a team first before selecting classes

eliminate123:
Why don't you abandon the class selection gui and make one from scratch using consumables and variables.

Have a limbo area where players select classes once match starts. Player clicks a consumable and a player var "current_class" is set to corresponding number. Each team has an array called "current_classes" and each index represents a different class and however many positions are open. 
(Ex: I click consumable and select archer class. Then <set var="current_class" value="${ARCHER}"). After X amount of seconds everyone who selected a class gets tpd into the actual play space on the map.

Let's say only one person per team is allowed to be archer. Once I select archer do <set var="team_classes" index="${ARCHER}" value="team_classes[${ARCHER}] - 1".
since the limit was originally 1, now its 0 you can do a few things (I'll suggest what I think is best) in the consumable action have a nested action at the start thats <action filter="not(team_classes[$ARCHER]=0"> with all the variable setting inside. Then have another nested action for when =0 and have it send a message that says "Class full". Now players cat select a class that's full while still keeping the class preview 


You can also have a variable tracking the players previous class. Just have their current_class default value=-1 when they first spawn. Then they select a class it does <set var="previous_class" value="current_class" then set their current class. Variables are way better imo bc they reset when switching teams. 

You will just need to add a system for resetting variables when a player is no longer playing.




PS I typed this comment while cooking pizza image

technodono:
> You can also have a variable tracking the players previous class. Just have their current_class default value=-1 when they first spawn. Then they select a class it does <set var="previous_class" value="current_class" then set their current class. Variables are way better imo bc they reset when switching teams.

this is a much better approach to take plus the logistics behind a pr like this remain for the most part tied up. Goign to close this issue in the meantime

PR #1682, opened 2026-06-02
Built filter

Javggg:
Adds a new `<built>` filter that checks whether the blocks in the world at a given origin match a defined structure.
Usage:
```xml
<built id="my-filter" structure="my-structure" origin="15,13,-18"/>
```
The name built is open to discussion. Alternatives like `structure-match` or just `structure` (following the precedent of the variable filter) are also reasonable.


cswhite2000:
I assume that this is for a gamemode where players would try to rebuild a structure from memory.

If so, then this filter not being dynamic could pose a problem. If you were to say run this filter every tick or so via a pulse filter, we'd be going over every block in the filter each time, which could get rather expensive performance wise.

Javggg:
I intentionally made it non-dynamic for this reason. Re-evaluating on every block change within the area could be very costly for large structures. Keeping it non-dynamic gives map authors precise control over when the filter is evaluated.
That said, this also means it can be misused, for example, placing it inside a pulse filter that runs every tick would be just as expensive, if not worse. I went with this approach as the safer default, but I'm open to making it dynamic if that's preferred.

Javggg:
I made it dynamic with a limit of 1024 blocks. I talked with Pablo about this and he agreed on this approach.

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+    implements ReactorFactory<BuiltFilter.Reactor> {
+
+  private static final int MAX_BLOCKS = 1024;
+  private static final String OVERSIZED_ERR =
+      "Structure '%s' used in 'built' filter has %d blocks, exceeds limit of %d. Filter will always return false";
+
+  private final FeatureReference<StructureDefinition> structure;
+  private final BlockVector origin;
+
+  private final Map<Match, Reactor> reactors = new ConcurrentHashMap<>();
```
Pablete1234: This is not how you hold per-match state, look at other stateful filters (eg: MonostableFilter )

```suggestion
```

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+  private Boolean oversized;
+
+  public BuiltFilter(FeatureReference<StructureDefinition> structure, BlockVector origin) {
+    this.structure = structure;
+    this.origin = origin;
+  }
+
+  private boolean isOversized() {
```
Pablete1234: This should be checked at parse time, not at runtime.

Make it a validation in the parsing stage that the structure has to be of this size max

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+          .log(
+              Level.SEVERE,
+              String.format(OVERSIZED_ERR, structure.get().getId(), count, MAX_BLOCKS));
+    }
+    return oversized;
+  }
+
+  @Override
+  public boolean isDynamic() {
+    return !isOversized();
```
Pablete1234:
```suggestion
    return true;
```
this filter should always be dynamic, and always enforce the max size

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Override
+  public boolean matches(MatchQuery query) {
+    if (isOversized()) return false;
+
+    Match match = query.getMatch();
+    return reactors.computeIfAbsent(match, m -> new Reactor(m, null)).evaluate();
```
Pablete1234:
```suggestion
    return query.state(this).matches(query);
```
get the reactor from the match context, also prefer `matches` over `evaluate`, `boolean evaluate(...)` makes it sound like the true/false is if it evaluated successfully or failed, not the *result* of the check

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Override
+  public Class<? extends MatchQuery> queryType() {
+    return MatchQuery.class;
+  }
+
+  @Override
+  public Reactor createReactor(Match match, FilterMatchModule fmm) {
+    return reactors.computeIfAbsent(match, m -> new Reactor(m, fmm));
```
Pablete1234:
```suggestion
    return new Reactor(m, fmm);
```
PGM only calls this method once per match, and caches the reactor for you in the match state, from where you can get it back, handling the lifetime and most importantly, cleaning it up once the match is over so that it doesn't memory leak like the current version does

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public boolean evaluate() {
+      if (cachedResult != null) return cachedResult;
+
+      Structure struct = structure.get().getStructure(match);
+      cachedMismatch = struct.findMismatch(match.getWorld(), origin);
+      cachedResult = cachedMismatch == null;
+      return cachedResult;
+    }
```
Pablete1234: rename to `matches` as discussed above
Keep in mind this can be called VERY often, depending on the configured dynamic filters on the map, this may be called effectively every tick as a result of player moves, so this needs to be as fast as possible.

turn `Structure` into a field in the reactor, populate once on the MatchLoadEvent (which happens after structures have loaded, so you should be always able to find it). For that you may need to change the scope from running to Loaded on the listener. If structure is null you should just return false, which will only happen during match load.

Also note, according to the other comment making mismatch a final field, you'd instead here do something like:
```java
if (lastResult != null) return lastResult;

var mismatch = struct.findMismatch(match.getWorld(), offset, lastMismatch);
if (mismatch != null) lastMismatch.set(mismatch);
return lastResult = mismatch == null;
```
Note that this changes `findMismatch` such that it takes an offset, instead of an origin, meaning it doesn't need to compute it every time, and also an optional "firstCheck" block vector, which if passed, it will check first.

core/src/main/java/tc/oc/pgm/filters/parse/FilterParser.java
```diff
@@ -626,6 +641,14 @@ public Filter parseOffsetFilter(Element el) throws InvalidXMLException {
     return LocationQueryModifier.of(child, vector);
   }
 
+  @MethodParser("built")
+  public Filter parseBuiltFilter(Element el) throws InvalidXMLException {
+    var structure = parser.reference(StructureDefinition.class, el, "structure").required();
```
Pablete1234:
```suggestion
    var structure = parser.reference(StructureDefinition.class, el, "structure")
        .validate(...)
        .required();
```
should validate the structure size here

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,135 @@
...
+        BlockVector offset =
+            origin.clone().subtract(struct.getDefinition().getOrigin()).toBlockVector();
+        Location relative =
+            loc.clone().subtract(offset.getBlockX(), offset.getBlockY(), offset.getBlockZ());
+        relevant = struct.getRegion().contains(relative);
+      }
+
+      if (relevant) {
+        cachedResult = null;
+        cachedMismatch = null;
```
Pablete1234:
```suggestion
```
I think losing the mismatch isn't great, its still valuable information and can be used in `matches` to check this first, as it was the last "known bad" block, and if we check it, and its still bad, we can skip checking all the other potentially "good" blocks on the region.

Overall, this is what i'd change of this reactor:

Initialize these at the very beginning:
```java
private Boolean lastResult = null;
private final BlockVector lastMismatch = origin.clone();
private Structure struct;
private Region region;
private BlockVector offset;
```
In the match load event, setup as class fields:
```java
struct = structure.get().getStructure(match);
region = struct.getRegion();
offset = origin.clone().subtract(struct.getDefinition().getOrigin()).toBlockVector();
```

Make the `matches` function check that `lastMismatch` block first, if its a mismatch you can just set cachedResult to false and you're done, quick check. If it does pass, then you do need to do the whole search for a new mismatch (or none, and then its a pass!).

Also you need a special pass only if the filter was passing, if it wasn't passing, modifications to other blocks than the mismatch aren't relevant.

Consider that if struct == null, you can just return false too (and the listener can also no-op) and consider stuff not initialized.

Listener should look something like:
```java
if (lastResult == null || struct == null) return;

var pos = BlockVectors.position(e.getBlock());

// Either its the mismatch, or the filter was passing, meaning
//   any change in the region invalidates it, set modified as first to check
if (lastMismatch.equals(pos) || 
    (lastResult &&  region.contains(pos.subtract(offset)))) {
  lastResult = null;
  lastMismatch.copy(pos);
  invalidate(...)  
}
```

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -63,4 +56,20 @@ public void placeAbsolute(BlockVector vector, boolean update) {
...
+      BlockMaterialData snapshotMaterial = snapshot.getOriginalMaterial(pos);
+      BlockMaterialData worldMaterial = MaterialData.block(blockData.getBlock(world, offset));
+
+      if (snapshotMaterial.encoded() != worldMaterial.encoded()) {
+        return blockData.getBlock(world, offset).getLocation().toVector().toBlockVector();
+      }
+    }
+
+    return null;
+  }
```
Pablete1234: When making functions like this one, we need to be very caureful with allocations, this is the opposite of caureful.

The idea behind `snapshot.getMaterials(region)` is that it returns an iterator that keeps returning itself, so for iterating 1024 materials it does not allocate 1k objects, it always keeps the same object. Also calling encode may not be the cheapest way to check that they are the same, and you shouldn't be afraid to implement methods for checking that two `BlockMaterialData`s are the same, or that a `BlockData` is the same as a block material data.

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -63,4 +56,20 @@ public void placeAbsolute(BlockVector vector, boolean update) {
     vector.subtract(getRegion().getBounds().getBlockMin());
     place(vector, update);
   }
+
+  public BlockVector findMismatch(World world, BlockVector origin) {
+    BlockVector offset = origin.clone().subtract(getDefinition().getOrigin()).toBlockVector();
+
+    for (BlockData blockData : snapshot.getMaterials(region)) {
+      BlockVector pos = blockData.getBlockVector();
+      BlockMaterialData snapshotMaterial = snapshot.getOriginalMaterial(pos);
```
Pablete1234: literally use `blockData` as your `snapshotMaterial`, blockData implements BlockMaterialData.

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -63,4 +56,20 @@ public void placeAbsolute(BlockVector vector, boolean update) {
...
+      BlockMaterialData worldMaterial = MaterialData.block(blockData.getBlock(world, offset));
+
+      if (snapshotMaterial.encoded() != worldMaterial.encoded()) {
```
Pablete1234: Implement a close to allocationless, `blockData.sameMaterialAs(offset)`, where internally it can have different impls per version and does the check for you using primitives as much as possible to avoid additional allocations

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -63,4 +56,20 @@ public void placeAbsolute(BlockVector vector, boolean update) {
...
+        return blockData.getBlock(world, offset).getLocation().toVector().toBlockVector();
```
Pablete1234:
```suggestion
        return blockData.getBlockVector();
```
no need for 4 allocations, this makes 0 allocs

Pablete1234 (changes requested):

Javggg:
I just updated the filter with the suggestions, sorry for the delay!
Let me know if it needs more work.

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -63,4 +56,27 @@ public void placeAbsolute(BlockVector vector, boolean update) {
...
   }
+
+  public boolean findMismatch(World world, BlockVector offset, BlockVector out) {
+    BlockMaterialData snapshotMaterial = snapshot.getOriginalMaterial(
+        out.getBlockX() - offset.getBlockX(),
+        out.getBlockY() - offset.getBlockY(),
+        out.getBlockZ() - offset.getBlockZ());
+    if (snapshotMaterial.encoded()
+        != MaterialData.block(world.getBlockAt(out.getBlockX(), out.getBlockY(), out.getBlockZ()))
+            .encoded()) {
```
Pablete1234, line 67: encoded form is not the correct way to compare materials, lift sameMaterialAs from BlockData to BlockMaterialData and use it here too

core/src/main/java/tc/oc/pgm/structure/Structure.java
```diff
@@ -63,4 +56,27 @@ public void placeAbsolute(BlockVector vector, boolean update) {
...
+      return true;
+    }
+
+    for (BlockData blockData : snapshot.getMaterials(region)) {
+      if (!blockData.sameMaterialAs(world, offset)) {
+        BlockVector sv = blockData.getBlockVector();
+        out.setX(sv.getBlockX() + offset.getBlockX());
+        out.setY(sv.getBlockY() + offset.getBlockY());
+        out.setZ(sv.getBlockZ() + offset.getBlockZ());
+        return true;
```
Pablete1234, line 77: instead of using an `out` param, return the BlockVector of where the first mismatch is, null means no mismatch.
`return true` -> `return new BlockVector(x, y,z);`
`return false` -> `return null;`

util/src/main/java/tc/oc/pgm/util/block/BlockData.java
```diff
@@ -35,4 +36,16 @@ default Block getBlock(World world, BlockVector offset) {
...
+  /**
+   * Check if this block data has the same material as the block at this position with the given
+   * offset in the world.
+   *
+   * @param world the world to check the block in
+   * @param offset the offset from the original position
+   * @return true if the materials match
+   */
+  default boolean sameMaterialAs(World world, BlockVector offset) {
+    return encoded() == MaterialData.block(getBlock(world, offset)).encoded();
```
Pablete1234, line 49: again, encoded form is not the way of comparing, but having it be part of the interface is the right start. I think it probably needs to be in BlockMaterialData so that it can be used in the first check too (as there you have just a material, not a block+material which is what BlockData is).

You need each implementation of BlockMaterialData to do the comparison, which is a very platform-dependant thing

core/src/main/java/tc/oc/pgm/filters/parse/FilterParser.java
```diff
@@ -626,6 +642,29 @@ public Filter parseOffsetFilter(Element el) throws InvalidXMLException {
...
+    var structure = parser
+        .reference(StructureDefinition.class, el, "structure")
+        .validate((FeatureValidation<StructureDefinition>) (def, node) -> {
+          int count = 0;
+          var it = def.getRegion().getStatic().getBlockVectorIterator();
+          while (it.hasNext()) {
+            it.next();
+            count++;
+          }
+          if (count > 1024)
```
Pablete1234, line 656: Do not count regions by actual block count, count by total volume of the bounds which is much easier to test.

If that's an issue for mapdevs because they're making absurdly big regions where very few blocks are actually part of the iterable we can revisit it later, but generally iterating a region implies iterating the whole bounds and filtering for what is contained, so it better reflects the cost.

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public boolean isDynamic() {
+    return true;
+  }
+
+  @Override
+  public boolean matches(MatchQuery query) {
+    return ((Reactor) query.state(this)).matches(query);
```
Pablete1234, line 37: pretty sure the cast is redundant and the type is filled in by the generic

```suggestion
    return query.state(this).matches(query);
```

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,108 @@
...
+    private final Structure struct;
+    private Region.Static region;
+    private BlockVector offset;
+
+    public Reactor(Match match, FilterMatchModule fmm) {
+      super(match, fmm);
+      match.addListener(this, MatchScope.LOADED);
+
+      struct = structure.get().getStructure(match);
+      if (struct != null) {
```
Pablete1234, line 68: That you added this nullcheck leads me to believe you've encountered some load order under which this structure hasn't been created yet at the time the filter is created. 

I'm not sure if it can happen or not, but what i am sure of is that if that happens, the current code will end up with a broken filter that is never able to fix itself.

Consider making this an init method that gets called either on the reactor, or whenever struct == null in matches() such that it will *eventually* get initialized, ie:

```
boolean ensureInit() {
  if (struct != null) return true;
  struct = structure.get().getStructure(match);
  if (struct == null) return false;
  region = struct.getRegion().getStatic();
  offset = origin.clone().subtract(struct.getDefinition().getOrigin()).toBlockVector();
  return true;
}
```
then on `matches` you can `if (!ensureInit()) return false;`

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,108 @@
...
+      if (struct == null) return false;
+      if (lastResult != null) return lastResult;
+
+      boolean hasMismatch = struct.findMismatch(match.getWorld(), offset, lastMismatch);
+      return lastResult = !hasMismatch;
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onBlockTransform(BlockTransformEvent e) {
+      if (lastResult == null || struct == null) return;
```
Pablete1234, line 84: This may need ensureInit call too, otherwise a dynamic trigger would never happen if initialization failed.

Also note, if lastResult starts as null it also skips this check so it never invalidates and it also never triggers dynamically.

core/src/main/java/tc/oc/pgm/filters/matcher/block/BuiltFilter.java
```diff
@@ -0,0 +1,108 @@
...
+
+      if (isMismatch || inRegion) {
+        lastResult = null;
+        lastMismatch.setX(x);
+        lastMismatch.setY(y);
+        lastMismatch.setZ(z);
+        invalidate(BuiltFilter.this, match);
+      }
+    }
+  }
```
Pablete1234, line 107: If the filter was used in a purely dynamic way, nothing would ensure initialization, so lastResult would always be null and nothing would ever trigger invalidation. Also using non-primitives is fine here, this runs once at most per modification, its only on the potentially large loops where we need to really make sure to avoid them

```suggestion
    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
    public void onBlockTransform(BlockTransformEvent e) {
      if (!ensureInit()) return;

      var position = BlockVectors.position(e.getBlock());
      
      // Wasn't matching, and last mismatch position isn't here
      if (Boolean.FALSE.equals(lastResult) && !lastMismatch.equals(position)) return;
      if (!region.contains(position)) return;
      
      lastMismatch.copy(position);
      invalidate(BuiltFilter.this, match);
    }
```
i'm thinking something along these lines should work, where if last result is false we only invalidate if the last mismatch is exactly what changed, otherwise we invalidate if any block in the region changed

Pablete1234 (changes requested):

PR #1684, opened 2026-06-08
<launch-projectile/> action

eliminate123:
This PR introduces a new action type ```<launch-projectile/>```. The XML looks like this:
```xml
<projectiles>
    <projectile id="test-arrow" projectile="Arrow" damage="80" destroy-filter="stone"/>
    <projectile id="lazer" name="lazer" projectile="Snowball" velocity="3.5" damage="50" throwable="false" cooldown="5s"/>
</projectiles>
<actions>
    <action id="test-action" scope="match" expose="true">
        <!-- Using toward attribute -->
        <launch-projectile projectile="test-arrow" origin="0,0,0" toward="5,5,5"/>
        <!-- Using yaw/pitch attributes instead -->
        <launch-projectile projectile="test-arrow" origin="0,0,0" yaw="45" pitch="22.5"/>
        <!-- Will only kill blue players -->
        <launch-projectile projectile="lazer" origin="0,0,0" toward="5,5,5" damage-filter="blue-team"/> 
    </action>
</actions>
```

(**NOTE:** ```toward="x,y,z"``` _is the direction the projectile will be fired towards once spawned in. It does not mean the projectile will travel until it hits that exact point_)

With ```<launch-projectile/>``` it would be nice to be able to launch custom fireworks, but at the moment custom fireworks can only be used for kits. My first thought was that ```<firework/>``` could live inside of ```<projectiles></projectiles>``` where it would then be given an id to be used for ```<launch-projectile projectile="firework-id" .../>```, but I'm not entirely sure and this sounds like it could become a proto bump.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -27,6 +27,7 @@ import tc.oc.pgm.action.actions.EnchantItemAction;
 import tc.oc.pgm.action.actions.ExposedAction;
 import tc.oc.pgm.action.actions.FillAction;
 import tc.oc.pgm.action.actions.KillEntitiesAction;
+import tc.oc.pgm.action.actions.LaunchProjectileAction;
 import tc.oc.pgm.action.actions.MessageAction;
 import tc.oc.pgm.action.actions.OpenShop;
 import tc.oc.pgm.action.actions.PasteStructureAction;
@@ -61,6 +62,7 @@ import tc.oc.pgm.filters.operator.AllFilter;
 import tc.oc.pgm.flag.FlagDefinition;
 import tc.oc.pgm.kits.Kit;
 import tc.oc.pgm.modules.WeatherMatchModule;
+import tc.oc.pgm.projectile.ProjectileDefinition;
 import tc.oc.pgm.shops.Shop;
 import tc.oc.pgm.shops.ShopModule;
 import tc.oc.pgm.shops.menu.Payable;
@@ -564,5 +566,17 @@ public class ActionParser {
     return MatchPlayer.class.isAssignableFrom(scope)
         ? new ScheduleAction.Player(after, (Action<? super MatchPlayer>) action)
         : new ScheduleAction<>(scope, after, action);
+  }
+
+  @MethodParser("launch-projectile")
+  public <T extends Filterable<?>> LaunchProjectileAction<T> parseProjectile(Element el, Class<T> scope)
+      throws InvalidXMLException {
+    scope = parseScope(el, scope);
+    var projectile = parser.reference(ProjectileDefinition.class, el, "id").required();
+    var origin = parser.vector(el, "origin").required();
+    var destination = parser.vector(el, "destination").required();
+    Filter damageFilter = parser.filter(el, "damage-filter").orNull();
+
+    return new LaunchProjectileAction(scope, projectile, origin, destination, damageFilter);
   }
 }
--- a/core/src/main/java/tc/oc/pgm/action/ActionParser.java
+++ b/core/src/main/java/tc/oc/pgm/action/ActionParser.java
@@ -19,6 +19,7 @@ import net.kyori.adventure.text.Component;
 import net.kyori.adventure.title.Title;
 import org.bukkit.enchantments.Enchantment;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.util.Vector;
 import org.jdom2.Element;
 import org.jspecify.annotations.Nullable;
 import tc.oc.pgm.action.actions.ActionNode;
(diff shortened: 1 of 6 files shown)


core/src/main/java/tc/oc/pgm/modules/LaunchProjectileModule.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.logging.Logger;
+import org.jdom2.Document;
+import tc.oc.pgm.api.map.MapModule;
+import tc.oc.pgm.api.map.factory.MapFactory;
+import tc.oc.pgm.api.map.factory.MapModuleFactory;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.util.xml.InvalidXMLException;
+
+public class LaunchProjectileModule implements MapModule<LaunchProjectileMatchModule> {
```
cswhite2000, line 11: Why are you making a new module here when we already have a projectile module?

core/src/main/java/tc/oc/pgm/action/actions/LaunchProjectileAction.java
```diff
@@ -0,0 +1,64 @@
...
+      Location dirLoc = new Location(null, 0, 0, 0, yaw, pitch);
+      dir = dirLoc.getDirection();
+    }
+    dir.multiply(def.getVelocity());
+    Location loc = origin.toLocation(match.getWorld());
+
+    Entity proj = match.getWorld().spawn(loc, def.getProjectile());
+    proj.setVelocity(dir);
+    proj.setMetadata("projectileDefinition", new FixedMetadataValue(PGM.get(), def));
+    proj.setMetadata("launchProjectile", new FixedMetadataValue(PGM.get(), this));
```
cswhite2000, line 62: Please don't store the action itself in the entity

core/src/main/java/tc/oc/pgm/action/actions/LaunchProjectileAction.java
```diff
@@ -0,0 +1,64 @@
...
+
+    Vector dir;
+    if (toward != null) {
+      dir = toward.clone().subtract(origin).normalize();
+    } else {
+      Location dirLoc = new Location(null, 0, 0, 0, yaw, pitch);
+      dir = dirLoc.getDirection();
+    }
+    dir.multiply(def.getVelocity());
+    Location loc = origin.toLocation(match.getWorld());
```
cswhite2000, line 57: All of this stuff should be decided at parse time and save a simpler set of variables for where to launch.
This would be the correct location if you were using formulas though for this, but you aren't currently doing that.

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -565,4 +568,34 @@ public <B extends Filterable<?>> Action<?> parseSchedule(Element el, Class<B> sc
...
+      Element el, Class<T> scope) throws InvalidXMLException {
+    scope = parseScope(el, scope);
+    var projectile =
+        parser.reference(ProjectileDefinition.class, el, "projectile").required();
+    var origin = parser.vector(el, "origin").required();
+    Vector toward = XMLUtils.parseVector(Node.fromAttr(el, "toward"));
+    Node yawNode = Node.fromAttr(el, "yaw");
+    Node pitchNode = Node.fromAttr(el, "pitch");
+    float yaw = XMLUtils.parseNumber(yawNode, Float.class, 0F);
+    float pitch = XMLUtils.parseNumber(pitchNode, Float.class, 0F);
```
cswhite2000, line 583: why aren't you using the fluent parser for these ones?

core/src/main/java/tc/oc/pgm/action/actions/LaunchProjectileAction.java
```diff
@@ -0,0 +1,64 @@
...
+  @Override
+  public void trigger(T t) {
+    Match match = t.getMatch();
+    ProjectileDefinition def = projectileReference.get();
+
+    Vector dir;
+    if (toward != null) {
+      dir = toward.clone().subtract(origin).normalize();
+    } else {
+      Location dirLoc = new Location(null, 0, 0, 0, yaw, pitch);
```
cswhite2000, line 53: null world?

core/src/main/java/tc/oc/pgm/action/ActionParser.java
```diff
@@ -565,4 +568,34 @@ public <B extends Filterable<?>> Action<?> parseSchedule(Element el, Class<B> sc
...
+      }
+      if (pitchNode != null) {
+        throw new InvalidXMLException("'toward' and 'pitch' cannot both be defined", el);
+      }
+    }
+    if (toward == null && yawNode == null && pitchNode == null) {
+      throw new InvalidXMLException("Either 'toward' or 'pitch/yaw' must be defined", el);
+    }
+
+    Filter damageFilter = parser.filter(el, "damage-filter").orNull();
```
cswhite2000, line 596: Better to just add a damage filter to the projectile definition than in the action

cswhite2000 (changes requested):

Pablete1234:
I do wonder, what's the usecase for this one? Is it for repeatedly spawning these projectiles in some region or area?

To me it seems like something xml spawners should support spawning instead of actions

PR #1686, opened 2026-06-11
Bridge egg

eliminate123:
2026-06-11 14-23-40 (online-video-cutter com)(1)

This PR adds new attributes for ```<projectile>```:

```bridge-egg="true/false"```
```bridge-range="integer"``` (Distance bridge continues before stopping)
```bridge-material="filter"``` (Blocks used for bridge. If >1 block type, randomizes placement)
```team-color="true/false"``` (Dyes dyeable blocks)
```silent="true/false"``` (Mutes bridge noises)

```xml
<projectiles>
    <projectile id="bridge" projectile="Egg" click="right" bridge-egg="true" bridge-range="30" bridge-material="bridge-block-filter" team-color="true" velocity="2"/>
</projectiles>
<filters>
    <any id="bridge-block-filter">
        <material>stone</material>
        <material>dirt</material>
        <material>diamond block</material>
        <material>wool</material>
    </any>
</filters>
```
If a player leaves/obs/switches teams with an active bridge egg, its killed. If a bridge egg hits any region where building isnt enabled the bridge egg is killed. If the bridge egg hits vanilla height limit/<maxheightlimit> it gets killed. If a bridge egg hits the void it gets killed.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
@@ -4,7 +4,6 @@ import static tc.oc.pgm.util.material.ColorUtils.COLOR_UTILS;
 
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Future;
@@ -25,6 +24,7 @@ import tc.oc.pgm.api.player.ParticipantState;
 import tc.oc.pgm.events.ParticipantBlockTransformEvent;
 import tc.oc.pgm.util.TimeUtils;
 import tc.oc.pgm.util.bukkit.Sounds;
+import tc.oc.pgm.util.material.BlockMaterialData;
 import tc.oc.pgm.util.nms.NMSHacks;
 
 public class ActiveBridgeEgg {
@@ -39,7 +39,7 @@ public class ActiveBridgeEgg {
   private Location prevPos;
   private final Location throwOrigin;
   private final int bridgeRange;
-  private final List<Material> bridgeMaterial;
+  private final BlockMaterialData bridgeMaterial;
   private final boolean teamColor;
   private final boolean silent;
   private Future<?> runnableTask = null;
@@ -54,7 +54,7 @@ public class ActiveBridgeEgg {
       Entity proj,
       PlayerInteractEvent launch,
       int bridgeRange,
-      List<Material> bridgeMaterial,
+      BlockMaterialData bridgeMaterial,
       Location bridgePos,
       boolean teamColor,
       boolean silent) {
@@ -136,16 +136,16 @@ public class ActiveBridgeEgg {
 
       for (Location loc : toPlace) {
         Block block = loc.getBlock();
-        Material material = bridgeMaterial.get(match.getRandom().nextInt(bridgeMaterial.size()));
 
         if (block.getType() == Material.AIR) {
           ParticipantState state = player.getParticipantState();
           BlockTransformEvent bte = state != null
-              ? new ParticipantBlockTransformEvent(launch, block, material, state)
-              : new BlockTransformEvent(launch, block, material);
+              ? new ParticipantBlockTransformEvent(
+                  launch, block, bridgeMaterial.getItemType(), state)
+              : new BlockTransformEvent(launch, block, bridgeMaterial.getItemType());
           match.callEvent(bte);
           if (!bte.isCancelled()) {
-            block.setType(material);
+            bridgeMaterial.applyTo(block, true);
             if (!silent) {
               Sounds.play(player.getBukkit(), Sounds.BRIDGE_EGG, loc);
             }
--- a/core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
+++ b/core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
@@ -1,0 +1,174 @@
+package tc.oc.pgm.projectile;
+
+import static tc.oc.pgm.util.material.ColorUtils.COLOR_UTILS;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import org.bukkit.DyeColor;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+import org.bukkit.entity.Entity;
+import org.bukkit.event.player.PlayerInteractEvent;
+import org.bukkit.scheduler.BukkitRunnable;
+import org.bukkit.util.Vector;
+import tc.oc.pgm.api.event.BlockTransformEvent;
+import tc.oc.pgm.api.match.Match;
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.player.ParticipantState;
+import tc.oc.pgm.events.ParticipantBlockTransformEvent;
+import tc.oc.pgm.util.TimeUtils;
+import tc.oc.pgm.util.bukkit.Sounds;
+import tc.oc.pgm.util.nms.NMSHacks;
+
+public class ActiveBridgeEgg {
+
+  private final Match match;
+  private final MatchPlayer player;
+  private final UUID uuid;
+  private final HashMap<UUID, ActiveBridgeEgg> registry;
+  private final Entity proj;
+  private final PlayerInteractEvent launch;
+  private Location bridgePos;
+  private Location prevPos;
+  private final Location throwOrigin;
+  private final int bridgeRange;
+  private final List<Material> bridgeMaterial;
+  private final boolean teamColor;
+  private final boolean silent;
+  private Future<?> runnableTask = null;
+  private final BukkitRunnable runnable = new BridgeEggRunnable();
+  private int ticksElapsed = 0;
+
+  public ActiveBridgeEgg(
+      Match match,
+      MatchPlayer player,
+      UUID uuid,
+      HashMap<UUID, ActiveBridgeEgg> registry,
+      Entity proj,
+      PlayerInteractEvent launch,
+      int bridgeRange,
+      List<Material> bridgeMaterial,
+      Location bridgePos,
+      boolean teamColor,
+      boolean silent) {
+    this.match = match;
+    this.player = player;
+    this.uuid = uuid;
+    this.registry = registry;
+    this.proj = proj;
+    this.launch = launch;
+    this.throwOrigin = bridgePos.clone();
+    this.bridgeRange = bridgeRange;
+    this.bridgeMaterial = bridgeMaterial;
+    this.bridgePos = bridgePos;
+    this.prevPos = bridgePos.clone();
+    this.teamColor = teamColor;
+    this.silent = silent;
+  }
+
+  public boolean isOwnedBy(MatchPlayer player) {
+    return this.player.equals(player);
+  }
+
+  public void start() {
+    if (runnableTask != null && !this.runnableTask.isDone()) return;
+
+    this.runnableTask = match
+        .getExecutor(MatchScope.RUNNING)
+        .scheduleAtFixedRate(this.runnable, 0, TimeUtils.TICK, TimeUnit.MILLISECONDS);
+  }
+
+  private class BridgeEggRunnable extends BukkitRunnable {
+    @Override
+    public void run() {
+      if (proj.isDead()) {
+        stop();
+        return;
+      }
+
+      if (proj.getLocation().getY() <= NMSHacks.NMS_HACKS.getMinHeight(proj.getWorld()) - 16) {
+        stop();
+        return;
+      }
+
+      org.bukkit.Location currentPos = proj.getLocation();
+      Location targetPos = currentPos.clone();
+      targetPos.setY(currentPos.getY() - 2);
+      Vector dir = targetPos.toVector().subtract(bridgePos.toVector()).normalize();
+      bridgePos = bridgePos.add(dir);
+      double disTraveled = throwOrigin.distance(bridgePos);
+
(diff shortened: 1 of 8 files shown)


core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,176 @@
...
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.player.ParticipantState;
+import tc.oc.pgm.events.ParticipantBlockTransformEvent;
+import tc.oc.pgm.util.TimeUtils;
+import tc.oc.pgm.util.bukkit.Sounds;
+import tc.oc.pgm.util.platform.Platform;
+
+public class ActiveBridgeEgg {
+  // Void heights with a buffer for faster velocities
+  private static final int VOID_HEIGHT = Platform.isLegacy() ? -16 : -80;
```
Pablete1234: Currently all of pgm assumes min y is simply 0, which is wrong, but at least as long as all maps are built with min y = 0 (and modern worlds support this great via datapacks customizing the dimension) it all works as intended.

If you want to have different min heights, the min needs to depend on the worlds' min height, not on the platform, otherwise you can probably default to just 0.
for diff min heights, make a platform-specific nms method like `getMinHeight(World world)` which on modern returns the true min height, and on legacy just defaults to 0

technodono:
Oh golly do I already have a map that could use this 😈

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -27,12 +29,11 @@
 import tc.oc.pgm.util.xml.Node;
 import tc.oc.pgm.util.xml.XMLUtils;
 
-public class ProjectileModule implements MapModule<ProjectileMatchModule> {
-  private final ImmutableSet<ProjectileDefinition> projectileDefinitions;
+public record ProjectileModule(ImmutableSet<ProjectileDefinition> projectileDefinitions)
+    implements MapModule<ProjectileMatchModule> {
```
Pablete1234: records forcibly expose internal state, we don't want all modules' internal data to be exposed as it makes it harder to change in the future if other plugins depend on it

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -77,6 +78,79 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
             XMLUtils.parseBoolean(projectileElement.getAttribute("throwable"), true);
         boolean precise = XMLUtils.parseBoolean(projectileElement.getAttribute("precise"), true);
 
+        boolean bridgeEgg =
+            XMLUtils.parseBoolean(projectileElement.getAttribute("bridge-egg"), false);
```
Pablete1234: instead of `bridge-egg="true"` i'd expect `projectile="bridge-egg"`, which goes in line with what i mentioned about supporting other projectile types that aren't vanilla entities but rather more expansive concepts

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -77,6 +78,79 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+          }
+        } else {
+          if (bridgeRange != 0) {
+            throw new InvalidXMLException(
+                "'bridge-range' is only supported for 'bridge-egg'", projectileElement);
+          }
+          if (!bridgeMaterials.isEmpty()) {
+            throw new InvalidXMLException(
+                "'bridge-material' is only supported for 'bridge-egg'", projectileElement);
+          }
```
Pablete1234: This is pain copypaste slop of mutually exclusive tags, many of them aren't even necessarily mutually exclusive they're just unused depending on what is being parsed. So bridge eggs just shouldn't be reading stuff like power, and non-bridge eggs shouldn't be reading stuff like bridge-range. 

PGM will report them as unused nodes and that's all you really need here.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
eliminate123: I made it so it now uses projectile="bridge-egg"

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
eliminate123: I updated NMSHACKS. I still added a -16 buffer on the bridge eggs.

core/src/main/java/tc/oc/pgm/projectile/ProjectileDefinition.java
```diff
@@ -22,6 +23,10 @@ public class ProjectileDefinition extends SelfIdentifyingFeatureDefinition {
   protected boolean throwable;
   protected boolean precise;
   protected BlockMaterialData blockMaterial;
+  protected final int bridgeRange;
+  protected List<Material> bridgeMaterial;
+  protected boolean teamColor;
+  protected boolean silent;
```
Pablete1234: Maybe i didn't make a clear enough comment in the first review, but what i mean is that the definition shouldn't have stuff not relevant to it, so i'd expect this to instead work more like:



```suggestion
  protected PgmProjectile projectile;
  protected List<PotionEffect> potion;
  protected Filter destroyFilter;
  protected Duration coolDown;
  protected boolean throwable;
  protected boolean precise;
  protected BlockMaterialData blockMaterial;
```

where that `PgmProjectile` can be a direct entity can be something along the lines of:

```java
interface PgmProjectile {

  public void spawn(...);

  public void despawn(...);

}
```

and you can have differing implementations like:
`record EntityProjectile(Class<? extends Entity> projectile) implements PgmProjectile {...}` (current behavior) 
`record BridgeEggProjectile(int range, List<Material> material, boolean teamColor, boolean silent) implements PgmProjectile { ... }`

That means that attributes that are specific to one type are only parsed by that type, so in the parsing code you can have something like:

```
PgmProjectile projectile = switch (projType) {
  case "bridge-egg" -> new BridgeEggProjectile(
      parser.parseInt(el, "bridge-range"),
      parser.parseMaterial(...),
      parser.parseBool(el, "bridge-team-color"),
      parser.parseBool(el, "bridge-silent"));
  default -> new EntityProjectile(...);
}
```

This way the attributes that are specific to bridge-eggs only get parsed if bridge egg is relevant, and are passed along with it.



You can see a big portion of this same idea implemented in #1502 (https://github.com/PGMDev/PGM/pull/1502/changes#diff-079be93d9cad6596ef0f505dad1789ed6b2f93ac49776622e21b4e82f3df6ae5R19)

Pablete1234: I've gone ahead and implemented alot of the scaffolding to make this possible, if you rebase your changes ontop of latest pgm you should see it becomes alot more natural to implement new projectile types now

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
```diff
@@ -88,26 +93,51 @@ public ProjectileModule parse(MapFactory factory, Logger logger, Document doc)
...
+      }
+      if (bridgeMaterialEl != null) {
+        for (Element child : bridgeMaterialEl.getChildren()) {
+          Material material = XMLUtils.parseMaterial(new Node(child));
+          if (!material.isBlock()) {
+            throw new InvalidXMLException(
+                "'" + material + "' is not a valid block for 'bridge-material'", child);
+          }
+          bridgeMaterials.add(material);
+        }
```
Pablete1234: This is NOT how you parse a list of materials.
You should not need `Document` being passed into this parser at all, and you should not be reaching into a <filters> child of the root node to try to search for your id. A filter can be defined in many diff places in the xml, and this code has like at least 15 different reasons why it would not be able to find or use something like:
```xml
<map>
  <regions>
    <apply>
        <block id="good-blocks">
          <allow>
            <any>
              <material>wool</material>
              <material>stained clay</material>
             </any>
          </allow>
        </block>
    </apply>
    </apply block="never" message="only bridge egg materials allowed idk"/>
  </regions>
</map>
```
You're assuming that there's going to be some filter with a list of materials, but that's not how filters work. A filter `always` should in theory mean all materials, but it would not work like this with current code, and any filter that has any amount of layers or conditions, or is defined anywhere else than the root, will not be found.

What you're actually looking for is not a **filter**, because realistically you cannot handle one for this purpose; its a list of blocks or materials.
there's `XMLUtils.parseBlockData` which is used by several features that need to place blocks in the world based on an xml definition (objective modes, fill actions, blockdrop replacements). You can simply parse something like `bridge-material="wool;stained clay"` by splitting over `;` and passing it thru that xml utils method i mentioned above. i do wonder tho, is there a use-case for multiple materials here? cause if there isn't, parsing a single material would be much easier.

Pablete1234 (changes requested):
On the parsing side, already so much better than before, but the use of a filter is not right for this task.

On the impl side, the point of having stuff like `BridgeEggProjectile` separate is such that the logic that is specific to it can be moved there too, or be generified in ways that can be more reusable, instead of all fitting in "ProjectileMatchModule".

ProjectileMatchModule should deal with interfaces that can be implemented by projectiles, instead of with specific classes.

Think of, instead of `if (bridgeEgg) -> create new ActiveBridgeEgg()` generify it to `if (tickingProjectile) -> create ticker()`, and the logic of how to create this ticker can be in the PgmProjectile (ie: implemented by BridgeEgg) itself, in ways that can be reused for future projectiles instead of hard-coding to bridge-eggs

core/src/main/java/tc/oc/pgm/projectile/ProjectileModule.java
eliminate123: It would probably be fine with parsing only 1 material type. Ill try to update when I get a chance

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,174 @@
...
+import tc.oc.pgm.api.match.MatchScope;
+import tc.oc.pgm.api.player.MatchPlayer;
+import tc.oc.pgm.api.player.ParticipantState;
+import tc.oc.pgm.events.ParticipantBlockTransformEvent;
+import tc.oc.pgm.util.TimeUtils;
+import tc.oc.pgm.util.bukkit.Sounds;
+import tc.oc.pgm.util.material.BlockMaterialData;
+import tc.oc.pgm.util.nms.NMSHacks;
+
+public class ActiveBridgeEgg {
```
Pablete1234, line 30: Should be moved inside BridgeEggProjectile

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,174 @@
...
+
+  private final Match match;
+  private final MatchPlayer player;
+  private final UUID uuid;
+  private final HashMap<UUID, ActiveBridgeEgg> registry;
```
Pablete1234, line 35: One bridge egg holds a registry of N bridge eggs? that makes no sense

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,174 @@
...
+        stop();
+        return;
+      }
+
+      if (proj.getLocation().getY() <= NMSHacks.NMS_HACKS.getMinHeight(proj.getWorld()) - 16) {
+        stop();
+        return;
+      }
+
+      org.bukkit.Location currentPos = proj.getLocation();
```
cswhite2000, line 101:
```suggestion
      Location currentPos = proj.getLocation();
```

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,174 @@
...
+    if (runnableTask != null && !this.runnableTask.isDone()) return;
+
+    this.runnableTask = match
+        .getExecutor(MatchScope.RUNNING)
+        .scheduleAtFixedRate(this.runnable, 0, TimeUtils.TICK, TimeUnit.MILLISECONDS);
+  }
+
+  private class BridgeEggRunnable extends BukkitRunnable {
+    @Override
+    public void run() {
```
cswhite2000, line 90: I'd split this method up

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,174 @@
...
+  private Location bridgePos;
+  private Location prevPos;
+  private final Location throwOrigin;
+  private final int bridgeRange;
+  private final BlockMaterialData bridgeMaterial;
+  private final boolean teamColor;
+  private final boolean silent;
+  private Future<?> runnableTask = null;
+  private final BukkitRunnable runnable = new BridgeEggRunnable();
+  private int ticksElapsed = 0;
```
cswhite2000, line 47: You shoudn't need this many fields here. 
Also, registry should stay in the match module and not be passed in here

core/src/main/java/tc/oc/pgm/projectile/ActiveBridgeEgg.java
```diff
@@ -0,0 +1,174 @@
...
+  private final PlayerInteractEvent launch;
+  private Location bridgePos;
+  private Location prevPos;
+  private final Location throwOrigin;
+  private final int bridgeRange;
+  private final BlockMaterialData bridgeMaterial;
+  private final boolean teamColor;
+  private final boolean silent;
+  private Future<?> runnableTask = null;
+  private final BukkitRunnable runnable = new BridgeEggRunnable();
```
Pablete1234, line 46: You may as well just implement runnable directly

Pablete1234 (changes requested):
A few of my older comments are still not addressed, projectile match module should not have a hard dependency on exactly bridge eggs directly, it should have dependencies on generic interfaces that multiple projectiles can implement

PR #1690, opened 2026-06-15
Implement pickups

### LINKED ISSUE #1018: Reimplement Pickups 🔮
"Pickups are used to create a pickup-able kit at a specific location. While somewhat similar to a kit applied using a region, pickups are represented by a physical object and have a refresh time."
https://pgm.dev/docs/modules/gear/pickups
(the only object that was accepted was ender crystal)

idk if this is the old code or what the what if it's not someone correct me please
https://github.com/OvercastNetwork/ProjectAres/tree/master/PGM/src/main/java/tc/oc/pgm/pickup

probably more viable for arcade or maybe even comp maps 😋i stumbled upon it in the documentation and i thought it's cool and wanted to use it then i got disappointed. couldn't find any issue for reimplementation yet.

calcastor:
Closes https://github.com/PGMDev/PGM/issues/1018

~~Additional support for entities that are not end crystals is implemented; TNT is excluded specifically due to legacy clientside issues.~~ Additional support added for mobs as pickup entities.

Additional support for mob attributes in the following form:

```xml
    <pickup id="item-box-1" name="`k1`r`l-`r`d`lItem Box`r`l-`r`k1" spawn-filter="always" pickup-filter="always" respawn-time="1m" pickup-time="5s" kit="shield-kit" effects="true" sounds="true">
        <appearance type="bee" glowing="true"/>
        <region>
            <point>410.5,62,-609.5</point>
        </region>
    </pickup>
```

built on the implementation in https://github.com/PGMDev/PGM/pull/1664


core/src/main/java/tc/oc/pgm/pickup/PickupMatchModule.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class PickupMatchModule implements MatchModule {
+
+  private final Match match;
+  private final List<Pickup> pickups;
+
+  public PickupMatchModule(Match match, List<Pickup> pickups) {
+    this.match = match;
+    this.pickups = pickups;
+    match.addTickable(new PickupTickTask(pickups), MatchScope.LOADED);
```
Pablete1234: You can just implement Tickable and Listener directly in the match module, there's little reason to keep them separate in this case

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/impl/ModernEffects.java
```diff
@@ -84,4 +87,37 @@ public void blockBreak(Location location, BlockMaterialData material) {
...
+          world,
+          random(box.getMinX(), box.getMaxX()),
+          random(box.getMinY(), box.getMaxY()),
+          random(box.getMinZ(), box.getMaxZ()));
+      world.spawnParticle(particle, location, 3, 1d, 1d, 1d, 0d, null, true);
+    }
+  }
+
+  private double random(double min, double max) {
+    return min == max ? min : ThreadLocalRandom.current().nextDouble(min, max);
```
Pablete1234: any reason to be using a concurrent random? why not just create a `Random` instance on this class once and reuse it?

platform/platform-sportpaper/src/main/java/tc/oc/pgm/platform/sportpaper/impl/SpNMSHacks.java
```diff
@@ -77,6 +79,43 @@ public void freezeEntity(Entity entity) {
...
+    var box = nmsEntity.getBoundingBox();
+    double centerX = (box.a + box.d) / 2d;
+    double centerY = (box.b + box.e) / 2d;
+    double centerZ = (box.c + box.f) / 2d;
+    double boxRadius = Math.sqrt(Math.pow((box.d - box.a) / 2d, 2)
+        + Math.pow((box.e - box.b) / 2d, 2)
+        + Math.pow((box.f - box.c) / 2d, 2));
+    return VectorUtils.segmentIntersectsSphere(
+        from, to, new Vector(centerX, centerY, centerZ), boxRadius + radius);
+  }
```
Pablete1234: Seems like this and the modern version are like 90% the same, and the only difference is how you obtain the bounding box.
I think you could simplify this to just `double getBoundingBoxRadius(Entity entity)` and keep only the specific part that needs to be version-specific. The world checks, and the intersection, are the same and can be put in the shared module as part of the pickup module itself

core/src/main/java/tc/oc/pgm/pickup/PickupModule.java
```diff
@@ -0,0 +1,116 @@
...
+        Region region = factory
+            .getRegions()
+            .parseRequiredProperty(el, "region", RandomPointsValidation.INSTANCE);
+        Kit kit = factory.getKits().parseKitProperty(el, "kit", KitNode.EMPTY);
+        Duration respawn =
+            XMLUtils.parseDuration(Node.fromAttr(el, "respawn-time"), Duration.ofSeconds(3));
+        Duration cooldown =
+            XMLUtils.parseDuration(Node.fromAttr(el, "pickup-time"), Duration.ofSeconds(3));
+        boolean effects = XMLUtils.parseBoolean(Node.fromAttr(el, "effects"), true);
+        boolean sounds = XMLUtils.parseBoolean(Node.fromAttr(el, "sounds"), true);
```
Pablete1234: This could really make use of modern parsing using the fluent parser in `factory.getParser()`

core/src/main/java/tc/oc/pgm/pickup/PickupModule.java
```diff
@@ -0,0 +1,116 @@
...
+
+      EntityType appearance = EntityTypes.getByName(node.getValueNormalize());
+      if (appearance == null || appearance == EntityTypes.UNKNOWN) {
+        throw new InvalidXMLException("Invalid pickup appearance '" + node.getValue() + "'", node);
+      }
+      if (appearance.getEntityClass() == null || !appearance.isSpawnable()) {
+        throw new InvalidXMLException(
+            "Pickup appearance '" + node.getValue() + "' is not spawnable", node);
+      }
+      return appearance;
```
Pablete1234: i wonder if it makes sense to reuse what cs implemented to define entities and properties for them here, so you could for example have a horse of a specific variant be the pickup.

core/src/main/java/tc/oc/pgm/pickup/Pickup.java
```diff
@@ -0,0 +1,159 @@
...
+    this.entity = entity;
+    this.location = location.clone();
+    this.spawnAtTick = -1;
+    playEffects(true, entity);
+  }
+
+  private void setupEntity(Entity entity) {
+    NMS_HACKS.freezeEntity(entity);
+    NMS_HACKS.setupPickup(entity);
+    if (entity instanceof TNTPrimed tnt) tnt.setFuseTicks(Integer.MAX_VALUE);
```
Pablete1234: i'm pretty sure this is not gonna solve it for older clients, they'll just see the tnt vanish after 4s even if its still there serverside
idk that there's much we can do tho

core/src/main/java/tc/oc/pgm/pickup/Pickup.java
```diff
@@ -0,0 +1,159 @@
...
+  public PickupDefinition getDefinition() {
+    return definition;
+  }
+
+  private boolean isSpawned() {
+    return entity != null;
+  }
+
+  private void spawn() {
+    Location location = definition.getRegion().getRandom(match).toLocation(match.getWorld());
```
Pablete1234: If you're going to use the region just to pick points from it, maybe a point provider makes more sense, it'd allow xml to set pitch/yaw, as well as creating sequential points etc

core/src/main/java/tc/oc/pgm/pickup/Pickup.java
```diff
@@ -0,0 +1,159 @@
+package tc.oc.pgm.pickup;
+
+import static tc.oc.pgm.util.TimeUtils.isInfinite;
+import static tc.oc.pgm.util.TimeUtils.toTicks;
```
Pablete1234: Generally speaking its best to avoid static imports of functions since it hides where they're coming from. We have an exception for this for components because of how prevalent they are, but i don't think these need the static imports

core/src/main/java/tc/oc/pgm/pickup/Pickup.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+    if (definition.hasSounds()) {
+      match.playSound(spawn ? Sounds.PICKUP_APPEAR : Sounds.PICKUP_DISAPPEAR, location);
+    }
+  }
+
+  private void playPickupEffects(Entity entity) {
+    if (definition.hasEffects()) EFFECTS.pickupEffect(entity, PickupEffect.PICKUP);
+    if (definition.hasSounds()) match.playSound(Sounds.PICKUP_DISAPPEAR, entity.getLocation());
+  }
```
Pablete1234:
```suggestion
  private void playEffects(Entity entity, PickupEffect effect) {
    if (definition.hasEffects()) EFFECTS.pickupEffect(entity, effect);
    if (definition.hasSounds()) match.playSound(soundFor(effect), entity.getLocation());
  }
```
Feels like these 2 functions are effectively encoding the 3 reasons to play effects, and would be better served by just one function where you can say which effect to play instead.

the `soundFor` function would be something like:


```java
public Sound soundFor(PickupEffect effect) {
  return switch (effect) {
    case SPAWN -> Sounds.PICKUP_APPEAR;
    case ... -> ...;
    case ... -> ...;
  }
}
```
if the sounds could be encoded into the enum that'd be even better so you could just `effect.sound()` instead

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/pickup/PickupMatchModule.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public void tick(Match match, Tick tick) {
+    for (Pickup pickup : pickups) {
+      pickup.tick(match, tick);
+    }
+  }
+
+  @EventHandler(ignoreCancelled = true)
+  public void onMove(PlayerMoveEvent event) {
```
cswhite2000, line 51: Should this be a coarse move event?
Alternatively could we at least early exit if say only the yaw and pitch of the new locations differ?

Also, will this trigger for a teleport, and say pickup anything between the teleport location and the destination?

calcastor: It shouldn't, as far as I can tell this won't listen on the teleport event despite its inheritance of it.

I've made changes to address your other comment and also added a listener for the teleport event so it doesn't pickup anything between from and to locations.

core/src/main/java/tc/oc/pgm/pickup/PickupAppearance.java
```diff
@@ -0,0 +1,24 @@
+package tc.oc.pgm.pickup;
+
+import org.bukkit.Location;
+import org.bukkit.entity.EntityType;
+import tc.oc.pgm.entity.SpawnableEntity;
+
+sealed interface PickupAppearance permits PickupAppearance.Entity, PickupAppearance.Mob {
```
Pablete1234: Given we now have a way to spawn custom mobs, it should probably use that (ie: SpawnableMob using the syntax from spawners)

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/entity/EntitySpawnFactory.java
```diff
@@ -0,0 +1,9 @@
+package tc.oc.pgm.entity;
+
+import org.bukkit.Location;
+import org.bukkit.entity.Entity;
+
+@FunctionalInterface
+public interface EntitySpawnFactory {
```
Pablete1234: Not entirely sure why this exists, just use `SpawnableEntity` directly, if we need that to change in the future we can look into that, but atm it seems like there's a single impl

core/src/main/java/tc/oc/pgm/pickup/PickupModule.java
```diff
@@ -0,0 +1,167 @@
...
+        return parseAppearanceElement(node.getElement(), factory);
+      }
+
+      return new PickupAppearance.Entity(parseAppearanceType(node));
+    }
+
+    private PickupAppearance parseAppearanceElement(Element el, MapFactory factory)
+        throws InvalidXMLException {
+      EntityType appearance = parseAppearanceType(Node.fromRequiredAttr(el, "type"));
+      if (LivingEntity.class.isAssignableFrom(appearance.getEntityClass())) {
```
Pablete1234: Instead of trying to workaround living entity only limitation in SpawnableEntity, just expand SpawnableEntity to allow arbitrary entities; if they're not LivingEntity then just no kit/properties are applied, and if anything, limit spawners to only allow living entity types instead of limiting all uses

PR #1697, opened 2026-07-09
Fix one filter operator not correctly evaluating allow

mitchts:
Currently the `<one> ... </one>` filter operator only returns allow if the last child allows and the proceeding child denies or all siblings abstain. This does not achieve the filter's purpose of returning allow when only one child allows, as demonstrated in this video.

https://github.com/user-attachments/assets/965c0994-2b49-4e81-b990-00bdd1b3cedf

This PR corrects this behaviour to do what it says on the tin.

https://github.com/user-attachments/assets/247e5056-f506-4545-8128-fab58a22d89c

[Demo map](https://github.com/user-attachments/files/29847256/one_filter.zip)

### DIFF
--- a/core/src/main/java/tc/oc/pgm/filters/operator/OneFilter.java
+++ b/core/src/main/java/tc/oc/pgm/filters/operator/OneFilter.java
@@ -12,17 +12,18 @@ public class OneFilter extends MultiFilterFunction {
   @Override
   public QueryResponse query(Query query) {
     // returns true if exactly one of the filters match
+    boolean hasAllow = false;
     QueryResponse response = QueryResponse.ABSTAIN;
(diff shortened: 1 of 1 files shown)


cswhite2000:
Do any maps rely on the current behavior?

calcastor:
> Do any maps rely on the current behavior?

The OCC repos contain one unaffected usage of the `one` filter; this PR seems to resolve an issue that has prevented broader usage of the filter in the past.

cswhite2000:
Please sign off the commit https://github.com/PGMDev/PGM/blob/dev/docs/CONTRIBUTING.md
ie `git commit -S -s`

Pablete1234 (approved):

PR #1698, opened 2026-07-14
fix: don't send legacy color codes through `player()` component

TTtie:
Fixes a practically invisible oversight where the code path for getting a component for a flair from the config file would emit components with legacy color codes, later on being sent over to players using the `player()` component.

This behavior is obscured by both the Minecraft client (legacy and modern alike) for players, and by `adventure-plaform-bukkit` which falls through to legacy string-based `sendMessage` API for consoles. 

It may be observed, for example, by swapping `adventure-platform-bukkit` on a modern server to an alternative ABI-compatible implementation that proxies Adventure calls through the PGM's component renderer directly to Paper's Adventure implementation:

image

### DIFF
--- a/core/src/main/java/tc/oc/pgm/api/Config.java
+++ b/core/src/main/java/tc/oc/pgm/api/Config.java
@@ -1,5 +1,6 @@
 package tc.oc.pgm.api;
 
+import static net.kyori.adventure.text.Component.empty;
 import static net.kyori.adventure.text.Component.newline;
 import static net.kyori.adventure.text.Component.text;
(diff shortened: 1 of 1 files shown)


cswhite2000 (dismissed):

Pablete1234:
Please doublecheck that this doesn't add more levels of nesting, and produces just something like `{"text":"*","color":"yellow"}`, and not something like `{"color":"yellow",extra: [{"text":"*"}]}`

It has been a problem in the past that too many levels of nesting of components lead to one of pgm's biggest hotspots being name rendering (in tablist)

Pablete1234 (approved):

PR #1699, opened 2026-07-16
Address potential memory leaks in modern

calcastor:
1. Remove a duplicate `WorldLoadEvent` call; this happens in `prepareLevel` already. Not likely to be a memory leak though, but still an unnecessary duplicate event call.
2. Use `OnlinePlayerUUIDMapAdapter` in `AfkTracker` to avoid holding direct, stale `Player` references in previous match worlds, which would prevent them from unloading fully.
3. Replace direct `Chunk` references in `ModernBlockPhysicsListener` with a world UUID and chunk coordinate record, and ensure proper cleanup occurs by listening on `WorldUnloadEvent` to remove chunks in unloaded worlds from the tracking set.

### DIFF
--- a/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernBlockPhysicsListener.java
+++ b/platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernBlockPhysicsListener.java
@@ -4,11 +4,14 @@ import com.destroystokyo.paper.event.block.BlockDestroyEvent;
 import com.destroystokyo.paper.event.server.ServerTickEndEvent;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.UUID;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.level.block.state.BlockState;
+import org.bukkit.Bukkit;
 import org.bukkit.Chunk;
+import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.craftbukkit.CraftWorld;
 import org.bukkit.craftbukkit.block.data.CraftBlockData;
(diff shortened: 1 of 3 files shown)


platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/listeners/ModernBlockPhysicsListener.java
```diff
@@ -86,4 +97,17 @@ private void updateShapeWithoutDestroying(Block block) {
...
+  private record LoadingChunk(UUID worldId, int x, int z) {
+    private static LoadingChunk of(Chunk chunk) {
+      return new LoadingChunk(chunk.getWorld().getUID(), chunk.getX(), chunk.getZ());
+    }
+
+    private boolean isLoading() {
+      World world = Bukkit.getWorld(worldId);
+      return world != null
+          && world.isChunkLoaded(x, z)
+          && world.getChunkAt(x, z).getLoadLevel() == Chunk.LoadLevel.BORDER;
```
cswhite2000: Strictly speaking there are two other load levels that could be classified as loading. Mainly `INACCESSIBLE` and `TICKING` with `ENTITY_TICKING` being the only fully loaded load level

Pablete1234:
Not a fan of having to get rid of online player map adapter and replace it with uuid based one. If this is the only way we can fix the leak, we may as well update online player map adapter to internally use uuids for storage but allow put/get operations with Player objects

calcastor:
> Not a fan of having to get rid of online player map adapter and replace it with uuid based one. If this is the only way we can fix the leak, we may as well update online player map adapter to internally use uuids for storage but allow put/get operations with Player objects

8791a7c10f870519cef308d4862b7510a742c517 adds additional event handlers to remove potentially stale entries from the map rather than changing `AfkTracker` to use the UUID-based map. This did not seem to be the primary memory leak in the incident that spurred this PR, so I'm not sure that a change to UUIDs internally is necessary right now.

Pablete1234 (approved):

PR #1700, opened 2026-07-20
Add support for giving ingame class picker items

cswhite2000:
Add support for the class picker to be used in game.

Mapmakers define an item as a picker (class-menu) like so:

```xml
<kit id="spawn-kit">
    <clear/>
    <item slot="4" material="nether star" name="`aSelect Class" class-menu="true"/>
</kit>
```

Then it opens the class picker when right clicked by the player. Same behaviour as `/class` in that is cues the class change for the next respawn.


eliminate123:
IMO I think class-menu="true" would be a better syntax

TheMolkaPL:
`picker` is too generic. `class-picker` or `class-menu` would be better.

Pablete1234 (approved):

PR #1701, opened 2026-07-20
Refactor regions for validation and duplication

cswhite2000:
Add missing parse time validation to:
- `LootableModule`

Refactor validation classes into enum.
Refactor `CompositeRegion` superclass out of `Union` and `Intersect`
Deduplicate normal/origin parsing between `<half>` and `<mirror>`

Fixes:
- `CylindricalRegion.isBlockBounded()` ignored infinite `height`
- `NegativeRegion.getBounds()` threw instead of returning unbounded
- `Complement.toString()` mismatched bracket

Removes dead `Regions` and `CuboidValidation`.


core/src/main/java/tc/oc/pgm/blockdrops/BlockDropsModule.java
```diff
@@ -62,7 +63,8 @@ public BlockDropsModule parse(MapFactory factory, Logger logger, Document doc)
           ImmutableSet.of("block-drops", "blockdrops"),
           ImmutableSet.of("rule"))) {
         Filter filter = filterParser.parseFilterProperty(elRule, "filter");
-        Region region = regionParser.parseRegionProperty(elRule, "region");
+        Region region =
+            regionParser.parseProperty(elRule, "region", RegionValidation.BLOCK_BOUNDED);
```
Pablete1234: Im pretty sure you can just do blockdrops with region="everywhere", no need for block bounded

core/src/main/java/tc/oc/pgm/loot/LootableModule.java
```diff
@@ -80,34 +81,32 @@ public LootableModule parse(MapFactory factory, Logger logger, Document doc)
...
+            loot, fillableFilter, refillTrigger, refillInterval, cleanBeforeRefill);
         factory.getFeatures().addFeature(filler, fillerDefinition);
         fillers.add(fillerDefinition);
       }
 
       for (Element cache : XMLUtils.flattenElements(doc.getRootElement(), "lootables", "cache")) {
-        Region region = factory.getRegions().parseRequiredRegionProperty(cache, "region");
+        Region region = factory
+            .getRegions()
+            .parseRequiredProperty(cache, "region", RegionValidation.BLOCK_BOUNDED);
```
Pablete1234: No, lootable regions do not need to be block bounded, you can create a lootable with region="everywhere" just fine.

I believe an earlier version of the pr was also limiting blockdrops but that seems gone now

core/src/main/java/tc/oc/pgm/regions/CylindricalRegion.java
```diff
@@ -39,7 +39,7 @@ public boolean canGetRandom() {
 
   @Override
   public boolean isBlockBounded() {
-    return !Double.isInfinite(radius);
+    return !Double.isInfinite(radius) && !Double.isInfinite(height);
```
Pablete1234: Given the nature of minecraft, infinite height cuboid is still block bounded (blocks from min y to max y), and this is technically a breaking change.

Also yes, i am aware it is inconsistent with how rectangle and a cuboid with infinite y works (both of those will not be considered block bounded) but also Bounds has separate isFinite and isBlockFinite methods to support a bit wider of a block bounded definition.

Pablete1234 (changes requested):

Pablete1234 (approved):
the PR no longer has the "gameplay affecting" changes:
- `Add missing parse time validation to: LootableModule`
- `CylindricalRegion.isBlockBounded() ignored infinite height`

But even if it now doesn't add any "gameplay" changes its still a nice code cleanup

PR #1702, opened 2026-07-22
Cleanup spawner module, fixes a few bugs

Pablete1234:
Fixes a bunch of minor and less minor issues with spawners:
 - Instead of spawners rolling their own player tracking, reuse the existing player tracker from controlpoints
   - This fixes things like dead players within spawner range still counting towards activating spawners until they moved again, or even if they joined obs so long as they did not move.
 - There were ways for entities to vanish without being "uncounted" from spawned entities, leading to spawners which would never spawn items again.
   - Some of these include making the item go into the void, or being picked up by hoppers or hopper minecarts
   - Coincidentally, also found a bug in player transfer item event (missing event handlers) which caused handling of hopper picking up items to fail, as listeners for player transfer events were registered to the item transfer event, and failed to cast to the right type to get the player for dynamic filters.
 - The region was forced to be specified, even when "everywhere" is a sane default and is used in several maps
 - The filter is a "player-filter", meaning it was tested for every player until one passed. 
   - This is unnecessary most of the times!
   - Many maps are never filtering on players themselves, they filter for match time or some variable that is also match global, since they basically use everywhere regions for the players anyways.
   - Made it so PGM tries to figure out if it really needs to check each player, or if it can get away with just checking the match.
   - The filter default is "always" (as it already was) which can also be optimized away in checking for individual players.
 - Setting min-delay and max-delay to the same number, would make the default delay (10s) be used instead.
   - On the definition, instead of minDelay, maxDelay and delay, we just have min and max. If they're the same, it won't be random.
   - On the parsing side, specifying delay means setting min and max to the same. They're still mutually exclusive.
 - Spawners were checking all the players for the filter even when the spawner was on cooldown.
   - Order of checks in the tick is now more optimized to skip checking filters whenever possible and short-circuit earlier.
 - Pgm didn't check that the spawn region supported random points at parse time. It would still break if it ever tried to spawn.

### DIFF
--- a/core/src/main/java/tc/oc/pgm/action/actions/TeleportAction.java
+++ b/core/src/main/java/tc/oc/pgm/action/actions/TeleportAction.java
@@ -33,9 +33,8 @@ public class TeleportAction extends AbstractAction<MatchPlayer> {
 
   @Override
   public void trigger(MatchPlayer player) {
-    var location = this.region
-        .map(r -> r.getRandom(player.getMatch()).toLocation(player.getWorld()))
-        .orElseGet(player::getLocation);
+    var location =
+        this.region.map(r -> r.getRandomLoc(player.getMatch())).orElseGet(player::getLocation);
 
     xformula.ifPresent(f -> location.setX(f.applyAsDouble(player)));
     yformula.ifPresent(f -> location.setY(f.applyAsDouble(player)));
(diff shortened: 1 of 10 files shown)


cswhite2000 (approved):

arcadeboss:
Spawner max-entities issue resolved now with subsequent updates.

PR #1706, opened 2026-07-23
Mannequins

eliminate123:
**TLDR:** This PR introduces new features allowing for extensive mannequin customization including customizable AI behavior.

Everything in the PR is for modern only. Mannequins were introduced in 1.21.9, and unlike a FakePlayer mannequins aren't a packet illusion. Mannequins are a LivingEntity, but vanilla AI behavior is limited to just Mobs. When a mannequin has a defined ```<behavior>```, an off world NMS Zombie is created (I refer to it as "ghost" in the code). This ghost has one job, use its AI behavior as a trade off for mannequins not having any. The ghost is used for pathfinding, attacking players, looking at players etc. 

This is a large PR but it can be broken into 4 sections. I think its easiest for everyone involved if this could reviewed section by section. I also created markdown tables of each section with descriptions of all the new xml additions. The sections go as follows:
1) ```<mannequins> </mannequins>``` module
2) ```<waypoints> </waypoints>``` module
3) ```<spawn-mannequin/>``` + ```<modify-mannequin/>``` actions
4) ```<behaviors> </behaviors>``` module **(most in depth section)**

### **SECTION 1:** Mannequins Module

<details>
<summary>Mannequin Module Documentation</summary>

#### Mannequins Element

| Element | Description |
|---|---|
| `<mannequins> </mannequins>` | Node containing the mannequins definitions. |

| Sub-elements ||
|---|---|
| `<mannequin> </mannequin>` | An individual mannequin. |

#### Mannequin Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `id` | <span className =" badge badge--danger " > Required </span><br /> Unique identifier used to reference this mannequin from other places in the XML. | <span className =" badge badge--primary " > String </span> |  |
| `name` | Sets the primary name tag text displayed above the mannequin's head. | <span className =" badge badge--primary " > Formatted Text </span> |  |
| `description` | Sets an additional secondary test line displayed directly beneath the mannequin's main name tag. | <span className =" badge badge--primary " > Formatted Text </span> |  |
| `silent` | Prevents the mannequin from producing sounds. | <span className =" badge badge--primary " > true/false </span> | false |
| `invulnerable` | Prevents the mannequin from being killed. | <span className =" badge badge--primary " > true/false </span> | false |
| `glowing` | Applies a white glowing outline effect around the mannequin. | <span className =" badge badge--primary " > true/false </span> | false |
| `health` | Health values in ½ heart increments. | <span className =" badge badge--primary " > Number </span> | 20.0 |
| `immovable` | Prevents the mannequin from being moved/pushed. | <span className =" badge badge--primary " > true/false </span> | false |
| `main-hand` | Sets the mannequins main-hand used when attacking players. | <br /> ` left ` <br /> ` right ` <br /> | right |
| `gravity` | Toggles whether gravity affects the mannequin's falling physics. | <span className =" badge badge--primary " > true/false </span> | true |
| `physics` | Toggles environmental physics, such as block collision or pushing, for the mannequin. | <span className =" badge badge--primary " > true/false </span> | true |
| `on-fire` | Sets the mannequin to have visual fire, which does not deal fire tick damage. | <span className =" badge badge--primary " > true/false </span> | false |
| `click-action` | Run the specified action upon right-clicking the mannequin. | Action ID |  |
| `waypoint` | Displays a pre-defined waypoint marker at the mannequin's location on the user's locator bar. | Waypoint ID |  |
| `behavior` | Applies a pre-defined AI behavior pattern to the mannequin. | Behavior ID |  |

#### Profile Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `uuid` | <span className =" badge badge--danger " > Required </span>A player's UUID used by the mannequin.   <br />**Note:** `#player#` can be used to automatically use the UUID of the player triggering either a player-scoped `<spawn-mannequin/>` or `<modify-mannequin/>` action. | <span className =" badge badge--primary " > String </span> |  |
| `remove-layers` | Hides the specified skin layers for the mannequin. | `all`, `arms`, `legs`, `cape`, `jacket`, `left_sleeve`, `right_sleeve`, `left_pants_leg`, `right_pants_leg`, or `hat` |  |
| `pose` | Change the pose used by the mannequin. | `standing`, `crouching`, `swimming`, `crawling`, or `sleeping` | standing |

### Profile Sub-elements

| Sub-elements | Description | Value |
|---|---|---|
| `<skin> </skin>` | <span className =" badge badge--danger " > Required </span>A player's skin texture used by the mannequin.   <br />**Note:** `#player#` can be used to automatically use the skin texture of the player triggering either a player-scoped `<spawn-mannequin/>` or `<modify-mannequin/>` action. | <span className =" badge badge--primary " > Base64 </span>  #|

</details>

Here is what an example XML looks like:
```xml
<mannequins>
    <mannequin id="example-mannequin" name="`cCubist" description="Im glowing!" health="20" glowing="true" click-action="some-action-id" waypoint="some-waypoint-id" behavior="some-behavior-id">
        <profile uuid="3fbec7dd-0a5f-40bf-9d11-885a54507112" remove-layers="hat" pose="crouching">
            <skin>eyJ0aW1lc3RhbXAiOjE0NDY0MDgwOTExNzQsInByb2ZpbGVJZCI6IjNmYmVjN2RkMGE1ZjQwYmY5ZDExODg1YTU0NTA3MTEyIiwicHJvZmlsZU5hbWUiOiJDdWJpc3QiLCJ0ZXh0dXJlcyI6eyJTS0lOIjp7InVybCI6Imh0dHA6Ly90ZXh0dXJlcy5taW5lY3JhZnQubmV0L3RleHR1cmUvOTRlZGExMjg4NWIzYmE2ZGY2ODMyZGZkMTIzNGEyNjc5MmQwNDI2ZDkyMDM2ZWVlYzc1M2ZiZmM2NmRiIn19fQ==</skin> 
        </profile>
    </mannequin>
</mannequins>
```
**NOTE:** I plan on adding kit support once mob-kits is finalized and merged. Also worth mentioning, in vanilla mannequins can never drop items when killed.

### **SECTION 2:** Waypoints Module

<details>
<summary>Waypoint Module Documentation</summary>

#### Waypoints Element

| Element | Description |
|---|---|
| `<waypoints> </waypoints>` | Node containing the waypoints definitions. |

| Sub-elements ||
|---|---|
| `<waypoint> </waypoint>` | An individual waypoint. |

#### Waypoint Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `id` | <span className =" badge badge--danger " > Required </span><br /> Unique identifier used to reference this kit from other places in the XML. | <span className =" badge badge--primary " > String </span> |  |
| `style` | Icon style displayed on the user's locator bar.  | `default` or `bowtie` | `default` |
| `color` | <span className=" badge badge--danger " > Required </span><br /> Color of the waypoint icon. | <span className =" badge badge--primary " > Hex Color </span> |  |
| `transmit-range` | Sets the maximum distance at which the waypoint is displayed on a user's locator bar. | <span className =" badge badge--primary " > Number </span> | 256 |

</details>

Here is what an example XML looks like:
```xml
<waypoints>
    <waypoint id="example-waypoint" style="bowtie" color="1D1D21" transmit-range="80"/>
</waypoints>
```

**NOTE:** I originally planned for on making a ```<waypoint/>``` player-scoped action for applying a waypoint by id, or removing it. Entities can only have 1 waypoint set on themselves. I temporarily removed the player action for now. I plan on finishing it.

### **SECTION 3:** Spawn-mannequin and Modify-mannequin Actions

<details>
<summary>Spawn-mannequin Documentation</summary>

### Spawn-Mannequin Attributes

| Attribute | Description | Value |
|---|---|---|
| `mannequin` | <span className =" badge badge--danger " > Required </span><br /> The mannequin to spawn. | Mannequin ID |
| `x` | <span className =" badge badge--danger " > Required </span><br /> The X coordinate, measured in east-west. | <span className="badge badge--primary">Expression</span> |
| `y` | <span className =" badge badge--danger " > Required </span><br /> The Y coordinate, measured in altitude. | <span className="badge badge--primary">Expression</span> |
| `z` | <span className =" badge badge--danger " > Required </span><br /> The Z coordinate, measured in north-south. | <span className="badge badge--primary">Expression</span> |
| `yaw` | The horizontal angle the mannequin looks to. | <span className="badge badge--primary">Expression</span> |
| `pitch` | The vertical angle the mannequin looks to. | <span className="badge badge--primary">Expression</span> |

</details>

Here is what an example XML looks like:
```xml
<action id="spawn-example" scope="match">
    <spawn-mannequin mannequin="example-mannequin" x="10" y="10" z="10" yaw="90"/>
</action>
```

<details>
<summary>Modify-mannequin Documentation</summary>

**NOTE:** Modify-mannequin supports all regular ```<mannequin>``` attributes as well as some new ones listed below.

### Modify-Mannequin Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `mannequin` | <span className =" badge badge--danger " > Required </span><br /> The spawned in mannequin to modify. | Mannequin ID |
| `despawn` | Despawns the selected mannequin. | <span className =" badge badge--primary " > true/false </span> | false |
| `name` | Change the mannequin's current name. | Formatted Text |  |
| `description` | Change the mannequin's current description. | Formatted Text |  |
| `hide-description` | Removes the mannequin's current description. | <span className =" badge badge--primary " > true/false </span> | false |
| `hide-titles` | Removes the mannequin's current name and description. | <span className =" badge badge--primary " > true/false </span> | false |
| `remove-click-action` | Removes the mannequin's current click-action. | <span className =" badge badge--primary " > true/false </span> | false |
| `remove-waypoint` | Removes the mannequin's current waypoint. | <span className =" badge badge--primary " > true/false </span> | false |
| `remove-behavior` | Removes the mannequin's current behavior. | <span className =" badge badge--primary " > true/false </span> | false |

### Teleport Sub-elements

| Sub-elements | Description | Value |
|---|---|---|
| `x` | <span className =" badge badge--danger " > Required </span><br /> The X coordinate, measured in east-west. | <span className="badge badge--primary">Expression</span> |
| `y` | <span className =" badge badge--danger " > Required </span><br /> The Y coordinate, measured in altitude. | <span className="badge badge--primary">Expression</span> |
| `z` | <span className =" badge badge--danger " > Required </span><br /> The Z coordinate, measured in north-south. | <span className="badge badge--primary">Expression</span> |
| `yaw` | The horizontal angle the mannequin looks to. | <span className="badge badge--primary">Expression</span> |
| `pitch` | The vertical angle the mannequin looks to. | <span className="badge badge--primary">Expression</span> |

</details>

Here is what an example xml looks like:
```xml
<action id="modify-example" scope="player">
    <modify-mannequin mannequin="example-mannequin" hide-titles="true" invulnerable="true">
        <profile uuid="#player#" pose="sleeping">
            <skin>#player#</skin>
        </profile>
        <teleport x="player.x" y="player.y" z="player.z"/>
    </modify-mannequin>
</action>
```

### **SECTION 4:** Behaviors Module

<details>
<summary>Behaviors Documentation</summary>

#### Behaviors Element

| Element | Description |
|---|---|
| `<behaviors> </behaviors>` | Node containing the behaviors definitions. |

| Sub-elements ||
|---|---|
| `<behavior> </behavior>` | An individual mannequin behavior. |

### Behavior Attributes
| Attribute | Description | Value | Default |
|---|---|---|---|
| `avoid-danger` | The mannequin will try to avoid taking contact damage from blocks while pathfinding. | <span className =" badge badge--primary " > true/false </span> | true |
| `open-doors` | Allows a mannequin to open wooden doors when trying to attack a player, or walking along a defined path. | <span className =" badge badge--primary " > true/false </span> | true |

### Behavior Elements
| Type | Description |
|---|---|
| `<combat/>` | Controls whether the mannequin is passive, neutral, or hostile, and how it attacks. |
| `<home/>` | Defines a mannequin's home region used by multiple attributes. |
| `<evade/>` | Controls when a mannequin should avoid a player. |
| `<look-at/>` | Defines what the mannequin should look at. |
| `<pathing/>` | Creates a custom path for a mannequin to walk by defining nodes along the way. |

### Combat Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `hostility` | Sets a mannequin's hostility. | `passive`, `neutral`, or `hostile` | passive |
| `aggro-duration` | The duration a mannequin remains aggressive towards its current target. <br />**Note:** If unset, the mannequin stays aggressive until its target dies, leaves the match, or exits the mannequin's home region or leash range. | Time Period | Indefinite |
| `attack-range` | Modifies the mannequin's attack range. | <span className="badge badge--primary">Number</span> | 3.0 |
| `attack-interval` | Amount of seconds before the mannequin can attack again. | Time Period | 1s |

### Home Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `region` | The region a mannequin will use as its home. <br />**Note:** `<home/>` cannot be used in combination with `<pathing/>`. | Region |  |
| `wander` | When enabled the mannequin will walk to random points within its home region. Once it reaches its destination it will idle for a random amount of time before repeating the process. <br />**Note:** If the mannequin is not `passive`, it will resume wandering once no longer aggrod onto a player. | <span className="badge badge--primary"> true/false </span> | false |
| `stray-distance` | The distance a mannequin can be from the center of its home `region` before it attempts to walk back. | <span className="badge badge--primary">Number</span> <br /> or `outside` |  |
| `return-after` | How long a mannequin must be outside of its `stray-distance` before walking back. | Time Period |  |

### Evade Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `panic` | When the mannequin takes damage it will run away in a random direction. <br />**Note:** Only supported for `passive` mannequins. | <span className="badge badge--primary"> true/false </span> | false |
| `panic-duration` | Duration a mannequin can be panicking for. <br />**Note:** Only supported for `passive` mannequins. | Time Period | 5s |
| `avoid-range` | Distance a mannequin will try to keep between itself and a player. <br />**Note:** Only supported for `passive` mannequins. | <span className =" badge badge--primary " > Number </span> |  |
| `avoid-filter` | A filter used to specify which players a mannequin will try to avoid. <br />**Note:** Only supported for `passive` mannequins. | Filter |  |

### Look-at Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `target` | What the mannequin should be looking at. | <span className="badge badge--primary">X,Y,Z</span> or  `player` | player |
| `rotation` | `body` rotates the mannequin's whole body. <br /> `head` limits the mannequin to only turning its head.  | `head` or `body` | `body` |
| `visibility-range` | Amount of blocks a player can be from the mannequin while still being tracked. | <span className =" badge badge--primary " > Number </span> | 6.0 |
| `rest-yaw` | The mannequin's yaw to return to when no players are within its `visibility-range`. | <span className="badge badge--primary"> Expression </span> |  |
| `rest-pitch` | The mannequin's pitch to return to when no players are within its `visibility-range`. | <span className="badge badge--primary"> Expression </span> |  |

### Pathing Attributes

| Attribute | Description | Value | Default |
|---|---|---|---|
| `start` | The start of the mannequin's defined path. This is **not** a goal. <br />**Note:** If not defined, the starting position will be assumed to be the position in the mannequin's initial `<spawn-mannequin/>` action. | <span className="badge badge--primary">X,Y,Z</span> |  |
| `loop` | If enabled, when the mannequin reaches its final `<goal/>` it will walk back to its starting point and start over. <br />**Note:** Looping does not trigger `completion-action`s on its path back. | <span className="badge badge--primary"> true/false </span> | false |
| `leash-distance` | How far the mannequin can chase a target from the point where it first aggroed before giving up and resuming its path from its last reached <goal/>. <br />**Note:** Acts as a dynamic `<home/>` region. A circle of radius `leash-distance` is created around the mannequin's position each time it acquires a target, and is cleared once the target is dropped and pathing resumes. If not defined, the mannequin does not stop pathing when aggrod. | <span className="badge badge--primary"> Number </span> |  |

### Pathing Sub-elements
| Attribute | Description | Value |
|---|---|---|
| `<goal/>` | <span className =" badge badge--danger " > Required </span><br /> An individual node for a mannequin to walk to. <br />**Note:** `<pathing/>` requires at least 1 goal to be defined. Goals are followed in sequential order following whats defined in the XML, with `start` as its starting point. | Goal Attributes |
| `<if-stuck/>` | Conditions for what happens if a mannequin gets stuck while walking between goals. | If-stuck Attributes |

### Goal Attributes
| Attribute | Description | Value | Default |
|---|---|---|---|
| `destination` | <span className =" badge badge--danger " > Required </span><br /> A single block located at `X,Y,Z`. <br />**Note:** A mannequin will always walk to the center of the block.  | <span className="badge badge--primary"> X,Y,Z </span> |  |
| `idle` | Amount of time the mannequin will wait at a goal before trying to move to the next. | Time Period | 0 |
| `completion-action` | The action to trigger upon successfully reaching this goal. <br />**Note:** The starting position does not receive a completion-action, even when loop is set to true. | <span className="badge badge--primary"> Action </span> |  |
| `goal-radius` | The radius around the `destination` within which the mannequin is considered to have reached this goal. <br />**Note:** The minimum goal-radius is `0.5` which is the same as standing directly in the center of the block. | <span className="badge badge--primary"> Number </span> | 0.5 |

### If-stuck Attributes
| Attribute | Description | Value | Default |
|---|---|---|---|
| `after` | How long the mannequin must go without making progress before it's considered stuck. | Time Period | 5s |
| `stuck-action` | Action triggered when the mannequin becomes stuck. <br />**Note:** Fires before any other resolution attributes. | <span className="badge badge--primary"> Action </span> |  |
| `move-to` | Where the mannequin should relocate to when stuck. <br />**Note:** Each goal is automatically assigned an index following the order defined in the XML. The starting position is always index 0. | `start`, `end`, `previous`, `next`, or goal index |  |
| `method` | Defines the method `move-to` will use to relocate the mannequin. <br />**Note:** `method="teleport"` is required when using `move-to="next"`. | `walk` or `teleport` |  |
| `give-up` | When stuck the mannequin will completely stop trying to path find and stay where it is. | <span className="badge badge--primary"> true/false </span> | false |
| `despawn` | When stuck the mannequin will be despawned. | <span className="badge badge--primary"> true/false </span> | false |

</details>

Here is what an example xml looks like:
```xml
<behaviors>
    <behavior id="passive-behavior" avoid-danger="true" open-doors="true">
        <combat hostility="passive"/>
        <home region="base" wander="true" stray-distance="outside" return-after="5s"/> 
        <evade panic="true" panic-duration="10s"/> 
    </behavior>
    <behavior id="neutral-behavior" avoid-danger="true" open-doors="true">
        <combat hostility="neutral" attack-range="3" attack-interval="1s"/>
        <pathing start="10.5,10,10.5" loop="true" leash-distance="8">
            <goal destination="15.5,10,10.5"/>
            <goal destination="15.5,10,15.5"/>
            <goal destination="20.5,10,15.5"/>
            <goal destination="20.5,10,20.5" completion-action="some-action"/>
            <if-stuck after="5s" move-to="next" method="teleport"/> 
        </pathing>
    </behavior>
</behaviors>
```

### **Potential Future Additions**

Heres a list of some things I think would be cool to add in the future.

- [ ] Kit support (previously mentioned)
- [ ] waypoint set=""/waypoint remove="" (previously mentioned)
- [ ] Ranged attack support. Vanilla has bow attack/crossbow attack mob goals.
- [ ] Customizable walking speed (for panic, wander, pathing etc)


eliminate123:
Also worth noting mannequins cant punch on 1.21.11. On 1.26.1 they got the ability to. My localhost is 1.21.11, so I havent gotten the chance to double check that the swing animation plays properly

Pablete1234 (changes requested):
the mobs being allowed on spawners gave PGM a syntax for spawning entities, that should work for both legacy and modern entities (given you're on the right version for them).

Mannequins should extend that feature instead of trying to make a completely separate system.

eliminate123:
> the mobs being allowed on spawners gave PGM a syntax for spawning entities, that should work for both legacy and modern entities (given you're on the right version for them).
> 
> Mannequins should extend that feature instead of trying to make a completely separate system.

I thought about about using spawners but at what point do spawners lose their original purpose. Also with mannequins the whole point is that they get an id so they can be manipulated later. Spawners seem to be fire and forget.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/waypoint/WaypointMatchModule.java
```diff
@@ -81,14 +83,35 @@ public void onFlagPickup(FlagStateChangeEvent event) {
...
     if (attr == null) return;
+    var nmsPlayer = ((CraftPlayer) player.getBukkit()).getHandle();
+    waypointManager.untrackWaypoint(nmsPlayer);
     if (color != null) {
-      var nmsPlayer = ((CraftPlayer) player.getBukkit()).getHandle();
-      // Ensures they're newly registered so the color updates
-      waypointManager.untrackWaypoint(nmsPlayer);
       nmsPlayer.waypointIcon().color = Optional.of(color.asRGB());
       attr.setBaseValue(256);
+      waypointManager.trackWaypoint(nmsPlayer);
```
Pablete1234, line 92: Why?
the untracking and re-tracking is only needed to update the color, otherwise just setting transmit range to 0 already makes it not visible.

eliminate123: I believe this line in particular is a line of dead code from when I was working on a waypoint action. I got the action set up at the start but couldnt get it working for players. Since it wasnt directly related to mannequins i decided to work on it at the very end. But i ultimately removed the code for the action at the end so i could fix it at a later date. 

I could be wrong, Ive been working on this for a while

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/waypoint/WaypointMatchModule.java
```diff
@@ -81,14 +83,35 @@ public void onFlagPickup(FlagStateChangeEvent event) {
...
+      nms.waypointIcon().style = def.getStyle();
+      nms.waypointIcon().color = Optional.of(def.getColor().asRGB());
+      attr.setBaseValue(def.getTransmitRange() != null ? def.getTransmitRange() : 256f);
+      waypointManager.trackWaypoint(nms);
+    } else {
+      nms.waypointIcon().style = WaypointStyleAssets.DEFAULT;
+      nms.waypointIcon().color = Optional.empty();
+      attr.setBaseValue(0);
+    }
+  }
```
Pablete1234, line 117: i don't believe changes to the waypoint module to be necessary at all for this, living entities can just be given the attribute and will show up without needing to pass thru this module, this module exists to track objectives on the waypoint system, and since those objectives are sometimes players themselves (ie: carried flag), we need to toggle them too. 
Other entities are free to show up without needing to pass thru this

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/action/actions/ModifyMannequinAction.java
```diff
@@ -0,0 +1,196 @@
...
+  private final @Nullable FeatureReference<WaypointDefinition> waypoint;
+  private final boolean removeWaypoint;
+  private final @Nullable FeatureReference<BehaviorDefinition> behavior;
+  private final boolean removeBehavior;
+  private final @Nullable Formula<B> xformula;
+  private final @Nullable Formula<B> yformula;
+  private final @Nullable Formula<B> zformula;
+  private final Optional<Formula<B>> yawFormula;
+  private final Optional<Formula<B>> pitchFormula;
+  private final @Nullable Boolean despawn;
```
Pablete1234, line 54: This is not a viable way to define an action. These need to be broken down into simpler actions.

And honestly, half of these already have a way of working with the whole mob properties system that was created as part of mobs in spawners, you should reuse that.

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/action/actions/SpawnMannequinAction.java
```diff
@@ -0,0 +1,72 @@
...
+import tc.oc.pgm.util.skin.Skin;
+
+public class SpawnMannequinAction<B extends Filterable<?>> extends AbstractAction<B> {
+
+  private final FeatureReference<MannequinDefinition> mannequinRef;
+  private final Formula<B> xformula;
+  private final Formula<B> yformula;
+  private final Formula<B> zformula;
+  private final Optional<Formula<B>> yawFormula;
+  private final Optional<Formula<B>> pitchFormula;
```
Pablete1234, line 22: We have a way to spawn arbitrary mobs via spawners, i think its ok to have a new action that will also just spawn a mob so it can be a one-off trigger instead of requiring a spawner, but i see no reason as to why it needs to be mannequin specific and not just spawn a mob in general, using the same logic used to define mobs in spawners atm

Pablete1234:
> I thought about about using spawners but at what point do spawners lose their original purpose. Also with mannequins the whole point is that they get an id so they can be manipulated later. Spawners seem to be fire and forget.

I'm not talking about using spawners, i'm talking about using the *xml syntax* and all the logic that spawners added to include mob spawning, all of that can be reused in a way that both allows spawners to spawn mannequins, as well as allowing an action to spawn any kind of mob, that a spawner could (including a mannequin)

If you want to have "tagged entities" so that you can run actions on them at a later point in time (ie: like you're saying mannequins have an id) that's great, but i see no reason why it can't also be made part of the whole mob syntax and support working for other mobs too

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/action/actions/ModifyMannequinAction.java
eliminate123: Can you give an example of what these simpler actions would look like. Just so were both on the same page.

core/src/main/java/tc/oc/pgm/action/actions/ActionParserProvider.java
TTtie, line 1: This platform module doesn't have a SportPaper implementation and will crash map loading there if unimplemented

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/behavior/combat/CombatBehavior.java
TTtie, line 1: Should be a record (along with other *Behavior classes)

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/behavior/combat/CombatBehavior.java
```diff
@@ -0,0 +1,38 @@
+package tc.oc.pgm.platform.modern.modules.behavior.combat;
+
+import java.time.Duration;
+import javax.annotation.Nullable;
```
TTtie: Prefer JSpecify annotations, along with JetBrains annotations for what can't be expressed using JSpecify.


```suggestion
import org.jspecify.annotations.Nullable;
```

TTtie (changes requested):

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/behavior/combat/CombatBehavior.java
eliminate123: Fixed

platform/platform-modern/src/main/java/tc/oc/pgm/platform/modern/modules/behavior/combat/CombatBehavior.java
eliminate123: I believe ive fixed most of this, ive been fixing them as I go. I might have 1-2 left.

eliminate123:
I was away for a week and unable to finish what I was working on before I left. Basically I started a new behavior called `<tempt/>`

Tempt has these flavors:
```xml
<!-- Follow the first player to enter its 10 block range, supports following a player outside of a mannequin's home region -->
<tempt follow="true" range="10" leave-home="true"/>

<!-- Follow a player holding a diamond -->
<tempt>
    <holding>diamond</holding>
</tempt>
```

Tempt has a lower priority over other other behaviors. 

The other main thing i worked on was another attribute for `<behavior>` called enter-vehicles="t/f". It defaults to false and prevents mannequins from getting inside of boats/minecarts.


Im still working on addressing other concerns.

eliminate123:
Ive gone ahead and made some changes mostly to how looping was handled for `<pathing>`. I also added an attack-filter="..." for the `<combat/>` behavior child. This was on my todo list. 

The main changes were moving loop="t/f" out of `<pathing loop="t/f"> </pathing>`. Now loop is its own child. One of my biggest gripes before was that each goal gets a completion-action but not for the starting position. SO when loop="true" and the mannequin has successfully reached its last goal it was impossible to fire a completion-action when it returned to its starting point. This is a little visual of what it used to be like.

image

By having loop be its own child I also had the ability to add some other optional attributes. This is what the loop child now can look like:

```xml
<behavior id="...">
    <combat hostility="neutral" attack-filter="red-team"/>
    <pathing start="10.5,10,10.5" leash-distance="8">
        <goal destination="15.5,10,10.5" completion-action="some-action-1"/>
        <goal destination="15.5,10,15.5" completion-action="some-action-2"/>
        <loop random="true" times="5" completion-action="reach-start-action"/>
    </pathing>
</behavior>
```

### Loop Attributes
| Attribute | Description | Value | Default |
|---|---|---|---|
| `random` | Shuffle the order of defined goals. Once all nodes have been reached, reshuffle and continue.  | true/false | false |
| `reverse` | Instead of walking from the last goal directly to the starting node, walk the order backwards. Where 0 is the starting node and 2 is the last goal: 0 -> 1 -> 2 -> 1 -> 0 <br />**Note:** Cannot be combined with `random`  | true/false | false |
| `times` | The amount of times to loop. | Expression |  |
| `completion-action` | The action to trigger upon successfully finishing a loop. **Note:** The starting position does not receive a completion-action, even when loop is set to true. | Action |  |

loop doesn't require any of these attributes and can just be used as `<loop/>`. There is a small bug with how shuffle works atm but I'm going to come back and fix it later.

eliminate123:
> If you want to have "tagged entities" so that you can run actions on them at a later point in time (ie: like you're saying mannequins have an id) that's great, but i see no reason why it can't also be made part of the whole mob syntax and support working for other mobs too

I've gone ahead and made a separate branch for the mob tagging stuff (will rebase if approved later). This is change is still a wip draft but I just want to get your onions before I continue. I made it so LivingEntity(s) can be given an ID. I havent incorporated this into spawners yet as I wanted to get your opinion on this first. I updated the mannequin referencing code to use the new mob tagging system. 

This branch has two features:
1) New mob tagging
2) New animation action

The new animation action is still heavily WIP but it looks like this `<animation entity-id="..." play="..."/>`
where it uses the new mob tagging system to play things like swing animations, sheep eating grass, allay dancing etc on demand. All the animations are stored in an enum. The goal for animations is creating something thats scalable, however I'm only really focused on shipping with mannequin swing animations



https://github.com/Javggg/PGM/compare/mannequins...animation

PR #1708, opened 2026-07-24
ci: update actions in the pipeline

TTtie:
Updates actions in the CI pipeline to their latest versions to get rid of the Node.js v20 EOL warning. A couple of things to note:
- The .jar artifact is no longer zipped, allowing people to download PGM builds with less friction
- ~~The v6 of the `setup-gradle` action has changed their caching mechanism to a [proprietary license](https://github.com/gradle/actions/releases/tag/v6.0.0). Currently, this PR opts into the "basic" caching mechanism (released later on), which is open source, but well… basic and isn't as granular as the (now) proprietary component. Although – per their words – the proprietary caching mechanism is free to use with public repositories (and thus probably OK to use), I'm keeping this as a draft until the way to go is decided.~~
  The `setup-gradle` action has been pinned to v5.0.2, which supports Node.js v24 and is the last version before the licensing changes. This should give us some time, during which PGM can use better caching (compared to v6 basic caching) under the MIT license (thus doesn't require agreeing to the commercial license and PGM having to trust a 2 MB obfuscated JavaScript blob) until it goes EOL in 2028.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
     - id: checkout
       name: Checkout
-      uses: actions/checkout@v6
+      uses: actions/checkout@v7
       with:
         fetch-depth: 0
     - id: setup-java
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,11 +23,11 @@ jobs:
       with:
         java-version: 21
         distribution: temurin
-    # Configures gradle with caching
+    # Configures gradle with caching. Intentionally pinned to v5.0.2 instead of v6
+    # to benefit from better caching without having to agree to a commercial license,
+    # along with having to trust a 2 MB obfuscated JS blob.
     - name: Setup Gradle
-      uses: gradle/actions/setup-gradle@v6
-      with:
-        cache-provider: basic
+      uses: gradle/actions/setup-gradle@0723195856401067f7a2779048b490ace7a47d7c # v5.0.2
     # Run "gradlew publish" for origin/dev and "gradlew build" for PRs or elsewhere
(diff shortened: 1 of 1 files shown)


cswhite2000 (approved):

Pablete1234:
question, didn't just using the basic caching also solve this without needing to pin to an older version?

TTtie:
Yes, but I believe it introduces a new problem. It seems like the cache data for the basic cache[^2] is saved only once[^1] until the gradle files change. I believe that this might be problematic for build times (although, I don't have any data to back this up with), especially with respect to modern, as paperweight will always have to download the latest build and do its thing whenever a new Paper build drops after the point the data is first cached.

[^1]: [`sources/src/cache-service-basic.ts`, line 65 to 78 in setup-gradle@v6.2.0](https://github.com/gradle/actions/blob/3f131e8634966bd73d06cc69884922b02e6faf92/sources/src/cache-service-basic.ts#L65-L78)
[^2]: basic caching is, sans some additional configuration options, equivalent to not using the `setup-gradle` action at all and just using `actions/setup-java` caching options (compare [`sources/src/cache-service-basic.ts` in setup-gradle@v6.2.0](https://github.com/gradle/actions/blob/3f131e8634966bd73d06cc69884922b02e6faf92/sources/src/cache-service-basic.ts) vs. [`src/cache.ts` in setup-java@v5.6.0](https://github.com/actions/setup-java/blob/03ad4de0992f5dab5e18fcb136590ce7c4a0ac95/src/cache.ts#L73-L96)); whereas the cache mechanism that [v5 line of the action uses](https://github.com/gradle/actions/blob/0723195856401067f7a2779048b490ace7a47d7c/docs/setup-gradle.md#how-gradle-user-home-caching-works) is nearly identical to [v6 "enhanced" caching](https://github.com/gradle/actions/blob/3f131e8634966bd73d06cc69884922b02e6faf92/docs/setup-gradle.md#enhanced-caching).

Pablete1234 (approved):

PR #1710, opened 2026-07-25
Shop xml changes and bump proto

cswhite2000:
Bumps the proto to 1.5.2

- register shop `<category>` as features
- Allow referencing categories
- ids for categories must now be unique though (why it has a proto bump)
- Fix armor stand shop keepers
- Add support for some mob properties on shopkeepers

Example:
```xml
<shops>
  <!-- e.g. from an include -->
  <category id="ammo" name="Ammo" material="arrow" currency="gold nugget" payment-color="aqua">
    <item price="1" material="arrow" amount="16"/>
    <item price="3" currency="emerald" material="tnt"/>
  </category>

  <shop id="main">
    <category id="ammo"/>
  </shop>
</shops>
```

### DIFF
--- a/core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
+++ b/core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
@@ -3,6 +3,8 @@ package tc.oc.pgm.shops;
 import static tc.oc.pgm.util.bukkit.BukkitUtils.colorize;
 import static tc.oc.pgm.util.nms.NMSHacks.NMS_HACKS;
 
+import java.util.List;
+import java.util.function.Consumer;
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
 import org.bukkit.entity.Entity;
@@ -10,6 +12,7 @@ import org.bukkit.entity.LivingEntity;
 import org.bukkit.metadata.FixedMetadataValue;
 import org.jetbrains.annotations.Nullable;
 import tc.oc.pgm.api.PGM;
+import tc.oc.pgm.api.feature.FeatureReference;
 import tc.oc.pgm.api.match.Match;
 import tc.oc.pgm.points.PointProvider;
 import tc.oc.pgm.util.bukkit.MetadataUtils;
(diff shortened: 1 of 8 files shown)


core/src/main/java/tc/oc/pgm/shops/ShopKeeper.java
```diff
@@ -21,18 +24,24 @@ public class ShopKeeper {
...
-  private final Shop shop;
+  private final List<Consumer<Entity>> properties;
+  private final FeatureReference<Shop> shop;
 
   public ShopKeeper(
-      @Nullable String name, PointProvider location, Class<? extends Entity> type, Shop shop) {
+      @Nullable String name,
+      PointProvider location,
+      Class<? extends Entity> type,
+      List<Consumer<Entity>> properties,
```
Pablete1234: Should be SpawnableMob

cswhite2000: SpawnableMob is a bit spawner scoped. Made a change to use SpawnableEntity instead.

eliminate123:
These are nice additions. If proto is going to be bumped for shops maybe some of this could be addressed to save the need for future proto bumps/PRs. 2 birds 1 stone.

- Would be cool if the category indicator (green glass pane) color could be customized to other colors. Some sort of <shop> attribute maybe divider-color="" and prevent gray being used. Green would stay the default. 

- Each category supports 21 items. In an example like this where a single variable is used as a filter for multiple items:
```xml
<category id="" material="">
    <item material="" filter="some_var=0" currency="" price=""/>
    <item material="" filter="some_var=1" currency="" price=""/>
    <item material="" filter="some_var=2" currency="" price=""/>
    <item material="" filter="some_var=3" currency="" price=""/>
</category>
```
Even though only one item will appear at a time, this still takes up 4/21 slots. It would be nice if shops only counted the items currently displayed towards the x/21 slots. I talked with pablo about this months ago. This is one of my biggest gripes with shops, its very limiting and annoying to min-max.

In scenarios where a filter ISNT a variable ie `<item ... filter="match-running"/>` maybe item would get an optional slot="0-20" attribute, where its first come first serve for situations like this. Just my 2 cents. Being able to define an item's slot would be extremely useful as it would remove the need to use light gray glass panes as padding like in this example:
image

- Shops dont support custom player head parsing. I also talked to pablo about this previously and he said that custom player heads would need their own proto bump. One of the biggest quirks of `<head></head>` is that name="..." really behaves as username="..." where special symbols, color codes etc cannot be used. IMO if shops are already getting a protobump this seems like the best time to address this.

Obviously changing how `<item filter=".."/>` is handled/refactoring head parsing is more extensive than whats included in this PR. But maybe we could hold this PR open for a bit to see if we could address some of this?

TheRealPear:
It would be nice to allow referencing items in the same manner as one reference categories. [The bedwars include currently has a "quick buy" category](https://github.com/OvercastCommunity/PublicMaps/blob/e175d72dff47c239414e1c94faf8844301d26d71/includes/4-team-bedwars.xml#L1224-L1263) that is meant to be the first thing player see, since it includes items from the other categories that I expect to be frequently purchased.

There should also be a filter that controls when to spawn or despawn a shopkeeper, so they're more like the vanilla wandering traders.

PR #1711, opened 2026-07-26
Move listeners to better match ownership

cswhite2000:
Move listeners around to better match ownership

Remove FormattingListener, reduce PGMListener.


core/src/main/java/tc/oc/pgm/gamerules/GameRulesMatchModule.java
Pablete1234: Advance time is already being set here

Pablete1234 (approved):

PR #1715, opened 2026-08-04
fix: remap ender pearl entity to ender pearl item

TTtie:
Remaps ender pearl entities to the ender pearl item as ender pearls can kill, and they don't have a translation on 1.8.


cswhite2000 (approved):

cswhite2000:
You'll need to update this branch, I can't seem to do so to merge it.

PR #1723, opened 2026-08-26
Optimize block lookups on large destroyables

calcastor:


Pablete1234:
Added further changes to:
 - Only attempt to process destroyable block changes at monitor if we marked as "affects destroyables" at the high
 - Don't iterate destroyables and check for regions twice at highest, instead do a single iteration/contains check. The contains inside testBlockChange is also removed as it's done in the parent.
 - For destroyables using multi-stage blocks, optimize their memory use by recycling the material -> health maps computed on cycle.
 - Fix a bug where multi-stage block destroyables would have incorrect max health if any of the blocks was less-than-max.

cswhite2000 (dismissed):

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -92,15 +90,15 @@ public void testBlockChange(BlockTransformEvent event) {
    */
   @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
   public void handleBlockChange(BlockTransformEvent event) {
-    if (this.match.getWorld() != event.getWorld() || !this.anyDestroyableAffected(event)) {
-      return;
-    }
+    if (this.match.getWorld() != event.getWorld() || !breakAffectsDestroyable) return;
+
+    breakAffectsDestroyable = false; // Consider handled
```
TTtie: Multi-block breaks (think TNT) fire block transform event handlers for every priority [for the whole batch](https://github.com/PGMDev/PGM/blob/0d0ce31ef518fbdfde14e2efbbe393768999a305/core/src/main/java/tc/oc/pgm/listeners/BlockTransformListener.java#L105-L143) (i.e. a `HIGHEST` priority event is dispatched for all destroyed blocks first, then a `MONITOR` event gets dispatched for all destroyed blocks). This effectively causes all monument breakages involving TNT to be counted only as if a single block broke despite it actually breaking multiple, effectively making some of the maps unfinishable.

TTtie (changes requested):

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -90,15 +92,18 @@ private boolean shouldCancel(Event cause, ParticipantState player) {
...
+    long pos = event.getPos();
+    BlockState oldState = event.getOldState();
     ParticipantState player = ParticipantBlockTransformEvent.getPlayerState(event);
+    BlockState newState = null;
 
     for (Destroyable destroyable : this.destroyables) {
-      destroyable.handleBlockChange(event.getOldState(), event.getNewState(), player, pos);
+      if (destroyable.isDestroyed() || !destroyable.getBlockRegion().containsPos(pos)) continue;
+
+      if (newState == null) newState = event.getNewState();
```
Pablete1234: what are you even gaining by delaying calling getNewState, its a getter not a complex operation with side effects.

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -90,15 +92,18 @@ private boolean shouldCancel(Event cause, ParticipantState player) {
...
 
-    long pos = BlockVectors.encodePos(event.getOldState());
+    long pos = event.getPos();
+    BlockState oldState = event.getOldState();
     ParticipantState player = ParticipantBlockTransformEvent.getPlayerState(event);
+    BlockState newState = null;
 
     for (Destroyable destroyable : this.destroyables) {
-      destroyable.handleBlockChange(event.getOldState(), event.getNewState(), player, pos);
+      if (destroyable.isDestroyed() || !destroyable.getBlockRegion().containsPos(pos)) continue;
```
Pablete1234: Given this always needs to be done, just leave it where it was inside the handleBlockChange.

The only reason it was lifted up from the testBlockChange method is that we needed the parent to do it to know if a cancel may be warranted or not

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -50,31 +52,31 @@ public Collection<Destroyable> getDestroyables() {
   public void testBlockChange(BlockTransformEvent event) {
     if (this.match.getWorld() != event.getWorld()) return;
 
-    long pos = BlockVectors.encodePos(event.getBlock());
+    long pos = event.getPos();
+    BlockState oldState = event.getOldState();
     ParticipantState player = ParticipantBlockTransformEvent.getPlayerState(event);
+    boolean shouldCancel = shouldCancel(event.getCause(), player);
```
Pablete1234: testing for cancellation when we don't know if any destroyable is affected at all is doing more work than it needs to, there's a reason i wrote it like i did

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -50,31 +52,31 @@ public Collection<Destroyable> getDestroyables() {
...
-      if (!anyDestroyableAffected && shouldCancel(event.getCause(), player)) {
+      if (shouldCancel) {
         event.setCancelled(true);
         return;
       }
-      anyDestroyableAffected = true;
 
-      String reasonKey =
-          destroyable.testBlockChange(event.getOldState(), event.getNewState(), player, pos);
+      if (newState == null) newState = event.getNewState();
```
Pablete1234: again, you save nothing by avoiding calling getNewState

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
```diff
@@ -90,15 +92,18 @@ private boolean shouldCancel(Event cause, ParticipantState player) {
    */
   @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
   public void handleBlockChange(BlockTransformEvent event) {
-    if (this.match.getWorld() != event.getWorld() || !breakAffectsDestroyable) return;
-
-    breakAffectsDestroyable = false; // Consider handled
```
Pablete1234: instead of removing the affects destroyable, make it depend on the cause event instead of a boolean yes/no, so instead of clearing it just check if your cause is the same cause we're handling. If it fails on the test method set it to null.

Pablete1234 (changes requested):

core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
```diff
@@ -241,9 +248,8 @@ protected void buildMaterialHealthMap() {
 
         long pos = BlockVectors.encodePos(block);
         this.blockMaterialHealth.put(pos, materialHealthMap);
-        var blockHealth = this.getBlockHealth(block.getState(), pos);
-        this.maxHealth += blockHealth;
-        this.health += blockHealth;
+        this.maxHealth += blockMaxHealth;
+        this.health += this.getBlockHealth(block.getState(), pos);
       }
```
Pablete1234, line 250: why is this being reverted, this change fixed a bug, you cannot add more max health than you add health or you end up with destroyables that start at <100% health.

core/src/main/java/tc/oc/pgm/destroyable/Destroyable.java
```diff
@@ -256,7 +261,7 @@ protected void buildMaterialHealthMap() {
   }
 
   protected int buildBlockMaterialHealthMap(
-      Block block,
+      BlockState state,
```
Pablete1234: This was fine as it was, you are passing a block AND a material, meaning that you have to assume that block as if it was made of that material. Passing a blockstate means there's a redundant amount of set, both the material in the blockstate and the actual material elsewhere in the method.

If the only reason for this is querying the blockDropRuleSet requiring a state, that's what you should be looking to change or ad an alternative to instead.

Note that internally getBlocks ends up doing
`block = BlockStates.cloneWithMaterial(block.getBlock(), material);` where `block` is the input `BlockState`.

You could simply pass a block to the function and get rid of the blockstate input all together.

util/src/main/java/tc/oc/pgm/util/block/BlockVectors.java
```diff
@@ -84,40 +84,46 @@ static BlockVector getRelative(BlockVector base, BlockFace dir, @Nullable BlockV
...
-    }
+  long X_MASK = (1L << X_BITS) - 1;
+  long Y_MASK = (1L << Y_BITS) - 1;
+  long Z_MASK = (1L << Z_BITS) - 1;
 
-    return packed;
+  /** Decode a single component from the packed coordinates */
+  private static long unpack(long packed, int shift, int bits) {
+    return (packed << (64 - shift - bits)) >> (64 - bits);
   }
```
Pablete1234, line 98: This change is breaking negative coordinates.

core/src/main/java/tc/oc/pgm/regions/FiniteBlockRegion.java
```diff
@@ -25,38 +25,62 @@
...
+  public boolean containsPos(long encodedPos) {
+    int x = BlockVectors.unpackX(encodedPos);
+    if (x < minX || x > maxX) return false;
+    int y = BlockVectors.unpackY(encodedPos);
+    if (y < minY || y > maxY) return false;
+    int z = BlockVectors.unpackZ(encodedPos);
+    if (z < minZ || z > maxZ) return false;
+
+    return positions.contains(encodedPos);
+  }
```
Pablete1234, line 82: Have you profiled doing all this being actually faster than the direct contains check against the hashmap? i have my suspicion that all the added branching likely makes it slower than the direct check

Pablete1234:
I've fixed the destroyable part i believe, reverting a bunch of your added changes that were actually breaking new things.
There are things that are plainly broken by this pr, like negative coordinates and positive coordinates encoding to the same values, meaning an objective centered around 0,0 would not be able to have all its blocks accounted for.

I have not looked into the core side yet, but the amount of changes being done is concerning when they have so many small broken things, like the handling of max health.

If you're using an AI to code these, please refrain from accepting line changes you don't fully understand the exact rationale behind changing, all those small changes they are breaking more than they are fixing.

core/src/main/java/tc/oc/pgm/destroyable/DestroyableMatchModule.java
calcastor: Would it not behave as more than a getter with a block-drops rule applied to a monument? At least one map, Blocks II, did this for cores at one point; I can't think of a DTM map which does it off the top of my head but it's my understanding that sort of functionality was supported intentionally.

util/src/main/java/tc/oc/pgm/util/block/BlockVectors.java
calcastor: What scenarios does it break them in?

